PR #17, opened 2022-04-20
Update ydb-kubernetes-operator.md

fireharp:
Cluster creation manifest should be updated 

`dataStore` – expecting array
`erasure` – is required


mvgorbunov (commented):
Hi!
We really appreciate your help, but you don't need to write this yaml configuration manually - we've added examples for both storage and database and already fixed it in documentation.
https://github.com/ydb-platform/ydb-kubernetes-operator/tree/master/samples
Try to use these configuration from examples.

PR #158, opened 2023-03-20
quick fix ----min-partitions

indiff:
quick fix ----min-partitions
图片


alexv-smirnov:
Hi! This typo was fixed by the other member of our team already, so closing this PR, but appreciate the further ones, as we now can handle them faster.

PR #160, opened 2023-03-22
Fix the warning 'RWF_APPEND' macro redefined

heckad:


dcherednik:
We have solve it a bit different way. https://github.com/ydb-platform/ydb/commit/215150b85af0221f3ccf0e8681f8adef1efe6ca9

The reason is no guarantee the representation of this macro is same for all linux version, but numeric value should be same - it is part of ABI.

Thanks!

PR #173, opened 2023-04-09
docs -> yql -> types -> optional : typo

AlexeyBukin:
I have found those zeros confusing, checked the code snippet in the current version of ydb, and got a different result (that makes more sense).

I hereby agree to the terms of the CLA available at:  https://yandex.ru/legal/cla/?lang=en


alexv-smirnov:
KIKIMR-17945

PR #189, opened 2023-05-03
Improve performance of import

kitaisreal:
Improve performance of import

### DIFF
--- a/ydb/public/lib/ydb_cli/import/import.cpp
+++ b/ydb/public/lib/ydb_cli/import/import.cpp
@@ -21,6 +21,7 @@
 #include <util/stream/file.h>
 #include <util/stream/length.h>
 #include <util/string/builder.h>
+#include <util/system/thread.h>
 
 #include <contrib/libs/apache/arrow/cpp/src/arrow/api.h>
 #include <contrib/libs/apache/arrow/cpp/src/arrow/io/api.h>
@@ -34,6 +35,7 @@
 
 #if defined(_win32_)
 #include <windows.h>
+#include <io.h>
 #elif defined(_unix_)
 #include <unistd.h>
 #endif
@@ -82,9 +84,18 @@ FHANDLE GetStdinFileno() {
 #endif
 }
 
+bool IsStdoutInteractive() {
+#if defined(_win32_)
+    return _isatty(_fileno(stdout));
+#elif defined(_unix_)
+    return isatty(fileno(stdout));
+#endif
+    return true;
+}
+
 class TMaxInflightGetter {
 public:
-    TMaxInflightGetter(ui64 totalMaxInFlight, std::atomic<ui64>& filesCount) 
+    TMaxInflightGetter(ui64 totalMaxInFlight, std::atomic<ui64>& filesCount)
         : TotalMaxInFlight(totalMaxInFlight)
         , FilesCount(filesCount) {
     }
@@ -134,7 +145,7 @@ private:
     };
 
 public:
-    TCsvFileReader(const TString& filePath, const TImportFileSettings& settings, TString& headerRow, 
+    TCsvFileReader(const TString& filePath, const TImportFileSettings& settings, TString& headerRow,
                    TMaxInflightGetter& inFlightGetter) {
         TFile file;
         if (filePath) {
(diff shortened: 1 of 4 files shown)


ydb/public/lib/ydb_cli/import/import.cpp
```diff
@@ -378,28 +432,51 @@ TStatus TImportFileClient::UpsertCsv(IInputStream& input, const TString& dbPath,
...
-            inFlightRequests.push_back(UpsertCsvBuffer(dbPath, buffer));
-            buffer = headerRow;
+        if (batchBytes < settings.BytesPerRequest_) {
+            continue;
+        }
+
+        if (inputSizeHint && progressCallback)
+        {
+            double progress = static_cast<double>(readBytes) / *inputSizeHint;
+            progressCallback(static_cast<size_t>(progress * 100));
```
4ertus2, line 464: Could we remove magic const 100 from here? I.e. we could call
progressCallback(readBytes, inputSizeHint)

ydb/public/lib/ydb_cli/import/import.cpp
```diff
@@ -245,42 +266,70 @@ TStatus TImportFileClient::Import(const TVector<TString>& filePaths, const TStri
...
+            }
 
-            if (resultStatus != EStatus::SUCCESS) {
-                return MakeStatus(EStatus::SCHEME_ERROR,
-                    TStringBuilder() <<  result.GetIssues().ToString() << dbPath);
+            ProgressCallbackFunc progressCallback;
+
+            if (isStdoutInteractive)
+            {
+                size_t oldProgress = 0;
```
4ertus2, line 324: Do we have guaranty that oldProgress allocated on stack would not be reused out of if () scope?

ydb/public/lib/ydb_cli/import/import.cpp
```diff
@@ -585,49 +691,61 @@ TStatus TImportFileClient::UpsertParquet([[maybe_unused]] const TString& filenam
...
+                        // Single row or fits into the byte limit.
+                        auto value = UpsertParquetBuffer(dbPath, NYdb_cli::NArrow::SerializeBatchNoCompression(rowsBatch), strSchema);
+                        auto status = value.ExtractValueSync();
+                        if (!status.IsSuccess())
+                            return status;
+
+                        uploadedRowsCallback(rowsBatch->num_rows());
+                    } else {
+                        // Split current slice.
+                        rowsToSendBatches.push(rowsBatch->Slice(rowsBatch->num_rows() / 2));
```
4ertus2, line 735: minor:

i64 halfLen = rowsBatch->num_rows() / 2;
rowsToSendBatches.push(rowsBatch->Slice(halfLen));
rowsToSendBatches.push(rowsBatch->Slice(0, halfLen));

ydb/public/lib/ydb_cli/import/import.cpp
```diff
@@ -585,49 +691,61 @@ TStatus TImportFileClient::UpsertParquet([[maybe_unused]] const TString& filenam
...
-                    // Single row or fits into the byte limit.
-                    auto status = WaitForQueue(inFlightGetter.GetCurrentMaxInflight(), inFlightRequests);
-                    if (!status.IsSuccess()) {
-                        return status;
+                    // Nothing to send. Continue.
+                    if (rowsBatch->num_rows() == 0) {
+                        continue;
                     }
+                    // Logarithmic approach to find number of rows fit into the byte limit.
+                    if (rowsBatch->num_rows() == 1 || NYdb_cli::NArrow::GetBatchDataSize(rowsBatch) < settings.BytesPerRequest_) {
```
4ertus2, line 725: It's better to make SerializeBatchNoCompression() and check result data size over settings.BytesPerRequest_. Then GetBatchDataSize() is not needed.

ydb/public/lib/ydb_cli/import/import.cpp
kitaisreal: We catch `oldProgress` in lambda by value, and then pass this lambda by value into upsert function.

ydb/public/lib/ydb_cli/import/import.cpp
kitaisreal: If we serialize batch using `SerializeBatchNoCompression` and result size will be `>= settings.BytesPerRequest_`, then we must somehow split batch, and throw away serialized result. 
Method `GetBatchDataSize` is not heavy, so it does not lead to some performance issues.

ydb/public/lib/ydb_cli/import/import.cpp
kitaisreal: Updated.

4ertus2 (approved):

PR #193, opened 2023-05-04
Fix stacktraces builds

kitaisreal:
Fix stacktraces in address sanitizer build


dcherednik:
Does -fno-omit-frame-pointer has performance impact?

-g ok.

kitaisreal:
@dcherednik updated.

PR #243, opened 2023-06-07
CLI added progress bar for import command

kitaisreal:
CLI added progress bar for import command


ydb/public/lib/ydb_cli/import/import.cpp
```diff
@@ -277,11 +270,12 @@ TStatus TImportFileClient::Import(const TVector<TString>& filePaths, const TStri
     std::mutex progressWriteLock;
     std::atomic<ui64> globalProgress{0};
 
+    TProgressBar progressBar(100);
```
Gazizonoki, line 273: Maybe we will print current processed file size, like: [200Mb/1000Mb]. It may be more useful info for user. We anyway print current percentage in a progress bar

Gazizonoki (approved):

PR #262, opened 2023-06-14
Added prompt and --force flag in ydb config profile replace command

Gazizonoki:


ydb/public/lib/ydb_cli/commands/ydb_profile.cpp
```diff
@@ -1043,11 +1043,12 @@ int TCommandUpdateProfile::Run(TConfig& config) {
 }
 
 TCommandReplaceProfile::TCommandReplaceProfile()
-            : TCommandProfileCommon("replace", {}, "Create new configuration profile or delete and re-configure existing one")
+            : TCommandProfileCommon("replace", {}, "Deletes profile and creates a new with the same name")
```
pnv1, line 1046: `Deletes profile and creates a new one with the same name and new property values from provided options. If no options are provided, a profile will be replaced with an empty one`
Not sure about the last sentence so its up to you

ydb/public/lib/ydb_cli/commands/ydb_profile.cpp
```diff
@@ -1058,7 +1059,15 @@ void TCommandReplaceProfile::Parse(TConfig& config) {
 
 int TCommandReplaceProfile::Run(TConfig& config) {
     auto profileManager = CreateProfileManager(config.ProfileFile);
-    profileManager->RemoveProfile(ProfileName);
+    if (profileManager->HasProfile(ProfileName)) {
+        if (!Force) {
+            Cout << "Profile data will be permanently removed. Continue? (y/n): ";
```
pnv1, line 1064: `Current profile will be replaced with a new one. All current profile data will be lost. Continue?`

Otherwise its sounds too dangerous

ydb/public/lib/ydb_cli/commands/ydb_profile.cpp
```diff
@@ -1058,7 +1059,15 @@ void TCommandReplaceProfile::Parse(TConfig& config) {
...
+            if (!AskYesOrNo()) {
+                return EXIT_SUCCESS;
```
pnv1, line 1066: It won't be success I guess...

pnv1 (changes requested):

pnv1 (approved):

alexv-smirnov (approved):

PR #275, opened 2023-06-26
SerializedCellVec improve serialization

kitaisreal:
SerializedCellVec improve serialization

### DIFF
--- a/ydb/core/scheme/scheme_tablecell.cpp
+++ b/ydb/core/scheme/scheme_tablecell.cpp
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
     };
 }
 
+namespace {
+
+#pragma pack(push,4)
+    struct TValue {
+        ui32 Size : 31;
+        ui32 IsNull : 1;
+    };
+#pragma pack(pop)
+
+    static constexpr ui8 CellVecCurrentSerializationVersion = 1;
+
+    Y_FORCE_INLINE void SerializeCellVec(const TConstArrayRef<TCell>& cells, TString & resultBuffer, TVector<TCell> * resultCells) {
+        if (cells.empty())
+            return;
+
+        size_t size = sizeof(ui16) + sizeof(ui8) + sizeof(ui16);
+        for (auto& cell : cells) {
+            size += sizeof(TValue) + cell.Size();
+        }
+
+        resultBuffer.resize(size);
+        char * resultBufferData = const_cast<char *>(resultBuffer.data());
+
+        ui16 header = std::numeric_limits<ui16>::max();
+        WriteUnaligned<ui16>(resultBufferData, header);
+        resultBufferData += sizeof(header);
+
+        ui8 serialization_version = CellVecCurrentSerializationVersion;
+        WriteUnaligned<ui8>(resultBufferData, serialization_version);
+        resultBufferData += sizeof(serialization_version);
+
+        ui16 cells_size = cells.size();
+        WriteUnaligned<ui16>(resultBufferData, cells_size);
+        resultBufferData += sizeof(cells_size);
+
+        for (size_t i = 0; i < cells_size; ++i) {
+            TValue header;
+            header.Size = cells[i].Size();
+            header.IsNull = cells[i].IsNull();
+            memcpy(resultBufferData, &header, sizeof(header));
+            resultBufferData += sizeof(header);
+        }
+
+        if (resultCells) {
+            resultCells->resize(cells.size());
+        }
+
+        for (size_t i = 0; i < cells_size; ++i) {
+            const auto & cell = cells[i];
+            memcpy(resultBufferData, cell.Data(), cell.Size());
+            resultBufferData += cell.Size();
+
+            if (resultCells) {
+                (*resultCells)[i] = TCell(resultBuffer.data() + resultBuffer.Size() - cell.Size(), cell.Size());
+            }
+        }
+    }
+
+    Y_FORCE_INLINE bool TryDeserializeCellVec(const TString & data, TString & resultBuffer, TVector<TCell> & resultCells) {
+        resultCells.clear();
+        if (data.empty())
+            return true;
+
+        const char* buf = data.data();
+        const char* bufEnd = data.data() + data.size();
+        if (bufEnd - buf < static_cast<i64>(sizeof(ui16)))
+            return false;
+
+        ui16 header = ReadUnaligned<ui16>(buf);
+        buf += sizeof(header);
+
+        if (header != std::numeric_limits<ui16>::max())
+        {
+            ui16 cells_size = header;
+            resultCells.resize(cells_size);
+
+            for (ui32 i = 0; i < cells_size; ++i) {
+                if (bufEnd - buf < (long)sizeof(TValue))
+                    return false;
+
+                const TValue cellHeader = ReadUnaligned<TValue>((const TValue*)buf);
+                if (bufEnd - buf < (long)sizeof(TValue) + cellHeader.Size)
+                    return false;
+
+                resultCells[i] = cellHeader.IsNull ? TCell() : TCell((const char*)((const TValue*)buf + 1), cellHeader.Size);
+                buf += sizeof(TValue) + cellHeader.Size;
+            }
+
+            resultBuffer = data;
+            return true;
+        }
+
+        if (bufEnd - buf < static_cast<i64>(sizeof(ui8) + sizeof(ui16)))
+            return false;
+
+        ui8 serialization_version = ReadUnaligned<ui8>(buf);
+        buf += sizeof(serialization_version);
+        Y_VERIFY(serialization_version == CellVecCurrentSerializationVersion);
+
+        ui16 cells_size = ReadUnaligned<ui16>(buf);
+        buf += sizeof(cells_size);
+
+        if (bufEnd - buf < static_cast<i64>(cells_size * sizeof(TValue))) {
+            return false;
+        }
+
+        const TValue * cellsHeaders = reinterpret_cast<const TValue *>(buf);
+        buf += sizeof(cellsHeaders[0]) * cells_size;
+
+        i64 size = 0;
+        for (size_t i = 0; i < cells_size; ++i) {
+            TValue cellHeader = cellsHeaders[i];
+            size += cellHeader.Size;
+        }
+
+        i64 availableBufferSpace = bufEnd - buf;
+        if (availableBufferSpace - size != 0) {
+            return false;
+        }
+
+        resultBuffer = data;
+        resultCells.resize(cells_size);
+
+        for (size_t i = 0; i < cells_size; ++i) {
+            TValue cellHeader = cellsHeaders[i];
+            resultCells[i] = cellHeader.IsNull ? TCell() : TCell(buf, cellHeader.Size);
+            buf += cellHeader.Size;
+        }
+
+        return true;
+    }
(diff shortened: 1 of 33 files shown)


ydb/core/scheme/scheme_tablecell.h
```diff
@@ -415,6 +415,8 @@ static_assert(std::is_nothrow_default_constructible_v<TOwnedCellVec>, "Expected
 // When loading from a buffer the cells will point to the buffer contents
 class TSerializedCellVec {
 public:
+    explicit TSerializedCellVec(const TConstArrayRef<TCell>& cells);
```
snaury, line 418: I'm on a fence about this one. I think we avoided this constructor in the past, because it's very expensive and we didn't want to use it unintentionally, but it's probably not a problem in practice. But in any case `TConstArrayRef` should always be passed by value, never by reference.

ydb/core/scheme/scheme_tablecell.h
```diff
@@ -466,41 +468,16 @@ class TSerializedCellVec {
...
-            return TString();
-
-        size_t sz = sizeof(ui16);
-        for (auto& c : cells) {
-            sz += sizeof(TValue) + c.Size();
-        }
-
-        TString res;
-        res.reserve(sz);
+    static void Serialize(TString& res, const TConstArrayRef<TCell>& cells);
```
snaury, line 478: `TConstArrayRef` must be passed by value, it's a simple struct with ptr and size and you don't want double pointer indirection.

snaury (changes requested):

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+    }
+
+    Y_FORCE_INLINE bool TryDeserializeCellVec(const TString & data, TString & resultBuffer, TVector<TCell> & resultCells) {
+        resultCells.clear();
+        if (data.empty())
+            return true;
+
+        const char* buf = data.data();
+        const char* bufEnd = data.data() + data.size();
+        if (bufEnd - buf < static_cast<i64>(sizeof(ui16)))
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+        ui32 Size : 31;
+        ui32 IsNull : 1;
+    };
+#pragma pack(pop)
+
+    static constexpr ui8 CellVecCurrentSerializationVersion = 1;
+
+    Y_FORCE_INLINE void SerializeCellVec(const TConstArrayRef<TCell>& cells, TString & resultBuffer, TVector<TCell> * resultCells) {
+        if (cells.empty())
+            return;
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
     };
 }
 
+namespace {
+
+#pragma pack(push,4)
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+            resultBuffer = data;
+            return true;
+        }
+
+        if (bufEnd - buf < static_cast<i64>(sizeof(ui8) + sizeof(ui16)))
+            return false;
+
+        ui8 serialization_version = ReadUnaligned<ui8>(buf);
+        buf += sizeof(serialization_version);
+        Y_VERIFY(serialization_version == CellVecCurrentSerializationVersion);
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+        buf += sizeof(cellsHeaders[0]) * cells_size;
+
+        i64 size = 0;
+        for (size_t i = 0; i < cells_size; ++i) {
+            TValue cellHeader = cellsHeaders[i];
+            size += cellHeader.Size;
+        }
+
+        i64 availableBufferSpace = bufEnd - buf;
+        if (availableBufferSpace - size != 0) {
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.h
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.h
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+        resultBufferData += sizeof(serialization_version);
+
+        ui16 cells_size = cells.size();
+        WriteUnaligned<ui16>(resultBufferData, cells_size);
+        resultBufferData += sizeof(cells_size);
+
+        for (size_t i = 0; i < cells_size; ++i) {
+            TValue header;
+            header.Size = cells[i].Size();
+            header.IsNull = cells[i].IsNull();
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+    };
+#pragma pack(pop)
+
+    static constexpr ui8 CellVecCurrentSerializationVersion = 1;
+
+    Y_FORCE_INLINE void SerializeCellVec(const TConstArrayRef<TCell>& cells, TString & resultBuffer, TVector<TCell> * resultCells) {
+        if (cells.empty())
+            return;
+
+        size_t size = sizeof(ui16) + sizeof(ui8) + sizeof(ui16);
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,160 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+        }
+
+        resultBuffer.resize(size);
+        char * resultBufferData = const_cast<char *>(resultBuffer.data());
+
+        ui16 header = std::numeric_limits<ui16>::max();
+        WriteUnaligned<ui16>(resultBufferData, header);
+        resultBufferData += sizeof(header);
+
+        ui8 serialization_version = CellVecCurrentSerializationVersion;
```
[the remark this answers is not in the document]
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,187 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
 
+namespace {
+
+    struct TCellHeader {
+        TCellHeader() = default;
+
+        TCellHeader(ui32 rawValue) : RawValue(rawValue) {}
+
+        TCellHeader(ui32 cellSize, bool isNull)
+            : RawValue(cellSize | (static_cast<ui32>(isNull) << 31ULL))
```
snaury, line 74: You shouldn't use `ULL` on the right operand of bitwise shift, just use a plain `31`.

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -63,6 +63,187 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
+                                         bool useOldSerialization = false) {
+        resultBuffer.clear();
+        if (resultCells)
+            resultCells->clear();
+
+        if (cells.empty())
+            return;
+
+        size_t size = 0;
+        if (!useOldSerialization) {
```
snaury, line 98: I think two separate functions would be cleaner than a bunch of conditionals, and they should be available from user code, e.g. user should be able to call and test `TSerializedCellVec::Serialize_v0` vs `TSerializedCellVec::Serialize_v1`. Then constructor and default serialize could call either v0 or v1 variant.

Also there should be a test (using public apis) that serializes and deserializes both formats irrespective of current default.

ydb/core/scheme/scheme_tablecell_ut.cpp
```diff
@@ -193,13 +193,68 @@ Y_UNIT_TEST_SUITE(Scheme) {
...
+        UNIT_ASSERT_EQUAL(cellHeader.IsNull(), isNull);
+
+        return cellHeader.RawValue;
+    }
+
+    Y_UNIT_TEST(TSerializedCellVecCellHeader) {
+        std::vector<size_t> sizes = {0, 1, 10, 20, 50, 100, 200, 500, 1000, 1500, 2000, 5000, std::numeric_limits<ui32>::max() >> 1};
+        for (size_t size : sizes)
+        {
+            UNIT_ASSERT_EQUAL(GetCellHeaderRawValueOld(size, true), GetCellHeaderRawValueNew(size, true));
```
snaury, line 246: You're not testing any real code here. If you want to make sure `TCellHeader` in scheme_tablecell.cpp behaves a certain way, you should probably make that class publicly available and tests its results. The result of `TCellHeader` is just math, so result can be compared with a precomputed expected value (so if serialization changes test would fail instead passing because two changes cancelled each other). The result of bitfields could be verified using memcpy or WriteUnaligned/ReadUnaligned using some buffer and again compared with the expected ui32 value, although it's not clear why, new code is not using it anyway. Even better if tests would have some preserialized examples that we expect to deserialize a certain way (so it would fail on big endian, since current format is not portable between little and big endian).

Testing round trip in binary serialization is often not enough to catch compatibility issues and changes.

snaury (changes requested):

ydb/core/scheme/scheme_tablecell.cpp
kitaisreal: Updated.

kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell.cpp
kitaisreal: Updated.

ydb/core/scheme/scheme_tablecell_ut.cpp
kitaisreal: Updated.

snaury (approved):

PR #294, opened 2023-07-05
Datashard part iterator improve performance

kitaisreal:
Datashard part iterator improve performance

### DIFF
--- a/ydb/core/tablet_flat/flat_part_iter_multi.h
+++ b/ydb/core/tablet_flat/flat_part_iter_multi.h
@@ -692,7 +692,8 @@ namespace NTable {
             size_t KeyCellDefaultsSize = KeyCellDefaults->Size();
             Key.resize(KeyCellDefaultsSize);
 
-            for (size_t pos = 0; pos < KeyCellDefaultsSize; ++pos) {
+            size_t infoSize = Part->Scheme->Groups[0].ColsKeyData.size();
+            for (size_t pos = infoSize; pos < KeyCellDefaultsSize; ++pos) {
                 Key[pos] = (*KeyCellDefaults)[pos];
             }
         }
@@ -1152,10 +1153,11 @@ namespace NTable {
 
             KeyInitialized = true;
 
-            auto& info = Part->Scheme->Groups[0].ColsKeyData;
-            size_t infoSize = info.size();
-            for (size_t pos = 0; pos < infoSize; ++pos) {
-                Key[pos] = Main.GetRecord()->Cell(info[pos]);
+            auto it = Key.begin();
+            auto* record = Main.GetRecord();
+            for (auto & info : Part->Scheme->Groups[0].ColsKeyData) {
+                *it = record->Cell(info);
+                ++it;
             }
         }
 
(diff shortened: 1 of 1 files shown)


ydb/core/tablet_flat/flat_part_iter_multi.h
```diff
@@ -690,6 +688,13 @@ namespace NTable {
                 Groups.emplace_back(Part->GetGroupIndex(groupId), groupId);
                 GroupRemap[group] = idx;
             }
+
+            size_t KeyCellDefaultsSize = KeyCellDefaults->Size();
+            Key.resize(KeyCellDefaultsSize);
+
+            for (size_t pos = 0; pos < KeyCellDefaultsSize; ++pos) {
```
snaury, line 695: It should start from `Part->Scheme->Groups[0].ColsKeyData.size()` (data page key schema), otherwise it's just a waste of cpu. We usually have it equal to `KeyCellDefaultsSize`, i.e. all key columns are present in the sst. This is used to fill missing key columns with a default value when adding a new key column.

ydb/core/tablet_flat/flat_part_iter_multi.h
```diff
@@ -1247,7 +1250,8 @@ namespace NTable {
         mutable TSmallVec<TPartGroupRowIt> Groups;
 
         // Key is lazily initialized so needs to be mutable
-        mutable TSmallVec<TCell> Key;
```
snaury, line 1250: Using TSmallVec is an important optimization, since we often have a very small number of key columns (e.g. 1) and we'd like to avoid extra allocations in those cases.

ydb/core/tablet_flat/flat_part_iter_multi.h
```diff
@@ -1142,17 +1147,15 @@ namespace NTable {
...
-                }
+            KeyInitialized = true;
 
-                for (size_t pos = info.size(); pos < KeyCellDefaults->Size(); ++pos) {
-                    Key.push_back((*KeyCellDefaults)[pos]);
-                }
+            auto& info = Part->Scheme->Groups[0].ColsKeyData;
+            size_t infoSize = info.size();
+            for (size_t pos = 0; pos < infoSize; ++pos) {
+                Key[pos] = Main.GetRecord()->Cell(info[pos]);
```
snaury, line 1158: Makes me wonder if this could be more efficient:

```
auto it = Key.begin();
auto* record = Main.GetRecord();
for (auto& info : Part->Schema->Groups[0].ColsKeyData) {
    *it++ = record->Cell(info);
}
```

I think iterations using pointers are often more efficient than indexes?

snaury (changes requested):

ydb/core/tablet_flat/flat_part_iter_multi.h
kitaisreal: Updated.

ydb/core/tablet_flat/flat_part_iter_multi.h
kitaisreal: Updated.

ydb/core/tablet_flat/flat_part_iter_multi.h
kitaisreal: Updated.

snaury (approved):

PR #305, opened 2023-07-10
SerializedCellVec improve serialize peformance

kitaisreal:
SerializedCellVec improve serialize peformance

### DIFF
--- a/ydb/core/scheme/scheme_tablecell.h
+++ b/ydb/core/scheme/scheme_tablecell.h
@@ -121,8 +121,8 @@ public:
     const char* Data() const                        { Y_VERIFY_DEBUG(!IsInline_); return Ptr; }
 #endif
 
-    void CopyData(char * dst) const {
-        if (CanInline(DataSize_)) {
+    void CopyDataInto(char * dst) const {
+        if (IsInline_) {
             switch (DataSize_) {
                 case 8: memcpy(dst, &IntVal, 8); break;
                 case 7: memcpy(dst, &IntVal, 7); break;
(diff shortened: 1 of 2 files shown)


ydb/core/scheme/scheme_tablecell.h
```diff
@@ -120,6 +120,26 @@ struct TCell {
     const char* InlineData() const                  { Y_VERIFY_DEBUG(!IsInline_); return Ptr; }
     const char* Data() const                        { Y_VERIFY_DEBUG(!IsInline_); return Ptr; }
 #endif
+
+    void CopyData(char * dst) const {
+        if (CanInline(DataSize_)) {
```
snaury, line 125: It should be `IsInline_` not `CanInline`. Then you should handle fixed size cases (including 0), and offload to memcpy everything else, using `Data()` instead of `Ptr`. I'd like code to be robust to unrelated changes.

ydb/core/scheme/scheme_tablecell.h
```diff
@@ -120,6 +120,26 @@ struct TCell {
     const char* InlineData() const                  { Y_VERIFY_DEBUG(!IsInline_); return Ptr; }
     const char* Data() const                        { Y_VERIFY_DEBUG(!IsInline_); return Ptr; }
 #endif
+
+    void CopyData(char * dst) const {
```
snaury, line 124: Name like `CopyData` lacks direction, e.g. could it maybe copy data into the cell? Maybe name it `CopyDataTo(dst)`?

snaury (changes requested):

ydb/core/scheme/scheme_tablecell.h
kitaisreal: Updated to use `IsInline_`. Althought usage of `Data()` will produce additional branches.

ydb/core/scheme/scheme_tablecell.h
kitaisreal: Updated.

snaury (approved):

PR #319, opened 2023-08-04
[docs+deploy] Add an option in quickstart to deploy a single-node instance with a real disk drive

blinkov:


blinkov:
This has actually been merged but not closed for some reason.

PR #352, opened 2023-08-29
Simplify cmake build

maximyurchuk:
I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en

Cmake build was simplified:

1) All specific flags are moved to the toolchain file;
2) The number of ccache parameters necessary to build is decreased.

I've checked compiler's calls after my changes, and they are the same as before the changes.

For some reason the build instruction requires `-O2` for ccache, looks like a mistake after  https://github.com/ydb-platform/ydb/commit/35367ac2df1649b19d9061dc41eed7c6fcad060c but I've kept such behavior.


maximyurchuk:
> Unfortunately, CMakeLists.txt files are generated. We should patch original code

Are you sure, that https://github.com/ydb-platform/ydb/blob/main/clang.toolchain is an auto-generated file? It doesn't have a standard header for autogenerated files, and it is not an ordinary `CMakeLists.txt` file

fomichev3000:
> Are you sure, that https://github.com/ydb-platform/ydb/blob/main/clang.toolchain is an auto-generated file? It doesn't have a standard header for autogenerated files, and it is not an ordinary `CMakeLists.txt` file

My mistake, it's not a good idea to review code at meetings :)
LGTM, but I ask @alexv-smirnov to double check!

alexv-smirnov:
There are no generated files in the PR, so we can take it. Checking, will approve shortly.

nikitka (commented):
LGTM

PR #467, opened 2023-12-01
Disable coordinator if no ingress tasks

kardymonds:
Do nothing if there is no one to send  barriers,

before this fix log was: `Checkpoint successfully created, going to inject barriers to 0 actor(s)"`


kardymonds:
broke a branch, i'll make a new one branch

PR #472, opened 2023-12-05
Disable coordinator if no ingress tasks

kardymonds:
Do nothing if there is no one to send barriers,

before this fix log was: `Checkpoint successfully created, going to inject barriers to 0 actor(s)`


ydb/core/fq/libs/checkpointing/checkpoint_coordinator.cpp
```diff
@@ -103,6 +103,12 @@ void TCheckpointCoordinator::Handle(NYql::NDqs::TEvReadyState::TPtr& ev) {
         AllActorsSet.insert(actorId);
     }
 
+    CC_LOG_I("ActorsToTrigger count: " << ActorsToTrigger.size() << ", ActorsToNotify count: " << ActorsToNotify.size() << ", ActorsToWaitFor count: " << ActorsToWaitFor.size());
```
dorooleg, line 106: 1. Do we need to use info instead of debug in this case? And the same below.
2. If we have different sources in the same graph ("PgSource", "S3Source") will it work fine?

kardymonds: 1. Changed to debug
2. I think yes, S3Source just not added to ActorsToTrigger (not tested)

dorooleg (approved):

CyberROFL:
@kardymonds Кажется, эти изменения сломали кучу тестов. Если так, откатите, пожалуйста.

PR #476, opened 2023-12-05
correctly handle the case with MaxInFlight > MaxMovements

vporyadke:


ijon (approved):
Can this change be tested, though?

vporyadke:
> Can this change be tested, though?

I verified it on a test cluster. This combination of settings is not meaningful enough to warrant a unit test, imo.

PR #485, opened 2023-12-07
Big checkpoint: state storage

kardymonds:
- saving any size checkpoint in several ydb-database lines


ydb/core/fq/libs/checkpoint_storage/storage_proxy.cpp
```diff
@@ -83,11 +83,15 @@ class TStorageProxy : public TActorBootstrapped<TStorageProxy> {
 static void FillDefaultParameters(NConfig::TCheckpointCoordinatorConfig& checkpointCoordinatorConfig, NConfig::TYdbStorageConfig& ydbStorageConfig) {
     auto& limits = *checkpointCoordinatorConfig.MutableStateStorageLimits();
     if (!limits.GetMaxGraphCheckpointsSizeBytes()) {
-        limits.SetMaxGraphCheckpointsSizeBytes(1099511627776);
+        limits.SetMaxGraphCheckpointsSizeBytes(1'099'511'627'776);
```
dorooleg, line 86: It's difficult to grep these kind of constants. Do we really want to use 1'099'511'627'776 instead of 1099511627776?

kardymonds: > It's difficult to grep these kind of constants. Do we really want to use 1'099'511'627'776 instead of 1099511627776?

Return back in other pr

PR #488, opened 2023-12-07
YQ-2651 use ChildPtr instead of ChildRef

zverevgeny:


dcherednik (approved):

rvu1024 (commented):
Use raw Child(). TExprNode guarantees all not null children

zverevgeny:
> Use raw Child(). TExprNode guarantees all not null children

fixed

rvu1024 (approved):

PR #491, opened 2023-12-08
[docs] try to enable single page and pdf

blinkov:


blinkov:
Doesn't seem to be working

PR #497, opened 2023-12-11
docs: make feature-parity table generic

piece-of-tart:


asmyasnikov (approved):

asmyasnikov (approved):

piece-of-tart:
Can someone look at the correctness of the build on CI? It seems to me that it is not correct, because in the version built by ydb/docs/build.sh the language is part of the url path, on the CI language transfer like a query parameter.

blinkov:
> Can someone look at the correctness of the build on CI? It seems to me that it is not correct, because in the version built by ydb/docs/build.sh the language is part of the url path, on the CI language transfer like a query parameter.

@piece-of-tart could you rebase your branch to fresh main? We've been changing the docs CI system recently and this PR doesn't look like the current version is supposed to work. There should be an automated comment about docs preview or issues. So this might have used an old version or there's some bug in CI.

P.S. How language argument is passed shouldn't matter. CI uses the same set of libraries and similar infrastructure to https://ydb.tech, but it is not supposed to be 100% identical to the production environment. The same story is applicable for `build.sh`, except it doesn't use any external infrastructure.

blinkov:
@piece-of-tart, could you rebase this on top of the current `main` branch? The documentation CI now posts comments with a build preview or errors, but it needs a newer commit to trigger.

maximyurchuk:
No movement for a long time, closing PR

PR #501, opened 2023-12-12
KIKIMR-20079: create/alter/drop group

VPolka:


ydb/core/kqp/gateway/actors/scheme.h
```diff
@@ -86,10 +86,17 @@ class TSchemeOpRequestHandler: public TRequestHandlerBase<
...
                         << ", TxId: " << response.GetTxId());
-                        
-                    TResult result;
-                    result.SetSuccess();
-                    Promise.SetValue(std::move(result));
+
+                    if (!response.GetIssues().empty()) {
+                        NYql::TIssues issues;
+                        NYql::IssuesFromMessage(response.GetIssues(), issues);
+                        Promise.SetValue(NYql::NCommon::ResultFromIssues<TResult>(NYql::TIssuesIds::SUCCESS, "", issues));
```
qrort, line 93: Should you put SUCCESS status here or maybe infer it from response.GetIssues()?

ydb/core/kqp/gateway/kqp_ic_gateway.cpp
```diff
@@ -1772,6 +1772,47 @@ class TKikimrIcGateway : public IKqpGateway {
...
+            auto& schemeTx = *ev->Record.MutableTransaction()->MutableModifyScheme();
+            schemeTx.SetWorkingDir(database);
+            schemeTx.SetOperationType(NKikimrSchemeOp::ESchemeOpAlterLogin);
+            auto& renameGroup = *schemeTx.MutableAlterLogin()->MutableRenameGroup();
+
+            renameGroup.SetGroup(settings.GroupName);
+            renameGroup.SetNewName(settings.NewName);
+
+            SendSchemeRequest(ev.Release()).Apply(
+                [renameGroupPromise](const TFuture<TGenericResult>& future) mutable {
```
qrort, line 1804: maybe [&renameGroupPromise]

ydb/core/kqp/host/kqp_gateway_proxy.cpp
```diff
@@ -1000,6 +1118,37 @@ class TKqpGatewayProxy : public IKikimrGateway {
...
+                }
+                case NYql::TAlterGroupSettings::EAction::RemoveRoles: {
+                    auto& alterGroup = *schemeTx.MutableAlterLogin()->MutableRemoveGroupMembership();
+                    alterGroup.SetGroup(group);
+                    alterGroup.SetMember(role);
+                    phyTx.MutableSchemeOperation()->MutableRemoveGroupMembership()->Swap(&schemeTx);
+                    break;
+                }
+                default:
+                    break;
```
qrort, line 1147: maybe throw here just in case

ydb/core/kqp/host/kqp_gateway_proxy.cpp
```diff
@@ -908,15 +908,133 @@ class TKqpGatewayProxy : public IKikimrGateway {
...
+        CHECK_PREPARED_DDL(CreateGroup);
+
+        auto createGroupPromise = NewPromise<TGenericResult>();
+
+        TString database;
+        if (!SetDatabaseForLoginOperation(database, GetDomainLoginOnly(), GetDomainName(), GetDatabase())) {
+            return MakeFuture(ResultFromError<TGenericResult>("Couldn't get domain name"));
+        }
+
+        NKikimrSchemeOp::TModifyScheme schemeTx;
```
UgnineSirdis, line 920: With these changes we get a new code that does the same things as in `kqp_ic_gateway.cpp`. Can we make some library with helpers that fill `TModifyScheme` proto and is used by both gateways?

ydb/core/tx/schemeshard/schemeshard__operation_alter_login.cpp
```diff
@@ -111,13 +114,39 @@ class TAlterLogin: public TSubOperationBase {
                     } else {
                         db.Table<Schema::LoginSidMembers>().Key(removeGroupMembership.GetGroup(), removeGroupMembership.GetMember()).Delete();
                         result->SetStatus(NKikimrScheme::StatusSuccess);
+                        if (response.Warning) {
```
qrort, line 117: AddGroup has `if Notice`, RemoveGroup has `if Warning`. Maybe there should be the same condition?

ydb/core/kqp/host/kqp_gateway_proxy.cpp
UgnineSirdis: It is not a strict requirement. I think we need to see if it will bring us a profit. Maybe in this particular cases protos are very simple. If so, maybe it would be better to leave this code in both gateways.

ydb/core/kqp/host/kqp_gateway_proxy.cpp
```diff
@@ -908,15 +908,133 @@ class TKqpGatewayProxy : public IKikimrGateway {
...
+
+        auto alterGroupPromise = NewPromise<TGenericResult>();
+
+        TString database;
+        if (!SetDatabaseForLoginOperation(database, GetDomainLoginOnly(), GetDomainName(), GetDatabase())) {
+            return MakeFuture(ResultFromError<TGenericResult>("Couldn't get domain name"));
+        }
+
+        if (!settings.Roles.size()) {
+            return MakeFuture(ResultFromError<TGenericResult>("No roles given for AlterGroup request"));
```
UgnineSirdis, line 958: I can't see the same check for `CreateGroup`. I think they should be equivalent in this case

ydb/core/kqp/gateway/kqp_ic_gateway.cpp
UgnineSirdis: `renameGroupPromise` is local variable. So it will be error to capture it by reference. Promises have reference counting, it is safe to copy `renameGroupPromise` into lambda.

ydb/core/tx/schemeshard/schemeshard.h
```diff
@@ -226,6 +227,16 @@ struct TEvSchemeShard {
             Record.SetReason(errStr);
         }
 
+        void AddWarningIssue(const TString& text) {
```
UgnineSirdis, line 230: Maybe better `AddWarning`/`AddNotice`?

ydb/library/login/login.cpp
```diff
@@ -160,7 +160,12 @@ TLoginProvider::TBasicResponse TLoginProvider::AddGroupMembership(const TAddGrou
     }
 
     TSidRecord& group = itGroupModify->second;
-    group.Members.insert(request.Member);
+
+    if (group.Members.count(request.Member)) {
+        response.Notice = "Role \"" + request.Member + "\" is already a member of role \"" + group.Name + "\"";
```
UgnineSirdis, line 165: It will be more effective to use `TStringBuilder` here

UgnineSirdis: And the same comment several lines below

ydb/core/kqp/gateway/actors/scheme.h
VPolka: Condition in case is already a status of response
This changes in ExecComplete status, so it means that result is success

ydb/core/kqp/host/kqp_gateway_proxy.cpp
VPolka: `CreateGroup` can be without any users, but before calling AddUsersToGroup I check that settings.Roles.size() is not 0
https://github.com/ydb-platform/ydb/pull/501/files#diff-fb95cf18d73802eec5aabc3807b82c4e98f0b3cfae62523917dab93336b0c5ebR933

ydb/core/tx/schemeshard/schemeshard__operation_alter_login.cpp
VPolka: I just made it like in pg, but yeah, maybe it is better to make it with the same levels

nepal (approved):

UgnineSirdis (approved):

PR #503, opened 2023-12-12
Do not update index record in case of upsert with value equal to already present one

dcherednik:
…ady present one


ydb/core/kqp/opt/physical/effects/kqp_opt_phy_upsert_index.cpp
```diff
@@ -542,13 +540,178 @@ TMaybeNode<TExprList> KqpPhyUpsertIndexEffectsImpl(TKqpPhyUpsertIndexMode mode,
...
+                                    .Body<TCoMap>()
+                                        .Input<TCoLookup>()
+                                            .Collection("collection")
+                                            .Lookup<TCoAsStruct>()
+                                                .Add(keyTuples)
+                                                .Build()
+                                            .Build()
+                                        .Lambda<TCoLambda>()
+                                            .Args({payload_table_row})
+                                            .Body<TExprList>()
```
qrort, line 650: Maybe put a comment here, something about that this code actually creates a tuple and what the values in it mean.

qrort (approved):

PR #513, opened 2023-12-13
KIKIMR-20100: Add background cleaning

### LINKED ISSUE #1303: KIKIMR-20100: Background удаление временных таблиц

shnikd:


ydb/core/kqp/executer_actor/kqp_scheme_executer.cpp
```diff
@@ -117,7 +118,8 @@ class TKqpSchemeExecuter : public TActorBootstrapped<TKqpSchemeExecuter> {
                 auto modifyScheme = schemeOp.GetDropTable();
                 if (Temporary) {
                     auto* dropTable = modifyScheme.MutableDrop();
-                    dropTable->SetName(dropTable->GetName() + SessionId);
+                    dropTable->SetTemporary(true);
+                    dropTable->SetSessionActorId(Target.ToString());
```
gridnevvvit, line 122: What we need here is to make it more abstract, let's say OwnerId. I assume not only sessions will create new temporary tables in the future.

ydb/core/protos/config.proto
```diff
@@ -1511,6 +1511,25 @@ message TCompactionConfig {
     optional TBorrowedCompactionConfig BorrowedCompactionConfig = 2;
 }
 
+message TBackgroundCleaningConfig {
+    optional double MaxRate = 1 [default = 0]; // unlimitted
+    optional uint64 InflightLimit = 2 [default = 10]; // TODO: consider more?
+
+    // After this interval we will try to restart
+    optional uint64 TimeoutSeconds = 3 [default = 15];
```
gridnevvvit, line 1519: We should reduce the complexity of the configuration a bit. It would be more accurate to say that the temporary table is deleted within 300 seconds of the node's disconnect. It's still acceptable if it is deleted after 350 seconds but it would be preferable to have a lower bound specified.

So, let's introduce TemporaryTableRemovalTimeout which by default will be 300 seconds.

ydb/core/protos/flat_scheme_op.proto
```diff
@@ -387,7 +389,9 @@ message TTableDescription {
     repeated TSequenceDescription Sequences = 39;
 
     optional TTableReplicationConfig ReplicationConfig = 40;
+
     optional bool Temporary = 41;
+    optional string SessionActorId = 42;
```
gridnevvvit, line 394: OwnerId

ydb/core/tx/schemeshard/schemeshard.h
```diff
@@ -92,6 +92,8 @@ struct TEvSchemeShard {
         EvProcessingRequest,
         EvProcessingResponse,
 
+        EvSessionActorAck,
```
gridnevvvit, line 95: Rename please.

ydb/core/protos/config.proto
shnikd: https://github.com/ydb-platform/ydb/pull/513/files#diff-2e781bd76ab9922cc0012cb685cbf5adcc215185dd89fb3b9f2566f407c62f3bR1530

shnikd: We will configure this parameter through the TRetrySettings.

ydb/core/tx/schemeshard/schemeshard__operation_side_effects.cpp
```diff
@@ -750,6 +771,89 @@ void TSideEffects::DoPersistDeleteShards(TSchemeShard *ss, NTabletFlatExecutor::
...
+            auto& nodeState = nodeStates[nodeId];
+            nodeState.Owners.insert(ownerActorId);
+            nodeState.RetryState.CurrentDelay =
+                TDuration::MilliSeconds(ss->BackgroundCleaningRetrySettings.GetStartDelayMs());
+        } else {
+            itNodeStates->second.Owners.insert(ownerActorId);
+        }
+
+        if (it == tempTablesByOwner.end()) {
+            ctx.Send(new IEventHandle(ownerActorId, ss->SelfId(),
```
ijon, line 795: I don't particularly like the idea to remember all session actors and probe node existence for every session individually.
All that temp table background cleaning required is to get notified about node disconnects -- disregard of actual number of sessions on that node and their specifics. And node existence probe could have used any actor (or may be service(?)) on that node, it doesn't have to be specifically session actor. 

May be there is a way to achieve that -- to remember, store and monitor nodes instead of all session actors?

ijon (changes requested):

gridnevvvit (dismissed):

ydb/core/protos/flat_scheme_op.proto
```diff
@@ -387,7 +389,9 @@ message TTableDescription {
     repeated TSequenceDescription Sequences = 39;
 
     optional TTableReplicationConfig ReplicationConfig = 40;
+
     optional bool Temporary = 41;
```
ijon, line 393: Also there is problem that temporary tables must be ignored by backups. Somehow.

ydb/core/tx/schemeshard/schemeshard_impl.cpp
```diff
@@ -6215,6 +6255,8 @@ void TSchemeShard::Handle(TEvSchemeShard::TEvNotifyTxCompletionResult::TPtr& ev,
         executed = true;
     }
 
+    HandleBackgroundCleaningCompletionResult(ev->Get()->Record.GetTxId());
```
ijon, line 6258: Inconsistent handling -- any BackgroundCleaningCompletion will result in warning `no able to determine destination ...`.
`BackgroundCleaningTxs` should be checked here the same way as other `TxIdTo...` mappings -- explicitly and with `executed` flag setting.

ydb/core/tx/schemeshard/schemeshard_impl.cpp
```diff
@@ -6162,6 +6200,8 @@ void TSchemeShard::Handle(TEvSchemeShard::TEvModifySchemeTransactionResult::TPtr
         return Execute(CreateTxReply(ev), ctx);
     }
 
+    HandleBackgroundCleaningTransactionResult(ev);
```
ijon, line 6203: Inconsistent handling -- any BackgroundCleaningTransaction will result in warning `no able to determine destination ...`
`BackgroundCleaningTxs` should be checked here the same way as other `TxIdTo...` mappings -- explicitly and with immediate return.

ijon (changes requested):
Not simplified enough after switching to single actor per node.
Also several problems with internal drop operations execution.

ydb/core/tx/schemeshard/schemeshard__background_cleaning.cpp
```diff
@@ -0,0 +1,281 @@
...
+    case NKikimrScheme::EStatus::StatusSuccess: {
+        BackgroundCleaningQueue->OnDone(pathId);
+        break;
+    }
+    case NKikimrScheme::EStatus::StatusAccepted:
+        Send(SelfId(), new TEvSchemeShard::TEvNotifyTxCompletion(record.GetTxId()));
+        break;
+    default: {
+        BackgroundCleaningQueue->OnDone(pathId);
+        EnqueueBackgroundCleaning(pathId);
```
ijon, line 257: Is there a potential for the eternal loop?

ijon (changes requested):
Minor changes are proposed. Otherwise in general looks good.

ydb/core/tx/schemeshard/schemeshard_impl.cpp
```diff
@@ -4334,7 +4362,12 @@ void TSchemeShard::StateInit(STFUNC_SIG) {
         HFunc(TEvPrivate::TEvConsoleConfigsTimeout, Handle);
 
     default:
-        StateInitImpl(ev, SelfId());
+        if (!HandleDefaultEvents(ev, SelfId())) {
+            ALOG_WARN(NKikimrServices::FLAT_TX_SCHEMESHARD,
+                       "StateInit:"
+                           << " unhandled event type: " << ev->GetTypeRewrite()
+                           << " event: " << ev->ToString());
+        }
```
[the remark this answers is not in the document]
shnikd: Here I subscribe to the nodes of the temporary tables that we uploaded at the start

ijon (approved):

CyberROFL:
@shnikd you broke schemeshard's schema tests, revert these changes

PR #564, opened 2023-12-19
update: remove russian version of feature-parity table

piece-of-tart:
If you don't figure out what problem with CI, then approve this pr


blinkov:
It appeared that our current docs infrastructure just doesn't have this feature of cross-language includes.

PR #609, opened 2023-12-20
KIKIMR-20523: VDisks balancing mvp

robdrynkin:
### Changelog entry 

New actor to send blob parts from handoffs to the main replicas and delete it from handoffs

### Changelog category 

* New feature

### Additional information

...


ydb/core/blobstorage/vdisk/hulldb/barriers/barriers_essence.cpp
```diff
@@ -91,8 +93,10 @@ namespace NKikimr {
             // is item is spread over multiple ssts?
             const bool itemIsSpreadOverMultipleSsts = recsMerged > 1;
 
+            const bool isUselessHandoffCopy = Top->IsHandoff(VDisk, id.Hash()) && ingress.GetVDiskHandoffVec(Top, VDisk, id).Empty();
```
alexvru, line 96: Решение по идее не здесь должно приниматься, а в процессе компакшена -- ведь у нас может быть несколько партов handoff, а не нужными могут стать только часть из них.

А в процессе компакшена останутся только те, которые нужны, а остальные удалятся. Если удалится всё, то блоб тоже удалится.

ydb/core/blobstorage/vdisk/balance/utils.cpp
```diff
@@ -0,0 +1,99 @@
+#include "utils.h"
+
+
+namespace NKikimr {
+
+    TVector<ui8> ApplyMask(NMatrix::TVectorType vec, NMatrix::TVectorType mask) {
```
alexvru, line 6: Название какое-то непонятное.

alexvru (changes requested):

ydb/core/blobstorage/vdisk/balance/handoff_map.h
```diff
@@ -202,117 +174,110 @@ namespace NKikimr {
...
+
+        const NMatrix::TVectorType delPlan(DelMap.at(Counter), vecSize);
+        const NMatrix::TVectorType delVec = delPlan & localVec;
+
+        if (delVec.Empty()) {
+            return defaultResult;
+        }
+
+        TIngress ingress = memRec->GetIngress(); // ingress we are going to change
+        NMatrix::TVectorType localParts = ingress.LocalParts(Top->GType);
```
alexvru, line 201: you already have localVec a few lines above, how is this different?

ydb/core/blobstorage/vdisk/balance/handoff_map.h
```diff
@@ -202,117 +174,110 @@ namespace NKikimr {
...
+        if (!RunHandoff) {
+            return defaultResult;
+        }
+
+        Y_DEBUG_ABORT_UNLESS(Counter < DelMap.size());
+        const NMatrix::TVectorType localVec = memRec->GetIngress().LocalParts(Top->GType);
+        ui8 vecSize = Top->GType.TotalPartCount();
+
+        const NMatrix::TVectorType delPlan(DelMap.at(Counter), vecSize);
+        const NMatrix::TVectorType delVec = delPlan & localVec;
```
alexvru, line 194: what's the meaning of this? DelMap already contains filtered (by local vector) parts, does it?

ydb/core/blobstorage/vdisk/ingress/blobstorage_ingress_matrix.h
```diff
@@ -389,6 +389,24 @@ namespace NKikimr {
...
+                Y_DEBUG_ABORT_UNLESS(End - Beg <= 2 * 8);
+                ui8 vec = 0;
+                TIterator it = Begin();
+                while (!it.IsEnd()) {
+                    vec <<= 1;
+                    bool firstBit = it.Get();
+                    it.Next();
+                    Y_DEBUG_ABORT_UNLESS(!it.IsEnd());
+                    bool secondBit = it.Get();
+                    Y_UNUSED(secondBit);
```
alexvru, line 402: you don't need it, why extract?

ydb/core/blobstorage/vdisk/balance/handoff_map.h
```diff
@@ -202,117 +174,110 @@ namespace NKikimr {
...
+        switch (memRec->GetType()) {
+            case TBlobType::DiskBlob: {
+                const TDiskBlob& blob = dataMerger->GetDiskBlobMerger().GetDiskBlob();
+
+                // mark deleted handoff parts in ingress
+                for (ui8 i = localParts.FirstPosition(); i != localParts.GetSize(); i = localParts.NextPosition(i)) {
+                    if (delVec.Get(i)) {
+                        const ui8 partId = i + 1;
+                        TLogoBlobID id(key.LogoBlobID(), partId);
+                        ingress.DeleteHandoff(Top.get(), HullCtx->VCtx->ShortSelfVDisk, id, true);
```
alexvru, line 212: DeleteHandoff() sets the deletion bit in handoff vector, which is already set, 'cause it is in delVec
this statement seems useless

ydb/core/blobstorage/vdisk/balance/handoff_map.h
```diff
@@ -202,117 +174,110 @@ namespace NKikimr {
...
+                TDataMerger newMerger;
+
+                Y_DEBUG_ABORT_UNLESS(oldMerger.SavedData().size() == localParts.CountBits());
+                Y_DEBUG_ABORT_UNLESS(oldMerger.SavedData().size() == oldMerger.GetCircaLsns().size());
+
+                for (ui8 i = localParts.FirstPosition(), j = 0; i != localParts.GetSize(); i = localParts.NextPosition(i), ++j) {
+                    if (delVec.Get(i)) {
+                        const ui8 partId = i + 1;
+                        TLogoBlobID id(key.LogoBlobID(), partId);
+                        ingress.DeleteHandoff(Top.get(), HullCtx->VCtx->ShortSelfVDisk, id, true);
```
alexvru, line 247: the same thing like with the DiskBlob one

ydb/core/blobstorage/vdisk/balance/handoff_map.h
```diff
@@ -202,117 +174,110 @@ namespace NKikimr {
...
+                        localParts.Clear(i);
+                        newMerger.AddDeletedHugeBlob(oldMerger.SavedData()[j]);
+                    } else {
+                        auto parts = NMatrix::TVectorType(0, localParts.GetSize());
```
alexvru, line 251: MakeOneHot()

ydb/core/blobstorage/vdisk/balance/handoff_map.h
```diff
@@ -202,117 +174,110 @@ namespace NKikimr {
...
+                        parts.Set(i);
+                        newMerger.AddHugeBlob(&oldMerger.SavedData()[j], &oldMerger.SavedData()[j] + 1, parts, oldMerger.GetCircaLsns()[j]);
+                    }
+                }
 
-        Counter++;
-        return result;
+                if (localParts.CountBits() == 0) {
```
alexvru, line 257: .Empty()

alexvru (changes requested):

ydb/core/blobstorage/vdisk/balance/handoff_map.h
robdrynkin: DelMap is constructed in BuildMap, it iterates and merge memRecs over whole lsm, not the specific level compacting now. That the reason why we could have bit set in delPlan and don't have certain bit in localVec

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,191 @@
...
+        std::pair<TQueue<TPartInfo>, TQueue<TPartInfo>> CollectKeys(const TActorContext &ctx) {
+            TQueue<TPartInfo> sendOnMainParts, tryDeleteParts;
+
+            for (It.SeekToFirst(); It.Valid(); It.Next()) {
+                TPartsCollectorMerger merger(Ctx->GInfo->GetTopology().GType);
+                It.PutToMerger(&merger);
+
+                for (ui8 partIdx: PartsToSendOnMain(Ctx->GInfo->GetTopology(), Ctx->VCtx->ShortSelfVDisk, It.GetCurKey().LogoBlobID(), merger.Ingress)) {
+                    if (!merger.Parts[partIdx - 1].has_value()) {
+                        LOG_WARN_S(ctx, NKikimrServices::BS_VDISK_BALANCING,
```
alexvru, line 51: Ctx->VCtx->VDiskLogPrefix
or else it would be impossible to distinguish problem disks
applies to all log places

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,191 @@
...
+            CreateQueuesForVDisks(*QueueActorMapPtr, SelfId(), Ctx->GInfo, Ctx->VCtx,
+                    Ctx->GInfo->GetVDisks(), Ctx->MonGroup.GetGroup(),
+                    queueClientId, NKikimrBlobStorage::EVDiskQueueId::GetAsyncRead,
+                    "DisksBalancing", interconnectChannel);
+        }
+
+        std::pair<TQueue<TPartInfo>, TQueue<TPartInfo>> CollectKeys(const TActorContext &ctx) {
+            TQueue<TPartInfo> sendOnMainParts, tryDeleteParts;
+
+            for (It.SeekToFirst(); It.Valid(); It.Next()) {
```
alexvru, line 45: this cycle seems taking too much time in single event processing (in System pool even more)

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,191 @@
...
+        void Bootstrap(const TActorContext &ctx) {
+            CreateVDisksQueues();
+            auto [sendOnMainParts, tryDeleteParts] = CollectKeys(ctx);
+            LOG_DEBUG_S(ctx, NKikimrServices::BS_VDISK_BALANCING,
+                Ctx->GInfo->GetTopology().GetOrderNumber(Ctx->VCtx->ShortSelfVDisk) << "$ "
+                << ctx.Now().MilliSeconds() << " Bootstrap" << ": "
+                << "sendOnMainParts size = " << sendOnMainParts.size() << "; tryDeleteParts size = " << tryDeleteParts.size());
+
+            SenderId = ctx.Register(CreateSenderActor(SelfId(), std::move(sendOnMainParts), QueueActorMapPtr, Ctx));
+            ActiveActors.Insert(SenderId, __FILE__, __LINE__, ctx, NKikimrServices::BLOBSTORAGE);
```
alexvru, line 176: you add actor here and take no operation when it dies and on PassAway
- on PassAway you should fan-out TEvPosion message to currently active actors
- when child actor dies, it should notify parent actor somehow (probably in TEvCompleted)

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,220 @@
...
+            ActiveActors.Insert(DeleterId, __FILE__, __LINE__, ctx, NKikimrServices::BLOBSTORAGE);
+
             Become(&TThis::StateFunc);
+            Schedule(TDuration::Seconds(1), new NActors::TEvents::TEvWakeup());
         }
 
+        void StartBalancing(const TActorContext &ctx) {
+            LOG_DEBUG_S(ctx, NKikimrServices::BS_VDISK_BALANCING,
+                        Ctx->VCtx->VDiskLogPrefix << "Ask repl token");
+            if (!Send(MakeBlobStorageReplBrokerID(), new TEvQueryReplToken(Ctx->VDiskCfg->BaseInfo.PDiskId))) {
```
alexvru, line 110: This is not very correct practice, however it is used in some other places.
The better method is to use ForwardOnNondelivery subsystem of IEventHandle and report TEvUndelivered instead of reply if actor is not found.
Typical race leading to incorrect behaviour here -- actor is running while the message is sent, but gets terminated before this message is processed.

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,220 @@
...
+                HandleReplToken(ctx);
+            }
+        }
+
+        void HandleReplToken(const TActorContext &ctx) {
+            LOG_DEBUG_S(ctx, NKikimrServices::BS_VDISK_BALANCING,
+                        Ctx->VCtx->VDiskLogPrefix << "Repl token acquired");
+            DoJobQuant(ctx);
```
alexvru, line 118: And, by the way, do we really want to interfere with actual replication?
Maybe we need to set up another queue (or priority queue) in the broker?
Otherwise this actor may speed down replication, which is much more relevant than blob balancing.

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,220 @@
...
+            Send(MakeBlobStorageReplBrokerID(), new TEvReleaseReplToken);
+            Schedule(TDuration::Seconds(1), new NActors::TEvents::TEvWakeup());
+        }
+
+        void Handle(NActors::TEvents::TEvCompleted::TPtr ev, const TActorContext &ctx) {
+            switch (ev->Get()->Id) {
+                case SENDER_ID: {
```
alexvru, line 125: Style, but why braces? :)

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,220 @@
...
+        }
+
+        void Handle(TEvVGenerationChange::TPtr ev) {
+            Cerr << "TEvVGenerationChange" << Endl;
+            // forward message to queue actors
+            TEvVGenerationChange *msg = ev->Get();
+            for (const auto& kv : *QueueActorMapPtr) {
+                Send(kv.second, msg->Clone());
+            }
+            Ctx->GInfo = msg->NewInfo;
```
alexvru, line 160: Storing something in Ctx this way is bad idea. It is assumed to be thread-safe. Either it is read-only, or it is stored locally. So it's better to keep group info as local variable.

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,220 @@
...
+            if (!Send(MakeBlobStorageReplBrokerID(), new TEvQueryReplToken(Ctx->VDiskCfg->BaseInfo.PDiskId))) {
+                HandleReplToken(ctx);
+            }
+        }
+
+        void HandleReplToken(const TActorContext &ctx) {
+            LOG_DEBUG_S(ctx, NKikimrServices::BS_VDISK_BALANCING,
+                        Ctx->VCtx->VDiskLogPrefix << "Repl token acquired");
+            DoJobQuant(ctx);
+            Send(MakeBlobStorageReplBrokerID(), new TEvReleaseReplToken);
```
alexvru, line 119: The job isn't actually done, so we can't just release this token, right? Or the whole broker gets useless here.

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -1,29 +1,220 @@
...
+            Schedule(TDuration::Seconds(1), new NActors::TEvents::TEvWakeup());
```
alexvru, line 120: The same thing -- you want multiple jobs quanta to perform concurrently?

ydb/core/blobstorage/vdisk/balance/deleter.cpp
```diff
@@ -0,0 +1,194 @@
...
+                    if (part.HasOnMain) {
+                        ++Stats.PartsDecidedToDelete;
+                        DeleteLocal(ctx, part.Key, part.Ingress);
+                    }
+                }
+            }
+            LOG_DEBUG_S(ctx, NKikimrServices::BS_VDISK_BALANCING,
+                        Ctx->VCtx->VDiskLogPrefix << "PartsRequester status " << ": " << (ui32)status << " " << Stats.PartsDecidedToDelete);
+            if (status == TPartsRequester::EState::FINISHED && Stats.PartsDecidedToDelete == Stats.PartsMarkedDeleted) {
+                Send(SelfId(), new NActors::TEvents::TEvPoison);
```
alexvru, line 131: PassAway()? There's no meaning to send yourself a message.

ydb/core/blobstorage/vdisk/balance/deleter.cpp
```diff
@@ -0,0 +1,194 @@
...
+            }
+        }
+
+        void DeleteLocal(const TActorContext &ctx, const TLogoBlobID& key, const TIngress& ingress) {
+            TLogoBlobID keyWithoutPartId(key, 0);
+
+            TIngress patchedIngress = ingress.CopyWithoutLocal(Ctx->GInfo->GetTopology().GType);
```
alexvru, line 138: I don't know where you get it, but the main idea is that every time you put something into database, you construct independent ingress containing the very local patch. In this case this is handoff deletion marker. And nothing more. It'll all be compacted to the single record.

ydb/core/blobstorage/vdisk/balance/deleter.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+        void Handle(TEvBlobStorage::TEvVGetResult::TPtr ev) {
+            ++Responses;
+            auto msg = ev->Get()->Record;
+            if (msg.GetStatus() != NKikimrProto::EReplyStatus::OK) {
+                return;
+            }
+            ui64 i = msg.GetCookie();
+            auto res = msg.GetResult().at(0);
+            for (ui32 partIdx: res.GetParts()) {
```
alexvru, line 95: Don't mix "PartId" and "PartIdx", they differ strictly by 1.

ydb/core/blobstorage/vdisk/balance/deleter.cpp
```diff
@@ -0,0 +1,194 @@
...
+                auto vDiskId = GetVDiskId(*GInfo, item.Key);
+                auto ev = TEvBlobStorage::TEvVGet::CreateExtremeIndexQuery(
+                    vDiskId, TInstant::Max(), NKikimrBlobStorage::EGetHandleClass::AsyncRead,
+                    TEvBlobStorage::TEvVGet::EFlags::None, i,
+                    {{item.Key.FullID(), 0, 0}}
+                );
+                TReplQuoter::QuoteMessage(
+                    Quoter,
+                    std::make_unique<IEventHandle>(QueueActorMapPtr->at(TVDiskIdShort(vDiskId)), ctx.SelfID, ev.release()),
+                    0
```
alexvru, line 71: This is useless unless you specify some valid weight, in bytes. The answer is really expected to consume some resources, isn't it?

ydb/core/blobstorage/vdisk/balance/utils.cpp
```diff
@@ -0,0 +1,101 @@
...
+    ) {
+        auto [_, delMask] = ingress.HandoffParts(&top, vdisk, key);
+        return GetRequiredParts(ingress.LocalParts(top.GType), delMask);
+    }
+
+    TVDiskID GetVDiskId(const TBlobStorageGroupInfo& gInfo, const TLogoBlobID& key) {
+        TBlobStorageGroupInfo::TOrderNums orderNums;
+        TBlobStorageGroupInfo::TVDiskIds vdisks;
+        gInfo.GetTopology().PickSubgroup(key.Hash(), orderNums);
+        for (const auto &x : orderNums) {
```
alexvru, line 40: Looks a bit trashy.
`return gInfo.GetVDiskId(orderNums[key.PartId() - 1]);`
?

ydb/core/blobstorage/vdisk/balance/utils.cpp
```diff
@@ -0,0 +1,101 @@
...
+        const TBlobStorageGroupInfo::TTopology& top,
+        const TVDiskIdShort &vdisk,
+        const TLogoBlobID &key,
+        const TIngress& ingress
+    ) {
+        auto [_, delMask] = ingress.HandoffParts(&top, vdisk, key);
+        return GetRequiredParts(ingress.LocalParts(top.GType), delMask);
+    }
+
+    TVDiskID GetVDiskId(const TBlobStorageGroupInfo& gInfo, const TLogoBlobID& key) {
```
alexvru, line 36: A bit confusing name. Suggest GetMainReplicaVDiskId, or something like that.

ydb/core/blobstorage/vdisk/balance/deleter.cpp
```diff
@@ -0,0 +1,194 @@
...
+                auto item = Parts.front();
+                Parts.pop();
+                Result[i] = TPartOnMain{
+                    .Key=item.Key,
+                    .Ingress=item.Ingress,
+                    .HasOnMain=false
+                };
+
+                auto vDiskId = GetVDiskId(*GInfo, item.Key);
+                auto ev = TEvBlobStorage::TEvVGet::CreateExtremeIndexQuery(
```
alexvru, line 63: I suggest batching multiple queries into single message. This would improve performance.

ydb/core/blobstorage/vdisk/balance/sender.cpp
```diff
@@ -0,0 +1,201 @@
...
+                ++ExpectedResponses;
+            }
+            return WAITING_PDISK_RESPONSES;
+        }
+
+        std::optional<TVector<TPart>> TryGetResults() {
+            if (ExpectedResponses != 0 && ExpectedResponses == Responses) {
+                ExpectedResponses = 0;
+                Responses = 0;
+                return Result;
```
alexvru, line 85: You can move it here. Or else it'll be copied.

ydb/core/blobstorage/vdisk/balance/sender.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+                    TReplQuoter::QuoteMessage(
+                        Quoter,
+                        std::make_unique<IEventHandle>(PDiskCtx->PDiskId, ctx.SelfID, ev.release()),
+                        diskPart.Size
+                    );
+                }
+                ++ExpectedResponses;
+            }
+            return WAITING_PDISK_RESPONSES;
```
alexvru, line 78: What if Responses == ExpectedResponses?

ydb/core/blobstorage/vdisk/balance/sender.cpp
```diff
@@ -0,0 +1,201 @@
...
+            Result.resize(Min(Parts.size(), BatchSize));
+            ExpectedResponses = 0;
+            for (ui64 i = 0; i < BatchSize && !Parts.empty(); ++i) {
+                auto item = Parts.front();
+                Parts.pop();
+                Result[i] = TPart{
+                    .Key=item.Key,
+                    .Ingress=item.Ingress
+                };
+                if (std::holds_alternative<TRope>(item.PartData)) {
```
alexvru, line 55: std::visit() suits this case better, I suppose. You may also use TOverloaded{} to distinguish between types.

ydb/core/blobstorage/vdisk/balance/sender.cpp
```diff
@@ -0,0 +1,201 @@
...
+            auto status = Reader.DoJobQuant(ctx);
+            if (auto batch = Reader.TryGetResults()) {
+                Stats.PartsRead += batch->size();
+                SendParts(*batch, ctx);
+            }
+            LOG_DEBUG_S(ctx, NKikimrServices::BS_VDISK_BALANCING,
+                        Ctx->VCtx->VDiskLogPrefix << "Reader status "
+                        << ": " << (ui32)status << " " << Stats.PartsRead << " " << Stats.PartsSent);
+            if (status == TReader::EReaderState::FINISHED && Stats.PartsRead == Stats.PartsSent) {
+                Send(SelfId(), new NActors::TEvents::TEvPoison);
```
alexvru, line 127: PassAway()

ydb/core/blobstorage/vdisk/balance/sender.cpp
```diff
@@ -0,0 +1,201 @@
...
+                TReplQuoter::QuoteMessage(
+                    Ctx->VCtx->ReplNodeRequestQuoter,
+                    std::make_unique<IEventHandle>(queue, SelfId(), ev.release()),
+                    part.PartData.size()
+                );
+            }
+        }
+
+        void Handle(TEvBlobStorage::TEvVPutResult::TPtr ev, const TActorContext &ctx) {
+            Stats.PartsSent += 1;
```
alexvru, line 155: I think we should have some indicators for erroneous replies. A counter may be.

alexvru (changes requested):
Suggest to get rid of const TActorContext& argument everywhere is new code.
And replace Die -> PassAway, HFunc -> hFunc, CFunc -> cFunc and so on.

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
robdrynkin: It is a copypaste from replication actor https://github.com/ydb-platform/ydb/blob/2380e383a99584b10ed0ac5ac570fd26a6593a62/ydb/core/blobstorage/vdisk/repl/blobstorage_repl.cpp#L545
Anyway it is not very clear what is a good way to update GInfo, because it is using in sender and deleter actors too

ydb/core/blobstorage/vdisk/balance/deleter.cpp
robdrynkin: I plan to optimize it in the next PR

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
robdrynkin: Probably you mean FlagTrackDelivery, not the ForwardOnNondelivery

alexvru (dismissed):

alexvru (approved):

PR #633, opened 2023-12-21
Ignore empty metrics in in-progress stats. Try-catch in-progress stats

Hor911:
+ fix time metrics and aggregates

### DIFF
--- a/ydb/core/fq/libs/compute/ydb/status_tracker_actor.cpp
+++ b/ydb/core/fq/libs/compute/ydb/status_tracker_actor.cpp
@@ -155,42 +155,46 @@ public:
     }
 
     void ReportPublicCounters(const Ydb::TableStats::QueryStats& stats) {
-        auto stat = GetPublicStat(GetV1StatFromV2Plan(stats.query_plan()));
-        auto publicCounters = GetPublicCounters();
+        try {
+            auto stat = GetPublicStat(GetV1StatFromV2Plan(stats.query_plan()));
+            auto publicCounters = GetPublicCounters();
 
-        if (stat.MemoryUsageBytes) {
-            auto& counter = *publicCounters->GetNamedCounter("name", "query.memory_usage_bytes");
-            counter = *stat.MemoryUsageBytes;
-        }
+            if (stat.MemoryUsageBytes) {
+                auto& counter = *publicCounters->GetNamedCounter("name", "query.memory_usage_bytes");
+                counter = *stat.MemoryUsageBytes;
+            }
 
-        if (stat.CpuUsageUs) {
-            auto& counter = *publicCounters->GetNamedCounter("name", "query.cpu_usage_us", true);
-            counter = *stat.CpuUsageUs;
-        }
+            if (stat.CpuUsageUs) {
+                auto& counter = *publicCounters->GetNamedCounter("name", "query.cpu_usage_us", true);
+                counter = *stat.CpuUsageUs;
+            }
 
-        if (stat.InputBytes) {
-            auto& counter = *publicCounters->GetNamedCounter("name", "query.input_bytes", true);
-            counter = *stat.InputBytes;
-        }
+            if (stat.InputBytes) {
+                auto& counter = *publicCounters->GetNamedCounter("name", "query.input_bytes", true);
+                counter = *stat.InputBytes;
+            }
 
-        if (stat.OutputBytes) {
-            auto& counter = *publicCounters->GetNamedCounter("name", "query.output_bytes", true);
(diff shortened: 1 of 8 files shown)


ydb/core/fq/libs/compute/ydb/status_tracker_actor.cpp
```diff
@@ -155,42 +155,46 @@ class TStatusTrackerActor : public TBaseComputeActor<TStatusTrackerActor> {
...
 
-        if (stat.RunningTasks) {
-            auto& counter = *publicCounters->GetNamedCounter("name", "query.running_tasks");
-            counter = *stat.RunningTasks;
+            if (stat.RunningTasks) {
+                auto& counter = *publicCounters->GetNamedCounter("name", "query.running_tasks");
+                counter = *stat.RunningTasks;
+            }
+        } catch(const NJson::TJsonException& ex) {
+            LOG_E("Error statistics conversion: " << ex.what());
```
dorooleg, line 197: Do we need to report this information to the user through transient issues?

ydb/library/yql/dq/actors/protos/dq_stats.proto
```diff
@@ -312,12 +316,21 @@ message TDqStageStats {
     TDqStatsAggr EgressBytes = 30;
     TDqStatsAggr EgressRows = 31;
 
-    TDqStatsMinMax FirstRowTimeMs = 13;
-    TDqStatsMinMax FinishTimeMs = 14;
-    TDqStatsMinMax StartTimeMs = 21;
-    uint64 DurationUs = 15; // microseconds from min(task_first_row_time) to max(task_finish_time)
+    reserved 13; // FirstRowTimeMs
```
dorooleg, line 319: Are we confident that this stat don't need here anymore?

Hor911: Long story here:

On Task level FirstRowTimeMs was never used actually. So, I've dropped this misleading field and left Start and Finish only, which seems to be self-descriptive. Task tries to report first input row timings as StartTime. "Old" StartTime (time when task was started) is reported if traffic timings are not available. It is not useful - all tasks start at about the same time, still is reported as some kind of approximation

On Stage level (aggregated stats) FirstRowTimeMs is dropped as well. Other metrics - Start and Finish are changed from Min/Max only to full set (+ Count and Sum). Sum has no meaning for timings, but Count has and Average = Sum/Count is also useful.

ydb/core/fq/libs/compute/ydb/status_tracker_actor.cpp
Hor911: Don't think so. Statistics is not UG, it is created by other part of our software, so bad format is likely to be our bug, user will get no useful information from the error and has no chances to help us or fix the error

resetius (changes requested):

ydb/core/kqp/executer_actor/kqp_executer_stats.cpp
```diff
@@ -64,26 +66,32 @@ void TStageExecutionStats::Resize(ui32 taskCount) {
...
+        target = source;
+    }
+}
+
 void TStageExecutionStats::UpdateAsyncStats(i32 index, TAsyncStats& aggrAsyncStats, const NYql::NDqProto::TDqAsyncBufferStats& asyncStats) {
-    aggrAsyncStats.Bytes[index] = asyncStats.GetBytes();
-    aggrAsyncStats.Rows[index] = asyncStats.GetRows();
-    aggrAsyncStats.Chunks[index] = asyncStats.GetChunks();
-    aggrAsyncStats.Splits[index] = asyncStats.GetSplits();
+    SetNonZero(aggrAsyncStats.Bytes[index], asyncStats.GetBytes());
```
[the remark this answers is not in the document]
Hor911: That's OK since:

1. I excplicitly call resize() for every std::vector() and zero is correct default
2. The main problem occurs when 0 overrides non-zero value. It should never occurs since all metrics have to grow monotonically. But is metric is missed for some reason, it is arrived as zero in protobuf abd we have to ignore it

resetius (approved):

ulya-sidorina (approved):

ydb/core/fq/libs/compute/ydb/status_tracker_actor.cpp
dorooleg: 1. How can user understand that the in progress statistics doesn't work?
2. How can we follow that we have a problem in this place? Do we need any sensors here or not?

PR #642, opened 2023-12-21
ActorSystem ticket lock use std atomic

kitaisreal:
ActorSystem ticket lock use std atomic

### DIFF
--- a/ydb/library/actors/util/ticket_lock.h
+++ b/ydb/library/actors/util/ticket_lock.h
@@ -1,12 +1,12 @@
 #pragma once
 
-#include "intrinsics.h"
+#include <util/system/spinlock.h>
 #include <util/system/guard.h>
 #include <util/system/yassert.h>
 
 class TTicketLock : TNonCopyable {
-    ui32 TicketIn;
-    ui32 TicketOut;
+    std::atomic<ui32> TicketIn;
+    std::atomic<ui32> TicketOut;
 
 public:
     TTicketLock()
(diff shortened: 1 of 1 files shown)


ydb/library/actors/util/ticket_lock.h
```diff
@@ -16,33 +16,34 @@ class TTicketLock : TNonCopyable {
...
-        AtomicUi32Increment(&TicketOut);
+        TicketOut.fetch_add(1, std::memory_order_release);
     }
 
     ui32 Acquire() noexcept {
         ui32 revolves = 0;
-        const ui32 ticket = AtomicUi32Increment(&TicketIn) - 1;
-        while (ticket != AtomicLoad(&TicketOut)) {
-            Y_DEBUG_ABORT_UNLESS(ticket >= AtomicLoad(&TicketOut));
+        const ui32 ticket = TicketIn.fetch_add(1, std::memory_order_seq_cst);
```
kruall, line 24: Why is std::memory_order_seq_cst used here instead of std::memory_order_acq_rel?

kitaisreal: Need to change to `std::memory_order_acq_rel`.

snaury: It can actually be relaxed, because `TicketIn` is used purely as a unique order counter. Since `Release()` uses release on `TicketOut` only acquire on `TicketOut` can synchronize with `Release()`.

PR #650, opened 2023-12-21
ActorSystem shared data ref count memory order fix

kitaisreal:
ActorSystem shared data ref count memory order fix


kruall (approved):

kruall:
I noticed that the test ydb/library/actors/util/ut/TRopeSharedDataNativeBackend.Insert failed. It is related to SharedData. Does it not fail locally?

PR #694, opened 2023-12-25
Support IF NOT EXISTS/IF EXISTS for tables, external tables and external data sources

UgnineSirdis:


ydb/library/yql/sql/v1/format/sql_format.cpp
```diff
@@ -398,7 +398,7 @@ friend struct TStaticData;
         } else {
             VisitAllFields(descr, msg);
         }
-
```
nepal, line 401: Need sql_format_ut test

nepal (approved):
sql part LGTM

PR #714, opened 2023-12-25
Keep Issue fields after censorship

Hor911:


ydb/core/fq/libs/common/util.cpp
```diff
@@ -50,7 +50,8 @@ class TIssueDatabaseRemover {
 
     TIntrusivePtr<NYql::TIssue> Run(const NYql::TIssue& issue) {
         auto msg = RemoveDatabaseFromStr(issue.GetMessage(), DatabasePath);
-        auto newIssue = MakeIntrusive<NYql::TIssue>(msg);
+        auto newIssue = MakeIntrusive<NYql::TIssue>(issue.Position, issue.EndPosition, msg);
```
dorooleg, line 53: Could we add the test here. We can change this field at https://a.yandex-team.ru/arcadia/ydb/tests/tools/fq_runner/kikimr_runner.py?rev=13154375#L379 to BM_DROP_WITH_WARNING in any existing test and check that the we don't any have problems in this case

Hor911: Done

MetaGigachad (approved):

dorooleg (approved):

PR #766, opened 2023-12-28
Fix OLAP stats

k-morozov:


k-morozov:
Im going to add some utests (for many tables in table store).

k-morozov:
There is a problem related with a large count tables in tablesstore. Some tables doesn't init by stats.

ydb/core/protos/tx_datashard.proto
```diff
@@ -806,6 +806,8 @@ message TEvPeriodicTableStats {
     optional uint64 TableOwnerId = 12;
 
     optional bool IsDstSplit = 13;
+
+    repeated TEvPeriodicTableStats ColumnTables = 14;
```
nsofya, line 810: Давай без Column. Статистика же общего вида

k-morozov: ок

ivanmorozov333 (approved):

ijon:
I don't understand, why this pr, which touches core mechanics of stats collection in schemeshard for all table types and not only for olap tables: 1) named highly inaccurate "fix olap", and 2) had been merged without approve from me while I explicitly registered my interest in reviewing it.

PR #795, opened 2023-12-28
Enable SELECT from views

jepett0:
KIKIMR-19567

The approach is the following:
- save query text specified in CREATE VIEW statement in SchemeShard,
- whenever we encounter a Read! operation from a view, rewrite it on the provider side. Such rewriting can only happen on the Kikimr (a. k. a. YDB) side, because YQL itself knows nothing about the view: the query text is stored in YDB.
- repeat some of the steps of the YQL transformer pipeline, so the query stored in the view can be evaluated. As a prime example: we need to rewrite all the reads from other views that we might encounter inside the query stored in the original view.

Rewriting a node happens during parsing of the YQL statement. At this moment the statement is expressed as a graph of TExprNode nodes. It is a better version of an AST of the statement. It is actually a DAG: some nodes appear multiple times in it.

Rewriting a read from a view consists of these steps:
1. Load view metadata (it is disguised as a generic table metadata, but with a view specific field present).
2. Get the query text from the metadata.
3. Read! from a view is expressed as a callable TExprNode, which we need to rewrite to correspond to a TExprNode, which provides results from the subquery, stored in the view. The rewritten node will replace the Read! from the view node in the execution graph of a SELECT from a view statement. The first step of the rewriting is to build the view query graph.
4. Then we need to correctly express the execution order dependencies between the operations, happening inside the view graph, and the operations outside the view graph. This consists of two parts:

a) inject dependency of the first operation in the view graph on the last operation before the Read! from the view. In more technical terms: replace all the pure worlds (i. e. worlds without any dependencies) inside the view graph with the outer world of the Read! from the view node.

b) output the last operation, happening in the view graph, outside the graph to make the outside nodes, whose operations should happen after the read from the view, to be dependent on the last operation of the view. In other words: return the last operation in the view graph as a result of a Left! query from the rewritten Read! from a view node. (The Left! query is a way to get from a node its current state of the world. And a world node is way we express time dependency between operations in our backend functional language (s-expressions).)


ydb/core/kqp/provider/yql_kikimr_datasource.cpp
```diff
@@ -673,58 +674,73 @@ class TKikimrDataSource : public TDataProviderBase {
...
-                                        .Add(ctx.NewCallable(node->Pos(), "Void", {}))
-                                        .Add(BuildExternalTableSettings(node->Pos(), ctx, tableDesc.Metadata->Columns, source, tableDesc.Metadata->ExternalSource.TableContent))
 
-                                    .Build()
-                                    .Done().Ptr();
-            auto retChildren = node->ChildrenList();
-            retChildren[0] = newRead;
-            return ctx.ChangeChildren(*node, std::move(retChildren));
+                const auto& query = tableDesc.Metadata->ViewPersistedData.QueryText;
+                return RewriteReadFromView(node, ctx, query, cluster);
```
jepett0, line 742: Forgot to add a feature flag check. I suppose it should go somewhere here. Will add in the next revision of this review.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+TExprNode::TPtr RewriteReadFromView(
+    const TExprNode::TPtr& node,
+    TExprContext& ctx,
+    const TString& query,
+    const TString& cluster
+) {
+    const auto readNode = node->ChildPtr(0);
+    const auto worldBeforeThisRead = readNode->ChildPtr(0);
+
+    TExprNode::TPtr queryGraph = FindSavedQueryGraph(readNode);
```
jepett0, line 143: I save the query graph as the last child of the Read! from the view node to reuse it for both:
1) rewriting of the Left! callable from the Read! from the view node,
2) rewriting of the Right! callable from the Read! from the view node.

I need all the generated Read! (and other nodes) of the view graph to be the same ("same" = same object) for both Left! and Right! rewrites, because [CheckTx](https://github.com/ydb-platform/ydb/blob/c288b66eb979d88e47d5c141f1fe9f853336e08f/ydb/core/kqp/provider/yql_kikimr_opt_build.cpp#L658) expects all the Read! nodes to be seen on the path from the root node by the first child.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+            }
+            return true;
+        }
+    );
+
+    if (!lastReadInTopologicalOrder) {
+        return nullptr;
+    }
+
+    Y_ENSURE(CheckTopLevelness(*lastReadInTopologicalOrder, queryGraph),
```
jepett0, line 124: Not sure if this is a good idea. This message is for developers, not for the user. However, I would like to check this assumption in the code somehow.

jepett0: On the other hand, if I messed up in choosing the correct Read! node in this function, then the user will get an even more cryptic error message:

"Failed to execute callable with name: ResWrite!, you possibly used cross provider/cluster operations or pulled not materialized result in refselect mode"

This message appears whenever KQP failed to substitute some ResWrite! node to a TKiExecDataQuery! node, which usually happens when there are nodes with side-effects (like Read!) not seen on the path by the first child from the root of the query graph (see [CheckTx](https://github.com/ydb-platform/ydb/blob/c288b66eb979d88e47d5c141f1fe9f853336e08f/ydb/core/kqp/provider/yql_kikimr_opt_build.cpp#L658) function).

ydb/core/kqp/provider/yql_kikimr_datasource.cpp
jepett0: Added the feature flag check in the next revision

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+            }
+            return true;
+        }
+    );
+
+    if (!lastReadInTopologicalOrder) {
+        return nullptr;
+    }
+
+    YQL_ENSURE(CheckTopLevelness(*lastReadInTopologicalOrder, queryGraph),
```
jepett0, line 124: Not sure if this is a good idea. This message is for developers, not for the user. However, I would like to check this assumption in the code somehow.

jepett0: On the other hand, if I messed up in choosing the correct Read! node in this function, then the user will get an even more cryptic error message:

"Failed to execute callable with name: ResWrite!, you possibly used cross provider/cluster operations or pulled not materialized result in refselect mode"

This message appears whenever KQP failed to substitute some ResWrite! node to a TKiExecDataQuery! node, which usually happens when there are nodes with side-effects (like Read!) not seen on the path by the first child from the root of the query graph (see [CheckTx](https://github.com/ydb-platform/ydb/blob/c288b66eb979d88e47d5c141f1fe9f853336e08f/ydb/core/kqp/provider/yql_kikimr_opt_build.cpp#L658) function).

gridnevvvit:
briefly pr looks okay to me, btw I'll  let peers to take a look a on this, if there are no other comments I will approve request.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+    NSQLTranslation::TTranslationSettings settings;
+
+    settings.DefaultCluster = cluster;
+    settings.ClusterMapping[cluster] = TString(NYql::KikimrProviderName);
+    settings.Mode = NSQLTranslation::ESqlMode::LIMITED_VIEW;
+
+    return settings;
+}
+
+TExprNode::TPtr CompileQuery(
```
spuchin, line 27: Давай наверное назовем тогда CompileView, а то есть уже такой метод в kqp_host, будет путаница.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+    const TString& query,
+    TExprContext& ctx,
+    const TString& cluster
+) {
+    TAstParseResult queryAst;
+    queryAst = NSQLTranslation::SqlToYql(query, CreateViewTranslationSettings(cluster));
+
+    ctx.IssueManager.AddIssues(queryAst.Issues);
```
spuchin, line 35: Нужно явно проверить в тесте что такие issues (включая warning) из view долетят до пользователя.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+}
+
+TExprNode::TPtr RewriteReadFromView(
+    const TExprNode::TPtr& node,
+    TExprContext& ctx,
+    const TString& query,
+    const TString& cluster
+) {
+    const auto readNode = node->ChildPtr(0);
```
spuchin, line 140: What kind of node we expect here? Use typed nodes or at least add comment.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+    const auto worldBeforeThisRead = readNode->ChildPtr(0);
+
+    TExprNode::TPtr queryGraph = FindSavedQueryGraph(readNode);
+    if (!queryGraph) {
+        queryGraph = CompileQuery(query, ctx, cluster);
+        if (!queryGraph || ContainsNullNode(queryGraph)) {
```
spuchin, line 146: I don't really get why we need a check for "contains null node"? If query has errors, it shouldn't compile?

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+    AddChild(carrier, ctx.NewCallable(payload->Pos(), QueryGraphNodeSignature, {payload}));
+}
+
+void InsertExecutionOrderDependencies(const TExprNode::TPtr& queryGraph, const TExprNode::TPtr& worldBefore) {
+    VisitExpr(
+        queryGraph,
+        nullptr,
+        [&worldBefore](const TExprNode::TPtr& node) {
+            if (node->ChildrenSize() > 0 && node->Child(0)->IsWorld()) {
+                ChangeChild(node, 0, worldBefore);
```
spuchin, line 86: I'm not sure changing nodes inside Visit is a good idea. We have ctx.ReplaceNodes method for this.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+        queryGraph = CompileQuery(query, ctx, cluster);
+        if (!queryGraph || ContainsNullNode(queryGraph)) {
+            ctx.AddError(TIssue(readNode->Pos(ctx),
+                         "The query stored in the view contains errors and cannot be compiled."));
+            return nullptr;
+        }
+        YQL_CLOG(TRACE, ProviderKqp) << "Expression graph of the query stored in the view:\n"
+                                     << NCommon::ExprToPrettyString(ctx, *queryGraph);
+
+        InsertExecutionOrderDependencies(queryGraph, worldBeforeThisRead);
```
spuchin, line 154: This looks a bit like a hack. Parser should be able to return you all "root" world dependences, so you don't have to search for the in query graph.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+        SaveQueryGraph(readNode, ctx, queryGraph);
```
spuchin, line 155: I'm not quite sure why do we need to save view AST to Read? Why not just rewrite the query immediately?

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    if (node->IsCallable(RightName)) {
+        return queryGraph;
+    }
+
+    const auto topLevelRead = FindTopLevelRead(queryGraph);
```
spuchin, line 162: I think you should at least combine InsertExecutionOrderDependencies and FindTopLevelRead methods. They serve the same purpose: get world topology roots and leafs.

ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -0,0 +1,169 @@
...
+    if (!topLevelRead) {
```
spuchin, line 163: How is that possible?

spuchin (changes requested):

ydb/core/kqp/gateway/kqp_metadata_loader.cpp
```diff
@@ -278,6 +278,26 @@ TTableMetadataResult GetExternalDataSourceMetadataResult(const NSchemeCache::TSc
     return result;
 }
 
+TTableMetadataResult GetViewMetadataResult(const NSchemeCache::TSchemeCacheNavigate::TEntry& schemeEntry,
```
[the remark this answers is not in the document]
jepett0: Could you, please, elaborate what changes you would like to see here? I don't see what's wrong with the current indentation. I align the parameters so each line has one parameter and they all begin at the same column

ydb/core/kqp/provider/yql_kikimr_datasource.cpp
```diff
@@ -673,58 +674,79 @@ class TKikimrDataSource : public TDataProviderBase {
...
-        }
-
-        if (key.GetKeyType() == TKikimrKey::Type::Table && tableDesc.Metadata->Kind == EKikimrTableKind::External  && tableDesc.Metadata->ExternalSource.SourceType == ESourceType::ExternalTable) {
-            const auto& source = ExternalSourceFactory->GetOrCreate(tableDesc.Metadata->ExternalSource.Type);
-            ctx.Step.Repeat(TExprStep::DiscoveryIO)
+        const TString cluster = source.Cluster().StringValue();
+        const TString tablePath = key.GetTablePath();
+        auto& tableDesc = SessionCtx->Tables().GetTable(cluster, tablePath);
+        if (key.GetKeyType() == TKikimrKey::Type::Table) {
+            if (tableDesc.Metadata->Kind == EKikimrTableKind::External) {
```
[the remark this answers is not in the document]
jepett0: I don't think it will improve readability here. I'm only interested in 2 out of 6 cases of this enum. I don't think switch will be more concise.

ydb/core/kqp/provider/yql_kikimr_datasource.cpp
```diff
@@ -673,58 +674,79 @@ class TKikimrDataSource : public TDataProviderBase {
...
+                }
+            } else if (tableDesc.Metadata->Kind == EKikimrTableKind::View) {
+                if (!SessionCtx->Config().FeatureFlags.GetEnableViews()) {
+                    ctx.AddError(TIssue(node->Pos(ctx),
+                                        "Views are disabled. Please contact your system administrator to enable the feature"));
+                    return nullptr;
+                }
+
+                ctx.Step
+                    .Repeat(TExprStep::ExprEval)
```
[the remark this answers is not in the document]
jepett0: I took this line from [yt implementation of views](https://github.com/ydb-platform/ydb/blob/5d59f2cda66c7fe1cc7b09717261af49b87da604/ydb/library/yql/providers/yt/provider/yql_yt_datasource.cpp#L693C49-L693C49).

Do you think expression evaluation inside views should be forbidden? I'm not sure, what it is exactly, but in my opinion the example from the [docs](https://yql.yandex-team.ru/docs/yt/builtins/basic#evaluate_expr_atom):
```sql
$now = CurrentUtcDate();
SELECT EvaluateExpr(
    DateTime::MakeDate(DateTime::StartOfWeek($now)
    )
);
```
looks fine and it is reasonable to allow it inside views.

ydb/core/kqp/provider/yql_kikimr_gateway.h
```diff
@@ -389,6 +390,10 @@ enum EMetaSerializationType : ui64 {
     Json = 2
 };
 
+struct TViewPersistedData {
+    TString QueryText;
```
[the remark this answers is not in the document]
jepett0: It is SQL. Ideally we would like to save the original text that the user specified at the moment of the CREATE VIEW call. However, I didn't find a way to save the original text (with the original formatting). I save [the query text that is recovered](https://github.com/ydb-platform/ydb/blob/5d59f2cda66c7fe1cc7b09717261af49b87da604/ydb/library/yql/sql/v1/sql_translation.cpp#L4425C52-L4425C52) from the protobuf of the select_stmt message.

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: Буду делать юнит тест

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: It is actually pretty easy to have a query stored in a view that does not contain any reads. For example,
```sql
SELECT 1
```

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: I explained the motivation in [this comment](https://github.com/ydb-platform/ydb/pull/795#discussion_r1438257676). The idea is that in most cases we use nodes that express Read!s from views in two ways: 
1) call Left! on the read node,
2) call Right! on that same read node.

I noticed that in order for all the nodes with side-effects (like Read! from tables and such) in the rewritten query graph of the select from the view statement to be linearly stacked on top of each other, we need to ensure that the query graph of the view query is exactly the same in both rewrites: a) of the Left! call and b) of the Right! call to the node reading from the view. Otherwise, the rewritten query will fail on CheckTx function.

I didn't find a better way to ensure that we call CompileExpr on the view query only once.

ydb/core/kqp/opt/logical/kqp_opt_log_ranges_predext.cpp
```diff
@@ -55,6 +55,7 @@ TMaybeNode<TExprBase> TryBuildTrivialReadTable(TCoFlatMap& flatmap, TKqlReadTabl
             break;
         case EKikimrTableKind::Olap:
         case EKikimrTableKind::External:
+        case EKikimrTableKind::View:
```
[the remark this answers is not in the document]
jepett0: Added [YQL_ENSURE(false)](https://github.com/ydb-platform/ydb/blob/586058fd4169e192e0902fd05db5e933c0a21c69/ydb/core/kqp/opt/logical/kqp_opt_log_ranges_predext.cpp#L61) in newer revision for the view case

ydb/core/kqp/gateway/kqp_metadata_loader.cpp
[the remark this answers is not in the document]
jepett0: [Changed](https://github.com/jepett0/ydb/blob/586058fd4169e192e0902fd05db5e933c0a21c69/ydb/core/kqp/gateway/kqp_metadata_loader.cpp#L281) indentation for the closing parenthesis to match the beginning of the function declaration

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: [Переименовал](https://github.com/jepett0/ydb/blob/586058fd4169e192e0902fd05db5e933c0a21c69/ydb/core/kqp/provider/rewrite_io_utils.cpp#L28) в последней ревизии кода

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: [Used](https://github.com/jepett0/ydb/blob/586058fd4169e192e0902fd05db5e933c0a21c69/ydb/core/kqp/provider/rewrite_io_utils.cpp#L135) TCoRead to express my intention better

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: [Removed](https://github.com/ydb-platform/ydb/pull/795/commits/586058fd4169e192e0902fd05db5e933c0a21c69#diff-bbf19f6d61e0ad71a303c0f72ccc83b3c5854d22463ae360f52f5e1dd9d6f085) the check

ydb/core/kqp/provider/yql_kikimr_gateway.h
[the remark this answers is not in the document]
nepal: You should save original query text, since you need to preserve user comments - we support optimiser hints via comments.

You can save original user query as following
1) remember position of first and last SELECT token
2) There is class TTextWalker somewhere in parser code that can help to translate position offsets to byte offset in original query

Also you need to save some parsers settings together with query text. At least ansi_lexer flag (otherwise you can not recompile query correctly)

nepal (changes requested):

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: I don't know of a way to do it. I will ask Vitaliy Stoyan to review this change too

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: One of the them (InsertExecutionOrderDependencies) is needed for both rewrites (rewrite of the Left! call to the read node and rewrite of the Right! call to the read node), while the other one (FindTopLevelRead) is needed only for the rewrite of the Left! call the read node. In addition, one of them does change the query graph, while the other is only searching for a node in it. I would like to keep them separated.

ydb/core/kqp/provider/yql_kikimr_gateway.h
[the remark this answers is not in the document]
jepett0: This will be done in a [separate ticket](https://st.yandex-team.ru/KIKIMR-20694) to speed up the development and split the big task into smaller ones for better workflow

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: I tried using both RemapExpr and OptimizeExpr and they don't seem to work for my case. I suspect the following [line](https://github.com/ydb-platform/ydb/blob/0a965144219a01dadef957ee280b78147d0667f7/ydb/library/yql/core/yql_expr_optimize.cpp#L94C1-L98C1):
```cpp
if (... || node->Type() == TExprNode::World) {
    // don't optimise node
}
```

So, I [ended up](https://github.com/ydb-platform/ydb/blob/504dad86b6f2606644cac43538b293c19ad15902/ydb/core/kqp/provider/rewrite_io_utils.cpp#L78) using ctx.ReplaceNode.

nepal (approved):

ydb/core/kqp/provider/rewrite_io_utils.cpp
nepal: ок

ydb/core/kqp/provider/yql_kikimr_datasource.cpp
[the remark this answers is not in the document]
jepett0: There is a bug in the way VIEWs evaluate expressions, see [this ticket](https://st.yandex-team.ru/KIKIMR-20819). In particular, SELECT EvaluateExpr(CurrentUtcTimestamp()) will be evaluated only once, if saved inside a VIEW.

spuchin (approved):

PR #802, opened 2023-12-29
Delete triggered simbols.

Pseudolukian:


blinkov:
@Pseudolukian is this still relevant?

PR #838, opened 2024-01-03
Support DDL operations (including IF EXISTS/IF NOT EXISTS) for objects with type secret

UgnineSirdis:


ydb/services/metadata/manager/generic_manager.h
```diff
@@ -42,35 +44,123 @@ class TGenericOperationsManager: public IObjectOperationsManager<T> {
...
+                operationProto = schemeOperation.MutableAlterObject();
+                break;
+            case EActivityType::Drop:
+                operationProto = schemeOperation.MutableDropObject();
+                operationProto->SetSuccessOnNotExist(settings.GetMissingOk());
+                break;
+            case EActivityType::Undefined:
+                return TYqlConclusionStatus::Fail("undefined action type");
+        }
+        Y_DEBUG_ABORT_UNLESS(operationProto);
```
qrort, line 108: We return Success in release mode, is that what we want? Maybe we should use Y_ENSURE here and abort the request?

ydb/services/metadata/manager/modification.h
```diff
@@ -152,15 +152,31 @@ template <class TObject>
...
     }
     virtual TString GetModifyType() const override {
         return "insert";
     }
+
+    void Handle(NRequest::TEvRequestFailed::TPtr& ev) override {
+        if (ev->Get()->GetStatus() == Ydb::StatusIds::PRECONDITION_FAILED && ExistingOk) {
+            NRequest::TDialogYQLRequest::TResponse resp;
+            this->Send(this->SelfId(), new NRequest::TEvRequestResult<NRequest::TDialogYQLRequest>(std::move(resp)));
+            this->Requests.clear(); // Remove history request
```
qrort, line 168: Why do we remove all requests, if one fails?

ydb/services/metadata/manager/generic_manager.h
```diff
@@ -42,35 +44,123 @@ class TGenericOperationsManager: public IObjectOperationsManager<T> {
...
+        }
+
+        auto promise = NThreading::NewPromise<TYqlConclusionStatus>();
+        auto controller = std::make_shared<TOperationsController>(promise);
+        IObjectModificationCommand::TPtr modifyObjectCommand;
+        TInternalModificationContext internalContext(context);
+        switch (schemeOperation.GetOperationCase()) {
+            case NKqpProto::TKqpSchemeOperation::kUpsertObject:
+                internalContext.SetActivityType(EActivityType::Upsert);
+                modifyObjectCommand = std::make_shared<TUpsertObjectCommand<T>>(BuildPatchFromProto(schemeOperation.GetUpsertObject()), manager, controller, internalContext);
```
qrort, line 135: We make a fresh TOperationsController pointer for each ObjectCommand, and we copy it inside. Also we copy it in TInsertObjectsActor constructor. Do we really need those copies, or maybe we could use unique pointer? Can ObjectCommand remain, if corresponding TOperationsController is destroyed?

ydb/services/metadata/manager/generic_manager.h
UgnineSirdis: We don't return Success here: we will simply segfault because we use operationProto :)
I think this is an internal logical error, because all options for ActivityType are listed here. Futher more, the compiler would tell us if additional option exists.

ydb/services/metadata/manager/generic_manager.h
UgnineSirdis: I only refactored this code and I don't know why it is shared. But as I see, it can be simplified as you suggested. Thank you, I will try :)

ydb/services/metadata/manager/modification.h
UgnineSirdis: It is rather complicated code :) This second request is here: https://github.com/ydb-platform/ydb/blob/main/ydb/services/metadata/manager/modification.h#L93
It is history insertion and we don't need it in case we haven't inserted main thing.

ydb/services/metadata/manager/generic_manager.h
UgnineSirdis: I tried to use unique_ptr and noticed it: https://github.com/ydb-platform/ydb/blob/main/ydb/services/metadata/initializer/accessor_init.h#L27
I am afraid that we need shared_ptr :(

UgnineSirdis: I avoided extra copying here using std::move - it is all that I can do :)

ydb/services/metadata/manager/generic_manager.h
qrort: I would prefer Y_ENSURE to get an exception in release instead of a segfault nonetheless

qrort (dismissed):

qrort (approved):

PR #861, opened 2024-01-06
Limit v2 compute load by CPU

Hor911:


ydb/core/fq/libs/compute/ydb/control_plane/compute_database_control_plane_service.cpp
```diff
@@ -38,6 +38,8 @@ struct TDatabaseClients {
         TActorId ActorId;
         NConfig::TComputeDatabaseConfig Config;
         TActorId DatabasesCacheActorId;
+        TActorId MonitoringClientActorId;
```
dorooleg, line 41: We don't use this variable. Can we get rid of it?

ydb/core/fq/libs/compute/ydb/control_plane/compute_database_control_plane_service.cpp
```diff
@@ -374,7 +416,38 @@ class TComputeDatabaseControlPlaneServiceActor : public NActors::TActorBootstrap
         Register(new TCreateDatabaseRequestActor(Clients, SynchronizationServiceActorId, Config, ev));
     }
 
+    void Handle(TEvYdbCompute::TEvCpuLoadRequest::TPtr& ev) {
+        auto actorId = GetMonitoringActorIdByScope(ev.Get()->Get()->Scope);
+        if (actorId != TActorId{}) {
```
dorooleg, line 421:
```
        if (actorId != TActorId{}) {
            Send(ev->Forward(actorId));
            return;
        }
        Send(ev->Sender, new TEvYdbCompute::TEvCpuLoadResponse(NYql::TIssues{NYql::TIssue{TStringBuilder{} << "Cluster load monitoring disabled"}}), 0, ev->Cookie);
```

to get rid of `else`

ydb/core/fq/libs/config/protos/compute.proto
```diff
@@ -17,10 +17,22 @@ message TSynchronizationService {
...
 
+message TLoadControlConfig {
+    bool Enable = 1;
+    string MonitoringRequestDelay = 2;     // default 0 == non-stop polling
+    string AverageLoadInterval = 3;        // default "10s", allowed min "1s"
+    uint32 MaxClusterLoadPercentage = 4;   // default 0 == no load control
+    uint32 DefaultQueryLoadPercentage = 5; // default 10
+    uint32 PendingQueueSize = 6;           // default 0 == instant decline if overloaded
+    bool Inherit = 7;                      // whether to apply global config to databases (override, NOT merge)
+    uint32 CpuNumber = 8;
```
dorooleg, line 28: Do we need this option? What does this option mean (per house, per cluster)?

ydb/core/fq/libs/config/protos/compute.proto
```diff
@@ -17,10 +17,22 @@ message TSynchronizationService {
 message TInPlaceCompute {
 }
 
+message TLoadControlConfig {
+    bool Enable = 1;
+    string MonitoringRequestDelay = 2;     // default 0 == non-stop polling
```
dorooleg, line 22: As i understand this call is heavy and we need to call it as little as possible. Is it possible to change the default option to several seconds so that there is no mistake during the configuration

dorooleg (changes requested):

ydb/core/fq/libs/config/protos/compute.proto
```diff
@@ -17,10 +17,22 @@ message TSynchronizationService {
...
+    string AverageLoadInterval = 3;        // default "10s", allowed min "1s"
+    uint32 MaxClusterLoadPercentage = 4;   // default 0 == no load control
+    uint32 DefaultQueryLoadPercentage = 5; // default 10
+    uint32 PendingQueueSize = 6;           // default 0 == instant decline if overloaded
+    bool Inherit = 7;                      // whether to apply global config to databases (override, NOT merge)
```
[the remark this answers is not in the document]
Hor911: It is not completely clean, how to merge global and local configs if both are present and partially filled. I think that current behaviour is more obvious and less misleading than any other

ydb/core/fq/libs/compute/ydb/status_tracker_actor.cpp
```diff
@@ -249,7 +255,12 @@ class TStatusTrackerActor : public TBaseComputeActor<TStatusTrackerActor> {
         pingTaskRequest.set_status(ComputeStatus);
         PrepareAstAndPlan(pingTaskRequest, QueryStats.query_plan(), QueryStats.query_ast());
         try {
-            pingTaskRequest.set_statistics(GetV1StatFromV2Plan(QueryStats.query_plan()));
+            TDuration duration = TDuration::MicroSeconds(QueryStats.total_duration_us());
```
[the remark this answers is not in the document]
Hor911: I've made this refactoring already with next "--stat_full" feature, will merge in next PR

ydb/core/fq/libs/compute/ydb/events/events.h
```diff
@@ -438,6 +444,58 @@ struct TEvYdbCompute {
...
+
+    struct TEvCpuQuotaAdjust : public NActors::TEventLocal<TEvCpuQuotaAdjust, EvCpuQuotaAdjust> {
+        TEvCpuQuotaAdjust(const TString& scope, TDuration duration, double cpuSecondsConsumed, double quota = 0.0)
+            : Scope(scope), Duration(duration), CpuSecondsConsumed(cpuSecondsConsumed), Quota(quota)
+        {}
+
+        TString Scope;
+        TDuration Duration;
+        double CpuSecondsConsumed;
+        double Quota; // if zero, default quota is used
```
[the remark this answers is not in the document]
Hor911: It was idea that client my override default expected load in relation to compiled query size (task count). Currently is not used

ydb/core/fq/libs/compute/ydb/control_plane/compute_database_control_plane_service.cpp
Hor911: It was with graceful shutdown/cleanup in mind.

ydb/core/fq/libs/compute/ydb/control_plane/compute_database_control_plane_service.cpp
Hor911: Why it is better?

ydb/core/fq/libs/config/protos/compute.proto
Hor911: It is number if CPU per cluster. It is needed to align two different "systems". Why I ask for cluster load with Monitoring Service, I've got ratio between 0.0 and 1.0 for complete cluster. Query reports its usage in number of CPUs. To convert one to another we need to know number of CPU per cluster.

ydb/core/fq/libs/config/protos/compute.proto
Hor911: Changed to "1s"

ydb/core/fq/libs/compute/ydb/control_plane/compute_database_control_plane_service.cpp
uzhastik: does not matter in this case imho
but yes, no "else" - simplier code

ydb/core/fq/libs/config/protos/compute.proto
dorooleg: Do we have another way to get this information? It's difficult to keep in consistent state two configs (ydb sls and fq proxy config). If we don't have another way than this solution is ok for me :)

ydb/core/fq/libs/config/protos/compute.proto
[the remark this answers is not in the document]
dorooleg: It's difficult to keep track of configuration changes (we need to check this `Inherit` flag). 
I'd like to expect the following behavior: if we don't have `Inherit` flag then:
1. If this section was specified for database then we use this config else global config.
2. If we want to disable this logic then we can add section with `Enable=false`. 

It's possible that i missed something. Could you give me an example of when you would use this flag?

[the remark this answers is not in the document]
Hor911: Enable=false is default value in proto 3 and is indistinguishable from missed Enable (empty section) unfortunately

ydb/core/fq/libs/config/protos/compute.proto
Hor911: No other sources is available. This param is not mandatory, it is required for "cashback" feature - decrease CPU usage estimation after query completion. Allows better CPU utilization for short light queries

ydb/core/fq/libs/config/protos/compute.proto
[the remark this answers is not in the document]
Hor911: Changed as agreed (Inherit has been removed, any level Enable activates load control)

dorooleg (approved):

PR #885, opened 2024-01-09
Write flat index last key setting

kungasc:
This new `WriteLastFlatIndexKey` setting is needed for comparison of Precharge logic: B-tree index doesn't have the last key, so it can't be used when we precharge some range [k1, k2] where k1 > the last key


snaury:
For the record: btree precharge may use rightmost slice's right boundary as the last key.

PR #894, opened 2024-01-09
Allow UPDATE ON, UPSERT with partial set of input...

dcherednik:
columns in case of unique index.

Example:
  table: pk, fk1, fk2. Uniq index: fk1, fk2.

To perform UPSERT INTO table (pk, fk1) we
need to read missed value fk2 from main table to
create lookup uniq index key.

Right now we have two lookups in to main table
- first one has been described above
- second we need to make to remove old value from index. It will be fixed in a next step.

### DIFF
--- a/ydb/core/kqp/opt/physical/effects/kqp_opt_phy_effects.cpp
+++ b/ydb/core/kqp/opt/physical/effects/kqp_opt_phy_effects.cpp
@@ -77,7 +77,6 @@ TMaybe<TCondenseInputResult> CondenseInput(const TExprBase& input, TExprContext&
     TVector<TCoArgument> stageArguments;
 
     if (IsDqPureExpr(input)) {
-        YQL_ENSURE(input.Ref().GetTypeAnn()->GetKind() == ETypeAnnotationKind::List, "" << input.Ref().Dump());
         auto stream = Build<TCoToStream>(ctx, input.Pos())
             .Input<TCoJust>()
                 .Input(input)
--- a/ydb/core/kqp/opt/physical/effects/kqp_opt_phy_indexes.cpp
+++ b/ydb/core/kqp/opt/physical/effects/kqp_opt_phy_indexes.cpp
@@ -48,7 +48,11 @@ TVector<TExprBase> CreateColumnsToSelectToUpdateIndex(
     return columnsToSelect;
 }
 
-TDqPhyPrecompute PrecomputeDict(const TCondenseInputResult& condenseResult, TPositionHandle pos, TExprContext& ctx) {
+} // namespace
+
+TDqPhyPrecompute PrecomputeCondenseInputResult(const TCondenseInputResult& condenseResult,
+    TPositionHandle pos, TExprContext& ctx)
+{
     auto computeDictStage = Build<TDqStage>(ctx, pos)
         .Inputs()
             .Add(condenseResult.StageInputs)
@@ -69,8 +73,6 @@ TDqPhyPrecompute PrecomputeDict(const TCondenseInputResult& condenseResult, TPos
             .Build()
         .Done();
 }
-
-} // namespace
 
 TVector<std::pair<TExprNode::TPtr, const TIndexDescription*>> BuildSecondaryIndexVector(
     const TKikimrTableDescription& table,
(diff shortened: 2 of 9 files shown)


ydb/core/kqp/opt/physical/effects/kqp_opt_phy_indexes.cpp
```diff
@@ -127,26 +129,79 @@ TSecondaryIndexes BuildSecondaryIndexVector(const TKikimrTableDescription& table
...
-        keyColumns, pos, ctx);
-
     auto lookupColumnsList = Build<TCoAtomList>(ctx, pos)
-        .Add(lookupColumns)
+        .Add(columnsList)
         .Done();
 
+    TExprNode::TPtr keys;
+
+    if (fixLookupKeys) {
```
spuchin, line 141: Добавь плз комментарий, что значит fixLookupKeys, но очевидно.

ydb/core/kqp/opt/physical/effects/kqp_opt_phy_upsert_index.cpp
```diff
@@ -328,30 +330,167 @@ TExprBase MakeUpsertIndexRows(TKqpPhyUpsertIndexMode mode, const TDqPhyPrecomput
...
     for (const auto& [_, indexDesc] : indexes) {
+        hasUniqIndex |= (indexDesc->Type == TIndexDescription::EType::GlobalSyncUnique);
         for (const auto& indexKeyCol : indexDesc->KeyColumns) {
-            if (inputColumns.contains(indexKeyCol) &&
-                (std::find(mainPk.begin(), mainPk.end(), indexKeyCol) == mainPk.end()))
-            {
-                usedIndexes.insert(indexDesc->Name);
-                break;
+            if (inputColumns.contains(indexKeyCol)) {
+                if (!usedIndexes.contains(indexDesc->Name) &&
```
qrort, line 355: we can omit `!usedIndexes.contains(indexDesc->Name)` check, the code will work the same

ydb/core/kqp/opt/physical/effects/kqp_opt_phy_upsert_index.cpp
```diff
@@ -328,30 +330,167 @@ TExprBase MakeUpsertIndexRows(TKqpPhyUpsertIndexMode mode, const TDqPhyPrecomput
...
+                            .Struct(missedFromMain)
+                            .Name().Build(x)
+                            .Build()
+                        .Done());
+
+            resNullCol.emplace_back(
+                Build<TCoNameValueTuple>(ctx, pos)
+                    .Name().Build(x)
+                    .Value<TCoNothing>()
+                        .OptionalType(NCommon::BuildTypeExpr(pos, *columnType, ctx))
```
qrort, line 412: I guess this does not work for Pg types, we'll need to somehow construct a pg null here (it corresponds to an empty `TUnboxedValuePod` with correct type)

ydb/core/kqp/opt/physical/effects/kqp_opt_phy_effects.cpp
```diff
@@ -77,7 +77,6 @@ TMaybe<TCondenseInputResult> CondenseInput(const TExprBase& input, TExprContext&
     TVector<TCoArgument> stageArguments;
 
     if (IsDqPureExpr(input)) {
-        YQL_ENSURE(input.Ref().GetTypeAnn()->GetKind() == ETypeAnnotationKind::List, "" << input.Ref().Dump());
```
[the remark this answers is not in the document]
dcherednik: Там про использовании CondenseInput из сложных оптимайзеров, еще type анотация не прошла. Оно просто тут сегфолтится

ydb/core/kqp/opt/physical/effects/kqp_opt_phy_indexes.cpp
dcherednik: ага. Продублирую тут
  // we need to left only table key columns to perform lookup
  // unfortunately we can't do it inside lookup stage

ydb/core/kqp/opt/physical/effects/kqp_opt_phy_upsert_index.cpp
dcherednik: Oh... And make sure we just return correct error in case of not null... Next pr?

qrort (approved):

PR #907, opened 2024-01-10
Fixed a problem with cardinality estimation for PK joins

pavelvelikhov:
- Fixed a problem with cardinality estimation for PK joins


resetius (approved):
OK

PR #925, opened 2024-01-10
YQ-2634: S3 runtime file listing

MetaGigachad:


ydb/library/yql/providers/s3/provider/yql_s3_provider.h
```diff
@@ -27,6 +37,8 @@ struct TS3State : public TThrRefBase
     TS3Configuration::TPtr Configuration = MakeIntrusive<TS3Configuration>();
     const NKikimr::NMiniKQL::IFunctionRegistry* FunctionRegistry = nullptr;
     ISecuredServiceAccountCredentialsFactory::TPtr CredentialsFactory;
+
+    TS3FileQueueParams FileQueueParams;
```
Hor911, line 41: Does it work with multiple S3 Sources per query?

Hor911 (changes requested):

ydb/library/yql/providers/s3/provider/yql_s3_provider.h
MetaGigachad: It does. Tested on query with 2 data sources from different buckets.

ydb/library/yql/providers/s3/provider/yql_s3_dq_integration.cpp
```diff
@@ -97,43 +107,74 @@ class TS3DqIntegration: public TDqIntegrationBase {
...
-                    it = parts.erase(to);
+        NActors::TActorId fileQueueActor;
+        if (useRuntimeListing) {
+            fileQueueActor = NActors::TActivationContext::ActorSystem()->Register(NDq::CreateS3FileQueueActor(
+                0ul,
+                parts.front(),
+                1000 * maxPartitions,
+                State_->Configuration->FileSizeLimit,
+                useRuntimeListing,
+                maxPartitions,
```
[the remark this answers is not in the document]
MetaGigachad: True, but it is the best assumption in this scope. In case there are less consumers FileQueue will die on timeout. Solution could be improved by getting the actual amount of tasks from executor, but I would suggest doing it in a separate PR, probably as a part of moving FileQueue creation to runtime.

CyberROFL:
dq tests broken by #1503

ydb/library/yql/providers/s3/provider/yql_s3_provider.h
```diff
@@ -27,6 +29,8 @@ struct TS3State : public TThrRefBase
     TS3Configuration::TPtr Configuration = MakeIntrusive<TS3Configuration>();
     const NKikimr::NMiniKQL::IFunctionRegistry* FunctionRegistry = nullptr;
     ISecuredServiceAccountCredentialsFactory::TPtr CredentialsFactory;
+    IHTTPGateway::TPtr Gateway;
+    TKikimrConfiguration::TPtr KikimrConfig;
```
dorooleg, line 33: Why do we need the kikimr configuration here? We usually have another deps from kqp to library/yql. Doest it correct here?

ydb/library/yql/providers/s3/provider/yql_s3_dq_integration.cpp
```diff
@@ -383,6 +405,107 @@ class TS3DqIntegration: public TDqIntegrationBase {
                 srcDesc.MutableSettings()->insert({"addPathIndex", "true"});
             }
 
+            auto useRuntimeListing = State_->Configuration->UseRuntimeListing.Get().GetOrElse(false);
+            srcDesc.SetUseRuntimeListing(useRuntimeListing);
+
+            auto fileQueueBatchSizeLimit = State_->Configuration->FileQueueBatchSizeLimit.Get().GetOrElse(1'000'000);
```
dorooleg, line 411: Could we use 1000000 instead of 1'000'000? it's difficult to grep this kind of consts

ydb/library/yql/providers/s3/proto/file_queue.proto
```diff
@@ -0,0 +1,30 @@
...
+}
+
+message TEvObjectPathBatch {
+    bool NoMoreFiles = 1;
+    repeated TObjectPath ObjectPaths = 2;
+
+    optional NYql.NDqProto.TMessageTransportMeta TransportMeta = 100;
+}
+
+message TEvObjectPathReadError {
```
dorooleg, line 20: Do we need a separate event for it? Here we have the common event: https://a.yandex-team.ru/arcadia/contrib/ydb/library/yql/dq/actors/protos/dq_events.proto?rev=r13368411#L75

dorooleg (changes requested):

ydb/library/yql/providers/s3/provider/yql_s3_provider.h
MetaGigachad: It is used to get MaxTasksPerStage, I can replace this with passing just this option

ydb/tests/fq/s3/test_yq_v2.py
```diff
@@ -56,7 +57,9 @@ def test_yqv2_enabled(self, kikimr, s3, client):
                                                  "csv_delimiter": ";"
                                              })
 
-        sql = R'''
+        sql = f'''
+            pragma s3.UseRuntimeListing="{str(runtime_listing).lower()}";
+            ''' + R'''
```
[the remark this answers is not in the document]
MetaGigachad: Made this just to be sure I haven't missed any unescaped characters in sql. I will remove it where possible

ydb/library/yql/providers/s3/proto/file_queue.proto
MetaGigachad: I would prefer it to be separate because it was separate before and has special logic around it already

dorooleg: Ok, let's leave it as it is

Alnen (approved):
LGTM

ydb/library/yql/providers/s3/provider/yql_s3_io_discovery.cpp
```diff
@@ -92,6 +92,7 @@ class TS3IODiscoveryTransformer : public TGraphTransformerBase {
               State_->Configuration->MinDesiredDirectoriesOfFilesPerQuery,
               State_->Configuration->MaxInflightListsPerQuery,
               State_->Configuration->AllowLocalFiles)) {
+        State_->Gateway = gateway;
```
Hor911, line 95: If we need Gateway in the state, it is better to assign it in InitS3Provider
https://github.com/ydb-platform/ydb/blob/main/ydb/core/kqp/host/kqp_host.cpp#L1470
and do not pass as param over dozen constructors.

Hor911 (approved):

dorooleg (approved):

ydb/library/yql/dq/actors/compute/retry_queue.h
```diff
@@ -120,6 +120,11 @@ class TRetryEventsQueue {
         }
         return false;
     }
+    
+    bool HasPendingEvents() {
+        RemoveConfirmedEvents(MyConfirmedSeqNo);
```
resetius, line 125: bad naming maybe
Why do you modify smth in method called Has... ?

ydb/library/yql/providers/s3/actors/yql_s3_read_actor.cpp
```diff
@@ -138,19 +142,82 @@ struct TS3ReadError : public yexception {
...
+        EvGetNextBatch,
+        EvObjectPathBatch,
+        EvObjectPathReadError,
+
+        EvEnd
+    };
+    static_assert(EvEnd < EventSpaceEnd(NKikimr::TKikimrEvents::ES_S3_FILE_QUEUE), 
+                  "expect EvEnd < EventSpaceEnd(TEvents::ES_S3_FILE_QUEUE)");
+    
+    struct TEvUpdateConsumersCount :
```
resetius, line 163: Why did you replace local events with non-local ones here?

resetius (approved):
Pls upload fixes in next PR

PR #931, opened 2024-01-11
Try to remove save/load() with traverse

kardymonds:


kardymonds:
Check in arc / used in https://a.yandex-team.ru/arcadia/yql/providers/rtmr/operations/lib/operations.cpp?rev=r12843017#L674
cannot be deleted

PR #1016, opened 2024-01-15
YQ-2630: fix and support detalization for v1

EgorkaZ:


ydb/core/fq/libs/control_plane_storage/internal/utils.cpp
```diff
@@ -304,10 +304,10 @@ TString GetPrettyStatistics(const TString& statistics) {
 
 namespace {
 
-const NJson::TJsonValue* GetStatisticsRoot(const NJson::TJsonValue& allStatistics) {
+const NJson::TJsonValue* GetStatisticsRoot(const NJson::TJsonValue& allStatistics, std::string_view v1Pointer = "Graph=0") {
```
Hor911, line 307: Both kind of statistics may include multiple graphs. To get correct stats it is needed to enumerate all of them and summarize total value

v1 graph will have root with name starting from
- Graph=
- Precompute=

v2 graph will have root with name starting from
- ResultSet
- Precompute
- Sink

v2 names may have no suffixes or suffixes like "_0", so Precompute can be distinguished from v1 by "=" symbol

dorooleg (approved):

PR #1053, opened 2024-01-16
KIKIMR-20740: QueryCache with temp tables

shnikd:


ydb/core/kqp/compile_service/kqp_compile_service.cpp
```diff
@@ -736,18 +745,26 @@ class TKqpCompileService : public TActorBootstrapped<TKqpCompileService> {
...
             if (compileResult->Status == Ydb::StatusIds::SUCCESS) {
-                if (QueryCache.FindByUid(compileResult->Uid, false)) {
-                    QueryCache.Replace(compileResult);
-                } else if (keepInCache) {
-                    if (QueryCache.Insert(compileResult, TableServiceConfig.GetEnableAstCache())) {
-                        Counters->CompileQueryCacheEvicted->Inc();
-                    }
-                    if (compileResult->Query && compileResult->Query->Settings.IsPrepareQuery) {
-                        if (InsertPreparingQuery(compileResult, compileRequest.KeepInCache)) {
+                if (!hasTempTables) {
```
VPolka, line 756: Очень большая вложенность получается везде, и сложнее будет понимать
Может стоит попробовать работу в кэшом вынести в отдельные функции, например, что то тип такого

bool CanUseCache(...) {}

CompileResult* FindInCache(...) {
    if !CanUseCache(...) return nullptr;
    ...
}

void InsertIntoCache(...) {
     if !CanUseCache(...) return;
    ....
}

VPolka (dismissed):

ydb/core/kqp/compile_service/kqp_compile_service.cpp
```diff
@@ -814,12 +814,60 @@ class TKqpCompileService : public TActorBootstrapped<TKqpCompileService> {
         StartCheckQueriesTtlTimer();
     }
 
+    TKqpCompileResult::TConstPtr WithCache(
```
gridnevvvit, line 817: Basically the naming of the function doesn't make sense for me.
I assume we need the following function:

HasTempTablesNameClashes that returns the true whenever we have the naming clashes between already prepared query and temp tables in session. this method should be const .  this method  can help us that we need to compile this query because it's cache miss now.

the meaning of `forInsert` flag is also not clear.

ydb/core/kqp/compile_service/kqp_compile_service.cpp
```diff
@@ -814,12 +814,60 @@ class TKqpCompileService : public TActorBootstrapped<TKqpCompileService> {
...
+            }
+            return compileResult;
+        }
+        if (!tempTablesState) {
+            return compileResult;
+        }
+        auto tables = compileResult->PreparedQuery->GetQueryTables();
+        auto tempTables = THashSet<TString>();
+        for (const auto& [path, info] : tempTablesState->TempTables) {
+            tempTables.insert(path.second);
```
gridnevvvit, line 839: why this hash set is not the part of temp tables state?

gridnevvvit (approved):

PR #1078, opened 2024-01-17
Safe topic reader KIKIMR-20673

CyberROFL:


ydb/core/tx/replication/ydb_proxy/ydb_proxy.cpp
```diff
@@ -160,8 +160,12 @@ class TBaseProxyActor: public TActor<TDerived> {
 
 class TTopicReader: public TBaseProxyActor<TTopicReader> {
     void Handle(TEvYdbProxy::TEvReadTopicRequest::TPtr& ev) {
+        WaitEvent(ev->Sender, ev->Cookie);
+    }
+
+    void WaitEvent(const TActorId& sender, ui64 cookie) {
         auto request = MakeRequest(SelfId());
-        auto cb = [request, sender = ev->Sender, cookie = ev->Cookie](const NThreading::TFuture<void>&) {
+        auto cb = [request, sender, cookie](const NThreading::TFuture<void>&) {
```
Enjection, line 168: Here you are copying TWeakRef and incrementing/decrementing atomic. It's better to do something like this
`auto cb = [request = MakeRequest(SelfId()), sender, cookie]...`

Enjection (changes requested):

Enjection (approved):

PR #1100, opened 2024-01-17
Migrating from antlr3 to antlr4

OrlovPavel:
### Changelog entry 

...

### Changelog category 

* Experimental feature
* Improvement

### Additional information

I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en.

### DIFF
--- a/cmake/antlr.cmake
+++ b/cmake/antlr.cmake
@@ -1,7 +1,7 @@
 function(ensure_antlr)
     if(NOT ANTLR3_EXECUTABLE)
         find_program(ANTLR3_EXECUTABLE
-                     NAMES antlr3)
+                     NAMES antlr4)
         if (NOT ANTLR3_EXECUTABLE)
             message(FATAL_ERROR "Unable to find antlr3 program. Please install antlr3 and make sure executable file present in the $PATH env.")
         endif()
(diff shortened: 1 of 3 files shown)


dcherednik:
Can you please set draft status for this pr?

OrlovPavel:
Currently, antlr4 can't generate protobuf. Here are all available targets - https://github.com/antlr/antlr4/blob/master/doc/targets.md 
But I am going to try adding protobuf target according to this instruction https://github.com/tunnelvisionlabs/antlr4/blob/master/doc/creating-a-language-target.md using your custom protobuf.stg. Unfortunately this means changing antlr4 source code and then rebuilding jar, but antlr3 in ydb was once changed the same way to generate protobuf.

dcherednik:
antlr3 hasn't been not patched. Moreover in case of cmake build find_library will try to use antlr3 from system. 
antlr3 uses custom stg file ydb/library/yql/parser/proto_ast/org/antlr/codegen/templates/protobuf/protobuf.stg.in

So the primary task is to find the way to do something similar for antlr4. 

> Currently, antlr4 can't generate protobuf. Here are all available targets - https://github.com/antlr/antlr4/blob/master/doc/targets.md But I am going to try adding protobuf target according to this instruction https://github.com/tunnelvisionlabs/antlr4/blob/master/doc/creating-a-language-target.md using your custom protobuf.stg. Unfortunately this means changing antlr4 source code and then rebuilding jar, but, as I know, antlr3 in ydb was once changed the same way to generate protobuf.

dcherednik:
I will close the pr. Please make the new one if found the way to use antlr4 to generate protobuf.

PR #1104, opened 2024-01-17
add sensors for metadata cache KIKIMR-20775

vporyadke:
### Changelog entry 

Added sensors for monitoring database metadata cache

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/core/grpc_services/db_metadata_cache.h
+++ b/ydb/core/grpc_services/db_metadata_cache.h
@@ -4,6 +4,7 @@
 #include <ydb/library/actors/core/actor_bootstrapped.h>
 #include <ydb/library/actors/core/hfunc.h>
 #include <ydb/core/base/appdata.h>
+#include <ydb/core/base/counters.h>
 #include <ydb/core/base/feature_flags.h>
 #include <ydb/core/base/domain.h>
 #include <ydb/core/base/statestorage.h>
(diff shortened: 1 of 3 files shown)


ydb/core/grpc_services/db_metadata_cache.h
```diff
@@ -44,6 +45,9 @@ class TDatabaseMetadataCache : public TActorBootstrapped<TDatabaseMetadataCache>
     TActorId PublishActor;
     TActorId SubscribeActor;
     bool RequestInProgress = false;
+    ::NMonitoring::TDynamicCounterPtr Counters;
+    static const inline TString HEALTHCHECK_REQUESTS_MADE_COUNTER = "DbMetadataCache/HealthCheckRequestsMade";
```
adameat, line 49: I think people are more used to terms like "hit" or "miss"

adameat (changes requested):

ydb/core/grpc_services/db_metadata_cache.h
vporyadke: True, but RequestsMade aren't cache misses in the traditional sense - they are made periodically by the cache itself. You don't think calling them cache misses would be misleading?

adameat: it would. okay then. maybe we should add some sensors on grpc side. is there any flag or status in response data which could tell us if the response was generated from a cache?

adameat (approved):

PR #1106, opened 2024-01-17
Add GLOBAL for udf service files (YQL-17551)

rvu1024:


zverevgeny:
Such approach will bloat client's executables. IMHO it's better to specify required public symbols in the linker script

zverevgeny (approved):

PR #1118, opened 2024-01-18
KIKIMR-20009: fix race condition on blobs removing

ivanmorozov333:
### Changelog entry 

split actions:
on execute we have to write into db only
on complete we have to write to memory storage
in other way we have problem in case execution not finished, but gc finished before and if binary killed in this time, we have problem with meta withno blobs

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/core/tx/columnshard/blob_manager.h
+++ b/ydb/core/tx/columnshard/blob_manager.h
@@ -90,7 +90,8 @@ public:
     }
 
     // Deletes the blob that was previously permanently saved
-    virtual void DeleteBlob(const TUnifiedBlobId& blobId, IBlobManagerDb& db) = 0;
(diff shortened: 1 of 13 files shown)


ydb/core/tx/columnshard/blob_manager.h
```diff
@@ -90,7 +90,8 @@ class IBlobManager {
     }
 
     // Deletes the blob that was previously permanently saved
```
k-morozov, line 92: It seems to me that comment relates to old DeleteBlob and we can write new comments for DeleteBlobOnExecute and DeleteBlobOnComplete (prefer) or remove old comment.

k-morozov (approved):

PR #1139, opened 2024-01-18
fix missing breaks CLOUD-174018

vporyadke:
### Changelog entry 

...

### Changelog category 

### Additional information

...


adameat (approved):
wow

PR #1152, opened 2024-01-19
(refactoring) Split TEvChangeExchange into two parts: common & DS KIKIMR-20673

CyberROFL:
### Changelog entry 

...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/core/base/events.h
+++ b/ydb/core/base/events.h
@@ -135,7 +135,7 @@ struct TKikimrEvents : TEvents {
         ES_HEALTH_CHECK,
         ES_DQ = NYql::NDq::TDqEvents::ES_DQ_COMPUTE, // 4212
         ES_YQ, // 4213
-        ES_CHANGE_EXCHANGE,
+        ES_CHANGE_EXCHANGE_DATASHARD,
         ES_DATABASE_SERVICE, //4215
         ES_SEQUENCESHARD, // 4216
         ES_SEQUENCEPROXY, // 4217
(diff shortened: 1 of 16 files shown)


ydb/core/base/events.h
```diff
@@ -172,6 +172,7 @@ struct TKikimrEvents : TEvents {
         ES_PQ_PARTITION_CHOOSER,
         ES_GRAPH,
         ES_REPLICATION_SERVICE,
+        ES_CHANGE_EXCHANGE,
```
Enjection, line 175: Maybe it's better to swap ES_CHANGE_EXCHANGE and ES_CHANGE_EXCHANGE_DATASHARD? If ES_CHANGE_EXCHANGE were used somewhere moving it would cause hidden issues.

ydb/core/tx/datashard/change_exchange.cpp
```diff
@@ -151,5 +32,4 @@ TString TEvChangeExchange::TEvRemoveSender::ToString() const {
     << " }";
 }
 
-} // NDataShard
-} // NKikimr
+}
```
Enjection, line 35: Minor, but it's good idea to leave comment `} // NKikimr::NDataShard`

ydb/core/tx/datashard/change_exchange.h
```diff
@@ -27,30 +26,17 @@ struct TEvChangeExchange {
...
-        EvEnqueueRecords,
-        // Request change record(s) by id
-        EvRequestRecords,
-        // Change record(s)
-        EvRecords,
-        // Remove change record(s) from local database
-        EvRemoveRecords,
-
         // Add new change sender
         EvAddSender,
```
Enjection, line 30: It looks dangerous to move message id's. Here i checked that all of them are local, but just as an idea for further development: maybe we should keep local message ids apart from remote one's?

Enjection (changes requested):
In general, PR seems okay, but perturbations involving event id's look dangerous.

Enjection (approved):

PR #1181, opened 2024-01-21
Simplify error handling

azevaykin:


ydb/core/engine/minikql/minikql_engine_host.cpp
```diff
@@ -908,14 +908,13 @@ ui64 TEngineHost::GetTableSchemaVersion(const TTableId&) const {
     return 0;
 }
 
-ui64 TEngineHost::LocalTableId(const TTableId& tableId) const {
+ui64 LocalTableId(const TTableId& tableId) {
+    // TODO: May be move to <ydb/core/scheme/scheme_tabledefs.h> ?
     return tableId.PathId.LocalPathId;
```
azevaykin, line 913: May be move to <ydb/core/scheme/scheme_tabledefs.h> ?

snaury: You cannot replace this method like that, because LocalTableId is NOT a LocalPathId. I think this is just a default implementation for local minikql transactions, where table ids are resolved into `(tabletId, localTableId)` pairs. Having that behavior baked into a public free function doesn't make sense, this is the exception, not the norm.

snaury (changes requested):

ydb/core/engine/mkql_engine_flat.cpp
```diff
@@ -798,7 +798,7 @@ class TEngineFlat : public IEngineFlat {
         return result;
     }
 
-    EResult Validate(TValidationInfo& validationInfo) override {
+    EResult ExtractKeys(TValidationInfo& validationInfo) override {
```
snaury, line 801: In the context of old engine this rename is wrong. This method does not extract keys, it actually validates whether it is possible to execute old minikql transaction, with validation info as output. Key extraction is only a useful side-effect.

snaury (approved):

PR #1218, opened 2024-01-23
Make ydb_root_common use only stderr to avoid command output pollution

Enjection:
### Changelog entry 

- make ydb_root_common use only stderr to avoid command output pollution

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/public/lib/ydb_cli/common/progress_bar.cpp
+++ b/ydb/public/lib/ydb_cli/common/progress_bar.cpp
@@ -26,7 +26,7 @@ void TProgressBar::AddProgress(size_t value) {
 
 TProgressBar::~TProgressBar() {
     if (!Finished) {
-        Cout << Endl;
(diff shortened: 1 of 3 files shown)


ydb/public/lib/ydb_cli/common/progress_bar.cpp
```diff
@@ -62,12 +62,12 @@ void TProgressBar::Render()
     output += TString("█") * filledBarLen;
     output += TString("░") * (barLen - filledBarLen);
     output += outputEnd;
-    Cout << output;
+    Cerr << output;
```
Gazizonoki, line 65: `import file` command outputs debug info in stdout and uses progress bar. I think mixing isn't good, better fix output in commands using progress bar too

Gazizonoki (approved):

PR #1219, opened 2024-01-23
Add ExecuteMode to YQL plugin for YT. Support validate and explain modes

Krock21:
### Changelog entry 

Issue: https://github.com/ytsaurus/ytsaurus/issues/223

Added ability to Validate and Explain(Optimize) YQL queries via YT YQL Plugin 
Added settingsLength to bridge to support settings with null-characters in them
...

### Changelog category 

* New feature

### DIFF
--- a/ydb/library/yql/yt/dynamic/impl.cpp
+++ b/ydb/library/yql/yt/dynamic/impl.cpp
@@ -12,7 +12,7 @@ extern "C" {
 
 ssize_t BridgeGetAbiVersion()
 {
-    return 1; // EYqlPluginAbiVersion::AbortQuery
+    return 2; // EYqlPluginAbiVersion
 }
 
 TBridgeYqlPlugin* BridgeCreateYqlPlugin(const TBridgeYqlPluginOptions* bridgeOptions)
@@ -74,8 +74,10 @@ TBridgeQueryResult* BridgeRun(
     const char* impersonationUser,
     const char* queryText,
     const char* settings,
+    int settingsLength,
     const TBridgeQueryFile* bridgeFiles,
-    int bridgeFileCount)
+    int bridgeFileCount,
+    int executeMode)
 {
     static const auto EmptyMap = TYsonString(TString("{}"));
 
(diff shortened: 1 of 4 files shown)


ydb/library/yql/yt/dynamic/impl.cpp
```diff
@@ -75,7 +75,8 @@ TBridgeQueryResult* BridgeRun(
     const char* queryText,
     const char* settings,
     const TBridgeQueryFile* bridgeFiles,
-    int bridgeFileCount)
+    int bridgeFileCount,
+    int executeMode)
```
Krisha11, line 79: You should also modify the bridge file
https://github.com/ydb-platform/ydb/blob/bbb26a84956dd242b2e8f6947e980f69b3b09035/ydb/library/yql/yt/bridge/interface.h#L93C24-L93C42

ydb/library/yql/yt/native/plugin.cpp
```diff
@@ -313,8 +314,45 @@ class TYqlPlugin
...
+            break;
+        default:
+            return TQueryResult{
+                .YsonError = MessageToYtErrorYson(Format("Unknown execution mode: %v", executeMode)),
+            };
+        }
+
         NYql::TProgram::TStatus status = NYql::TProgram::TStatus::Error;
-        status = program->RunWithConfig(impersonationUser, pipelineConfigurator);
+        switch (executeModeEnum) {
```
Krisha11, line 341: Maybe we can use just one switch and write something like:
```
case 0: // NYql::NEmbedded::EExecuteMode::Run:
    status = program->RunWithConfig(impersonationUser, pipelineConfigurator);
    break;
```

It won't change the logic, but it will simplify the code a little bit.

Krock21: Original idea was to transform `NYT::NYqlClient::EExecuteMode` to `NYql::NEmbedded::EExecuteMode`, and then cover all cases in "switch"

Seems like we don't need to use `NYql::NEmbedded::EExecuteMode` at all, so removed this logic and use one switch now

Krisha11 (approved):

vitstn (approved):

PR #1310, opened 2024-01-25
Support table changes observer KIKIMR-20853

snaury:
### Changelog entry 

Support table changes observer

### Changelog category 

### Additional information

This will allow datashard followers to inspect leader changes without re-reading tables at the start of every transaction.

### DIFF
--- a/ydb/core/tablet_flat/flat_executor.cpp
+++ b/ydb/core/tablet_flat/flat_executor.cpp
@@ -97,6 +97,32 @@ TTableSnapshotContext::~TTableSnapshotContext() = default;
 
 using namespace NResourceBroker;
 
+class TExecutor::TActiveTransactionZone {
+public:
+    explicit TActiveTransactionZone(TExecutor* self) noexcept
+        : Self(self)
+    {
+        Y_DEBUG_ABORT_UNLESS(!Self->ActiveTransaction);
+        Self->ActiveTransaction = true;
+        Active = true;
+    }
+
+    ~TActiveTransactionZone() noexcept {
+        Done();
+    }
+
+    void Done() noexcept {
+        if (Active) {
+            Self->ActiveTransaction = false;
+            Active = false;
+        }
+    }
+
+private:
+    TExecutor* Self;
+    bool Active = false;
+};
+
 TExecutor::TExecutor(
         NFlatExecutorSetup::ITablet* owner,
         const TActorId& ownerActorId)
@@ -877,8 +903,7 @@ void TExecutor::ApplyFollowerUpdate(THolder<TEvTablet::TFUpdateBody> update) {
         return;
 
     // Protect against recursive transactions in callbacks
-    Y_DEBUG_ABORT_UNLESS(!ActiveTransaction);
-    ActiveTransaction = true;
+    TActiveTransactionZone activeTransaction(this);
 
     TString schemeUpdate;
     TString dataUpdate;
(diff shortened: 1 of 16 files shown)


ydb/core/tablet_flat/flat_table.cpp
```diff
@@ -1338,6 +1341,11 @@ TCompactionStats TTable::GetCompactionStats() const
     return stats;
 }
 
+void TTable::SetTableObserver(ITableObserverPtr ptr) noexcept
+{
+    TableObserver = std::move(ptr);
```
kunga, line 1346: Let's check that current `TableObserver ` is null?

ydb/core/tablet_flat/flat_table_observer.h
```diff
@@ -0,0 +1,61 @@
...
+            ERowOp rop,
+            TArrayRef<const TRawTypeValue> key,
+            TArrayRef<const TUpdateOp> ops,
+            ui64 txId) = 0;
+    };
+
+    /**
+     * Smart pointer that can safely be used to call methods even when it's nullptr
+     */
+    class ITableObserverPtr : public TIntrusivePtr<ITableObserver> {
```
kunga, line 34: Do we really need that wrapper? It seems that just checking 

```
if (TableObserver) {
    TableObserver->OnUpdate(rop, key, ops, rowVersion);
}
```

is easier

ydb/core/tablet_flat/flat_executor.cpp
```diff
@@ -876,6 +876,10 @@ void TExecutor::ApplyFollowerUpdate(THolder<TEvTablet::TFUpdateBody> update) {
     if (update->IsSnapshot) // do nothing over snapshot after initial one
         return;
 
+    // Protect against recursive transactions in callbacks
+    Y_DEBUG_ABORT_UNLESS(!ActiveTransaction);
+    ActiveTransaction = true;
```
kunga, line 881: It's hard to track that there is no `return` in that method before `ActiveTransaction = false` line

Maybe we can use some shared pointer, copy it here to a local variable and check that it has zero ref count, so it will be automatically released on field destruction? like lock_guard semantic

kungasc (changes requested):

ydb/core/tablet_flat/flat_table.cpp
snaury: Why? I may want to install different observers depending on the lifecycle.

kunga: It's easy to double call `SetTableObserver` and expect that both observers will work

But if we really need such an option we can additionally have `ResetTableObserver` method, so anyone who call `SetTableObserver` should reset it before a next user

ydb/core/tablet_flat/flat_table_observer.h
kunga: Also it seems that someone may call `TableObserver->OnUpdate(rop, key, ops, rowVersion)` even now with TableObserver=nullptr

ydb/core/tablet_flat/flat_table.cpp
snaury: I would prefer a simple uncomplicated setter for now, then clients could just call `SetTableObserver(nullptr)` without the need for a separate reset method.

kunga: `SetTableObserver(nullptr)` is ok instead of  `ResetTableObserver`

ydb/core/tablet_flat/flat_executor.cpp
snaury: Refactored into a `TActiveTransactionZone` RAII.

kungasc (approved):

PR #1317, opened 2024-01-25
Sift PRAGMA TablePathPrefix down to the persisted query text of a view

jepett0:
KIKIMR-20656

Whenever a user SELECTs from a view, we compile the persisted query text of the view in an [almost blank context](https://github.com/ydb-platform/ydb/blob/0c915e616b63e7047189c91e523a47e6476bd3df/ydb/core/kqp/provider/rewrite_io_utils.cpp#L42), which doesn't have any connection with the context in which the corresponding CREATE VIEW statement was executed. This means that the tables mentioned in the query text have to be specified by their absolute path. I believe this is very inconvinient and unexpected for users and needs to be fixed.

The solution proposed in this PR is the following: [prepend](https://github.com/ydb-platform/ydb/blob/de947bf1d784a3f88d7f848b752367e1dae98264/ydb/library/yql/sql/v1/sql_translation.cpp#L4507) the persisted query text with the PRAGMA TablePathPrefix (if it is present in the context of the CREATE VIEW statement) and save it like this. This should be transparent to the user and moreover this solution enables cross YDB installation usage of VIEWs, i.e. the VIEWs created in YDB version #X should be usable (= SELECTable from) in YDB version #Y. The alternative solution (persist to the database not the query text, but the parsed AST of the query text, which is prepared using the context of the CREATE VIEW statement) might turn out to be incompatible across different YDB versions.

In the future we might found out that some other parts of the context in which the CREATE VIEW statement was executed need to be persisted in some way for later SELECTing from that view. I would like us to agree on the way we will do this and use this PR for the discussion.

### DIFF
--- a/ydb/core/kqp/ut/view/input/cases/multiple_tables/create_view.sql
+++ b/ydb/core/kqp/ut/view/input/cases/multiple_tables/create_view.sql
@@ -1,17 +1,14 @@
-CREATE VIEW `/Root/read_from_multiple_tables` WITH (security_invoker = TRUE) AS
+CREATE VIEW read_from_multiple_tables WITH (security_invoker = TRUE) AS
     SELECT
         *
-    FROM `/Root/series`
-        AS series
+    FROM series
     JOIN (
         SELECT
             seasons.title AS seasons_title,
             episodes.title AS episodes_title,
             seasons.series_id AS series_id
-        FROM `/Root/seasons`
-            AS seasons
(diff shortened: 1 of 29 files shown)


jepett0:
We decided with @nepal and @vitstn that it would be better to implement a way to serialise the necessary parts of the context of the parser at the moment of CREATE VIEW call to use later at the moment of SELECT from view call. Probably, we will design a special protobuf for this

PR #1360, opened 2024-01-26
YQL-17583: DELETE with alias crash fix

marsaly79:
### Changelog entry 
DELETE FROM tablename AS alias doesn't crash anymore. Still unsupported, though

### Changelog category 
* Bugfix


lll-phill-lll (approved):

resetius (approved):
OK

PR #1374, opened 2024-01-29
[dq] Don't return future with exception from ExecPlan call (YQL-17677)

rvu1024:
### Changelog entry

* Fix query assert/crash when DQ session is unexpectedly closed

### Changelog category

* Bugfix


resetius (approved):
OK

PR #1393, opened 2024-01-29
create statistics aggregator tablet during migration

alexd65536:
create statistics aggregator tablet during migration


the-ancient-1 (approved):
LGTM

PR #1404, opened 2024-01-30
Blocks for first stages

nikvas0:
### Changelog entry 

enable block computation for first stages

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/core/kqp/host/kqp_runner.cpp
+++ b/ydb/core/kqp/host/kqp_runner.cpp
@@ -194,29 +194,13 @@ public:
         YQL_ENSURE(IsIn({EKikimrQueryType::Query, EKikimrQueryType::Script}, TransformCtx->QueryCtx->Type));
         YQL_ENSURE(TMaybeNode<TKiDataQueryBlocks>(query));
 
-        const auto dataQueryBlocks = TKiDataQueryBlocks(query);
-        if (IsOlapQuery(dataQueryBlocks)) {
-            TypesCtx.UseBlocks = true;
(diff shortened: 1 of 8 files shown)


ydb/core/kqp/host/kqp_runner.cpp
```diff
@@ -192,14 +194,19 @@ class TKqpRunner : public IKqpRunner {
         YQL_ENSURE(IsIn({EKikimrQueryType::Query, EKikimrQueryType::Script}, TransformCtx->QueryCtx->Type));
         YQL_ENSURE(TMaybeNode<TKiDataQueryBlocks>(query));
 
+        TypesCtx.BlockEngineMode = NYql::EBlockEngineMode::Auto;
```
nepal, line 197: You may want to explicitly disable dq.UseWideChannels and dq.UseWideBlockChannels, since BlockEngineMode=auto will enable WideChannels and WideBlocksChannels unless they are disabled explicitly

nikvas0: We already set scalar channels only mode here https://github.com/ydb-platform/ydb/blob/97103370d3942b7301dfd3d042df52dcf0bc15b5/ydb/core/kqp/opt/kqp_opt_build_txs.cpp#L26

(similar to https://github.com/ydb-platform/ydb/blob/97103370d3942b7301dfd3d042df52dcf0bc15b5/ydb/library/yql/providers/dq/provider/exec/yql_dq_exectransformer.cpp#L287)

nepal (approved):

PR #1423, opened 2024-01-30
Change max size of DP table

Dark-Avery:
Changing default max size of DP table for DPccp algorithm
Also change size of bitset for graph


ydb/library/yql/dq/opt/dq_opt_join_cost_based.cpp
```diff
@@ -718,7 +718,7 @@ template <int N> std::bitset<N> TDPccpSolver<N>::NextBitset(const std::bitset<N>
     return res;
 
     // TODO: We can optimize this with a few long integer operations,
-    // but it will only work for 254 bit bitsets
+    // but it will only work for 128 bit bitsets
```
pavelvelikhov, line 721: This is a commented line, any the method below (which is commented out) works only for 64 bits. Can you change this comment back?

resetius (approved):
Better to define this constant somewhere as a constexpr

PR #1435, opened 2024-01-30
Move signals yql lib in os

Krisha11:
### Changelog entry 
Move signals yql lib in os

### Changelog category 


resetius (approved):
OK

PR #1488, opened 2024-01-31
Add unit tests for EvaluateExpr in VIEWs

jepett0:
KIKIMR-20819

Concerns were expressed that EvaluateExpr might behave differently if written inside VIEWs. The prime suspicion was that queries selecting from views would be loaded from the node's query cache and the result of the following query:
```sql
-- CREATE VIEW CurrentTimeView WITH (security_invoker = TRUE) AS SELECT CurrentUtcTimestamp();
SELECT * FROM CurrentTimeView;
```
would always be the same.

However, this suspicion turned out to be wrong. No code changes were necessary to ensure the same behaviour of EvaluateExpr written (1) in a view and (2) in a SELECT statement. This PR only adds unit tests.

### DIFF
--- a/ydb/core/kqp/ut/view/view_ut.cpp
+++ b/ydb/core/kqp/ut/view/view_ut.cpp
@@ -60,15 +60,23 @@ void ExecuteDataDefinitionQuery(TSession& session, const TString& script) {
                                           << script << "\nThe issues:\n" << result.GetIssues().ToString());
 }
 
-TDataQueryResult ExecuteDataModificationQuery(TSession& session, const TString& script) {
+TDataQueryResult ExecuteDataModificationQuery(TSession& session,
+                                              const TString& script,
+                                              const TExecDataQuerySettings& settings = {}
+) {
     const auto result = session.ExecuteDataQuery(
             script,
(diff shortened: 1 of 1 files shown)


ydb/core/kqp/ut/view/view_ut.cpp
```diff
@@ -377,3 +385,82 @@ Y_UNIT_TEST_SUITE(TSelectFromViewTest) {
...
+        queryExecutionSettings.KeepInQueryCache(true);
+        const auto executeTwice = [&](const TString& query) {
+            return TVector<TString>{
+                GetYsonResults(session, query, queryExecutionSettings),
+                GetYsonResults(session, query, queryExecutionSettings)
+            };
+        };
+        const auto viewResults = executeTwice(selectFromViewQuery);
+        const auto etalonResults = executeTwice(timeQuery);
+        UNIT_ASSERT_EQUAL_C(viewResults[0] < viewResults[1], etalonResults[0] < etalonResults[1],
```
jepett0, line 424: This might be counter intuitive, but etalonResult[0] is equal to etalonResult[1] at the moment of writing this unit test. @gridnevvvit confirmed that this behaviour is somewhat expected.

spuchin (approved):

Issue #1497, opened 2024-01-31
JSON import requires all columns to be provided


zinal:
YDB CLI's `ydb import file json` requires that all columns for the particular table are provided. This looks like an overkill, specifically for nullable columns, where the missing column in a row can be treated as NULL.

zinal:
Added pull request https://github.com/ydb-platform/ydb/pull/1498 to resolve the issue

PR #1498, opened 2024-01-31
#1497: treat missing JSON values as NULLs for import

zinal:
### Changelog entry 

YDB CLI now treats missing JSON values as NULLs for `ydb import file json` command.

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/public/lib/json_value/ydb_json_value.cpp
+++ b/ydb/public/lib/json_value/ydb_json_value.cpp
@@ -745,11 +745,15 @@ namespace {
                     const TString& memberName = TypeParser.GetMemberName();
                     const auto it = jsonMap.find(memberName);
                     if (it == jsonMap.end()) {
-                        ThrowFatalError(TStringBuilder() << "No member \"" << memberName
-                            << "\" in the map in json string for YDB struct type");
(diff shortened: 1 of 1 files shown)


Gazizonoki (approved):

ydb/public/lib/json_value/ydb_json_value.cpp
```diff
@@ -745,11 +745,15 @@ namespace {
                     const TString& memberName = TypeParser.GetMemberName();
                     const auto it = jsonMap.find(memberName);
                     if (it == jsonMap.end()) {
-                        ThrowFatalError(TStringBuilder() << "No member \"" << memberName
-                            << "\" in the map in json string for YDB struct type");
+                        // missing JSON value is treated as NULL on input
+                        ValueBuilder.AddMember(memberName);
+                        TypeParser.OpenOptional();
```
Gazizonoki, line 750: Better to check, that value kind is optional, because otherwise error will be confusing for user. For example, "Expected optional type because no member in the map in json string for YDB struct type"

Gazizonoki (changes requested):

PR #1536, opened 2024-02-02
KIKIMR-20945: simplify example configurations

blinkov:
### Changelog entry 

* Allow to specify `type: RAM` in configuration for in-memory disks.
* Allow to specify default disk type, erasure species, and TLS settings only once.
* Allow to skip explicitly specifying `vdisk_kind`.
* Assume all hosts are in the same datacenter but distinct bodies and racks, unless specified otherwise.
* Clean-up example configs.

### Changelog category 

* New feature

### Additional information

...


blinkov:
The `asan` findings seem to be unrelated.

Enjection:
Oh.. This code is under heavily refactoring right now, I would like to merge my refactoring first and port this pull request to new configuring method, because further increasing complexity of yaml parser will lead to completely unreadable and unsupportable code. Another reason for refactoring this part is a lack of error message for user when he tries to configure cluster.
I will port your PR myself after mine PR will be merged. Hope you don't mind.

PR #1558, opened 2024-02-02
KIKIMR-20234: Add access right for topic commit offset

molotkov-and:
Add check right SelectRow with check rights ReadAttributes and WriteAttributes for commit topic offset


adameat (approved):

antonkovalenko:
Let's merge to 24-1 otherwise auditor privilege is endangered! @molotkov-and

PR #1571, opened 2024-02-04
[ydb-configurator] added host_config_id and host_configs in config.yaml

shmel1k:
Add host_configs and host_config_id for hosts section in YDB configuration file

* New feature
* Experimental feature
* Improvement

### Additional information

...

### DIFF
--- a/ydb/tools/cfg/static.py
+++ b/ydb/tools/cfg/static.py
@@ -385,6 +385,9 @@ class StaticConfigGenerator(object):
         if self.table_service_config:
             normalized_config["table_service_config"] = self.table_service_config
 
+        if self.__cluster_details.host_configs is not None:
+            normalized_config["host_configs"] = self.__cluster_details.raw_host_configs
+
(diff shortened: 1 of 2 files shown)


ydb/tools/cfg/static.py
```diff
@@ -385,6 +385,9 @@ def get_yaml_format_config(self):
         if self.table_service_config:
             normalized_config["table_service_config"] = self.table_service_config
 
+        if self.__cluster_details.host_configs is not None:
+            normalized_config["host_configs"] = self.__cluster_details.raw_host_configs
```
kobzonega, line 389: We can't just copy this section `host_configs` as is, cause there are names collision

With `cluster.yaml` we are using field `drives` in `host_configs` section
https://github.com/ydb-platform/ydb/blob/main/ydb/tools/cfg/base.py#L475

But inside `config.yaml` we should use field `drive` in `host_configs` section
https://github.com/ydb-platform/ydb/blob/main/ydb/library/yaml_config/static_validator/builders.cpp#L18

PR #1584, opened 2024-02-05
Use std::atomic in shared data header

snaury:
### Changelog entry 

Use std::atomic in shared data header

### Changelog category 

* Improvement

### Additional information

This should fix performance issues where TSharedData is more expensive to copy than std::shared_ptr.


kungasc (approved):
Now copying of `TSharedData` works the same time as `std::shared_ptr<TString>`:
```
-------------------------------------------------------------|-------------   
Benchmark                                          Before    |    After    
-------------------------------------------------------------|------------- 
TPartIndexSeekFixture/PassIndexPage/0/0            0.018 us  |    0.013 us    
TPartIndexSeekFixture/PassIndexPage/1/0            0.013 us  |    0.013 us 
```

Full benchmarks results https://gist.github.com/kunga/02e7e9db74b09801774010cdc0cf1e2b

PR #1599, opened 2024-02-05
YQL-17542 move TaskRunner dependent Execute to TDqSyncComputeActorBase

zverevgeny:
pr over contrib: https://a.yandex-team.ru/review/5283356/details


ydb/library/yql/dq/actors/compute/dq_async_compute_actor.cpp
```diff
@@ -524,10 +523,6 @@ class TDqAsyncComputeActor : public TDqComputeActorBase<TDqAsyncComputeActor, TC
             }
         }
 
-        if (status != ERunStatus::Finished) {
```
Darych, line 527: Is it true that this code is not necessary here?

zverevgeny: It used to call a method from TDqComputeActorBase, that does nothing

Darych (approved):

PR #1637, opened 2024-02-06
store history of changes made to hive via http

vporyadke:
### Changelog entry 

You can now view history of changes made to hive settings through http requests (Only works for changes made after the introduction of this feature)

### Changelog category 

* New feature

### Additional information

...


vporyadke:
image

ydb/library/actors/protos/actors.proto
```diff
@@ -34,6 +34,7 @@ message TRemoteHttpInfo {
     optional bytes PostContent = 8;
     repeated THeader Headers = 9;
     optional string RemoteAddr = 7;
+    optional string UserToken = 10;
```
adameat, line 37: we didn't have a separate UserToken field here because token is always in Headers field

ydb/core/mind/hive/hive_schema.h
```diff
@@ -298,6 +298,15 @@ struct Schema : NIceDb::Schema {
         using TColumns = TableColumns<Node, TabletType, MaxCount>;
     };
 
+    struct OperationsLog : Table<21> {
+        struct Timestamp : Column<1, NScheme::NTypeIds::Uint64> {};
+        struct User : Column<2, NScheme::NTypeIds::String> {};
+        struct Operation : Column<3, NScheme::NTypeIds::String> { using Type = NKikimrHive::TOperation; };
```
adameat, line 304: I don't think you actually need these complex protobuf specs just for logging purposes. And think about migrations and rollbacks - it could became a huge pain...
I think just a free text would be more than enough. (or json if you need some structure to parse)

adameat (changes requested):

adameat (approved):

PR #1678, opened 2024-02-07
YQ-2824: Bring minimal schema inference to cloned clickhouse code

EgorkaZ:
### Changelog entry 

Bring minimal necessary code from ClickHouse for schema inference

### Changelog category 

### Additional information

* Generally code is taken from https://github.com/ClickHouse/ClickHouse/pull/32455
* It uses evaluation of constant expressions for inferring types, which is not present in our code and in ClickHouse was later replaced with more explicit inference by hand. The latter seemed as bringing less code, thus was used instead


dorooleg:
Больше не актуально

PR #1688, opened 2024-02-07
Fix ENUM_SERIALIZATION_ARGS_GEN_HEADER cmake support.

dcherednik:
### Changelog category 

* Bugfix


alexv-smirnov:
Wrong file, needs to be changed here: build/export_generators/hardcoded-cmake/cmake/common.cmake
It's a template for CMakeLists generation.

PR #1741, opened 2024-02-08
improve base stats propagation logic

alexd65536:
improve base stats propagation logic


the-ancient-1 (approved):
LGTM

PR #1792, opened 2024-02-12
Workload read without consumer and metrics fixes

niksaveliev:
### Changelog category

### Additional information

...


ydb/core/persqueue/read_balancer.cpp
```diff
@@ -536,7 +538,8 @@ void TPersQueueReadBalancer::Handle(TEvPersQueue::TEvUpdateBalancerConfig::TPtr
             Consumers[rr];
         }
     }
-
+    Consumers[CLIENTID_WITHOUT_CONSUMER];
```
alexnick88, line 541: Read without consumer doesn't use read balancer. Why do we need this line?

ydb/core/persqueue/ut/resources/counters_topics.html
```diff
@@ -31,6 +31,20 @@
                         name=topic.reserve.used_bytes: 0
                         name=topic.storage_bytes: 0
 
+                        consumer=$without_consumer:
```
alexnick88, line 34: This is not what we want to see. No label consumer is possible solution. But some sensors are useless in reading without consumer. Sensor topic.partition.committed_end_to_end_lag_milliseconds_max is undefined for example.

alexnick88 (approved):

PR #1872, opened 2024-02-13
YDB-1065 added proxying for render handler

StekPerepolnen:
added parameters for handler
`&direct=false&database=%2Fslice%2Fdb`

`direct` is false by default - means graph info will be collect on some differ dynamic node which belongs to specific database
if true - means that graph info must be collect on the same node (like before)


ydb/core/viewer/json_tenantinfo.h
```diff
@@ -389,9 +389,10 @@ class TJsonTenantInfo : public TViewerPipeClient<TJsonTenantInfo> {
                 RequestDone();
                 break;
             case NKikimrViewer::TEvViewerResponse::kQueryResponse:
-            case NKikimrViewer::TEvViewerResponse::kReserved14:
+            case NKikimrViewer::TEvViewerResponse::kRenderResponse:
```
adameat, line 392: maybe we could put default here?

ydb/core/viewer/json_tenantinfo.h
```diff
@@ -425,9 +426,10 @@ class TJsonTenantInfo : public TViewerPipeClient<TJsonTenantInfo> {
                     SendViewerSystemRequest(tenantId);
                     break;
                 case NKikimrViewer::TEvViewerRequest::kQueryRequest:
-                case NKikimrViewer::TEvViewerRequest::kReserved14:
+                case NKikimrViewer::TEvViewerRequest::kRenderRequest:
```
adameat, line 429: and here

ydb/core/viewer/viewer_request.cpp
```diff
@@ -74,9 +75,11 @@ IActor* CreateViewerRequestHandler(TEvViewer::TEvViewerRequest::TPtr& request) {
             return new TViewerWhiteboardRequest<TEvWhiteboard::TEvSystemStateRequest, TEvWhiteboard::TEvSystemStateResponse>(request);
         case NKikimrViewer::TEvViewerRequest::kQueryRequest:
             return new TJsonQuery(request);
-        case NKikimrViewer::TEvViewerRequest::kReserved14:
+        case NKikimrViewer::TEvViewerRequest::kRenderRequest:
```
adameat, line 78: but not here!

adameat (changes requested):

adameat (approved):

PR #1916, opened 2024-02-14
pg wire bypasses rpc layer and doesn't perform attach call

dcherednik:
so we can't requre attach currently

### Changelog category 
* Bugfix


adameat (approved):
feel bad approving this

PR #1960, opened 2024-02-15
Invalidate query compilation cache entries with outdated VIEWs

jepett0:
KIKIMR-21002

YDB caches query compilation results on the server side for efficiency. (For small queries compilation can take up to 100x times more time than the execution.) We search the cache entry by the text of the query. One can miss the cache just by adding a comment to the query. (Well, not anymore, because we have added cache by AST recently. Compilation takes much more time than AST building, so it makes sense.) The compilation result caching could theoretically cause problems even for the basic:
```sql
SELECT * FROM some_table_whose_name_stays_the_same_while_the_content_changes;
```
We could change the definition of this table in a separate session like this:
```sql
DROP TABLE some_table_...;
CREATE TABLE some_table_... ( /* different content */ );
```
and expect the select from this table to produce wrong results, because of the query cache, which will not notice the change in the definition of the table, because it is not apparent in the text of the query. However, there are some special mitigation mechanisms implemented for tables, which were missing for views up until this PR.

In this PR we add the following algorithm for invalidating cache entries for outdated VIEWs:
1) Store path ids and schema versions of the views that were used in the query in the cache entries, so they can be accessed later.
2) Whenever we retrieve a compilation result from cache, send a request for SchemeCache to check if the schema version of the views used in this query (if any) has not changed since we compiled this query.
3) Send a recompilation request if any view is outdated.

There are two important things to note about this solution:
- We make a SchemeCache request for each repeated query and there is a lot of these in an OLTP-focused database like YDB. However, we have already been sending these request for preliminary (this is not the last check of schema version mismatch (at least for tables)) cache invalidation for tables, so views should not incur an additional performance impact here.
- This solution does not guarantee strong consistency for queries using views, because query cache invalidation will not happen instantly after the view definition is updated. The node should get an update from the SchemeCache, which takes some time.

### DIFF
--- a/ydb/core/protos/kqp_physical.proto
+++ b/ydb/core/protos/kqp_physical.proto
@@ -508,5 +508,5 @@ message TKqpPhyQuery {
 
     string QueryDiagnostics = 10;
 
-    repeated TKqpTableInfo ViewsInfo = 11;
+    repeated TKqpTableInfo ViewInfos = 11;
 }
--- a/ydb/core/protos/kqp_physical.proto
+++ b/ydb/core/protos/kqp_physical.proto
@@ -507,4 +507,6 @@ message TKqpPhyQuery {
     bool HasUncommittedChangesRead = 9;
 
     string QueryDiagnostics = 10;
+
+    repeated TKqpTableInfo ViewsInfo = 11;
 }
--- a/ydb/core/kqp/provider/yql_kikimr_datasource.cpp
+++ b/ydb/core/kqp/provider/yql_kikimr_datasource.cpp
@@ -352,10 +352,11 @@ public:
                     return TStatus::Error;
                 }
 
-                if (res.Metadata->Kind == EKikimrTableKind::View) {
+                if (const auto& preparingQuery = SessionCtx->Query().PreparingQuery;
+                        preparingQuery
+                        && res.Metadata->Kind == EKikimrTableKind::View
+                ) {
                     const auto& viewMetadata = *res.Metadata;
-                    const auto& preparingQuery = SessionCtx->Query().PreparingQuery;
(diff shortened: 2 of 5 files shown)


ydb/core/protos/kqp.proto
```diff
@@ -111,17 +111,17 @@ message TQueryRequest {
 }
 
 message TKqpPathIdProto {
-    optional uint32 OwnerId = 1;
-    optional uint32 TableId = 2;
+    optional uint64 OwnerId = 1;
+    optional uint64 TableId = 2;
```
jepett0, line 115: The change from uint32 to uint64 produces binary compatible protobufs. See the docs of both [proto2](https://protobuf.dev/programming-guides/proto2/#updating) and [proto3](https://protobuf.dev/programming-guides/proto3/#updating), they have the same line on this topic:

```
int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility
```

ydb/core/protos/kqp_physical.proto
```diff
@@ -507,4 +507,6 @@ message TKqpPhyQuery {
     bool HasUncommittedChangesRead = 9;
 
     string QueryDiagnostics = 10;
+
+    repeated TKqpTableInfo ViewsInfo = 11;
```
ijon, line 511: `ViewsInfo` -> `ViewInfos`

1. the use of `ViewsInfo` is inconsistent with other intances of plural forms in that same message (and entire file)
2. compound nouns normally form the plural by adding -s to the last word of the compound

gridnevvvit (dismissed):

gridnevvvit (approved):

PR #2118, opened 2024-02-20
Fix kv init 'It is not allowed to create not null data column' 23-3 error

kungasc:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

Fix kv init 'It is not allowed to create not null data column' 23-3 error

KIKIMR-21089

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,4 @@ util/all_util.cpp
 util/charset/all_charset.cpp
 
 list_result.log
+bin/config.json
--- a/ydb/library/workload/kv_workload.cpp
+++ b/ydb/library/workload/kv_workload.cpp
@@ -78,13 +78,17 @@ void AddResultSet(const NYdb::TResultSet& resultSet, TVector<TRow>& rows) {
 
         for (size_t col = 0; col < parser.ColumnsCount(); col++) {
             auto& valueParser = parser.ColumnParser(col);
-            if (valueParser.GetKind() == NYdb::TTypeParser::ETypeKind::Optional) {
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -74,3 +74,4 @@ util/all_thread.cpp
 util/all_util.cpp
 util/charset/all_charset.cpp
 
+bin/config.json
```
kunga, line 77: запуск собранной ~/ydb/ydb/apps/ydb спавнит этот конфиг

```
{"last_check":1708449875,"outdated":false}
```

так что добавил его в игнор

ydb/library/workload/kv_workload.cpp
```diff
@@ -78,6 +78,9 @@ void AddResultSet(const NYdb::TResultSet& resultSet, TVector<TRow>& rows) {
         
         for (size_t col = 0; col < parser.ColumnsCount(); col++) {
             auto& valueParser = parser.ColumnParser(col);
+            if (valueParser.GetKind() == NYdb::TTypeParser::ETypeKind::Optional) {
+                valueParser.OpenOptional();
```
Gazizonoki, line 82: We shouldn't have a pair CloseOptional() below?

Gazizonoki (changes requested):

ydb/library/workload/kv_workload.cpp
kunga: I'm not sure, so ok, let's close it

But it seems to work without because we don't use `valueParser` anymore

Gazizonoki (dismissed):

fomichev3000 (approved):

PR #2198, opened 2024-02-22
Import from S3: check error is retriable KIKIMR-21102

CyberROFL:
### Changelog entry 

...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/core/tx/datashard/import_s3.cpp
+++ b/ydb/core/tx/datashard/import_s3.cpp
@@ -309,10 +309,6 @@ class TS3Downloader: public TActorBootstrapped<TS3Downloader> {
 
     }; // TUploadRowsRequestBuilder
 
-    enum class EWakeupTag: ui64 {
-        Restart,
-    };
-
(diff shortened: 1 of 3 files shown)


ydb/core/tx/datashard/import_s3.cpp
```diff
@@ -656,21 +652,35 @@ class TS3Downloader: public TActorBootstrapped<TS3Downloader> {
...
+    template <typename T>
+    void RetryOrFinish(const T& error) {
+        if (CanRetry(error)) {
+            Retry();
+        } else {
+            if constexpr (std::is_same_v<T, Aws::S3::S3Error>) {
+                Finish(false, TStringBuilder() << "S3 error: " << error.GetMessage().c_str());
+            } else {
+                Finish(false, error);
+            }
```
Enjection, line 683: Why did you use nested if here? `constexpr if` can be combined with regular `if`.

Enjection (approved):

PR #2217, opened 2024-02-23
YDB-1470 added database filter for time difference

StekPerepolnen:
rework interconnect sends clockskew info only for one tenant nodes. HC ignores clockskew for storage nodes now

### DIFF
--- a/ydb/library/actors/interconnect/interconnect_common.h
+++ b/ydb/library/actors/interconnect/interconnect_common.h
@@ -71,10 +71,10 @@ namespace NActors {
         bool Orange;
         bool Red;
         i64 ClockSkew;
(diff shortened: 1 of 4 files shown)


ydb/library/actors/interconnect/interconnect_common.h
```diff
@@ -71,8 +71,10 @@ namespace NActors {
         bool Orange;
         bool Red;
         i64 ClockSkew;
+        bool IsReportClockSkew;
```
alexvru, line 74: prefix "Is" makes this name very hard to process
either no prefix, or "Do"

alexvru (dismissed):

alexvru (approved):

PR #2286, opened 2024-02-27
Fix stale read of some acknowledged writes after a table split

snaury:
### Changelog entry 

Fix stale read of some acknowledged writes after a table split.

### Changelog category 

* Bugfix 

### Additional information

Some G-single-item-realtime anomalies were detected with Jepsen, which corresponded to a stale read immediately after a table split. Investigation showed several cases where a single-shard write could be acknowledged to clients by source shard, when destination shards would consider those versions unacknowledged due to lagging mediator time. The underlying races caused several unintended side-effects:

* Destination shards could attach to mediator time before they had all the necessary information:
  * Non-repeatable reads: destination shards could select a new write version which was supposed to be frozen by a repeatable snapshot read at their source shard during split
  * Stale reads: destination shards could select a new read version which was acknowledged by a single-shard write at their source shard during split
* Source shards could reply to writes after destination shards have fully initialized, which could cause stale reads due to mediator time lagging at their corresponding nodes

These issues are fixed by not starting mediator time restore until all snapshots are received by destination shards (this ensures destination shards await mediator time which is not less than the last theoretically observed by source shards at the time they sent their snapshots), and not sending delayed replies after a snapshot is prepared by source shards (this ensures destination shards may trust their local mediator time to determine write visibility).

Partially fixes KIKIMR-21065.

### DIFF
--- a/ydb/core/tx/datashard/create_table_unit.cpp
+++ b/ydb/core/tx/datashard/create_table_unit.cpp
@@ -86,6 +86,7 @@ EExecutionStatus TCreateTableUnit::Execute(TOperation::TPtr op,
         txc.DB.NoMoreReadsForTx();
         DataShard.SetPersistState(TShardState::Ready, txc);
         DataShard.CheckMvccStateChangeCanStart(ctx); // Recheck
+        DataShard.SendRegistrationRequestTimeCast(ctx);
     }
 
     return EExecutionStatus::DelayCompleteNoMoreRestarts;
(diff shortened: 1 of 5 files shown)


CyberROFL (dismissed):

snaury:
The `Cdc.ResolvedTimestamps` test exposed a `TxInit` race at split dst, need to find a proper non-racy way to initialize post split.

snaury:
Failing tests seem to be all KV/PQ, unrelated to datashards.

CyberROFL (approved):

PR #2301, opened 2024-02-28
[CLI] double output fixed

pashandor789:
### Changelog entry 

...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/public/lib/ydb_cli/common/format.cpp
+++ b/ydb/public/lib/ydb_cli/common/format.cpp
@@ -423,7 +423,7 @@ TString formatPrettyTableDouble(double value) {
     std::strstream stream;
 
     if (1e-3 < value && value < 1e8 || value == 0) {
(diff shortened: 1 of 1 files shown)


pashandor789:
before/after
before
after

ydb/public/lib/ydb_cli/common/format.cpp
```diff
@@ -416,6 +419,19 @@ TString replaceAll(TString str, const TString& from, const TString& to) {
     return str;
 }
 
+TString formatPrettyTableDouble(double value) {
+    std::strstream stream;
+
+    if (1e-3 < value && value < 1e8 || value == 0) {
+        stream << static_cast<int>(value) << '\0';
```
pavelvelikhov, line 426: This is not a correct rounding of doubles

Check out std::round

pavelvelikhov (approved):

PR #2356, opened 2024-02-29
Support for temporary tokens in yql_plugin

Krisha11:
### Changelog category 

* New feature

### Additional information

In this pull request, I am changing the authorization method in yql QT requests. Temporary tokens will be used instead of impersonation.

The algorithm for generating temporary tokens is:
1) We get all used clusters from the query (in yql_plugin)
2) Generating a temporary token on all clusters (in yql_agent)
3) Run the query (in yql_agent)

Unfortunately, in order to get all the clusters from a request, you must first compile it, and for this you must already have partial rights to execute it. Therefore, this step will be performed with superuser rights. (by yql_agent user)

yql_agent PR: https://github.com/ytsaurus/ytsaurus/pull/410

### DIFF
--- a/ydb/library/yql/yt/native/plugin.cpp
+++ b/ydb/library/yql/yt/native/plugin.cpp
@@ -75,6 +75,8 @@ namespace NNative {
 using namespace NYson;
 using namespace NKikimr::NMiniKQL;
 
+static const TString YqlAgent = "yql_agent";
+
 ////////////////////////////////////////////////////////////////////////////////
 
 std::optional<TString> MaybeToOptional(const TMaybe<TString>& maybeStr)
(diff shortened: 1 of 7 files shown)


ydb/library/yql/yt/native/plugin.cpp
```diff
@@ -523,10 +589,12 @@ class TYqlPlugin
                 };
             }
             if (!ActiveQueriesProgress_[queryId].Compiled) {
+                ActiveQueriesProgress_.erase(queryId);
                 return TAbortResult{
                     .YsonError = MessageToYtErrorYson(Format("Query %v is not compiled", queryId)),
                 };
```
Krock21, line 595: Shouldn't we just return success in this case?
Otherwise it is strange that we return an error while aborting query. This behaviour is strange:
```
1. First abort -> ERROR Query is not compiled
2. Second abort -> ERROR Query is not found
```

```suggestion
                return {};
```

ydb/library/yql/yt/native/plugin.cpp
```diff
@@ -554,10 +622,18 @@ class TYqlPlugin
...
+    TString YqlAgentToken_;
+
     YT_DECLARE_SPIN_LOCK(NThreading::TReaderWriterSpinLock, ProgressSpinLock);
     THashMap<TQueryId, TActiveQuery> ActiveQueriesProgress_;
     TVector<NYql::TDataProviderInitializer> DataProvidersInit_;
 
+    void RemoveQuery(TQueryId queryId)
+    {
+        auto guard = WriterGuard(ProgressSpinLock);
+        ActiveQueriesProgress_.erase(queryId);
```
Krock21, line 634: We didn't have lock between checking `contains` and this code. Someone else may have acquired the lock and erased the query.
I am not sure how erase works if there is no such element. I think it throws some exception
```suggestion
        if (ActiveQueriesProgress_.contains(queryId)) {
            ActiveQueriesProgress_.erase(queryId);
        }
```

Krisha11: Erase does not fail if there is no query. Therefore, I thought that this was not necessary.

But for clarity of the code, put it here

Krock21 (approved):
LGTM

rvu1024 (approved):

PR #2411, opened 2024-03-04
Support passing debug info from shards to grpc responses

snaury:
### Changelog entry 

Support passing debug info from shards to grpc responses

### Changelog category 

* Experimental feature

### Additional information

When investigating consistency failures (e.g. with jepsen) it is necessary to gather additional info about query execution, so it can be attached to operations in the history. This patch adds a feature flag that allows shards to generate debug info in json format and pass it via kqp to grpc responses.

### DIFF
--- a/ydb/core/grpc_services/rpc_commit_transaction.cpp
+++ b/ydb/core/grpc_services/rpc_commit_transaction.cpp
@@ -93,6 +93,8 @@ private:
                 FillQueryStats(*commitResult->mutable_query_stats(), kqpResponse);
             }
 
+            FillDebugInfo(*commitResult, kqpResponse);
+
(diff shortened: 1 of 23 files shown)


snaury:
Support meeting notes:

* Modifies public API, however not useful for end users. Intended use case is too specific: only for consistency checkers as part of bugs investigation
* Private forks currently use query_ast field for passing this info, but there's a query_plan field already, that contains json with various useful fields, better move debug info there
* Debug info should probably be part of full stats, returned only for special builds, and when full query stats are requested

snaury:
Пока закрою PR, о том как сделать лучше придумаю как-нибудь потом.

PR #2460, opened 2024-03-05
sessions system table

gridnevvvit:
### Changelog entry

implement the new introspection tool: the system view with the set of the query sessions.

### Changelog category

* Experimental feature

### Additional information

...

### DIFF
--- a/ydb/core/kqp/proxy_service/kqp_proxy_service_impl.h
+++ b/ydb/core/kqp/proxy_service/kqp_proxy_service_impl.h
@@ -88,6 +88,8 @@ struct TKqpSessionInfo {
     TNodeId AttachedNodeId;
     TActorId AttachedRpcId;
     bool PgWire;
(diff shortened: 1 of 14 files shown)


nikvas0 (dismissed):

ydb/core/kqp/proxy_service/kqp_proxy_service_impl.h
```diff
@@ -295,7 +308,7 @@ class TLocalSessionsRegistry {
         return LocalSessions.size();
     }
 
-    const TKqpSessionInfo* FindPtr(const TString& sessionId) const {
```
dcherednik, line 298: Lets make a method to set QueryText instead of removing const. It is unsafe to allow to change arbitrary fields for already created session. See AttachSession for example

dcherednik (dismissed):

dcherednik (approved):

PR #2495, opened 2024-03-06
Support Nebius IAM in server

UgnineSirdis:
### Changelog entry 

Nebius IAM support

### Changelog category 

* New feature

### Additional information

...


ydb/core/security/ticket_parser_impl.h
```diff
@@ -3,6 +3,7 @@
 #include <ydb/library/actors/core/actor_bootstrapped.h>
 #include <ydb/library/actors/core/hfunc.h>
 #include <library/cpp/digest/md5/md5.h>
+#include <ydb/library/nebius_cloud/impl/access_service.h>
```
adameat, line 6: maybe it would be better to call it ncloud, just to be even with existing ycloud ?

ydb/core/protos/auth.proto
```diff
@@ -49,6 +49,7 @@ message TAuthConfig {
     optional bool UseAccessServiceApiKey = 76 [default = false]; // Use IAM ApiKey
     optional string AsSignatureExpireTime = 77 [default = "1m"];
     optional bool UseBuiltinDomain = 78 [default = true];
+    optional string AccessServiceType = 79 [default = "Yandex"]; // For now the following values are supported: "Yandex", "Nebius"
```
adameat, line 52: maybe it's better to add a number as a version? just in case?

ydb/core/security/ticket_parser_impl.h
```diff
@@ -804,20 +960,132 @@ class TTicketParserImpl : public TActorBootstrapped<TDerived> {
...
     }
 
-    void Handle(NCloud::TEvAccessService::TEvBulkAuthorizeResponse::TPtr& ev) {
-        auto getResourcePathIdForRequiredPermissions = [] (const ::yandex::cloud::priv::accessservice::v2::Resource& resource_path) -> TString {
-            if (resource_path.type() == "resource-manager.folder") {
-                return " folder_id " + resource_path.id();
-            }
-            if (resource_path.type() == "resource-manager.cloud") {
-                return " cloud_id " + resource_path.id();
+    void Handle(NNebiusCloud::TEvAccessService::TEvAuthorizeResponse::TPtr& ev) {
```
adameat, line 963: it looks like a huge copy-pase code? am I wrong? is it possible to avoid or minimize copy of existing code?

ydb/core/security/ticket_parser_impl.h
```diff
@@ -1261,10 +1529,16 @@ class TTicketParserImpl : public TActorBootstrapped<TDerived> {
         for (const TString& sid : record.AdditionalSIDs) {
             token->AddGroupSID(sid);
         }
+        Cerr << "EnrichUserTokenWithBuiltins. Permissions: " << record.Permissions.size() << "\n";
```
adameat, line 1532: it looks like forgotten debug output

ydb/core/security/ticket_parser_impl.h
```diff
@@ -1261,10 +1529,16 @@ class TTicketParserImpl : public TActorBootstrapped<TDerived> {
...
         if (!record.Permissions.empty()) {
             TString subject;
             TVector<TString> groups;
             for (const auto& [permission, rec] : record.Permissions) {
+                Cerr << "Perm " << permission << " Subj " << rec.Subject << " Err \"" << rec.Error.Message << "\"";
```
adameat, line 1537: same here

adameat (changes requested):

UgnineSirdis (commented):
Fixed issues

ydb/core/security/ticket_parser_impl.h
UgnineSirdis: I've already tried to avoid copy and paste. But what I see is that API is similar, but not the same. For example, there is no UserAccountService/ServiceAccountService in Nebius, similar fields are taken from different places (for example, account field in authorization response is now in each result; results are in map, not in repeated field; there is additional field "authorized" etc). I can create complex templates to make the same code for both APIs, but I think it will be too complex in that case. I looked at the code one more time and refactored several places to that the copy & paste reduced. Look please at the new code and, if you'll see something more to be refactored, I can do it

ydb/core/protos/auth.proto
UgnineSirdis: Thank you! It's a good idea

adameat (approved):

molotkov-and (approved):

dorooleg (approved):

PR #2515, opened 2024-03-06
Use empty parsing context to build view's select in CREATE VIEW statement (#2049)

jepett0:
KIKIMR-20891

CREATE VIEW statement parses (and validates) the select statement saved in the view. It should be parsed in a context isolated from the statements executed before the CREATE VIEW statement (we haven't decided yet on the exact scope of the context of the view's select statement, see [KIKIMR-20656](https://st.yandex-team.ru/KIKIMR-20656)). It is pretty obvious that one should be able to execute the following statement in one go ("one go" = one press of a "run" button in YDB UI):
```sql
-- create view NecessaryInnerView with (security_invoker = true) as select 1;
-- create view ContextTestingView with (security_invoker = true) as select * from `/local/NecessaryInnerView`; -- where `/local/...` is your cluster name

drop view ContextTestingView;
create view ContextTestingView with (security_invoker = true) as select * from `/local/NecessaryInnerView`;
```

However, executing both drop view and create view in one go currently produces and error:
```
DropObject is not yet implemented for intent determination transformer
```
which indicates that the context of the inner query:
```sql
select * from `/local/NecessaryInnerView`
```
is polluted by the previous:
```sql
drop view ContextTestingView;
```
statement.

This problem is fixed by using an empty parsing context for parsing view's inner select statement during handling of CREATE VIEW statement.

### DIFF
--- a/ydb/core/kqp/ut/view/view_ut.cpp
+++ b/ydb/core/kqp/ut/view/view_ut.cpp
@@ -316,6 +316,24 @@ Y_UNIT_TEST_SUITE(TKQPViewTest) {
             UNIT_ASSERT(dropResult.GetIssues().ToString().Contains("Error: Path does not exist"));
         }
     }
+
+    Y_UNIT_TEST(ContextPollution) {
+        TKikimrRunner kikimr(TKikimrSettings().SetWithSampleTables(false));
+        EnableViewsFeatureFlag(kikimr);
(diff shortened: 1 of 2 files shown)


jepett0:
Couldn't get proper links to the test results of the linux-x86_64-relwithdebinfo build in testmo, so I decided to drop this pull request and open [a new one](https://github.com/ydb-platform/ydb/pull/2595).

PR #2663, opened 2024-03-12
implement ICodec and its support in clients

ildar-khisambeev:
### Changelog entry 

support ability to expand codec set

### Changelog category 

* New feature


alexnick88 (dismissed):

ydb/public/sdk/cpp/client/ydb_persqueue_core/impl/read_session.h
```diff
@@ -1029,6 +1041,13 @@ class TSingleClusterReadSessionImpl : public NPersQueue::TEnableSelfContext<TSin
         return Log;
     }
 
+    const NTopic::ICodec* TryGetCodecImpl(ECodecAlias<UseMigrationProtocol> codecId) const {
```
qyryq, line 1044: Just a thought. `TryGetSomething` function sounds like a "safe" version of `GetSomething` that returns an empty result in case of a failure. And only `GetSomething` throws an exception.

alexnick88 (approved):

PR #2678, opened 2024-03-12
Enable quoting disk space usage by storage pool kind

jepett0:
KIKIMR-18302

Users pay differently for HDD and SSD storage. They create storage pools differentiated by the underlying storage kind for their databases. Moreover, they can specify the preferred storage kind for each column in a table (see [column groups](https://ydb.tech/docs/en/yql/reference/syntax/create_table#column-family) in the docs for the CREATE TABLE statement).

However, up until this PR they didn't know, how much storage was used on each of the storage pool kinds. (And we didn't have storage pool kinds quotas to disable writes to the database, which exceeded the limit on one of its storage pools.)

We would like to provide users with an aggregate of the disk space usage of the database so they can order additional disks before the space is physically depleted. This is done by aggregating the [by channel disk space usage statistics](https://github.com/ydb-platform/ydb/blob/7a673cf01feefbe95bf5e7396d9179a5f283aeba/ydb/core/protos/table_stats.proto#L57) that the SchemeShard receives from the data shards (see [TEvPeriodicTableStats](https://github.com/ydb-platform/ydb/blob/7a673cf01feefbe95bf5e7396d9179a5f283aeba/ydb/core/protos/tx_datashard.proto#L789)). Channels are mapped to the corresponding storage pool kinds via the information that the SchemeShard has about the database (in code databases are subdomains) and the storage pools it was created with. Aggregation is done on two levels: by tables and by database. Aggregate by the table path can be seen in the UI in the path description of the table under the Describe -> PathDescription -> TableStats -> StoragePools field. Aggregate by the database can be seen in the UI in the Describe -> PathDescription -> DomainDescription -> DiskSpaceUsage -> StoragePoolsUsage field.

In addition, we implemented "storage_pools_quotas" that the user can specify in the "DatabaseQuotas" section of the config of the database that the user would like to create. There are 3 parameters in each [storage pool quota](https://github.com/jepett0/ydb/blob/a19c3b4dcc28fb1da6d04ecfb139ffdfe90c72fb/ydb/public/api/protos/ydb_cms.proto#L98):
- pool kind,
- hard quota (if any storage pool exceeds its hard quota, writes to the **whole** database (not just the storage pool that has exceeded the quota!) are restricted),
- soft quota (if all storage pools use less storage than the corresponding soft quota, then the database opens for writing again).

"storage_pools_quotas" can be used together with the existing ["data_size_hard_quota"](https://github.com/jepett0/ydb/blob/a19c3b4dcc28fb1da6d04ecfb139ffdfe90c72fb/ydb/public/api/protos/ydb_cms.proto#L82) and ["data_size_soft_quota"](https://github.com/jepett0/ydb/blob/a19c3b4dcc28fb1da6d04ecfb139ffdfe90c72fb/ydb/public/api/protos/ydb_cms.proto#L88) that do not differentiate between storage pools. Exceedance of __any__ hard quota (either the storage pool one, or the entire "data_size_hard_quota") disables writes to the database. To reenable writes, __all__ disk space usage (either the [storage pool one](https://github.com/jepett0/ydb/blob/a19c3b4dcc28fb1da6d04ecfb139ffdfe90c72fb/ydb/core/tx/schemeshard/schemeshard_info_types.h#L1460), or the aggregated [TotalSize](https://github.com/jepett0/ydb/blob/a19c3b4dcc28fb1da6d04ecfb139ffdfe90c72fb/ydb/core/tx/schemeshard/schemeshard_info_types.h#L1452)) must be below the corresponding soft quota.

One important thing to note about the storage pools usage statistics is that it is delivered to the SchemeShard with a considerable delay (about 1 minute). This means that the storage pools usage will be checked against the storage pools quotas with a delay and some data can be written above the hard limit. (And the other way around too: deleting some data to open the database for writes will be noticed by the SchemeShard with a considerable delay (about 420 seconds in my tests with a default compaction policy, I don't know where this number comes from). This is due to the fact that the new data is stored in the LSM tree (I guess) and is written to the appropriate storage pool later, after compaction.

### DIFF
--- a/ydb/core/cms/console/console__create_tenant.cpp
+++ b/ydb/core/cms/console/console__create_tenant.cpp
@@ -270,7 +270,26 @@ public:
             auto hardQuota = quotas.data_size_hard_quota();
             auto softQuota = quotas.data_size_soft_quota();
             if (hardQuota && softQuota && hardQuota < softQuota) {
-                return Error(Ydb::StatusIds::BAD_REQUEST, "Data size soft quota cannot be larger than hard quota", ctx);
+                return Error(Ydb::StatusIds::BAD_REQUEST,
+                    TStringBuilder() << "Overall data size soft quota (" << softQuota << ")"
+                                     << " of the database " << path
+                                     << " must be smaller than the hard quota (" << hardQuota << ")",
+                    ctx
+                );
+            }
+            for (const auto& storageQuota : quotas.storage_quotas()) {
+                const auto unitHardQuota = storageQuota.data_size_hard_quota();
+                const auto unitSoftQuota = storageQuota.data_size_soft_quota();
+                if (unitHardQuota && unitSoftQuota && unitHardQuota < unitSoftQuota) {
+                    return Error(Ydb::StatusIds::BAD_REQUEST,
+                        TStringBuilder() << "Data size soft quota (" << unitSoftQuota << ")"
+                                         << " for a " << storageQuota.unit_kind() << " storage unit "
+                                         << " of the database " << path
+                                         << " must be smaller than the corresponding hard quota (" << unitHardQuota << ")",
+                        ctx
+                    );
+                }
+
             }
             Tenant->DatabaseQuotas.ConstructInPlace(quotas);
         }
--- a/ydb/core/protos/subdomains.proto
+++ b/ydb/core/protos/subdomains.proto
@@ -61,8 +61,17 @@ message TDiskSpaceUsage {
         optional uint64 UsedReserveSize = 4;
     }
 
+    message TStoragePoolUsage {
+        // in bytes
+        optional string PoolKind = 1;
+        optional uint64 TotalSize = 2;
+        optional uint64 DataSize = 3;
+        optional uint64 IndexSize = 4;
+    }
+
     optional TTables Tables = 1;
     optional TTopics Topics = 2;
+    repeated TStoragePoolUsage StoragePoolsUsage = 3;
 }
 
 message TDomainState {
--- a/ydb/core/protos/table_stats.proto
+++ b/ydb/core/protos/table_stats.proto
@@ -16,6 +16,15 @@ message TChannelStats {
     optional uint64 IndexSize = 3;
 }
 
+message TStoragePoolsStats {
+    message TPoolUsage {
+        optional string PoolKind = 1;
+        optional uint64 DataSize = 2;
+        optional uint64 IndexSize = 3;
+    }
+    repeated TPoolUsage PoolsUsage = 1;
+}
+
 message TTableStats {
     optional uint64 DataSize = 1; // both inMem and ondisk
     optional uint64 RowCount = 2; // both inMem and ondisk
@@ -55,4 +64,6 @@ message TTableStats {
     optional bool HasLoanedParts = 29;
 
     repeated TChannelStats Channels = 30;
+
+    optional TStoragePoolsStats StoragePools = 31;
 }
--- a/ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
@@ -515,11 +515,11 @@ void TSchemeShard::Handle(TEvDataShard::TEvPeriodicTableStats::TPtr& ev, const T
                                                      << " rowCount " << rowCount
                                                      << " cpuUsage " << tabletMetrics.GetCPU()/10000.0);
 
-    LOG_DEBUG_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
+    LOG_TRACE_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
                 "Got periodic table stats at tablet " << TabletID()
                                                      << " from shard " << datashardId
                                                      << " pathId " << pathId
-                                                     << " raw table stats:\n" << tableStats.DebugString());
+                                                     << " raw table stats:\n" << tableStats.ShortDebugString());
 
     TStatsId statsId(pathId, datashardId);
 
--- a/ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
@@ -5,6 +5,32 @@
 #include <ydb/core/base/cputime.h>
 #include <ydb/core/protos/sys_view.pb.h>
 
+namespace {
+
+THashMap<ui32, TString> MapChannelsToStoragePoolKinds(const NActors::TActorContext& ctx,
+                                                      const NKikimr::TStoragePools& pools,
+                                                      const NKikimr::TChannelsBindings& bindings
+) {
+    THashMap<TString, TString> nameToKindMap(pools.size());
+    for (const auto& pool : pools) {
+        nameToKindMap.emplace(pool.GetName(), pool.GetKind());
+    }
+    THashMap<ui32, TString> channelsMapping(bindings.size());
+    for (ui32 channel = 0u; channel < bindings.size(); ++channel) {
+        if (const auto* poolKind = nameToKindMap.FindPtr(bindings[channel].GetStoragePoolName())) {
+            channelsMapping.emplace(channel, *poolKind);
+        } else {
+            LOG_DEBUG_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
+                        "MapChannelsToStoragePoolKinds: the subdomain has no info about the storage pool named "
+                            << bindings[channel].GetStoragePoolName()
+            );
+        }
+    }
+    return channelsMapping;
+}
+
+}
+
 namespace NKikimr {
 namespace NSchemeShard {
 
@@ -93,7 +119,7 @@ public:
     void ScheduleNextBatch(const TActorContext& ctx) override;
 
     template <typename T>
-    TPartitionStats PrepareStats(const TActorContext& ctx, const T& rec) const;
+    TPartitionStats PrepareStats(const TActorContext& ctx, const T& rec, const THashMap<ui32, TString>& channelsMapping = {}) const;
 };
 
 
@@ -125,7 +151,10 @@ THolder<TProposeRequest> MergeRequest(
 }
 
 template <typename T>
-TPartitionStats TTxStoreTableStats::PrepareStats(const TActorContext& ctx, const T& rec) const {
+TPartitionStats TTxStoreTableStats::PrepareStats(const TActorContext& ctx,
+                                                 const T& rec,
+                                                 const THashMap<ui32, TString>& channelsMapping
+) const {
     const auto& tableStats = rec.GetTableStats();
     const auto& tabletMetrics = rec.GetTabletMetrics();
 
@@ -137,6 +166,18 @@ TPartitionStats TTxStoreTableStats::PrepareStats(const TActorContext& ctx, const
     newStats.IndexSize = tableStats.GetIndexSize();
     newStats.LastAccessTime = TInstant::MilliSeconds(tableStats.GetLastAccessTime());
     newStats.LastUpdateTime = TInstant::MilliSeconds(tableStats.GetLastUpdateTime());
+    for (const auto& channelStats : tableStats.GetChannels()) {
+        if (const auto* poolKind = channelsMapping.FindPtr(channelStats.GetChannel())) {
+            auto& [dataSize, indexSize] = newStats.StoragePoolsStats[*poolKind];
+            dataSize += channelStats.GetDataSize();
+            indexSize += channelStats.GetIndexSize();
+        } else {
+            LOG_DEBUG_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
+                        "PrepareStats: the subdomain has no info about the datashard's channel "
+                            << channelStats.GetChannel()
+            );
+        }
+    }
 
     newStats.ImmediateTxCompleted = tableStats.GetImmediateTxCompleted();
     newStats.PlannedTxCompleted = tableStats.GetPlannedTxCompleted();
@@ -205,6 +246,12 @@ bool TTxStoreTableStats::PersistSingleStats(const TPathId& pathId,
     }
 
     TShardIdx shardIdx = Self->TabletIdToShardIdx[datashardId];
+    const auto* shardInfo = Self->ShardInfos.FindPtr(shardIdx);
+    Y_ABORT_UNLESS(shardInfo, "ShardInfos does not know about the shardIdx returned by TabletIdToShardIdx.");
+    auto subDomainInfo = Self->ResolveDomainInfo(pathId);
+    const auto channelsMapping = MapChannelsToStoragePoolKinds(ctx,
+                                                               subDomainInfo->EffectiveStoragePools(),
+                                                               shardInfo->BindedChannels);
 
     LOG_DEBUG_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
                 "TTxStoreTableStats.PersistSingleStats: main stats from"
@@ -212,7 +259,7 @@ bool TTxStoreTableStats::PersistSingleStats(const TPathId& pathId,
                     << ", pathId: " << pathId << ", pathId map=" << Self->PathsById[pathId]->Name
                     << ", is column=" << isColumnTable << ", is olap=" << isOlapStore);
 
-    const TPartitionStats newStats = PrepareStats(ctx, rec);
+    const TPartitionStats newStats = PrepareStats(ctx, rec, channelsMapping);
 
     LOG_INFO_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
                "Add stats from shard with datashardId(TabletID)=" << datashardId << ", pathId " << pathId.LocalPathId
@@ -290,7 +337,6 @@ bool TTxStoreTableStats::PersistSingleStats(const TPathId& pathId,
 
     if (updateSubdomainInfo) {
         auto subDomainId = Self->ResolvePathIdForDomain(pathId);
-        auto subDomainInfo = Self->ResolveDomainInfo(pathId);
         subDomainInfo->AggrDiskSpaceUsage(Self, newAggrStats, oldAggrStats);
         if (subDomainInfo->CheckDiskSpaceQuotas(Self)) {
             Self->PersistSubDomainState(db, subDomainId, *subDomainInfo);
@@ -468,6 +514,12 @@ void TSchemeShard::Handle(TEvDataShard::TEvPeriodicTableStats::TPtr& ev, const T
                                                      << " dataSize " << dataSize
                                                      << " rowCount " << rowCount
                                                      << " cpuUsage " << tabletMetrics.GetCPU()/10000.0);
+
+    LOG_DEBUG_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
+                "Got periodic table stats at tablet " << TabletID()
+                                                     << " from shard " << datashardId
+                                                     << " pathId " << pathId
+                                                     << " raw table stats:\n" << tableStats.DebugString());
 
     TStatsId statsId(pathId, datashardId);
 
--- a/ydb/core/tx/schemeshard/schemeshard_info_types.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_info_types.cpp
@@ -19,6 +19,40 @@
 
 #include <util/generic/algorithm.h>
 
+namespace {
+
+using TDiskSpaceQuotas = NKikimr::NSchemeShard::TSubDomainInfo::TDiskSpaceQuotas;
+using TQuotasPair = TDiskSpaceQuotas::TQuotasPair;
+using TStoragePoolUsage = NKikimr::NSchemeShard::TSubDomainInfo::TDiskSpaceUsage::TStoragePoolUsage;
+
+enum class EDiskUsageStatus {
+    AboveHardQuota,
+    InBetween,
+    BelowSoftQuota,
+};
+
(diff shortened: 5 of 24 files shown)


ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
```diff
@@ -5,6 +5,23 @@
 #include <ydb/core/base/cputime.h>
 #include <ydb/core/protos/sys_view.pb.h>
 
+namespace {
+
+TVector<TString> MapChannelsToStoragePoolKinds(const NKikimr::TStoragePools& pools, const NKikimr::TChannelsBindings& bindings) {
```
jepett0, line 10: @ijon Is it ok to calculate this mapping each time we receive TEvPeriodicTableStats? (I think yes, because there is usually only a handful of storage pools in a database and channels in a datashard.) The other option I see is to save it in TSubDomainInfo in SchemeShard.

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
```diff
@@ -5,6 +5,23 @@
...
+    THashMap<TString, TString> nameToKindMap(pools.size());
+    for (const auto& pool : pools) {
+        nameToKindMap.emplace(pool.GetName(), pool.GetKind());
+    }
+    TVector<TString> channelsMapping;
+    channelsMapping.reserve(bindings.size());
+    for (const auto& binding : bindings) {
+        channelsMapping.emplace_back(nameToKindMap.at(binding.GetStoragePoolName()));
```
jepett0, line 18: @snaury Could it happen that we have received a channels' binding of some datashard that contains some storage pool name that is not present in the database storage pools array? Is it ok to leave .at (and potentially throw an error) here?

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
```diff
@@ -137,6 +157,12 @@ TPartitionStats TTxStoreTableStats::PrepareStats(const TActorContext& ctx, const
     newStats.IndexSize = tableStats.GetIndexSize();
     newStats.LastAccessTime = TInstant::MilliSeconds(tableStats.GetLastAccessTime());
     newStats.LastUpdateTime = TInstant::MilliSeconds(tableStats.GetLastUpdateTime());
+    for (const auto& channelStats : tableStats.GetChannels()) {
+        const auto& poolKind = channelsMapping.at(channelStats.GetChannel());
```
jepett0, line 161: @snaury Could channelStats.GetChannel() be different from the index of the channel in the tableStats.GetChannels()? If so, could .GetChannel() be out of bounds of the Channels array?

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
```diff
@@ -205,14 +231,17 @@ bool TTxStoreTableStats::PersistSingleStats(const TPathId& pathId,
     }
 
     TShardIdx shardIdx = Self->TabletIdToShardIdx[datashardId];
+    const auto& shardInfo = Self->ShardInfos[shardIdx];
+    auto subDomainInfo = Self->ResolveDomainInfo(pathId);
+    const auto channelsMapping = MapChannelsToStoragePoolKinds(subDomainInfo->EffectiveStoragePools(), shardInfo.BindedChannels);
```
jepett0, line 236: @snaury Is shardInfo.BindedChannels the most up-to-date information on the binding of the channels to the particular storage pool? Does it follow the Hive's balancing of the channels by storage pools based on usage?

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
```diff
@@ -205,14 +231,17 @@ bool TTxStoreTableStats::PersistSingleStats(const TPathId& pathId,
     }
 
     TShardIdx shardIdx = Self->TabletIdToShardIdx[datashardId];
+    const auto& shardInfo = Self->ShardInfos[shardIdx];
```
jepett0, line 234: @ijon What if the datashard has migrated since the TEvPeriodicTableStats message was sent? Will we **always** get the correct shardInfo here?

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -46,6 +80,111 @@ void TSubDomainInfo::ApplyAuditSettings(const TSubDomainInfo::TMaybeAuditSetting
...
+    };
+
+    auto quotas = GetDiskSpaceQuotas();
+    if (!quotas) {
+        return changeSubdomainState(EDiskUsageStatus::BelowSoftQuota);
+    }
+
+    ui64 totalUsage = TotalDiskSpaceUsage();
+    const auto storagePoolsUsageStatus = CheckStoragePoolsQuotas(DiskSpaceUsage.Tables.StoragePoolsUsage, quotas.StoragePoolsQuotas);
+    if (quotas.HardQuota && totalUsage > quotas.HardQuota
```
jepett0, line 142: I hope this condition is easy to understand. In other words, we want to account for both:
1) quotas.HardQuota
and
2) storage pools quotas.
If any one of them is present, we check, if it was exceeded.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -46,6 +80,111 @@ void TSubDomainInfo::ApplyAuditSettings(const TSubDomainInfo::TMaybeAuditSetting
...
+            || !quotas.StoragePoolsQuotas.empty() && storagePoolsUsageStatus == EDiskUsageStatus::AboveHardQuota) {
+        return changeSubdomainState(EDiskUsageStatus::AboveHardQuota);
+    }
+
+    if ((!quotas.SoftQuota || totalUsage < quotas.SoftQuota)
```
jepett0, line 147: **All** usage (both totalUsage and storagePoolUsage) should be below the corresponding soft quota for the database to be opened for writing again.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -1436,6 +1575,30 @@ void TAggregatedStats::UpdateShardStats(TShardIdx datashardIdx, const TPartition
...
+            aggregatedStoragePoolStats->DataSize += newStoragePoolStats.DataSize - (oldStoragePoolStats ? oldStoragePoolStats->DataSize : 0u);
+            aggregatedStoragePoolStats->IndexSize += newStoragePoolStats.IndexSize - (oldStoragePoolStats ? oldStoragePoolStats->IndexSize : 0u);
+        } else {
+            Aggregated.StoragePoolsStats.emplace(poolKind, newStoragePoolStats);
+        }
+    }
+    for (const auto& [poolKind, oldStoragePoolStats] : oldStats.StoragePoolsStats) {
+        if (const auto* newStoragePoolStats = newStats.StoragePoolsStats.FindPtr(poolKind); !newStoragePoolStats) {
+            auto* aggregatedStoragePoolStats = Aggregated.StoragePoolsStats.FindPtr(poolKind);
+            Y_ENSURE(aggregatedStoragePoolStats); // How else could it be present in the oldStats?
```
jepett0, line 1594: @ijon Is it ok to potentially throw an error here? (I think this condition should never be violated.)

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
snaury: You can't save it in TSubDomainInfo, since TChannelsBinding is different for each tablet.

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
snaury: We don't support removing storage pools right now, and schemeshard should only add storage pools that it knows about. However I'm not sure what could happen in some legacy databases, where `/Root` doesn't have any storage pools, but where it has tables. I suspect such tablets wouldn't have channel bindings in such cases, but I'm not sure, they could maybe have storage pool copied from `channels.txt`. I would prefer you don't crash when possible, and simply don't count such unknown channels towards per-pool-kind quota.

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
snaury: Probably not, we don't support removing channels, so channel bindings may grow over time, but never shrink. However please note that new schemeshard generation may execute an alter that reconfigures channels, and datashard may theoretically connect to older schemeshard generation which doesn't know about new channels. Please handle any mismatches gracefully without crashing.

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
snaury: As far as I know it is the most up to date channel bindings, i.e. when altering table's column families we change `shardInfo.BindedChannels` when we prepare the operation, so before it is applied. However please note that older tablet generations may have stale info in memory.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -19,6 +19,40 @@
...
+                                         const THashMap<TString, TQuotasPair>& storagePoolsQuotas
+) {
+    bool softQuotaExceeded = false;
+    for (const auto& [poolKind, usage] : storagePoolsUsage) {
+        if (const auto* quota = storagePoolsQuotas.FindPtr(poolKind)) {
+            const auto totalSize = usage.DataSize + usage.IndexSize;
+            if (totalSize > quota->HardQuota) {
+                return EDiskUsageStatus::AboveHardQuota;
+            }
+            if (totalSize > quota->SoftQuota) {
```
snaury, line 44: It should be `totalSize >= quota->SoftQuota` for consistency with total quota. For regular quota we require usage to drop _below_ soft quota for the database to become read-write again. You also may need to handle the case where `SoftQuota == 0` (i.e. there is no soft quota and database becomes read-write when not exceeding hard quota).

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -19,6 +19,40 @@
...
+};
+
+EDiskUsageStatus CheckStoragePoolsQuotas(const THashMap<TString, TStoragePoolUsage>& storagePoolsUsage,
+                                         const THashMap<TString, TQuotasPair>& storagePoolsQuotas
+) {
+    bool softQuotaExceeded = false;
+    for (const auto& [poolKind, usage] : storagePoolsUsage) {
+        if (const auto* quota = storagePoolsQuotas.FindPtr(poolKind)) {
+            const auto totalSize = usage.DataSize + usage.IndexSize;
+            if (totalSize > quota->HardQuota) {
```
snaury, line 41: Should only be applied when `HardQuota > 0`. We use 0 to mean "no quota".

ydb/core/protos/table_stats.proto
```diff
@@ -55,4 +55,11 @@ message TTableStats {
     optional bool HasLoanedParts = 29;
 
     repeated TChannelStats Channels = 30;
+
+    message TStoragePoolStats {
+        optional string PoolKind = 1;
+        optional uint64 DataSize = 2;
+        optional uint64 IndexSize = 3;
+    }
+    repeated TStoragePoolStats StoragePools = 31;
```
jepett0, line 64: TO DO: add a comment stating that this field was added solely for the informative PathDescription in the UI.

ydb/public/api/protos/ydb_cms.proto
```diff
@@ -94,6 +94,13 @@ message DatabaseQuotas {
     // A minimum value of `TtlSettings.run_interval_seconds` that can be specified.
     // Default is 1800 (15 minutes).
     uint32 ttl_min_run_internal_seconds = 4;
+    
+    message StoragePoolQuotas {
+        string storage_pool_kind = 1;
+        uint64 data_size_hard_quota = 2;
+        uint64 data_size_soft_quota = 3;
+    }
+    repeated StoragePoolQuotas storage_pools_quotas = 6;
```
snaury, line 103: The name `storage_pools` is misleading, since we don't have storage pools in the public api, we have "storage units". And it's not even about storage pools, but about storage pool kinds. I think it should be simplified to `repeated StorageQuotas storage_quotas` and `unit_kind` instead of `storage_pool_kind` (consistent naming with `StorageUnits`).

jepett0:
TO DO: add a check to the alter (and create too?) subdomain in SchemeShard that the storage pool kinds used in the DatabaseQuotas are present in the StoragePools of this database.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
jepett0: TO DO: change the condition to be `totalSize >= quota->SoftQuota`.

As for the case when the `SoftQuota == 0`, we handle it in the same manner as the regular quotas do, i.e. [we set](https://github.com/jepett0/ydb/blob/a19c3b4dcc28fb1da6d04ecfb139ffdfe90c72fb/ydb/core/tx/schemeshard/schemeshard_info_types.cpp#L102) it to be equal to the `HardQuota` if it wasn't set by the user.

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
jepett0: @snaury Could it be that the channels of a datashard are reconfigured in such a way that the storage pool will have the same name as before, but a different kind? Or could it happen that the storage pool name is also changed in TChannelsBindings, but to some other storage pool that has a different kind than this channel had before?

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
jepett0: @snaury What do you mean by the "older tablet generations may have stale info in memory"? Does it mean that an old SchemeShard may have an outdated `shardInfo`? Is it going to be updated? Are there some messages that data shards send to SchemeShard to push updates of theirs infos?

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
jepett0: Discussed this privately with @snaury and I believe the answer is yes, channels can be reconfigured to use a different storage pool kind. However, currently nobody does this and channels are only added if a new configuration is needed, not changed.

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
jepett0: Discussed this privately with @snary and I believe the answer is yes, SchemeShard can have an outdated shardInfo, if it is on older generation tablet. However, such a tablet will not be able to change any object in the scheme, so the database will not be blocked for writing if the old SchemeShard decides that it has exceeded its quota based on the outdated information the old SchemeShard has on the channelsBinding. The worst thing that could happen is that the old SchemeShard will aggregate some table usage statistics that does not correspond to the real picture. However, this should not have any real impact on the cluster.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
jepett0: Changed the code to be more explicit and added a comment.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
jepett0: Changed the code to be more explicit.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
jepett0: Changed in the newer revision.

ydb/public/api/protos/ydb_cms.proto
jepett0: Changed to the proposed spelling in the newer revision.

jepett0:
TO DO: persist storage pool usage stats

ijon (changes requested):
More explicit tests on the permutation cases of (database, pool, hard, soft quotas, defined, not defined) would be nice

ydb/core/tx/schemeshard/ut_subdomain/ut_subdomain.cpp
```diff
@@ -2979,6 +3044,213 @@ Y_UNIT_TEST_SUITE(TSchemeShardSubDomainTest) {
...
+        }
+        CheckQuotaExceedance(runtime, "/MyRoot/SomeTenant", true, "after the compaction");
+
+        // step 5: drop the table
+        TestDropTable(runtime, ++txId, "/MyRoot/SomeTenant", "SomeTable");
+        env.TestWaitNotification(runtime, txId);
+        CheckQuotaExceedance(runtime, "/MyRoot/SomeTenant", false, "after the drop of the table");
+    }
+
+    Y_UNIT_TEST(StoragePoolsQuotas_ExternalSubdomain) {
```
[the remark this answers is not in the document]
jepett0: done ✅

ydb/core/cms/console/console__create_tenant.cpp
```diff
@@ -269,7 +269,13 @@ class TTenantsManager::TTxCreateTenant : public TTransactionBase<TTenantsManager
...
             auto softQuota = quotas.data_size_soft_quota();
-            if (hardQuota && softQuota && hardQuota < softQuota) {
+            if (hardQuota && softQuota && hardQuota < softQuota
+                    || AnyOf(quotas.storage_quotas(), [](const auto& storageQuota) {
+                        const auto hardQuota = storageQuota.data_size_hard_quota();
+                        const auto softQuota = storageQuota.data_size_soft_quota();
+                        return hardQuota && softQuota && hardQuota < softQuota;
+                    })
+            ) {
                 return Error(Ydb::StatusIds::BAD_REQUEST, "Data size soft quota cannot be larger than hard quota", ctx);
```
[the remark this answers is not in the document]
jepett0: Made the error message more verbose and mentioning the exact unit_kind of the storage quota that is malformed.

ydb/core/protos/subdomains.proto
```diff
@@ -51,6 +51,14 @@ message TDiskSpaceUsage {
...
         optional uint64 DataSize = 2;
         optional uint64 IndexSize = 3;
+
+        message TStoragePoolUsage {
+            optional string PoolKind = 1;
+            optional uint64 TotalSize = 2;
+            optional uint64 DataSize = 3;
+            optional uint64 IndexSize = 4;
+        }
+        repeated TStoragePoolUsage StoragePoolsUsage = 4;
```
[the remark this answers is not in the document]
jepett0: Changed the position of the storage pools usage statistics in the subdomain description

ydb/core/tx/schemeshard/schemeshard_info_types.h
```diff
@@ -1446,6 +1452,12 @@ struct TSubDomainInfo: TSimpleRefCount<TSubDomainInfo> {
             ui64 TotalSize = 0;
             ui64 DataSize = 0;
             ui64 IndexSize = 0;
+            
+            struct TStoragePoolUsage {
+                ui64 DataSize = 0;
+                ui64 IndexSize = 0;
+            };
+            THashMap<TString, TStoragePoolUsage> StoragePoolsUsage;
```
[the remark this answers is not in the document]
jepett0: Fixed

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
```diff
@@ -471,6 +517,12 @@ void TSchemeShard::Handle(TEvDataShard::TEvPeriodicTableStats::TPtr& ev, const T
                                                      << " rowCount " << rowCount
                                                      << " cpuUsage " << tabletMetrics.GetCPU()/10000.0);
 
+    LOG_DEBUG_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
+                "Got periodic table stats at tablet " << TabletID()
+                                                     << " from shard " << datashardId
+                                                     << " pathId " << pathId
+                                                     << " raw table stats:\n" << tableStats.DebugString());
+
```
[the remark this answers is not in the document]
jepett0: It was helpful during the development and I think, there is no such message in the debug logs now. I would like to leave it. It is debug level log, so it should not bother others much

ydb/public/api/protos/ydb_cms.proto
```diff
@@ -94,6 +94,14 @@ message DatabaseQuotas {
     // A minimum value of `TtlSettings.run_interval_seconds` that can be specified.
     // Default is 1800 (15 minutes).
     uint32 ttl_min_run_internal_seconds = 4;
+    
+    message StorageQuotas {
+        // in theory an arbitrary string, but in practice HDD or SSD
```
snaury, line 99: Только `"hdd" or "ssd"`, строчными буквами обычно.

snaury (commented):
Mostly LGTM.

ydb/public/api/protos/ydb_cms.proto
jepett0: Это же акроним. Мне кажется более правильным писать заглавными

jepett0: ааа... ты имеешь ввиду, что надо в кавычках записать. Это, пожалуй, сделаю

ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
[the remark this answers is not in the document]
jepett0: TO DO: change level to TRACE and make it ShortDebugString() to make it one line.

ydb/public/api/protos/ydb_cms.proto
jepett0: TO DO: should really be "hdd" or "ssd" to emphasize the real values of these strings that users can see in practice.

ijon (approved):

PR #2708, opened 2024-03-13
BTreeIndex Tests ON

### LINKED ISSUE #2547: Fix all the test with `EnableLocalDBFlatIndex = false` `EnableLocalDBBtreeIndex = true`

kungasc:
### Changelog entry 

...

### Changelog category 

### Additional information

Verified that almost all the test work with `WriteFlatIndex = false` and `EnableLocalDBFlatIndex = false` settings (except the test that check default values or check flat index behaviour). Then switched it off back.


snaury (approved):
LGTM, но я не понял что означает название PR.

PR #2782, opened 2024-03-14
(refactoring) Simplify child actors creation

CyberROFL:
### Changelog entry 

...

### Changelog category 

### Additional information

...


Enjection (approved):
Seems ok to me, but I would replace `ui64 tid, rid, id` to something more meaningful or at least strongly typed.

PR #2804, opened 2024-03-15
secure fatal operations

### LINKED ISSUE #2770: secure potentialy fatal operations in hive monitoring
we need to make sure that nobody could mistakenly call a unrevertable destroy operations

adameat:
secure fatal hive operations, make them harder to execute
closes #2770 

* Improvement

### DIFF
--- a/ydb/core/mind/hive/hive_impl.h
+++ b/ydb/core/mind/hive/hive_impl.h
@@ -243,6 +243,7 @@ protected:
 
     friend struct TStoragePoolInfo;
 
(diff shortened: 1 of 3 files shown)


vporyadke:
Maybe we should also refuse queries that set both (ownerid, owneridx) and tabletid?

adameat:
> Maybe we should also refuse queries that set both (ownerid, owneridx) and tabletid?

the intention of this PR is to make harder to call these two particular handlers

vporyadke (approved):

PR #2836, opened 2024-03-15
YQL-18027: Add block implementation for Remove UDFs

igormunkin:
Within this changeset `String::RemoveAll`, `String::RemoveFirst` and `String::RemoveLast` UDFs are rewritten to provide both scalar and block implementations.

It's worth to mention, that all `String::Remove*` functions work with bytes (i.e. ASCII charset), so using `std::unordered_set<char>` here looks excessive in terms of performance. Hence, the lookup for the chars to be removed from the given string is rewritten via `std::array`.

### Changelog category

* Improvement


ydb/library/yql/udfs/common/string/string_udf.cpp
```diff
@@ -266,7 +266,8 @@ namespace {
         return args[0];
     }
 
-    SIMPLE_STRICT_UDF(TRemoveAll, char*(TAutoMap<char*>, char*)) {
+
+    BEGIN_SIMPLE_STRICT_ARROW_UDF(TRemoveAll, char*(TAutoMap<char*>, char*)) {
         std::string input(args[0].AsStringRef());
         const std::string_view remove(args[1].AsStringRef());
```
vitstn, line 272: please use std::array<char, 256> for membership testing

vitstn (approved):

Issue #2856, opened 2024-03-17
SIGSEGV in S3 export on some systems


zinal:
On some systems, S3 export leads to SIGSEGV in YDB dynamic nodes with the following stack:

```
Thread 72 "ydbd-deb.Batch" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffd91bb700 (LWP 2631261)]
0x00000000571c5980 in tcmalloc::tcmalloc_internal::Static::cpu_cache_ ()
(gdb) bt
#0  0x00000000571c5980 in tcmalloc::tcmalloc_internal::Static::cpu_cache_ ()
#1  0x0000000047112fa4 in open_udp_socket (channel=0x44a790e4900, server=0x44af71e9800)
    at /home/conan/w/prod-v1/bsr/75071/dffdc/.conan/data/c-ares/1.19.1/_/_/build/61f6486eafa3466332751467b79fdc604dc107fc/src/src/lib/ares_process.c:1277
#2  0x000000004711244c in ares__send_query (channel=0x44a790e4900, query=0x44af5e81040, now=0x7fffd9183cd8)
    at /home/conan/w/prod-v1/bsr/75071/dffdc/.conan/data/c-ares/1.19.1/_/_/build/61f6486eafa3466332751467b79fdc604dc107fc/src/src/lib/ares_process.c:876
#3  0x0000000047116f55 in ares_send (channel=0x44a790e4900, qbuf=0x44a84904aa0 "\n'\001", qlen=41, callback=0x471153c0 <qcallback>, arg=0x44af63dd210)
    at /home/conan/w/prod-v1/bsr/75071/dffdc/.conan/data/c-ares/1.19.1/_/_/build/61f6486eafa3466332751467b79fdc604dc107fc/src/src/lib/ares_send.c:128
#4  0x000000004711535a in ares_query (channel=0x44a790e4900, name=0x44af428fc40 "storage.yandexcloud.net", dnsclass=1, type=1, callback=0x47107570 <host_callback>, arg=0x44af7331c80)
    at /home/conan/w/prod-v1/bsr/75071/dffdc/.conan/data/c-ares/1.19.1/_/_/build/61f6486eafa3466332751467b79fdc604dc107fc/src/src/lib/ares_query.c:103
#5  0x000000004710702c in next_dns_lookup (hquery=0x44af7331c80)
[...]

PR #2857, opened 2024-03-17
#2856 avoid non-initialized data in ares_options

zinal:
### Changelog entry 

`Curl_resolver_init()` function does not properly initialize the `ares_options` structure before passing it to `ares_init_options()`

### Changelog category 

* Bugfix 

### Additional information

See https://github.com/ydb-platform/ydb/issues/2856

...


zinal:
Seems that the fix should be different - two different Ares versions are comint into the build, which causes the issue.

PR #2860, opened 2024-03-17
Run & stop workers

CyberROFL:
### Changelog entry 

...

### Changelog category 

### Additional information

...


ydb/core/tx/replication/controller/controller.cpp
```diff
@@ -213,7 +215,7 @@ void TController::CreateSession(ui32 nodeId, const TActorContext& ctx) {
         << ": nodeId# " << nodeId);
 
     Y_ABORT_UNLESS(!Sessions.contains(nodeId));
-    Sessions.emplace(nodeId, TSessionInfo{});
+    Sessions.emplace(nodeId, TSessionInfo());
```
Enjection, line 218: If you are calling default constructor, maybe better just Sessions.emplace(nodeId). And instead of contains + emplace, you could use result of emplace to check either you emplaced new value or it was already there.

Enjection (approved):

PR #3013, opened 2024-03-20
Fix NodeWarden caching error

alexvru:
### Changelog entry 

Fix NodeWarden caching error

### Changelog category 

* Bugfix 

### Additional information

Fix NodeWarden caching error leading to impossibility to use configuration caching mechanism allowing storage operation without immediate BSC availability.


the-ancient-1 (approved):
LGTM

PR #3064, opened 2024-03-21
Fix update token interval calculation.

dcherednik:
We should chouse min value from expiresIn and RefreshPeriod_, not max one

### Changelog category 

* Bugfix 

### Additional information

...


adameat (approved):

dcherednik:
We should chouse min value from expiresIn and RefreshPeriod_, not max one
Also reduce server returned interval by 2 to prevent race at expiring time.

PR #3076, opened 2024-03-22
Fix bugs/flaws in harness library

maximyurchuk:
### Changelog entry 
Several fixes and flaws were fixed in harness library

### Changelog category 

* Improvement
* Bugfix 

### Additional information:
- use scp with same options as ssh
- don't ignore exceptions in  parallel execution (see `_run_on` function)
- trace process'es output as string not as binary (`process output` instead of `b'process output'`)
- don't show stack traces in case of `raise_on_error=False`
- Use `LogLevel=ERROR` in order to show less spam in logs from ssh/scp

### DIFF
--- a/ydb/tests/library/harness/daemon.py
+++ b/ydb/tests/library/harness/daemon.py
@@ -225,6 +225,7 @@ class ExternalNodeDaemon(object):
     def __init__(self, host):
         self._host = host
         self._ssh_username = param_constants.ssh_username
+        self._ssh_options = ["-o", "UserKnownHostsFile=/dev/null", "-o", "StrictHostKeyChecking=no", "-o", "LogLevel=ERROR"]
         self.logger = logger.getChild(self.__class__.__name__)
 
     @property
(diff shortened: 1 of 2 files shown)


ydb/tests/library/harness/daemon.py
```diff
@@ -253,10 +252,10 @@ def copy_file_or_dir(self, file_or_dir, target_path):
 
         transit_path = os.path.join(self._artifacts_path, os.path.basename(file_or_dir))
 
-        self.ssh_command(['sudo', 'rm', '-fr', target_path], raise_on_error=True)
+        self.ssh_command(['sudo', 'rm', '-rf', target_path], raise_on_error=True)
 
         self._run_in_subprocess(
-            ["scp", '-r', file_or_dir, self._path_at_host(transit_path)],
+            ["scp"] + self._ssh_options + [ '-r', file_or_dir, self._path_at_host(transit_path)],
```
maximyurchuk, line 258: Without that options there was prompt from scp (`Are you sure you want to continue connecting ...`) and the test process was hung.

gridnevvvit (approved):

PR #3100, opened 2024-03-22
YQL-18027: Add block implementation for padding formatters

igormunkin:
Apart from the block implementation for padding formatters (i.e. `String::LeftPad` and `String::RightPad` UDFs), there are several changes extending or enhancing Arrow UDFs creation.

---

The first patch makes the `SimpleSignatureType` builder handle the optional parameters and proxy it to the callableBuilder instance.

Before the patch, callableBuilder instance was generated without optional arguments even though the callable to be built via the particular FunctionTypeInfoBuilder specified the amount of its optionals.

Though the information about the optional parameters can be retrieved from the given FunctionTypeInfoBuilder, it was decided to preserve builder routine being pure functions, so its signature is extended with the parameter to obtain the number of the optional arguments for the callable.

---

In scope of the second patch, PrepareSimpleArrowUdf helper is adjusted to provide the support for the UDF's optional arguments. Except the relaxing the couple of assertions, there are two main changes made.

Unlikely to the required arguments, the type of the omitted argument has to be passed to the specialized UDF signature builder as an original `TOptional` parameter type. At the same time, all the omitted parameter types require to be wrapped with the Block type with the Scalar shape in the UDFKernel signature).

Finally, all the omitted arguments are substituted with the Null datum value to be passed to the fixed-arg kernel in scope of the `Run` routune.

---

The last preliminary patch introduces the new macro to generate Arrow UDF with optional arguments. As a result the signature of `<BEGIN_ARROW_UDF>` macro is extended to pass the number of the optional arguments in the particular UDF to `SimpleSignatureType` helper. Other auxiliary macros, generating Arrow UDFs, are adjusted respectively.

---

### Changelog category

* Improvement


ydb/library/yql/public/udf/arrow/udf_arrow_helpers.h
```diff
@@ -123,6 +124,12 @@ class TSimpleArrowUdfImpl : public TBoxedValue {
             for (ui32 i = 0; i < ArgArrowTypes_.size(); ++i) {
                 bool isScalar;
                 ui64 length;
+                // XXX: If no value is given to the UDF, pass the Null datum as
+                // an optional argument value to the fixed-arg kernel.
+                if (!args[i].HasValue()) {
```
vitstn, line 129: if !args[i]. HasValue has different sematic for nested Optionals

igormunkin: Thanks! [Fixed](https://github.com/ydb-platform/ydb/pull/3100/files#diff-0b96ada35f11c5251304bd4786016a2fe25410a591d3c904cba44911909bb967R131), left the comment.

vitstn (approved):

PR #3120, opened 2024-03-25
actually delete nodes if possible

vporyadke:
### Changelog entry 

Delete long-time inactive nodes from Hive without waiting for Hive to restart

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/core/mind/hive/hive_impl.cpp
+++ b/ydb/core/mind/hive/hive_impl.cpp
@@ -94,6 +94,18 @@ void THive::RestartPipeTx(ui64 tabletId) {
     }
 }
 
+bool THive::TryToDeleteNode(TNodeInfo* node) {
+    if (node->CanBeDeleted()) {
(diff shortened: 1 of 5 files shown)


ydb/core/mind/hive/hive_impl.cpp
```diff
@@ -3279,6 +3293,17 @@ void THive::Handle(TEvPrivate::TEvLogTabletMoves::TPtr&) {
     TabletMovesByTypeForLog.clear();
 }
 
+void THive::Handle(TEvPrivate::TEvDeleteNode::TPtr& ev) {
+    auto node = FindNode(ev->Get()->NodeId);
+    if (node == nullptr) {
+        return;
+    }
+    node->DeletionScheduled = false;
+    if (!node->IsAlive()) {
```
adameat, line 3302: I would suggest to check last connect time instead of current Alive status

adameat (approved):

PR #3160, opened 2024-03-26
Measure huge blob size threshold

### LINKED ISSUE #3161: Measure huge blob size used for writes
We need to measure the threshold which is used for huge blobs creation. 
This will allow us to distinguish vdisks for which a new threshold is applied to monitor the rollout of the https://github.com/ydb-platform/ydb/issues/2942

qjex:


ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
```diff
@@ -2329,6 +2329,7 @@ namespace NKikimr {
             // generation independent self VDisk Id
             auto genIndSelfVDiskId = SelfVDiskId;
             genIndSelfVDiskId.GroupGeneration = -1;
+            IFaceMonGroup->MinHugeBlobInBytes() = Config->MinHugeBlobInBytes;
```
alexvru, line 2332: This value is not the actual size of a huge blob; the actual size is aligned to the append block size and can be found by the strange name MinREALHugeBlobSize (inside HugeBlobCtx).

alexvru (changes requested):

ydb/core/blobstorage/vdisk/common/vdisk_mongroups.h
```diff
@@ -517,6 +517,8 @@ public:
 
                 COUNTER_INIT_IF_EXTENDED(PutTotalBytes, true);
                 COUNTER_INIT_IF_EXTENDED(GetTotalBytes, true);
+                
+                COUNTER_INIT(MinHugeBlobInBytes, false);
```
alexvru, line 521: Counter of this type is useless. We need a histogram-like counter, with one label indicating minimum huge blob size, and the sensor value of 1 for specific vdisk.

alexvru (changes requested):

alexvru (approved):

alexvru (changes requested):
Failed tests.

alexvru (approved):

Issue #3084, opened 2024-03-22
DataShard: abort all pending volatile transactions at split start


snaury:
I'm sometimes seeing long (i.e. dozens of seconds) stretches of unavailability under Jepsen, when randomly killing dynamic nodes (even when not killing storage nodes), and I think those correspond to mass overload errors due to split being in progress (which could block for up to 30 seconds waiting until propose window expires). There's not much we could do about it in the past, but with volatile transactions we could abort all of them (that haven't executed yet) without any waiting. I suspect this should either help with unavailability, or expose other problems, if any.

PR #3175, opened 2024-03-26
24-1: Cancel waiting volatile transactions on split and schema changes. Fixes #3084.

snaury:
### Changelog entry 

Cancel waiting volatile transactions on split and schema changes.

### Changelog category 

* Improvement

### Additional information

Proposed, but not yet planned transactions currently block splits and schema changes, during which shard becomes unavailable for up to 30 seconds until those operations time out. Since volatile transactions are in-memory and don't have any obligation to other participants until they are executed, shards may cancel them at any time, which they now do after split start or schema propose. This helps with long unavailability when nodes under load are suddenly killed.

Fixes #3084.

Merge from #3136 and #3170.


snaury:
Решили подождать, пока 24.1 катится до продов, чтобы не блокировать выкатку в случае если патч вызовет какие-то проблемы.

snaury:
Версия 24-1 уже не актуальна.

PR #3412, opened 2024-04-03
YDBDOCS-650 import from JDBC docs

zinal:
### Changelog entry 

...

### Changelog category 

### Additional information

...


ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+             the schema and table name of the source table. -->
+        <table-name-format>oraimp1/${schema}/${table}</table-name-format>
+        <!-- The template used to generate the name of the supplemental table
+             to store source BLOB field's data in YDB.
+             The values ${schema}, ${table} and ${field}
+             are used for source schema, table and BLOB field names. -->
+        <blob-name-format>oraimp1/${schema}/${table}_${field}</blob-name-format>
+        <!-- Date and timestamp data type values conversion mode.
+             Possible values: DATE (use YDB Date datatype, default), INT, STR.
+             DATE does not support input values before January, 1, 1970.
```
blinkov, line 164: 1. there's also an upper bound
2. what will happen with out-of-range values?

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+        <include-schemas regexp="true">.*</include-schemas>
+        <!-- Schemas to exclude (of those included), as literal or regular expression -->
+        <exclude-schemas>SOMESCHEMA</exclude-schemas>
+        <!-- The particular tables can be included or excluded using the
+             "include-tables" and "exclude-tables" tags,
+             using literals or regular expressions as well. -->
+    </table-map>
+    <!-- Table reference may refer to the particular table in the particular schema,
+         and optionally specify the query to execute. The latter option allows to import
+         the virtual tables which do not actually exist on the data source. -->
```
blinkov, line 188:
```suggestion
        <!-- Particular tables can be included or excluded using the
             "include-tables" and "exclude-tables" tags,
             using literals or regular expressions as well. -->
    </table-map>
    <!-- Table reference may refer to a particular table in a particular schema
         and optionally specify the query to execute. The latter option allows to import
         virtual tables that do not actually exist on the data source. -->
```

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+    <table-ref options="default">
+        <schema-name>ora$sys</schema-name>
+        <table-name>all_tables</table-name>
+        <!-- If the query text is defined, it is executed as shown. -->
+        <query-text>SELECT * FROM all_tables</query-text>
+        <!-- Primary key columns should be defined for the query.  -->
```
blinkov, line 194:
```suggestion
        <!-- Primary key column names should be defined for the query result.  -->
```

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+## Import tool limitations {#limitations}
+
+Right now the tested data sources include the following:
+
+* [PostgreSQL](https://www.postgresql.org/),
+* [MySQL](https://www.mysql.com/),
+* [Oracle Database](https://www.oracle.com/database/),
+* [Microsoft SQL Server](https://www.microsoft.com/sql-server/),
+* [IBM Db2](https://www.ibm.com/products/db2),
+* [IBM Informix](https://www.ibm.com/products/informix).
```
blinkov, line 210:
```suggestion
## Import tool limitations {#limitations}

The tested data sources include the following database management systems:

* [PostgreSQL](https://www.postgresql.org/)
* [MySQL](https://www.mysql.com/)
* [Oracle Database](https://www.oracle.com/database/)
* [Microsoft SQL Server](https://www.microsoft.com/sql-server/)
* [IBM Db2](https://www.ibm.com/products/db2)
* [IBM Informix](https://www.ibm.com/products/informix)
```

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+
+Other data sources will probably work, too, as the tool uses the generic JDBC APIs to retrieve data and metadata.
+
+Secondary indexes are not imported.
+
+Some data types and table structures are known to be unsupported:
+
+* the embedded tables for Oracle Database,
+* spatial data type for Microsoft SQL Server,
+* object data types for Informix.
```
blinkov, line 220:
```suggestion
Other data sources will likely work, too, as the tool uses the generic JDBC APIs to retrieve data and metadata.

Secondary indexes are not imported.

Some data types and table structures are known to be unsupported:

* the embedded tables for Oracle Database
* spatial data type for Microsoft SQL Server
* object data types for Informix
```

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+
+In case the source table has no primary key and no unique indexes, primary key columns can be configured using the import settings in the configuration file, as a sequence of `key-column` elements in the `table-ref` section (see below the example in the description of the configuration file format).
+
+When there is no primary key defined anywhere, the tool automatically adds the column `ydb_synth_key` to the target table, and uses it as the primary key. The values of the synthetic primary key are computed as "SHA-256" hash code over the values of all columns of the row, except the columns of the BLOB type.
+
+If the input table contains several rows with completely identical values, the destination table will have only one row per each set of duplicate input rows. This also means that there the tool cannot import the table in which all columns are of BLOB type.
+
+### Importing large objects (BLOB, XML) {#blob}
```
blinkov, line 232: Some of the things mentioned in the example config are clarified here, so it makes sense to move the config to the very bottom of the article so readers can make sense of it without scrolling back and forth.

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+
+In case the source table has no primary key and no unique indexes, primary key columns can be configured using the import settings in the configuration file, as a sequence of `key-column` elements in the `table-ref` section (see below the example in the description of the configuration file format).
+
+When there is no primary key defined anywhere, the tool automatically adds the column `ydb_synth_key` to the target table, and uses it as the primary key. The values of the synthetic primary key are computed as "SHA-256" hash code over the values of all columns of the row, except the columns of the BLOB type.
+
+If the input table contains several rows with completely identical values, the destination table will have only one row per each set of duplicate input rows. This also means that there the tool cannot import the table in which all columns are of BLOB type.
```
blinkov, line 230:
```suggestion
### Handling tables without the primary key {#nopk}

Each {{ ydb-short-name }} table must have a primary key. If a primary key (or at least a unique index) is defined on the source table, the tool creates the primary key for the target  {{ ydb-short-name }} table with the columns and order determined by the original primary key. When multiple unique indexes are defined on the source table, the tool prefers the index with the smallest number of columns.

In case the source table has no primary key and no unique indexes, primary key columns can be configured using the import settings in the configuration file as a sequence of `key-column` elements in the `table-ref` section (see below the example in the description of the configuration file format).

When no primary key is defined anywhere, the tool automatically adds the column `ydb_synth_key` to the target table and uses it as the primary key. The values of the synthetic primary key are computed as "SHA-256" hash code over the values of all row cells except the columns of the `BLOB` type.

If the input table contains several rows with completely identical values, the destination table will have only one row per each set of duplicate input rows. This also means that the tool cannot import the table in which all columns are of `BLOB` type.
```

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Importing large objects (BLOB, XML) {#blob}
+
+For each BLOB field in the source database the import tool creates an additional YDB target table (BLOB supplemental table), having the following structure:
```
blinkov, line 234:
```suggestion
For each BLOB field in the source database, the import tool creates an additional {{ ydb-short-name }} target table (BLOB supplemental table) with the following schema:
```

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+    `id` Int64,
+    `pos` Int32,
+    `val` String,
+    PRIMARY KEY(`id`, `pos`)
+)
+```
+
+The name of the BLOB supplemental table is generated based on the `table-options` / `blob-name-format` setting in the configuration file.
+
+Each BLOB field value is saved as a sequence if records in the BLOB supplemental table in YQB. Actual data is stored in the `val` field, storing no more than 64K bytes. The order of blocks stored is defined by the values of the `pos` column, containing the integer values `1..N`.
```
blinkov, line 247:
```suggestion
Each BLOB field value is saved as a sequence of records in the BLOB supplemental table in {{ ydb-short-name }}. Actual data is stored in the `val` field, storing no more than 64K bytes. The order of blocks stored is defined by the values of the `pos` column, containing the integer values `1..N`.
```

blinkov, line 247: So it sounds like there's never pos=0, or is it a special value for something?

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+
+A unique value of type `Int64` is generated for each source BLOB value, and stored in the `id` field of the BLOB supplemental table.
+This identifier is also stored in the "main" target table in the field having the same name as the BLOB field in the source table.
+
+For PostgreSQL working with the `EXTENSION lo` BLOBS may require extra permissions for the account used to connect to the data source. An alternative option is to enable the compatibility mode on the database level using the following statement:
```
blinkov, line 252:
```suggestion
For each source BLOB value, a unique value of type `Int64` is generated and stored in the `id` field of the BLOB supplemental table.
This identifier is also stored in the "main" target table in a field that has the same name as the BLOB field in the source table.

For PostgreSQL, working with the `EXTENSION lo` BLOBS may require extra permissions for the account used to connect to the data source. An alternative option is to enable the compatibility mode on the database level using the following statement:
```

blinkov (changes requested):

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+4. Row data is imported by reading from the source database using the SELECT statements, and written into the target YDB database using the Bulk Upsert mechanizm.
+
+All operations, including metadata extraction from the source database, table creation (and re-creation) in the target database, and row data import, are performed in parallel mode (thread-per-table), using multiple concurrent threads and multiple open connections to both source and target databases. The maximum degree of parallelizm is configurable, although the actual number of concurrent operations cannot exceed the number of tables (or custom SQL queries) being imported.
+
+## Import tool installation {#install}
+
+To run the import tool, JDK 8 or later is required, with the `java` executable available in the program search path.
+
+Import tool binary distribution is available on the [Releases page](https://github.com/ydb-platform/ydb-importer/releases). The import tool is provided as the ZIP archive, which needs to be unpacked into the local directory on the server where the tool is required to run.
```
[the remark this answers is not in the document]
zinal: Perhaps it is a separate task?

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+1. The tool connects to the source database, and determines the list of tables and custom SQL queries to be imported.
+2. The structure of target YDB tables is generated, and optionally saved as YQL script.
+3. Target database is checked for existence of the target tables. Missing tables are created, already existing ones may be left as is, or re-created.
+4. Row data is imported by reading from the source database using the SELECT statements, and written into the target YDB database using the Bulk Upsert mechanizm.
+
+All operations, including metadata extraction from the source database, table creation (and re-creation) in the target database, and row data import, are performed in parallel mode (thread-per-table), using multiple concurrent threads and multiple open connections to both source and target databases. The maximum degree of parallelizm is configurable, although the actual number of concurrent operations cannot exceed the number of tables (or custom SQL queries) being imported.
+
+## Import tool installation {#install}
+
+To run the import tool, JDK 8 or later is required, with the `java` executable available in the program search path.
```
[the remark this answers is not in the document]
zinal: Added extra text on JDK flavour used (OpenJDK 8)

ydb/docs/en/core/integrations/import-jdbc.md
```diff
@@ -0,0 +1,256 @@
...
+
+Import tool binary distribution is available on the [Releases page](https://github.com/ydb-platform/ydb-importer/releases). The import tool is provided as the ZIP archive, which needs to be unpacked into the local directory on the server where the tool is required to run.
+
+The import tool distribution archive contains the configuration file examples as `*.xml` files, sample tool startup script `ydb-importer.sh`, and the executable code for the tool and its dependencies (including the YDB Java SDK) as `lib/*.jar` files.
+
+Before running the import tool, the required JDBC drivers should be put (as `*.jar` files) into the `lib` subdirectory of the tool installation directory.
```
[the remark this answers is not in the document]
zinal: Made text more clear in the sense that only source JDBC drivers are needed.

ydb/docs/en/core/integrations/import-jdbc.md
zinal: > So it sounds like there's never pos=0, or is it a special value for something?

Yes, zero is never used.

fomichev3000 (approved):

blinkov (approved):

PR #3415, opened 2024-04-03
[YQL-17573] Fix handling groping sets + window functions/top level dstinct

nepal:
* [YQL-17573] Fix handling groping sets + window functions/top level dstinct

### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/library/yql/sql/v1/select.cpp
+++ b/ydb/library/yql/sql/v1/select.cpp
@@ -449,8 +449,9 @@ protected:
     TColumns Columns;
 };
 
-class IComposableSource {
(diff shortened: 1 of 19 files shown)


ydb/library/yql/sql/v1/select.cpp
```diff
@@ -449,6 +449,13 @@ class IRealSource: public ISource {
     TColumns Columns;
 };
 
+class IComposableSource {
```
vitstn, line 452: virtual destructor

vitstn (approved):

PR #3460, opened 2024-04-03
TFederatedWriteSession: take into account AllowFallback setting; use retry policy

qyryq:


ydb/public/sdk/cpp/client/ydb_federated_topic/impl/federated_write_session.cpp
```diff
@@ -129,29 +129,51 @@ void TFederatedWriteSessionImpl::OpenSubSessionImpl(std::shared_ptr<TDbInfo> db)
...
+        if (AsciiEqualsIgnoreCase(FederationState->SelfLocation, db->location())) {
+            // Don't return it immediately, as we may find preferred database on next iterations.
+            sameDatacenterDb = db;
         } else {
             availableDatabases.push_back(db);
             totalWeight += db->weight();
         }
     }
 
+    if (sameDatacenterDb) {
```
alexnick88, line 171: This is not right in case of unavailable preferrableDatabase.  We must route client by hash, but we will use sameDatacanterDb always. It will we correct if we will use desiredDb = preferrableDatabase ? preferableDatabase : sameDatacenterDb. Ignoring Available flag. And then check if desiredDb is available.

alexnick88 (dismissed):

alexnick88 (approved):

PR #3461, opened 2024-04-03
YQL-18053: Add block implementation for Member callable

igormunkin:
This patchset introduces block implementation for `Member` callable. The semantics of this callable differs from `Nth` callable only in the type of the object being processed (structure vs tuple). Hence, the first patch of the series turns `Nth` callable block implementation into generic `GetElem` helper, so the last patch only introduce its specialization for the Struct type.

### Changelog category

* Improvement

### DIFF
--- a/ydb/library/yql/tests/sql/suites/blocks/member.sql
+++ b/ydb/library/yql/tests/sql/suites/blocks/member.sql
@@ -1,0 +1,7 @@
+USE plato;
+/* XXX: Enable UseBlocks pragma and provide input to trigger block execution. */
+pragma UseBlocks;
+
+SELECT
+    val.a as a,
+FROM Input;
--- a/ydb/library/yql/core/peephole_opt/yql_opt_peephole_physical.cpp
+++ b/ydb/library/yql/core/peephole_opt/yql_opt_peephole_physical.cpp
@@ -5468,7 +5468,7 @@ bool CollectBlockRewrites(const TMultiExprType* multiInputType, bool keepInputCo
 
         TExprNode::TListType funcArgs;
         std::string_view arrowFunctionName;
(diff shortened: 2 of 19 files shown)


ydb/library/yql/tests/sql/suites/blocks/member.sql
```diff
@@ -0,0 +1,8 @@
+USE plato;
+/* XXX: Enable UseBlocks pragma and provide input to trigger block execution. */
+pragma UseBlocks;
+
+SELECT
+    TryMember(val, "a", 999) as a,
```
vitstn, line 6: just val.a

ydb/library/yql/tests/sql/suites/blocks/member.sql
```diff
@@ -0,0 +1,8 @@
...
+    TryMember(val, "b", 999) as b,
```
vitstn, line 7: we don't need to test block version of TryMember at all

ydb/library/yql/tests/sql/suites/blocks/member.sql
igormunkin: Thanks, [fixed](https://github.com/ydb-platform/ydb/pull/3461/commits/adea9ac29d90bd05e7b2b26b3addfa6b898b3918#diff-d2885c9862c3ac8939f76d9da5c2c8b977105e051d28d8d1442d5b19bff2bb45R6).

ydb/library/yql/tests/sql/suites/blocks/member.sql
igormunkin: Agree, removed.

vitstn (approved):

Issue #3430, opened 2024-04-03
[viewer api] return storage consumption and limit separated by storage type


antonkovalenko:

artemmufazalov:
We need following data:
- SSD Blob storage - used and limit
- HDD Blob storage - used and limit
- SSD Tablet storage - used and limit
- HDD Tablet storage - used and limit

Currently, we have following data structure from backend:
```JS
{
DatabaseQuotas: {
    "data_size_hard_quota": "53687091200" // Limit for tablet storage
},
Metrics: {
    "Storage": "3885878" // Tablet storage
},
StorageAllocatedSize: "12345", // Blob storage
StorageAllocatedLimit: "12345", // Blob storage limit
StorageUsage: [
    {
        "Type": "SSD",
        "Size": "12345", // SSD Blob storage
        "Limit": "12345", // SSD Blob storage limit
    },
    {
        "Type": "HDD"
    }
]
}
```
In the UI we check values, which are present and use them.

It will be better to have simplified structure, which could be used by UI without additional calculations:
```JS
{
"TabletStorage": [
     {
        "Type": "SSD",
        "Size": "12345", // SSD Tablet storage
        "Limit": "12345", // SSD Tablet storage limit (data_size_hard_quota)
                          // or SSD Blob storage limit if tablet limit isn't set
    },
    {
        "Type": "HDD",
        "Size": "12345", // HDD Tablet storage
        "Limit": "12345", // HDD Tablet storage limit 
                          // or HDD Blob storage limit if tablet limit isn't set
    },
],
"BlobStorage" : [
    {
        "Type": "SSD",
        "Size": "12345", // SSD Blob storage
        "Limit": "12345", // SSD Blob storage limit
    },
    {
        "Type": "HDD",
        "Size": "12345", // HDD Blob storage
        "Limit": "12345", // HDD Blob storage limit
    }
]
}

StekPerepolnen:
https://github.com/ydb-platform/ydb/pull/3463

StekPerepolnen:
the final result looks like
{
"StorageUsage": [
     {
        "Type": "SSD",
        "Size": "12345", // SSD Blob storage
        "Limit": "12345", // SSD Blob storage limit
    },
    {
        "Type": "HDD"
    },
],
"QuotaUsage" : [
    {
        "Type": "SSD",
        "Size": "12345", // Tablet storage
        "Limit": "12345", // Tablet limit (data_size_hard_quota)
    },
    {
[...]

PR #3463, opened 2024-04-04
 added QuotaStorage to tenantinfo response

StekPerepolnen:
https://github.com/ydb-platform/ydb/issues/3430


ydb/core/viewer/protos/viewer.proto
```diff
@@ -375,6 +375,8 @@ message TTenant {
     uint64 StorageAllocatedLimit = 41;
     Ydb.Cms.DatabaseQuotas DatabaseQuotas = 42;
     repeated TStorageUsage StorageUsage = 43;
+    repeated TStorageUsage TabletStorage = 44;
```
adameat, line 378: I think the idea was to make api where UI doesn't have to think. It just works with the only values it has.

StekPerepolnen: during the discussion we decided to leave the format with minor changes

adameat (approved):

PR #3485, opened 2024-04-04
YQL-18053: Add block implementation for AsStruct callable

igormunkin:
This patchset introduces block implementation for `AsStruct` callable. The semantics of this callable differs from `AsTuple` callable only in the resulting type of the object being processed (structure vs tuple). Hence, the first patch of the series turns `Nth` callable block implementation into generic `AsContainer` helper, being used by both `AsTuple` and `AsStruct` runtime node emitter.

Apart from the type annotation and runtime node emitter, there is a patch in the series that makes `CollectBlockRewrites` handle the rewrites related to the `AsStruct` arguments. `AsStruct` arguments (i.e. the members of the resulting structure object) are literal tuples, that don't propagate their child rewrites. As a result, if there is a block rewrite occurred in the expression graph below any of these lists, it takes no effect above this "list barrier". Hence, all `AsStruct` children should be traversed separately to either wrap the complete expressions, supported by the block engine, with `AsScalar` callable, or apply the necessary rewrites. If none of the conditions above are met, the block rewrite for `AsStruct` node is aborted, since one of the children being proceseed is rewritten neither to block, nor scalar, ergo, violating `BlockAsStruct` type requirements.

### Changelog category

* Improvement


ydb/library/yql/core/peephole_opt/yql_opt_peephole_physical.cpp
```diff
@@ -5468,7 +5468,7 @@ bool CollectBlockRewrites(const TMultiExprType* multiInputType, bool keepInputCo
 
         TExprNode::TListType funcArgs;
         std::string_view arrowFunctionName;
-        if (node->IsList() || node->IsCallable({"And", "Or", "Xor", "Not", "Coalesce", "If", "Just", "Nth", "ToPg", "FromPg", "PgResolvedCall", "PgResolvedOp"}))
+        if (node->IsList() || node->IsCallable({"And", "Or", "Xor", "Not", "Coalesce", "If", "Just", "Member", "Nth", "ToPg", "FromPg", "PgResolvedCall", "PgResolvedOp"}))
```
vitstn, line 5471: AsStruct?

igormunkin: The revision was irrelevant. [Force-pushed](https://github.com/ydb-platform/ydb/commit/568b6c8198a0e23471ffcd776180a50db98ce56a).

vitstn (approved):

PR #3552, opened 2024-04-08
KIKIMR-19090 Import CSV to table where all columns are in PK

zinal:
### Changelog entry 

Fix assertion when importing CSV data to a table where all columns are included in the primary key.

### Changelog category 

* Bugfix 

### Additional information

...


ivanmorozov333 (approved):

uzhastik:
Let's cancel this pr

PR #3615, opened 2024-04-09
RoundDown/RoundUp for bigdates

amikish:


ydb/library/yql/minikql/comp_nodes/mkql_round.cpp
```diff
@@ -213,6 +276,10 @@ IComputationNode* WrapRound(TCallable& callable, const TComputationNodeFactoryCo
         case EDataSlot::Timestamp:
             Y_ENSURE(GetDataTypeInfo(to).Features & DateType);
             return new TRoundDateTypeWrapper(ctx.Mutables, source, down, from, to);
+        case EDataSlot::Datetime64:
```
vitstn, line 279: what about Date32?

amikish: It seems there is no need to handle Date/Date32 here
Probably it is handled in https://github.com/ydb-platform/ydb/blob/main/ydb/library/yql/core/type_ann/type_ann_core.cpp#L11544
Calls `select Yql::RoundDown(date32('1970-1-2'), date32` and `select Yql::RoundDown(date32('1970-1-2'), datetime64` work well - will add these test cases to misc.sql

nepal (approved):

PR #3656, opened 2024-04-10
Fix grace join UUID encoding

yumkam:
### Changelog entry 

Fix Uuid handling in mkql_grace_join

### Changelog category 

* Bugfix 

### Additional information

I noticed something weird while reading the code: I believe there are missing `break;` statement in handling Uuid's; also note mismatch between Uuid handling and string-alike handling, this is yet another bug (mostly easily fixed by unifying string-alike and Uuid handling).
I added very crude test (just copied another and replaced strings with Uuid). It fails when applied alone, still fails with only `break` added, and passes with complete series applied.
Compile-tested, limited runtime-tested (my machine is too weak to handle even single round of compilation).

### DIFF
--- a/ydb/library/yql/minikql/comp_nodes/mkql_grace_join.cpp
+++ b/ydb/library/yql/minikql/comp_nodes/mkql_grace_join.cpp
@@ -285,6 +285,7 @@ void TGraceJoinPacker::Pack()  {
             auto str = TuplePtrs[i]->AsStringRef();
             TupleStrings[offset] = str.Data();
             TupleStrSizes[offset] = str.Size();
+            break;
         }
         case NUdf::EDataSlot::TzDate:
         {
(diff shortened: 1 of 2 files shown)


yumkam:
Somewhat tested.
I added a very crude test as a first patch (just copied another test and replaced plain string with Uuid).
It fails with incomplete series applied (I guess this confirms code *was* buggy, and that the final patch *was* necessary and also fixes a bug).
Given recent history of this file (mismatches between Pack/Unpack), I think it would be good idea to add tests for all other types too (but I won't without feedback/guidance first).

yumkam:
Also, why was not this bug (well, *first half* of this bug) caught in by static analysis (aka warning): it seems missing `-Wimplicit-fallthrough` (implied by `-Wextra` in gcc, but not in clang).

aakulaga-ydb (approved):

PR #3866, opened 2024-04-18
Refactoring to make cardinality estimation provider-specific

alephonea:
### Changelog entry 

...

### Changelog category 

* New feature
* Experimental feature
* Improvement
* Performance improvement
* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/library/yql/core/cbo/cbo_optimizer_new.cpp
+++ b/ydb/library/yql/core/cbo/cbo_optimizer_new.cpp
@@ -109,4 +109,123 @@ void TJoinOptimizerNode::Print(std::stringstream& stream, int ntabs) {
     RightArg->Print(stream, ntabs+1);
 }
 
+bool IsPKJoin(const TOptimizerStatistics& stats, const TVector<TString>& joinKeys) {
+    if (stats.KeyColumns.size()==0) {
+        return false;
+    }
+
+    for(size_t i=0; i<stats.KeyColumns.size(); i++){
+        if (std::find(joinKeys.begin(), joinKeys.end(), stats.KeyColumns[i]) == joinKeys.end()) {
+            return false;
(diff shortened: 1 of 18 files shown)


pavelvelikhov (changes requested):

ydb/library/yql/core/cbo/cbo_optimizer_new.cpp
```diff
@@ -109,4 +109,123 @@ void TJoinOptimizerNode::Print(std::stringstream& stream, int ntabs) {
...
+
+    for(size_t i=0; i<stats.KeyColumns.size(); i++){
+        if (std::find(joinKeys.begin(), joinKeys.end(), stats.KeyColumns[i]) == joinKeys.end()) {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool TDummyProviderContext::IsJoinApplicable(const std::shared_ptr<IBaseOptimizerNode>& left,
```
pavelvelikhov, line 125: Lets rename TDummyProviderContext to something like TBaseProviderContext or TDefaultProviderContext.

Its not that dummy anymore if we're including the standard cardinality estimation in there

pavelvelikhov (changes requested):
We also need to fix our KqpProviderContext to subclass from TDummyProviderContext (renamed to something less insulting)

pavelvelikhov (approved):

ydb/library/yql/core/cbo/cbo_optimizer_new.cpp
alephonea: Done.

alephonea:
> We also need to fix our KqpProviderContext to subclass from TDummyProviderContext (renamed to something less insulting)

PR #3882, opened 2024-04-18
implicit transaction in per statement mode

### LINKED ISSUE #3423: [QS] Use implicit transaction for multi-statement queries
When no explicit transaction is specified (empty TxControl and no BEGIN/COMMIT) we should use implicit transaction for the entire query.

```
SELECT ... FROM table1;
SELECT ... FROM table2;
```

Such query should be executed in single transaction.

VPolka:
implicit transaction in per statement mode

### Changelog entry 

make 1 implicit transaction for all multi statement query

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/core/kqp/session_actor/kqp_query_state.h
+++ b/ydb/core/kqp/session_actor/kqp_query_state.h
@@ -33,6 +33,22 @@ namespace NKikimr::NKqp {
 // common case).
 class TKqpQueryState : public TNonCopyable {
 public:
+    class TQueryTxId {
+    public:
+        TQueryTxId() = default;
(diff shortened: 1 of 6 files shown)


ydb/core/kqp/session_actor/kqp_query_state.h
```diff
@@ -33,6 +33,22 @@ namespace NKikimr::NKqp {
 // common case).
 class TKqpQueryState : public TNonCopyable {
 public:
+    class TQueryTxId {
```
spuchin, line 36: А чем это отличается от std::optional<TTxId>?

spuchin (approved):

PR #3942, opened 2024-04-19
Refactor acceptance_run.yml (use actions directly)

maximyurchuk:
### Changelog category 

### DIFF
--- a/.github/workflows/acceptance_run.yml
+++ b/.github/workflows/acceptance_run.yml
@@ -22,6 +22,9 @@ on:
       put_build_results_to_cache:
         type: boolean
         default: true
+      commit_sha:
+        type: string
+        default: ""
 
 jobs:
   main:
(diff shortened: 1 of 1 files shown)


.github/workflows/acceptance_run.yml
```diff
@@ -4,18 +4,47 @@ on:
...
-    secrets: inherit
-    with:
-      runner_label: auto-provisioned
-      build_preset: relwithdebinfo
-      build_target: ydb/tests/acceptance
-      run_build: true
-      run_tests: true
+  main:
+    name: Build and test ${{ inputs.build_preset }}    
+    runs-on: [ self-hosted, "${{ inputs.runner_label }}"]
```
nikitka, line 32: Apart from `auto-provisioned`, you should use the `build-preset-<preset_name>` label here. You can look at how this label is generated [here](https://github.com/ydb-platform/ydb/blob/main/.github/workflows/build_and_test_ya_provisioned.yml#L104). It is needed because different presets require different VM configurations..

nikitka (changes requested):

.github/workflows/acceptance_run.yml
maximyurchuk: Done

nikitka (approved):

PR #3977, opened 2024-04-22
YQ-3116 Add restoring issues to query status

kardymonds:
### Changelog entry 

...

### Changelog category 

### Additional information

...


dorooleg (dismissed):

ydb/core/fq/libs/checkpointing/checkpoint_coordinator.cpp
```diff
@@ -301,9 +302,10 @@ void TCheckpointCoordinator::Handle(const NYql::NDq::TEvDqCompute::TEvRestoreFro
     }
 
     if (status != NYql::NDqProto::TEvRestoreFromCheckpointResult_ERestoreStatus_OK) {
-        CC_LOG_E("[" << checkpoint << "] Can't restore: " << statusName);
+        auto msg = TStringBuilder() << "Can't restore: " << statusName << ", " << record.GetIssues();
```
UgnineSirdis, line 305: It is not good in common to have multiline string with issues in log. There is special logging method in TIssues, that is better to use here: https://github.com/ydb-platform/ydb/blob/main/ydb/library/yql/public/issue/yql_issue.h#L320

But these issues should be subissues in checkpoint coordinator error message and TEvDqFailure

UgnineSirdis (changes requested):

UgnineSirdis (approved):

PR #4014, opened 2024-04-22
Aggregate storage pool usage statistics by pool name instead of a pool kind

jepett0:
Changes: https://github.com/ydb-platform/ydb/pull/2678
[Draft ticket](https://github.com/orgs/ydb-platform/projects/31/views/6?pane=issue&itemId=53688221)
Legacy ticket: KIKIMR-18302

We decided to change the way separate storage pool quotas for databases work.

Initially we quoted a particular storage pool kind, because when the database is created (for example, using `ydbd admin database create` command), the user specifies only the `kind` of a storage pool they would like to create for this database to use, not the storage pool `name`. The `name` of the storage pool is autogenerated based on the specified `kind` and the path to the subdomain (a.k.a. database) like this:
`/Root/dedicated:hdd`, where `/Root/dedicated` is the path to the database, `hdd` is the storage pool `kind`, `:` is a separator.

However, quoting by storage pool `kind` is meaningless. If the physical storage of a particular storage pool was depleted, then it would <ins>not</ins> start writing its data to another storage pool with the same kind. The writes will just stop. Abruptly. To prevent this you need to set quotas not on the storage pool `kind`, but on the particular storage pool `name`.

Code changes in this PR are mainly of the following form:
- `unit_kind` (a.k.a. storage pool `kind`) -> `storage_unit` (a.k.a. storage pool `name`)
- `PoolKind` -> `PoolName`

### DIFF
--- a/ydb/core/cms/console/console__create_tenant.cpp
+++ b/ydb/core/cms/console/console__create_tenant.cpp
@@ -283,7 +283,7 @@ public:
                 if (unitHardQuota && unitSoftQuota && unitHardQuota < unitSoftQuota) {
                     return Error(Ydb::StatusIds::BAD_REQUEST,
                         TStringBuilder() << "Data size soft quota (" << unitSoftQuota << ")"
-                                         << " for a " << storageQuota.unit_kind() << " storage unit "
+                                         << " for a " << storageQuota.storage_unit() << " storage unit "
                                          << " of the database " << path
                                          << " must be smaller than the corresponding hard quota (" << unitHardQuota << ")",
                         ctx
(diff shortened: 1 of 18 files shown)


jepett0:
We figured out that storage pool kinds are unique for a particular database, because 1) storage pool names are auto-generated from specified kinds and 2) storage pool names [are unique](https://st.yandex-team.ru/KIKIMR-4504#5ac738775af046001a32b9f3). We will use storage pool kind as a unique storage pool identifier within a particular database. The changes in this PR are not needed.

PR #4059, opened 2024-04-24
Support AlterSequence in schemeshard

### LINKED ISSUE #3847: Support AlterSequence in schemeshard

shnikd:


ydb/core/tx/schemeshard/schemeshard__operation_create_sequence.cpp
```diff
@@ -246,6 +246,63 @@ class TPropose: public TSubOperationState {
...
+    }
+    result.SetIncrement(increment);
+
+    i64 minValue = 1;
+    i64 maxValue = Max<i64>();
+    if (increment < 0) {
+        maxValue = -1;
+        minValue = Min<i64>();
+    }
+
```
ijon, line 267: Why this arithmagic is necessary in the PR about adding the alter sequence op?

And perhaps it deserves a line or two with the explanation in the code.

ydb/core/tx/schemeshard/schemeshard__operation_alter_sequence.cpp
```diff
@@ -0,0 +1,486 @@
...
+                << " at tablet " << ssId);
+            return false;
+        }
+
+        context.OnComplete.UnbindMsgFromPipe(OperationId, tabletId, txState->TargetPathId);
+
+        if (txState->ShardsInProgress.empty()) {
+            NIceDb::TNiceDb db(context.GetDB());
+            context.SS->ChangeTxState(db, OperationId, TTxState::Propose);
+            context.OnComplete.ActivateTx(OperationId);
```
ijon, line 80: Call to `context.OnComplete.ActivateTx()` is unnecessary here -- on returning `true` (that is, on completing the state) it would be called automatically.

ijon (changes requested):

ydb/core/tx/schemeshard/schemeshard__operation_create_sequence.cpp
shnikd: Before that, the sequence scheme was not used and was not filled with default values, now it was necessary to add logic for filling the scheme with default values

comment added

ijon (approved):

PR #4087, opened 2024-04-24
Fix reordering of change records

### LINKED ISSUE #4034: Extra rows in the asynchronous index

CyberROFL:
### Changelog entry 

FIxed reordering of change records.

### Changelog category 

* Bugfix

### Additional information

...

### DIFF
--- a/ydb/core/change_exchange/change_sender_common_ops.cpp
+++ b/ydb/core/change_exchange/change_sender_common_ops.cpp
@@ -173,11 +173,11 @@ void TBaseChangeSender::SendRecords() {
             break;
         }
 
+        processedAtLeastOne = true;
+
         if (PendingBody && PendingBody.begin()->Order <= it->first) {
             break;
         }
-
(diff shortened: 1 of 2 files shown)


ydb/core/change_exchange/change_sender_common_ops.cpp
```diff
@@ -162,6 +162,14 @@ void TBaseChangeSender::SendRecords() {
     bool needToResolve = false;
 
     while (it != PendingSent.end()) {
+        if (Enqueued && Enqueued.begin()->Order <= it->first) {
```
snaury, line 165: I'm worried this may deadlock. When we previously sent records we removed them from `MemUsage`, and we might have loaded more records in the available space. Now we have reenqueued some records, but they might not be able to start loading, because `MemUsage` no longer has enough space. This would deadlock: we don't send ready records because we have some enqueued, and we don't load enqueued records because we don't have enough memory.

snaury (approved):

PR #4140, opened 2024-04-26
Add analytical nodes to kqp

### LINKED ISSUE #4141: Distinguish OLAP nodes in kqp

pixcc:
### Changelog category 

### DIFF
--- a/ydb/core/config/init/init.h
+++ b/ydb/core/config/init/init.h
@@ -316,9 +316,9 @@ private:
 };
 
 enum class EWorkload {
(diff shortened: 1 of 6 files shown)


gridnevvvit:
Overall i think that PR is okay generally, however i think that the usage of "ServedWorkload" field isn't clear. I guess that we should introduce "NodeTags" field instead so that the resource manager will have an ability to assign specific tasks and queries to the nodes with the specific tag?

how this field should be used in the future and how it can be extended in the future?

PR #4153, opened 2024-04-26
Add flexibility to sanitizer ignore in YQL UDFs test

Krisha11:
### Changelog category 

* Bugfix


Krisha11:
Currently, ``--sanitizer ignore thread`` is equivalent to ``--sanitizer ignore``.

Because of this, some of the internal tests are not running, but as I understand it, they are supposed to. To temporarily fix the problem (and make it possible to merge changes internally). I suggest temporarily turning off the breaking tests inside

https://nda.ya.ru/t/dzeL3xNE75cuPp

PR #4190, opened 2024-04-29
Fix pdisk restart race on vdisk log reading

SammyVimes:
### Changelog category 
* Bugfix


ydb/core/blobstorage/nodewarden/node_warden_pdisk.cpp
```diff
@@ -278,6 +282,24 @@ namespace NKikimr::NStorage {
...
+            // Restart is already in progress.
+            return;
+        }
+
+        const TPDiskKey pdiskKey(LocalNodeId, pdiskId);
+
+        const TVSlotId from(pdiskKey.NodeId, pdiskKey.PDiskId, 0);
+        const TVSlotId to(pdiskKey.NodeId, pdiskKey.PDiskId, Max<ui32>());
+
+        for (auto it = LocalVDisks.lower_bound(from); it != LocalVDisks.end() && it->first <= to; ++it) {
```
alexvru, line 297: should we really do it explicitly here? I thought it was expected that VDisks terminate in their own way when they find out that round in incorrect

alexvru (changes requested):

alexvru (approved):

PR #4198, opened 2024-04-30
Extract BuildStats for B-Tree index tables

kungasc:
### Changelog entry 

...

### Changelog category 

### Additional information

Extract `BuildStatsBTreeIndex` for further effective implementation of stats collecting in case when all parts have B-Tree index

Implement first part that calculates index and data sizes with by channel statistic


ydb/core/tablet_flat/flat_stat_table.cpp
```diff
@@ -1,70 +1,29 @@
 #include "flat_part_laid.h"
```
kunga, line 1: По-сути переименование в flat_stat_table_mixed_index.h

ydb/core/tablet_flat/flat_stat_table_btree_index.h
```diff
@@ -0,0 +1,198 @@
+#include "flat_stat_table.h"
```
kunga, line 1: По-сути переименование из flat_stat_table.cpp, можно не смотреть

snaury (approved):

PR #4212, opened 2024-05-01
Do not cache dns at the http_proxy actor level

Jorres:
### Changelog entry 

- remove dns queries caching at the http_proxy actor level (dns caching should be done by dns server)

### Changelog category 

* Bugfix 

### Additional information

It seems that http_actor had a special map `Hosts` that basically re-implemented caching for dns requests. This PR basically removes the check:
```
        auto it = Hosts.find(host);
        if (it == Hosts.end() || it->second.DeadlineTime > ctx.Now()) { ...
```

which leads the code into always calling 

```
TNetworkAddress addr(addressPart, portPart);
```

which calls `getaddrinfo` under the hood.


Jorres:
Discussed this implementation with adameat@, my view appears to be wrong, and there was actually no bug in this code.

PR #4238, opened 2024-05-02
Add slice tests 

maximyurchuk:
### Changelog entry 
Add tests:
- sample test there slice is used 
- ydb_serializable now runs on slice 

### Changelog category 


gridnevvvit (dismissed):

ydb/tools/ydbd_slice/ya.make
```diff
@@ -23,7 +23,6 @@ PY_SRCS(
 PEERDIR(
     ydb/tools/cfg
     ydb/public/sdk/python
-    ydb/public/sdk/python/enable_v3_new_behavior
```
maximyurchuk, line 26: Some tests were broken, so I decided to remove thus peerdir because it seems meaningless in ydbd_slice

gridnevvvit (approved):

PR #4331, opened 2024-05-06
Support arbiters in volatile transactions

### LINKED ISSUE #3918: DataShard: readset arbiter for volatile transactions
The total number of readsets grows quadratically with the number of affected shards in volatile transactions. For example, upsert into 100 shards requires almost 20k total readset messages. We need to implement arbiters, where a single shard is designated as the decider for the transaction. Basically it exchanges readsets with all other shards and all other shards only exchange readsets with the arbiter. Arbiter will need to generate outgoing readsets normally, but will only release them when the transaction is committed, i.e. when it receives commit decision from all other shards. We don't need to finish the commit, outgoing readsets could be sent as soon as we decide to commit the transaction (and before that commit is fully persisted), however aborting is a bit trickier, it will need to clear these outgoing readsets and make sure they are never sent after it aborted.

This additional hop will delay commits a bit, but as long as we don't have storage in the hot path it should be ok. We will need to choose one of writer shards as the arbiter, as long as the number of readsets is large enough (e.g. it doesn't reduce anything for 2 shards, but for 10 affected shards it goes down from 180 readset messages to just 36). The number of full mesh messages is 2n(n-1), while for the star it is 4(n-1), which means the cut-off (using arbiters is better) is any n > 2 in the number of messages, but the additional latency may not be worth it for a small n. We could try setting a cut-off for using arbiters at n=5 or n=10.

snaury:
### Changelog entry 

Support arbiters in volatile transactions, reducing the number of messages exchanged for distributed transactions with a large number of participants.

### Changelog category 

* New feature

### Additional information

Volatile distributed transactions often cause every shard to exchange readsets with every other shard, which leads to the total number of messages growing quadratically. Selecting a single arbiter shard changes a full mesh graph to a star, which reduces the number of messages at the expense of possibly increased latency. This patch implements arbiters for volatile transactions which may be enabled using the `EnableVolatileTransactionArbiters` feature flag.

Fixes #3918.

### DIFF
--- a/ydb/core/tx/datashard/datashard_ut_write.cpp
+++ b/ydb/core/tx/datashard/datashard_ut_write.cpp
@@ -2,10 +2,13 @@
 #include "datashard_ut_read_table.h"
 #include <ydb/core/tx/data_events/payload_helper.h>
 #include <ydb/core/tx/datashard/ut_common/datashard_ut_common.h>
+#include <ydb/core/base/tablet_pipecache.h>
+#include "datashard_ut_common_kqp.h"
 
 namespace NKikimr {
 
 using namespace NKikimr::NDataShard;
+using namespace NKikimr::NDataShard::NKqpHelpers;
 using namespace NSchemeShard;
 using namespace Tests;
 using namespace NDataShardReadTableTest;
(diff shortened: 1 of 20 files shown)


CyberROFL (approved):

ydb/core/tx/datashard/datashard_ut_write.cpp
```diff
@@ -716,5 +719,452 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
...
+        maxStep = Min(maxStep, msg->Record.GetMaxStep());
+        UNIT_ASSERT_VALUES_EQUAL(msg->Record.DomainCoordinatorsSize(), 1);
+        if (coordinator == 0) {
+            coordinator = msg->Record.GetDomainCoordinators(0);
+        } else {
+            UNIT_ASSERT_VALUES_EQUAL(coordinator, msg->Record.GetDomainCoordinators(0));
+        }
+    }
+
+    void SendProposeToCoordinator(
```
azevaykin, line 781: There is a similar method [SendProposeToCoordinator](https://github.com/ydb-platform/ydb/blob/293e40077ccd74eba4b96a089fb830a918dea1b3/ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp#L2101)

It should be better either use a single method, or move this overloaded method to datashard_ut_common.cpp

ydb/core/tx/datashard/datashard_ut_write.cpp
```diff
@@ -716,5 +719,452 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
...
+
+        RunUpsertWithArbiter(
+            runtime, tableId, shards,
+            /* txId */ 1000001,
+            // arbiter will send 6 readsets (3 decisions + 3 expectations)
+            // shards will send 2 readsets each (decision + expectation)
+            /* expectedReadSets */ 6 + 3 * 2);
+
+        Cerr << "========= Checking table =========" << Endl;
+        UNIT_ASSERT_VALUES_EQUAL(
```
azevaykin, line 911: Everywhere in this file another construction is used:
```
            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
            UNIT_ASSERT_VALUES_EQUAL(tableState, expectedTableState);
```

I suggest using a single approach.

snaury: ReadTable uses persistent snapshots and very heavyweight scans with extra waits for history to be committed. It should never be used unless that's what you're actually testing.

ydb/core/tx/datashard/datashard_ut_write.cpp
snaury: I prefer local functions unless they are useful in multiple translation units. The existing function may have the same name, but it has the wrong arguments and doesn't work correctly, fixing existing tests was out of scope of this pr.

ydb/core/tx/datashard/datashard_ut_write.cpp
azevaykin: Is it preferable to change all the previous cases to `KqpSimpleExec` ?

snaury: Either that, or make helper functions that would read using EvRead.

PR #4419, opened 2024-05-09
EnforceUserTokenCheckRequirement option: require token check if token was specified

UgnineSirdis:
EnforceUserTokenCheckRequirement option: require token check if token was specified. Valid only when EnforceUserTokenRequirement option is false.

This behavour is expected by users: when they specify token, they expect that they use authorized access to database. But for historical reasons unauthorized tokens now are accepted as if user not specified token at all. In this mode it is hard to check correctness of authorization: both in server and in client.

### Changelog category 

* New feature

### Additional information

...

### DIFF
--- a/ydb/core/protos/config.proto
+++ b/ydb/core/protos/config.proto
@@ -232,6 +232,7 @@ message TDomainsConfig {
 
     message TSecurityConfig {
         optional bool EnforceUserTokenRequirement = 1 [default = false];
+        optional bool EnforceUserTokenCheckRequirement = 7 [default = false]; // Check if a token was specified // If not, or if the token was incorrect or access was denied, the request will be handled as if no token was provided
         repeated string MonitoringAllowedSIDs = 2;
         repeated string AdministrationAllowedSIDs = 3;
         repeated string DefaultUserSIDs = 4;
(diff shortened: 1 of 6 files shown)


ydb/core/protos/config.proto
```diff
@@ -232,6 +232,7 @@ message TDomainsConfig {
 
     message TSecurityConfig {
         optional bool EnforceUserTokenRequirement = 1 [default = false];
+        optional bool EnforceUserTokenCheckRequirement = 7 [default = true]; // Check token if it was specified // If false and token was incorrect or access was denied, the request will be treated as if there was no token
```
adameat, line 235: // Check if a token was specified
// If not, or if the token was incorrect or access was denied, the request will be handled as if no token was provided

ydb/core/security/secure_request.h
```diff
@@ -137,7 +141,9 @@ class TSecureRequestActor : public TBase {
 
 public:
     bool IsTokenRequired() const {
-        return GetEnforceUserTokenRequirement() || (RequireAdminAccess && !GetAdministrationAllowedSIDs().empty());
+        return GetEnforceUserTokenRequirement()
```
adameat, line 144: don't you think it would be better to rewrite it without multiple ORs. just 3 if's with corresponding comments?

adameat (approved):

PR #4434, opened 2024-05-10
Object storage listing

SammyVimes:
### Changelog category 

* New feature

# Proposed docs

## Object Storage Listing API

Objects in (S3-like) object storage have a name column. Although object storage is unstructured, users often give their objects name similar to file path, e.g. /foo/baz.txt, /foo/bar/test.txt. One of the common operations is listing. For example, listing folder /foo/ should yield:

/bar/
/foo/baz.txt

#### Table name
Name of the table to list objects from

#### Key prefix
Limits the response to keys that begin with the specified key prefix. keyColumn0 ... path column (exclusive) values.

Example in Go:
```
Ydb.TypedValue{
    Type: &Ydb.Type{
        Type: &Ydb.Type_TupleType{
            TupleType: &Ydb.TupleType{
                Elements: []*Ydb.Type{
                    {
                        Type: &Ydb.Type_TypeId{
                            TypeId: Ydb.Type_UINT64,
                        },
                    },
                },
            },
        },
    },
    Value: &Ydb.Value{
        Items: []*Ydb.Value{
            {
                Value: &Ydb.Value_Uint64Value{
                    Uint64Value: 100,
                },
            },
        },
    },
}
```

This is a Ydb.TypedValue that contains key column values in PK order. Must have all columns from first until the "path" column (excluding it). 

#### Path column prefix
Limits the response to keys that begin with the specified path prefix (e.g. /foo)

#### Path column delimiter
A delimiter is a character that you use to group keys. For example, if path is /foo/bar/baz.txt, then delimiter should be "/"

#### Start after key suffix
Key's suffix column values: path column (inclusive) ... columnN values.
StartAfterKeySuffix is where you want to start listing from. This API starts listing after this specified key. StartAfterKeySuffix can be any existing or non-existing key

Example in Go:
```
Ydb.TypedValue{
    Type: &Ydb.Type{
        Type: &Ydb.Type_TupleType{
            TupleType: &Ydb.TupleType{
                Elements: []*Ydb.Type{
                    {
                        Type: &Ydb.Type_TypeId{
                            TypeId: Ydb.Type_UTF8,
                        },
                    },
                },
            },
        },
    },
    Value: &Ydb.Value{
        Items: []*Ydb.Value{
            {
                Value: &Ydb.Value_TextValue{
                    TextValue: "/Foo/bar/",
                },
            },
        },
    },
}
```

This is a Ydb.TypedValue that contains key column values in PK order.

#### Matching filter
Column values to filter out rows

Here is a filter that filters out all rows with "deleted" column's value equal to true.
Example in Go:
```
Ydb.TypedValue{
    Type: &Ydb.Type{
        Type: &Ydb.Type_TupleType{
            TupleType: &Ydb.TupleType{
                Elements: []*Ydb.Type{
                    {
                        Type: &Ydb.Type_ListType{
                            ListType: &Ydb.ListType{
                                Item: &Ydb.Type{
                                    Type: &Ydb.Type_TypeId{
                                        TypeId: Ydb.Type_STRING,
                                    },
                                },
                            },
                        },
                    },
                    {
                        Type: &Ydb.Type_ListType{
                            ListType: &Ydb.ListType{
                                Item: &Ydb.Type{
                                    Type: &Ydb.Type_TypeId{
                                        TypeId: Ydb.Type_UINT32,
                                    },
                                },
                            },
                        },
                    },
                    {
                        Type: &Ydb.Type_TupleType{
                            TupleType: &Ydb.TupleType{
                                Elements: []*Ydb.Type{
                                    {
                                        Type: &Ydb.Type_TypeId{
                                            TypeId: Ydb.Type_BOOL,
                                        },
                                    },
                                },
                            },
                        },
                    },
                },
            },
        },
    },
    Value: &Ydb.Value{
        Items: []*Ydb.Value{
            {
                Items: []*Ydb.Value{
                    {
                        Value: &Ydb.Value_TextValue{
                            TextValue: "deleted",
                        },
                    },
                },
            },
            {
                Items: []*Ydb.Value{
                    {
                        Value: &Ydb.Value_Uint32Value{
                            Uint32Value: uint32(Ydb_ObjectStorage.ListingRequest_EQUAL),
                        },
                    },
                },
            },
            {
                Items: []*Ydb.Value{
                    {
                        Value: &Ydb.Value_BoolValue{
                            BoolValue: true,
                        },
                    },
                },
            },
        },
    },
}
```

This is a Ydb.TypedValue that contains list of column names, list of matching types (either EQUAL or NOT_EQUAL) and list of values for the columns.

#### Continuation token
ContinuationToken indicates that the list is being continued on with a token. ContinuationToken is obfuscated and is not a real key. You can use this ContinuationToken for pagination of the list results

#### Columns to return
Which columns of the row to return. Path column and filter columns are added automatically 

#### Max keys
Sets the maximum number of keys returned in the response. By default, the action returns up to 1,000 key names. The response might contain fewer keys but will never contain more

#### Status
YDB operation status code

#### Issues
YDB operation issue messages

#### Common prefixes
All the keys (up to 1_000) that share the same prefix are grouped together. When counting the total numbers of returns by this API operation, this group of keys is considered as one item.  
A response can contain CommonPrefixes only if you specify a delimiter.  
CommonPrefixes contains all (if there are any) keys between Prefix and the next occurrence of the string specified by a delimiter.  

#### Contents
Returned rows with specified columns (path and filter columns are added automatically)

#### Next continuation token
NextContinuationToken is sent when isTruncated is true, which means there are more keys that can be listed. The next list requests can be continued with this NextContinuationToken. NextContinuationToken is obfuscated and is not a real key

### Is truncated
Set to false if all of the results were returned. Set to true if more keys are available to return

### DIFF
--- a/ydb/core/tx/datashard/datashard.h
+++ b/ydb/core/tx/datashard/datashard.h
@@ -241,8 +241,8 @@ struct TEvDataShard {
         EvGetTableStatsResult,
         EvPeriodicTableStats,
 
-        EvS3ListingRequest,
-        EvS3ListingResponse,
+        EvObjectStorageListingRequest,
+        EvObjectStorageListingResponse,
 
         EvUploadRowsRequest,
         EvUploadRowsResponse,
@@ -1383,20 +1383,20 @@ struct TEvDataShard {
         }
     };
 
-    struct TEvS3ListingRequest
-        : public TEventPB<TEvS3ListingRequest,
-                            NKikimrTxDataShard::TEvS3ListingRequest,
-                            TEvDataShard::EvS3ListingRequest> {
-        TEvS3ListingRequest() = default;
+    struct TEvObjectStorageListingRequest
+        : public TEventPB<TEvObjectStorageListingRequest,
+                            NKikimrTxDataShard::TEvObjectStorageListingRequest,
+                            TEvDataShard::EvObjectStorageListingRequest> {
+        TEvObjectStorageListingRequest() = default;
     };
 
-    struct TEvS3ListingResponse
-         : public TEventPB<TEvS3ListingResponse,
-                            NKikimrTxDataShard::TEvS3ListingResponse,
-                            TEvDataShard::EvS3ListingResponse> {
-        TEvS3ListingResponse() = default;
+    struct TEvObjectStorageListingResponse
+         : public TEventPB<TEvObjectStorageListingResponse,
+                            NKikimrTxDataShard::TEvObjectStorageListingResponse,
+                            TEvDataShard::EvObjectStorageListingResponse> {
+        TEvObjectStorageListingResponse() = default;
 
-        explicit TEvS3ListingResponse(ui64 tabletId, ui32 status = NKikimrTxDataShard::TError::OK) {
+        explicit TEvObjectStorageListingResponse(ui64 tabletId, ui32 status = NKikimrTxDataShard::TError::OK) {
             Record.SetTabletID(tabletId);
             Record.SetStatus(status);
         }
--- a/ydb/core/tx/datashard/datashard.h
+++ b/ydb/core/tx/datashard/datashard.h
@@ -1383,6 +1383,25 @@ struct TEvDataShard {
         }
     };
 
+    struct TEvS3ListingRequest
+        : public TEventPB<TEvS3ListingRequest,
+                            NKikimrTxDataShard::TEvS3ListingRequest,
+                            TEvDataShard::EvS3ListingRequest> {
+        TEvS3ListingRequest() = default;
+    };
+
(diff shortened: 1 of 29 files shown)


ydb/core/tx/datashard/datashard.h
```diff
@@ -1383,6 +1383,25 @@ struct TEvDataShard {
         }
     };
 
+    struct TEvObjectStorageListingRequest
```
azevaykin, line 1386: Could you make some tests in `ydb/core/tx/datashard` folder? Without `ydb/core/client`.

dcherednik (approved):

PR #4436, opened 2024-05-10
Reduce CPU reporting threshold delta from 10% to 2%

kungasc:
### Changelog entry 

...

### Changelog category 

### Additional information

Reporting CPU usage with 10% threshold is too confusing, simple key-value reading may consume around 5%

Also if we have a huge cluster with hundred thousand tablets, treating 7% CPU usage as zero seems wrong

Before that fix UI shows:

telegram-cloud-photo-size-2-5326022372509211720-y

Hive shows:

telegram-cloud-photo-size-2-5326022372509211727-y

And restarts didn't help to fix tablet 72075186224037908 usage in Hive

### DIFF
--- a/ydb/core/tablet/tablet_metrics.h
+++ b/ydb/core/tablet/tablet_metrics.h
@@ -59,7 +59,7 @@ public:
     bool TryUpdate(TResourceMetricsValues& src, const TActorContext& ctx);
 
 protected:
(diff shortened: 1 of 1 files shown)


kunga:
Hacked with https://github.com/ydb-platform/ydb/pull/4446

PR #4540, opened 2024-05-15
rewrote begintx to notx in generic query type. At the moment dataquer…

flown4qqqq:
### Changelog entry 
rewrote begintx to notx in generic query type. At the moment dataquery notx is not supported
...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/public/sdk/cpp/client/ydb_query/tx.h
+++ b/ydb/public/sdk/cpp/client/ydb_query/tx.h
@@ -65,7 +65,7 @@ struct TTxSettings {
 
     FLUENT_SETTING(TTxOnlineSettings, OnlineSettings);
 
-    ETransactionMode getMode() const {
+    ETransactionMode GetMode() const {
         return Mode_;
     }
 private:
--- a/ydb/public/sdk/cpp/client/ydb_query/tx.h
+++ b/ydb/public/sdk/cpp/client/ydb_query/tx.h
@@ -63,12 +63,16 @@ struct TTxSettings {
         TS_SNAPSHOT_RO
     };
 
-    const ETransactionMode Mode_;
     FLUENT_SETTING(TTxOnlineSettings, OnlineSettings);
 
+    ETransactionMode getMode() const {
+        return Mode_;
+    }
 private:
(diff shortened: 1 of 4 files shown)


ydb/public/sdk/cpp/client/ydb_query/tx.h
```diff
@@ -63,12 +63,16 @@ struct TTxSettings {
         TS_SNAPSHOT_RO
     };
 
-    const ETransactionMode Mode_;
     FLUENT_SETTING(TTxOnlineSettings, OnlineSettings);
 
+    ETransactionMode getMode() const {
```
Gazizonoki, line 68: Wrong style, functions should begin with capital letter
```suggestion
    ETransactionMode GetMode() const {
```

ydb/public/sdk/cpp/client/ydb_query/tx.h
```diff
@@ -63,12 +63,16 @@ struct TTxSettings {
         TS_SNAPSHOT_RO
     };
 
-    const ETransactionMode Mode_;
```
Gazizonoki, line 66: This may break backward compatibility, users can already use Mode_ public member ih their code

ydb/public/sdk/cpp/client/ydb_query/tx.h
```diff
@@ -63,12 +63,16 @@ struct TTxSettings {
...
     FLUENT_SETTING(TTxOnlineSettings, OnlineSettings);
 
+    ETransactionMode getMode() const {
+        return Mode_;
+    }
 private:
     TTxSettings(ETransactionMode mode)
         : Mode_(mode) {}
+
+    ETransactionMode Mode_;
```
Gazizonoki, line 75: This member is not changing, better make it const

flown4qqqq: image

If to return const qualifier, such errors will come out (the fact is that I use a constructor that sets Mode_, which is why it cannot be const)

Gazizonoki (approved):

PR #4559, opened 2024-05-15
Yql-18391 gracejoin refactoring for spilling

lll-phill-lll:
### Changelog entry 

Small refactoring that makes its easier to add spilling to grace join

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
+++ b/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
@@ -82,7 +82,7 @@ void TTable::AddTuple(  ui64 * intColumns, char ** stringColumns, ui32 * strings
     std::vector<char, TMKQLAllocator<char>> & stringVals = TableBuckets[bucket].StringsValues;
     KeysHashTable & kh = TableBuckets[bucket].AnyHashTable;
 
-    ui64 offset = keyIntVals.size();  // Offset of tuple inside the keyIntVals vector
+    ui32 offset = keyIntVals.size();  // Offset of tuple inside the keyIntVals vector
 
     keyIntVals.push_back(hash);
     keyIntVals.insert(keyIntVals.end(), intColumns, intColumns + NullsBitmapSize_);
--- a/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
+++ b/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
@@ -4,9 +4,7 @@
 #include <ydb/library/yql/utils/log/log.h>
 
 #include <contrib/libs/xxhash/xxhash.h>
-#include <chrono>
 #include <string_view>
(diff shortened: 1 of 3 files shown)


ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
```diff
@@ -65,52 +63,43 @@ void TTable::AddTuple(  ui64 * intColumns, char ** stringColumns, ui32 * strings
...
         }
     }
 
-
-    TableBuckets[bucket].TuplesNum++;
+    TableBucketsStats[bucket].TuplesNum++;
 
     if (NumberOfStringColumns || NumberOfIColumns ) {
-        stringsOffsets.push_back(offset); // Adding offset to tuple in keyIntVals vector
-        stringsOffsets.push_back(stringVals.size());  // Adding offset to string values
```
lll-phill-lll, line 113: these 2 sizes will be less than full vector size because of spilling. Vectors partly will be spilled.

ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
```diff
@@ -140,17 +127,18 @@ void TTable::AddTuple(  ui64 * intColumns, char ** stringColumns, ui32 * strings
...
         stringVals.insert(stringVals.end(), *(dataStringsColumns + i), *(dataStringsColumns + i) + currStringSize);
     }
 
     for ( ui64 i = 0; i < NumberOfDataIColumns; i++) {
         stringVals.insert( stringVals.end(), IColumnsVals[NumberOfKeyIColumns + i].begin(), IColumnsVals[NumberOfKeyIColumns + i].end());
-
     }
 
-
+    TableBucketsStats[bucket].KeyIntValsTotalSize += keyIntVals.size() - offset;
```
Darych, line 140: Why do we reduce size by offset? Maybe we can just use assignment?
`TableBucketsStats[bucket].KeyIntValsTotalSize = keyIntVals.size()`

lll-phill-lll: In `TableBucketsStats[bucket].KeyIntValsTotalSize` we want to have total size of `keyIntVals` vector, because we use this size to calculate correct indexes (offsets).

After spilling the bucket `keyIntVals.size()` will be equal to 0. But we want to keep indexing like there are no spilling. Because after restoring all the vectors from spilling indexes must be correct. 

That is why here I increase the size of  `KeyIntValsTotalSize` by the number of added elements. Number of added elements is computed like this:

```
offset = keyIntVals.size();

keyIntVals.AddElements();

KeyIntValsTotalSize  += keyIntVals.size() - offset;
```

I believe it would be more clear if i used `InitialSize` instead of `offset`

Darych (dismissed):

Darych (approved):

PR #4568, opened 2024-05-15
IGNIETFERRO-2059: Fix use-after-free error in the TDqExecuter

ubyte:
### Changelog entry 

Fix use-after-free error caused by constructing an input stream over a temporary string object

### Changelog category 

* Bugfix

### DIFF
--- a/ydb/library/yql/providers/dq/actors/executer_actor.cpp
+++ b/ydb/library/yql/providers/dq/actors/executer_actor.cpp
@@ -99,7 +99,8 @@ private:
         Yql::DqsProto::TWorkerFilter pragmaFilter;
         if (Settings->WorkerFilter.Get()) {
             try {
-                TStringInput inputStream1(Settings->WorkerFilter.Get().GetOrElse(""));
+                TString inputString = Settings->WorkerFilter.Get().GetOrElse("");
+                TStringInput inputStream1(inputString);
                 ParseFromTextFormat(inputStream1, pragmaFilter);
             } catch (...) {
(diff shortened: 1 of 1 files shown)


maximyurchuk:
@vitstn please, take a look

ydb/library/yql/providers/dq/actors/executer_actor.cpp
```diff
@@ -99,7 +99,8 @@ class TDqExecuter: public TRichActor<TDqExecuter>, NYql::TCounters {
         Yql::DqsProto::TWorkerFilter pragmaFilter;
         if (Settings->WorkerFilter.Get()) {
             try {
-                TStringInput inputStream1(Settings->WorkerFilter.Get().GetOrElse(""));
+                TString inputString = Settings->WorkerFilter.Get().GetOrElse("");
```
zverevgeny, line 102: Now, you make a copy of string, that has a perf penalty. Guess, that there should be a correct way of passing the string to the parser without making a copy, because all operations are in a local context

ubyte: Here is a fix that eliminates excessive copying of strings in this and many other cases: https://nda.ya.ru/t/wltpi_kf75waar

Unfortunately, I can't merge it before this PR because the compiler starts detecting a problem in the `GetPragmaFilter` function and refuses to compile it:  https://nda.ya.ru/t/QEfNSPQY75waZ6

I would prefer to use simpler code here rather than write performance-optimized code that will become unnecessary as soon as the second pull request is merged.

zverevgeny (approved):

PR #4590, opened 2024-05-16
Disable spilling in LastWideCombiner if input or state is not serializable (currently has Resource type).

Darych:
### Changelog entry 

Disable spilling in LastWideCombiner if input or state is not serializable (currently has Resource type).

### Changelog category 

* Bugfix
* 
### Additional information

...

### DIFF
--- a/ydb/library/yql/minikql/comp_nodes/mkql_wide_combine.cpp
+++ b/ydb/library/yql/minikql/comp_nodes/mkql_wide_combine.cpp
@@ -344,7 +344,7 @@ public:
         TMemoryUsageInfo* memInfo,
         const TCombinerNodes& nodes, IComputationWideFlowNode *const flow, size_t wideFieldsIndex,
         const TMultiType* usedInputItemType, const TMultiType* keyAndStateType, ui32 keyWidth,
-        const THashFunc& hash, const TEqualsFunc& equal
+        const THashFunc& hash, const TEqualsFunc& equal, bool allowSpilling
     )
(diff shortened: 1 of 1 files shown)


ydb/library/yql/minikql/comp_nodes/mkql_wide_combine.cpp
```diff
@@ -1523,6 +1530,10 @@ using TBaseComputation = TStatefulWideFlowCodegeneratorNode<TWideLastCombinerWra
 #endif
 };
 
+bool IsTypeSerializable(const TType* type) {
+    return !type->IsResource();
```
lll-phill-lll, line 1534: As far as I understand there are more non-serializable types. 
Here are all possible types: https://github.com/ydb-platform/ydb/blob/f2efea6e0b234506f42443226951d4944452ce69/ydb/library/yql/minikql/mkql_node.h#L131
And all the types that are not listed here are non-serializable: https://github.com/ydb-platform/ydb/blob/f2efea6e0b234506f42443226951d4944452ce69/ydb/library/yql/minikql/computation/mkql_computation_node_pack.cpp#L167

Also, I think we can use `ctx.Mutables.SerializableValues` to get indexes of serializable values

Darych: Thank you and I agree with you. Will add.

lll-phill-lll (approved):

PR #4715, opened 2024-05-21
Always add 'u-' prefix to the dedicated YDB database endpoint during resolving

vitalyisaev2:
* Always add 'u-' prefix to the dedicated YDB database endpoint during resolving (https://github.com/ydb-platform/ydb/pull/4688)


kardymonds (approved):

dcherednik (changes requested):
Please make an explanation about this patch. It looks like some yc specific logic. How does this code work outside Yandex?

uzhastik (approved):

dcherednik (approved):
I still do not understand why this code is in OSS. But if it ok for FQ - ok.

Issue #3275, opened 2024-03-28
add separate mon sensors for every storage type


antonkovalenko:
According to https://ydb.tech/docs/en/troubleshooting/monitoring#resources currently YDB provides public sensors for quota and usage that are combined for all storage types:

resources.storage.used_bytes
resources.storage.limit_bytes

There's an idea to add label to current sensor to distinguish storage type. Migration and numerous dashboards and alerts must be addressed. (Label might be bad idea)

In order to make separated storage consumtion and quota info available in embedded ui, let's extend DiskSpaceUsage section with consumtion by type and DatabaseQuotas section with quotas by type.

Old quotas and consumption must be kept for migration purposes.

![Image](https://github.com/ydb-platform/ydb/assets/692649/d3b52f21-39d7-4eb0-8966-0a6c2d17dbd9)

jepett0:
PR:
https://github.com/ydb-platform/ydb/pull/3798

PR #4727, opened 2024-05-21
HDD and SSD used bytes counters

jepett0:
#3275

In this PR two new counters are added:
- `resources.storage.used_bytes.ssd`
- `resources.storage.used_bytes.hdd`

Necessary subcounters like `resources.storage.used_bytes.table.ssd` (note the added `table` part) are also added. These counters are needed to monitor SSD and HDD storage usage separately. 

In the follow-up [PR](https://github.com/ydb-platform/ydb/pull/4765) `.ssd` and `.hdd` versions of `resources.storage.limit_bytes` are added too.

Probably the most interesting part of this PR is the "hacky" [mapping](https://github.com/ydb-platform/ydb/blob/4fca557e2646a1de873613694c2adfd10c780b2f/ydb/core/tx/schemeshard/schemeshard_info_types.cpp#L54) of storage pool kinds to `ssd` and `hdd` counters. It is discussed in more detail in these comments:
- https://github.com/ydb-platform/ydb/pull/4727#discussion_r1608650304
- https://github.com/ydb-platform/ydb/pull/4727#discussion_r1608656793

The data for the `used_bytes` counters is sourced from the SchemeShard's counters:
- `SchemeShard/DiskSpaceTablesTotalBytesOnSsd`
- `SchemeShard/DiskSpaceTablesTotalBytesOnHdd`

which are updated each time the SchemeShard receives TEvPeriodicTableStats from DataShards.

Currently, only row-oriented tables report disk space usage separately by channels, so effectively:
- `resources.storage.used_bytes.ssd` is equal to `resources.storage.table.used_bytes.ssd` (note the added `table` part)
- `resources.storage.used_bytes.hdd` is equal to `resources.storage.table.used_bytes.hdd` (note the added `table` part)

In the future, we will start accounting for HDD and SSD usage separately for both column-oriented tables and topics. Consequently, `resources.storage.used_bytes.ssd` will more accurately represent the total SSD space used by all entities in the database.

P.S. For the best reviewing experience, I recommend reviewing the [commits](https://github.com/ydb-platform/ydb/pull/4727/commits) individually.

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__operation_common_subdomain.h
+++ b/ydb/core/tx/schemeshard/schemeshard__operation_common_subdomain.h
@@ -6,10 +6,11 @@
 namespace NKikimr {
 namespace NSchemeShard {
 
-inline bool CheckStoragePoolsInQuotas(const Ydb::Cms::DatabaseQuotas& quotas,
-                                      const TVector<TStoragePool>& pools,
-                                      const TString& path,
-                                      TString& error
+inline bool CheckStoragePoolsInQuotas(
+    const Ydb::Cms::DatabaseQuotas& quotas,
+    const TVector<TStoragePool>& pools,
+    const TString& path,
+    TString& error
 ) {
     TVector<TString> quotedKinds;
     quotedKinds.reserve(quotas.storage_quotas_size());
--- a/ydb/core/tx/schemeshard/schemeshard__operation_common_subdomain.h
+++ b/ydb/core/tx/schemeshard/schemeshard__operation_common_subdomain.h
@@ -6,33 +6,43 @@
 namespace NKikimr {
 namespace NSchemeShard {
 
-inline bool CheckStorageQuotasKinds(const Ydb::Cms::DatabaseQuotas& quotas,
-                                    const TVector<TStoragePool>& pools,
-                                    const TString& path,
-                                    TString& error
+inline bool CheckStoragePoolsInQuotas(const Ydb::Cms::DatabaseQuotas& quotas,
+                                      const TVector<TStoragePool>& pools,
+                                      const TString& path,
+                                      TString& error
 ) {
     TVector<TString> quotedKinds;
+    quotedKinds.reserve(quotas.storage_quotas_size());
     for (const auto& storageQuota : quotas.storage_quotas()) {
         quotedKinds.emplace_back(storageQuota.unit_kind());
     }
     Sort(quotedKinds);
-    const auto uniqueEnd = Unique(quotedKinds.begin(), quotedKinds.end());
-    if (uniqueEnd != quotedKinds.end()) {
+    if (const auto equalKinds = AdjacentFind(quotedKinds);
+        equalKinds != quotedKinds.end()
+    ) {
         error = TStringBuilder()
-            << "Malformed subdomain request: storage quotas' unit kinds must be unique, but "
-            << *uniqueEnd << " appears twice in the storage quotas definition of the " << path << " subdomain.";
+            << "Malformed subdomain request: storage kinds in DatabaseQuotas must be unique, but "
+            << *equalKinds << " appears twice in the quotas definition of the " << path << " subdomain.";
         return false;
     }
 
-    for (const auto& quotedKind : quotedKinds) {
-        if (!AnyOf(pools, [&quotedKind](const TStoragePool& pool) {
-            return pool.GetKind() == quotedKind;
-        })) {
-            error = TStringBuilder()
-                << "Malformed subdomain request: cannot set a " << quotedKind << " storage quota, "
-                << "because no storage pool in the subdomain " << path << " has the specified kind.";
-            return false;
-        }
+    TVector<TString> existingKinds;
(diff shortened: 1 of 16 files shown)


ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -1577,19 +1577,21 @@ void TAggregatedStats::UpdateShardStats(TShardIdx datashardIdx, const TPartition
...
+        const auto* oldStoragePoolStats = oldStats.StoragePoolsStats.FindPtr(poolKind);
         if (aggregatedStoragePoolStats) {
-            const auto* oldStoragePoolStats = oldStats.StoragePoolsStats.FindPtr(poolKind);
-
             // Missing old stats for a particular storage pool are interpreted as if this data
             // has just been written to the datashard and we need to increment the aggregate by the entire new stats' sizes.
             aggregatedStoragePoolStats->DataSize += newStoragePoolStats.DataSize - (oldStoragePoolStats ? oldStoragePoolStats->DataSize : 0u);
             aggregatedStoragePoolStats->IndexSize += newStoragePoolStats.IndexSize - (oldStoragePoolStats ? oldStoragePoolStats->IndexSize : 0u);
         } else {
+            Y_ABORT_UNLESS(!oldStoragePoolStats, "Old stats are present, but they haven't been aggregated.");
```
jepett0, line 1604: I added this Y_ABORT_UNLESS for greater similarity with the for loop below (which I have added before).

I believe that the state of the TAggregatedStats in which:
1) the aggregated stats do not contain an entry for a particular storage pool kind
2) there is an old stats entry for this storage kind

cannot be correctly handled silently. The semantics of TEvPeriodicStats messages implies that the aggregated stats are incremented by `newStats - oldStats (or 0 if they are absent)`. If some old stats, that haven't been aggregated, are present, it signals an error in aggregation, which should be investigated.

I suppose it is better to discover these kinds of errors during testing deployment than to silently ignore and calculate an aggregate using some stale oldStats.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -50,6 +51,16 @@ EDiskUsageStatus CheckStoragePoolsQuotas(const THashMap<TString, TStoragePoolUsa
             : EDiskUsageStatus::BelowSoftQuota;
 }
 
+EUserFacingStorageType GetUserFacingStorageType(const TString& poolKind) {
+    if (poolKind.StartsWith("ssd")) {
```
jepett0, line 55: This should cover "ssd" and "ssdencrypted". This hack will certainly be documented.

ydb/core/tx/schemeshard/schemeshard_info_types.h
```diff
@@ -1281,6 +1281,12 @@ struct TSchemeQuotas : public TVector<TSchemeQuota> {
     mutable size_t LastKnownSize = 0;
 };
 
+enum class EUserFacingStorageType {
```
jepett0, line 1284: This enum highlights that we categorize numerous storage pool kinds into just two reported storage types: HDD and SSD.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
ijon: Also there should be LOUD (but short if possible) indication and explanation that this is a workaround.
That makes sense only in a limited case when there are no multiple storage-pools with the same kind and storage-pool kinds are all in (ssh*, hdd*, rot*, nvme*) set.

ydb/core/tx/schemeshard/schemeshard__operation_common_subdomain.h
```diff
@@ -6,33 +6,43 @@
...
 namespace NSchemeShard {
 
-inline bool CheckStorageQuotasKinds(const Ydb::Cms::DatabaseQuotas& quotas,
-                                    const TVector<TStoragePool>& pools,
-                                    const TString& path,
-                                    TString& error
+inline bool CheckStoragePoolsInQuotas(const Ydb::Cms::DatabaseQuotas& quotas,
+                                      const TVector<TStoragePool>& pools,
+                                      const TString& path,
+                                      TString& error
```
ijon, line 12: No need to align args to the opening brace, double indentation from the left is enough.

ijon (changes requested):

ydb/core/tx/schemeshard/schemeshard__operation_common_subdomain.h
jepett0: I hope I got your idea right

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
ijon: But also we don't want to lose entire node (with schemeshard and all other tablets) to a cycle of the same aborts. `Y_ABORT_UNLESS` are too risky in that regard.

ijon (changes requested):

ijon (approved):

PR #4734, opened 2024-05-21
Pass flat index pages to followers

kungasc:
### Changelog entry 

...

### Changelog category 

### Additional information

In order to avoid increased latency and index reading from disk on each compaction


snaury (approved):

kunga:
Requires https://github.com/ydb-platform/ydb/issues/4744 at first

PR #4736, opened 2024-05-21
YQL-17072: fix Coalesce optimization on nested Optional

fedor-miron:
### Changelog entry 

Coalesce has one of the following signatures:
```
COALESCE(T?, T)->T
COALESCE(T?, T?)->T?
```
It is incorrect to guess it's signature from checking if the last type is Optional. We have to compare argument types

### Changelog category 

* Bugfix 

### Additional information

...


fedor-miron:
Already fixed in https://github.com/ydb-platform/ydb/pull/4056/files

PR #4758, opened 2024-05-22
the race between TEvLockStatus and TEvProposeTransaction

Alek5andr-Kotov:
### Changelog entry 

The PQ tablet may lose the beginning of the data recorded in the transaction.

An auxiliary partition is created at the first message with a WriteId and is deleted either when the transaction completes or when a TEvLockStatus message arrives with a status other than STATUS_SUBSCRIBED. A situation may arise where the partition is deleted, and then the next message with the same WriteId will arrive and the recording will continue to another partition. As a result, the recorded data will be partially lost.

It is proposed to save a number in KQP for each partition in a transaction, and then send the accumulated number to TEvProposeTransaction. The identifier of the auxiliary partition can be used as such a number.

### Changelog category 

 Bugfix 


ydb/core/kqp/topics/kqp_topics.cpp
```diff
@@ -88,6 +89,8 @@ void TTopicPartitionOperations::AddOperation(const TString& topic, ui32 partitio
         Partition_ = partition;
     }
 
+    SupportivePartition_ = SupportivePartition_ ? Max<ui32>() : supportivePartition;
```
alexnick88, line 92: Is it correct? In case of different SupportivePartition_ we must set it's value to "incorrect supportive partition", not blindly rewrite.

ydb/core/persqueue/pq_impl.cpp
```diff
@@ -2648,6 +2653,14 @@ void TPersQueue::HandleEventForSupportivePartition(const ui64 responseCookie,
                                                    sender);
         }
     } else {
+        if (!firstWrite) {
+            ReplyError(ctx,
+                       responseCookie,
+                       NPersQueue::NErrorCode::BAD_REQUEST,
+                       "lost messages");
```
alexnick88, line 2660: Code means that client doing smthng wrong. But in this case it os not his fault. PRECONDITION_FAILED or some other code is preferable.

ydb/core/persqueue/writer/writer.cpp
```diff
@@ -313,6 +319,14 @@ class TPartitionWriter: public TActorBootstrapped<TPartitionWriter>, private TRl
         cmd.SetForce(true);
 
         SetWriteId(request);
+        // While waiting for a response from KQP, the recording session could
+        // send the following messages. They are in the "Pending" queue and 
+        // the FirstWrite flag will be reset
+        if (Pending) {
```
alexnick88, line 325: Why can some GetOwnership requests be with FirstWrite == false?

ydb/core/persqueue/writer/writer.cpp
```diff
@@ -344,7 +360,39 @@ class TPartitionWriter: public TActorBootstrapped<TPartitionWriter>, private TRl
...
+    }
+
+    void HandlePartitionIdSaved(NKqp::TEvKqp::TEvQueryResponse::TPtr& ev, const TActorContext&) {
+        auto& record = ev->Get()->Record.GetRef();
+        switch (record.GetYdbStatus()) {
+        case Ydb::StatusIds::SUCCESS:
+            break;
+        case Ydb::StatusIds::SESSION_BUSY:
+        case Ydb::StatusIds::PRECONDITION_FAILED: // see TKqpSessionActor::ReplyBusy
+            return Retry(record.GetYdbStatus());
```
alexnick88, line 391: Retry will lead to infinite loop. It should stop after several attempts.

ydb/core/kqp/topics/kqp_topics.cpp
Alek5andr-Kotov: Еrror has been fixed
https://github.com/ydb-platform/ydb/pull/4758/commits/32cbf0b47cb8ad5df42195cd5d5eab7448e0d71b

ydb/core/persqueue/writer/writer.cpp
Alek5andr-Kotov: Fixed
https://github.com/ydb-platform/ydb/pull/4758/commits/7b56087e43f8fe11e87371e5f3d79292aff5717a

ydb/core/persqueue/writer/writer.cpp
Alek5andr-Kotov: Here I rely on the fact that the KQP session lives for a limited time

ydb/core/persqueue/pq_impl.cpp
Alek5andr-Kotov: Fixed
https://github.com/ydb-platform/ydb/pull/4758/commits/083ea2518891ac980926be5e9c317679380ec934

ydb/public/api/protos/draft/persqueue_error_codes.proto
```diff
@@ -50,5 +50,7 @@ enum EErrorCode {
 
     UNKNOWN_TXID = 30;
 
+    PRECONDITION_FAILED = 31;
```
alexnick88, line 53: New code. In Write session this code is converted to YdbStatus. You need to update this code too.

alexnick88 (dismissed):

ydb/public/api/protos/draft/persqueue_error_codes.proto
Alek5andr-Kotov: it is in the file `persqueue_utils.cpp`

qyryq (approved):

PR #4763, opened 2024-05-22
only give counter resource to tablets that never had normal metrics

vporyadke:
### Changelog entry 

only give counter resource to tablets that never had normal metrics

### Changelog category 

* Improvement

### Additional information

Counter-balancing was introduced with columnshards in mind and may be undesirable for other tablets.

### DIFF
--- a/ydb/core/mind/hive/hive_ut.cpp
+++ b/ydb/core/mind/hive/hive_ut.cpp
@@ -4672,6 +4672,16 @@ Y_UNIT_TEST_SUITE(THiveTest) {
         auto newDistribution = getDistribution();
 
         UNIT_ASSERT_EQUAL(initialDistribution, newDistribution);
+
+        {
+            auto request = std::make_unique<TEvHive::TEvRequestHiveDomainStats>();
+            request->Record.SetReturnMetrics(true);
+            runtime.SendToPipe(hiveTablet, senderA, request.release());
+            TAutoPtr<IEventHandle> handle;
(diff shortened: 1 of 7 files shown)


ydb/core/mind/hive/hive_ut.cpp
```diff
@@ -4594,6 +4594,86 @@ Y_UNIT_TEST_SUITE(THiveTest) {
...
+        }
+
+        {
+            TDispatchOptions options;
+            options.FinalEvents.emplace_back(NHive::TEvPrivate::EvBalancerOut);
+            runtime.DispatchEvents(options, TDuration::Seconds(10));
+        }
+
+        // Check that balancer moved no tablets
+        auto newDistribution = getDistribution();
```
adameat, line 4672: maybe it could be wise to also check the sum of all resources to see that the counter is zero. of course if it easy to obtain the sum

ydb/core/util/metrics.h
```diff
@@ -395,6 +395,9 @@ class TMaximumValueVariableWindowUI64 : public NKikimrMetricsProto::TMaximumValu
     using TProto = NKikimrMetricsProto::TMaximumValueUI64;
 
     void SetValue(TType value, TInstant now = TInstant::Now()) {
+        if (TProto::GetAllTimeMaximum() > 0 || MaximumValue > 0) { // ignoring initial value
```
adameat, line 398: why won't you put it into the end of the function?

adameat (dismissed):

adameat (approved):

PR #4765, opened 2024-05-23
Add HDD and SSD limit_bytes counters

jepett0:
https://github.com/ydb-platform/ydb/issues/3275

This is a follow-up PR for the https://github.com/ydb-platform/ydb/pull/4727.

In this PR two new counters are added:
- `resources.storage.limit_bytes.ssd`
- `resources.storage.limit_bytes.hdd`

The most important things to note are:
- `limit_bytes` metric corresponds to the SOFT database quota. Hard quota is not reported via metrics at all (as far as I know).
- Consider the following situation:
a database has "ssd" and "ssdencrypted" storage pools and the DatabaseQuotas look like this:
```proto
DatabaseQuotas {
    storage_quotas {
        unit_kind: "ssd"
        data_size_hard_quota: 2
        data_size_soft_quota: 1
    }
    storage_quotas {
        unit_kind: "ssdencrypted"
        data_size_hard_quota: 20
        data_size_soft_quota: 10
    }
}
```
The question is: "What will the value of `resources.storage.limit_bytes.ssd` be?"

The answer is 11, because the quotas for all storage pools with kinds prefixed by "ssd" are summed up to determine the value of `resources.storage.limit_bytes.ssd`.

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard_info_types.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_info_types.cpp
@@ -174,6 +174,26 @@ bool TSubDomainInfo::CheckDiskSpaceQuotas(IQuotaCounters* counters) {
     return false;
 }
 
+void TSubDomainInfo::CountDiskSpaceQuotas(IQuotaCounters* counters, const TDiskSpaceQuotas& quotas) {
+    if (quotas.HardQuota != 0) {
+        counters->ChangeDiskSpaceHardQuotaBytes(quotas.HardQuota);
+    }
+    if (quotas.SoftQuota != 0) {
+        counters->ChangeDiskSpaceSoftQuotaBytes(quotas.SoftQuota);
+    }
+}
+
+void TSubDomainInfo::CountDiskSpaceQuotas(IQuotaCounters* counters, const TDiskSpaceQuotas& prev, const TDiskSpaceQuotas& next) {
+    i64 hardDelta = i64(next.HardQuota) - i64(prev.HardQuota);
+    if (hardDelta != 0) {
+        counters->ChangeDiskSpaceHardQuotaBytes(hardDelta);
+    }
+    i64 softDelta = i64(next.SoftQuota) - i64(prev.SoftQuota);
+    if (softDelta != 0) {
+        counters->ChangeDiskSpaceSoftQuotaBytes(softDelta);
+    }
+}
+
 void TSubDomainInfo::AggrDiskSpaceUsage(IQuotaCounters* counters, const TPartitionStats& newAggr, const TPartitionStats& oldAggr) {
     DiskSpaceUsage.Tables.DataSize += (newAggr.DataSize - oldAggr.DataSize);
     counters->ChangeDiskSpaceTablesDataBytes(newAggr.DataSize - oldAggr.DataSize);
--- a/ydb/core/tx/schemeshard/schemeshard_info_types.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_info_types.cpp
@@ -181,6 +181,11 @@ void TSubDomainInfo::CountDiskSpaceQuotas(IQuotaCounters* counters, const TDiskS
     if (quotas.SoftQuota != 0) {
         counters->ChangeDiskSpaceSoftQuotaBytes(quotas.SoftQuota);
     }
+    for (const auto& [poolKind, poolQuotas] : quotas.StoragePoolsQuotas) {
+        if (poolQuotas.SoftQuota != 0) {
+            counters->ChangeDiskSpaceSoftQuotaBytes(GetUserFacingStorageType(poolKind), poolQuotas.SoftQuota);
+        }
+    }
 }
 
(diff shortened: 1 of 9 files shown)


ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -186,6 +191,23 @@ void TSubDomainInfo::CountDiskSpaceQuotas(IQuotaCounters* counters, const TDiskS
     if (softDelta != 0) {
         counters->ChangeDiskSpaceSoftQuotaBytes(softDelta);
     }
+    for (const auto& [poolKind, newPoolQuotas] : next.StoragePoolsQuotas) {
+        const auto* oldPoolQuotas = prev.StoragePoolsQuotas.FindPtr(poolKind);
+        i64 delta = i64(newPoolQuotas.SoftQuota) - (oldPoolQuotas ? i64(oldPoolQuotas->SoftQuota) : 0);
+        if (delta != 0) {
```
ijon, line 197: Expression `i64 delta = i64(a) - i64(b)` will have unexpected result if `a > INT64_MAX` and `b < INT64_MAX` (and vice versa). So all that delta calculations effectively means that quota values must not be set to values greater than INT64_MAX. And this is something that users would never expect.

I think its time to right the wrong and change all delta expressions to `i64 delta = a - b`. It is the least that we could do here.

ijon (changes requested):

ydb/tests/functional/tenants/test_db_counters.py
```diff
@@ -129,3 +129,239 @@ def test_case(self):
...
+            "unit_kind": "hdd1",
+            "data_size_hard_quota": "20",
+            "data_size_soft_quota": "10",
+        }
+
+        slot_mon_port = self.cluster.slots[1].mon_port
+        # Note 1: limit_bytes is equal to the database's SOFT quota
+        # Note 2: .hdd counter aggregates quotas across all storage pool kinds with prefix "hdd", i.e. "hdd" and "hdd1"
+        # Note 3: 200 seconds can sometimes be not enough
+        self.check_counters(slot_mon_port, {"resources.storage.limit_bytes.hdd": 11}, retries=60, sleep_duration=5)
```
jepett0, line 303: Please make sure you are ok with this

jepett0: Also, it is important to note that the wait time must be at least 200 seconds according to my tests on a heavily loaded local machine. It is already equal to [150 seconds](https://github.com/ydb-platform/ydb/blob/f835e78f443414d6158c60b63c9e790e763b55ec/ydb/tests/functional/tenants/test_db_counters.py#L119) for the other (previously existing) test in this file.

ijon: > Please make sure you are ok with the reported value being 11, i.e. a sum of the soft quotas of "hdd" and "hdd1" storage pools.

That is ok for now.

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -174,6 +174,48 @@ bool TSubDomainInfo::CheckDiskSpaceQuotas(IQuotaCounters* counters) {
...
+            counters->ChangeDiskSpaceSoftQuotaBytes(GetUserFacingStorageType(poolKind), delta);
+        }
+    }
+    for (const auto& [poolKind, oldPoolQuotas] : prev.StoragePoolsQuotas) {
+        if (const auto* newPoolQuotas = next.StoragePoolsQuotas.FindPtr(poolKind);
+            !newPoolQuotas
+        ) {
+            i64 delta = -i64(oldPoolQuotas.SoftQuota);
+            if (delta != 0) {
+                counters->ChangeDiskSpaceSoftQuotaBytes(GetUserFacingStorageType(poolKind), delta);
```
ijon, line 213: This will not get expected result either.
Instead of addition of the negative value, there should be proper subtraction of the positive value.

ydb/tests/functional/tenants/test_db_counters.py
```diff
@@ -129,3 +129,239 @@ def test_case(self):
...
+        logger.debug(f"Create database {self.database}")
+        self.cluster.create_database(self.database, storage_pool_units_count={"hdd": 1, "hdd1": 1})
+
+        self.cluster.register_and_start_slots(self.database, count=1)
+        self.cluster.wait_tenant_up(self.database)
+
+    def teardown_method(self, method=None):
+        logger.debug(f"Remove database {self.database}")
+        self.cluster.remove_database(self.database)
+        self.database = None
```
ijon, line 146: Consider switching to using fixtures from `ydb/tests/library/harness/ydb_fixtures.py`. And switching to standalone test functions.

ijon (changes requested):

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
jepett0: I changed the signature of this function to accept ui64 as a parameter. It should make all the calculations in ui64 and if we assume that all the deltas are correct, then it should never be "negative" (i.e. a very large unsigned number) and should always display correct and expected value.

ijon (dismissed):

ijon (approved):

PR #4774, opened 2024-05-23
Make max commit size configurable with icb

### LINKED ISSUE #4766: Add ICB control for max commit size
We sometimes encounter bugs where tablets try to commit a very large redo log and fail due to a hard-coded limit of 256MB. When that happens we have to build a special build that raises this limit temporarily, which takes a lot of time and effort. We need to add this setting to immediate control board, so we can tune it for a specific node/database without rebuilds and restarts.

snaury:
### Changelog entry 

Make max commit size configurable with ICB.

### Changelog category 

### Additional information

Fixes #4766.

### DIFF
--- a/ydb/core/protos/config.proto
+++ b/ydb/core/protos/config.proto
@@ -1256,6 +1256,14 @@ message TImmediateControlsConfig {
             DefaultValue: 10 }];
     }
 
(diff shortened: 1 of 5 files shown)


the-ancient-1 (approved):
LGTM

PR #4791, opened 2024-05-23
Fix ReadLogContinue race on stop

alexvru:
### Changelog entry 

Fix ReadLogContinue race on stop

### Changelog category 

* Bugfix 

### Additional information

Plain pointer was replaced by shared/weak pointer.


SammyVimes (approved):
LGTM!

PR #4796, opened 2024-05-23
YQL-18419 create callback that checks RSS and limits memory usage

lll-phill-lll:
### Changelog entry 
After this pr it will be possible to set an RSS limit for the process and enable spilling according to current RSS value. 
...

### Changelog category 

* New feature

### Additional information

...

### DIFF
--- a/ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
+++ b/ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
@@ -131,7 +131,6 @@ namespace NYql::NDq {
 
     private:
         void RequestExtraMemory(ui64 memory, NKikimr::NMiniKQL::TScopedAlloc* alloc) {
-            alloc->GetAllocated();
             memory = std::max(AlignMemorySizeToMbBoundary(memory), MemoryLimits.MinMemAllocSize);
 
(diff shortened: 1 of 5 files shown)


ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
```diff
@@ -55,6 +56,23 @@ namespace NYql::NDq {
             }
         }
 
+        // This callback is created for testing purposes and will be enabled only with spilling.
+        // Most likely this callback will be removed after KIKIMR-21481.
+        void TrySetIncreaseMemoryLimitCallbackWithRSSControl(NKikimr::NMiniKQL::TScopedAlloc* alloc) {
+            if (!CanAllocateExtraMemory) return;
+            const ui64 limitRSS = std::numeric_limits<ui64>::max();
```
Darych, line 63: Why do we use max ui64?

Darych (approved):

ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
lll-phill-lll: This value is expected to be changed manually while testing. Here I used max just as a placeholder

Issue #4643, opened 2024-05-17
Use per-node pipe cache for readset exchanges


snaury:
Currently datashards use private pipe caches for sending readsets to other shards, which as it turned out had a bug where it would reopen pipes after any inactivity instead of caching them. Connects to tablets use a handshake with several roundtrips, which should be avoided as much as possible. And ideally there should be no private per tablet caches, and instead there should be a per-node cache that basically keeps connections forever. Such a cache already exists, but it's currently mostly used for requests and transaction proposals.

Unfortunately current per-node cache cannot be used, because it has very small connection timeout with almost no retries, while persistent messages need a more robust channels that don't fail too fast and often. Moreover, persistent messages need dead tablet detection, which complicates things.

I think we need to add a special persistent mode for per-node cache forwarding, which would use robust channels with dead tablet detection, and then switch datashards to use that channel.

PR #4798, opened 2024-05-23
Use persistent pipe cache for readset exchanges

snaury:
### Changelog entry 

Using the per-node pipe cache allows multiple shards to reuse the same connection to other shards.

### Changelog category 

### Additional information

Part of #4643.

### DIFF
--- a/ydb/core/base/tablet_pipecache.h
+++ b/ydb/core/base/tablet_pipecache.h
@@ -74,13 +74,23 @@ struct TEvPipeCache {
 
     struct TEvDeliveryProblem : public TEventLocal<TEvDeliveryProblem, EvDeliveryProblem> {
         const ui64 TabletId;
+        const bool Connected;
         const bool NotDelivered;
         const bool IsDeleted;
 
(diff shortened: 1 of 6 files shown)


ydb/core/base/tablet_pipecache.h
```diff
@@ -74,14 +74,24 @@ struct TEvPipeCache {
 
     struct TEvDeliveryProblem : public TEventLocal<TEvDeliveryProblem, EvDeliveryProblem> {
         const ui64 TabletId;
+        const NKikimrProto::EReplyStatus ConnectStatus;
```
azevaykin, line 77: More grammatically correct: ConnectionStatus

snaury: Well, it's the status of an initial connect, not of any connection.

snaury: But I'm going to replace it with `bool Connected` anyway.

azevaykin (dismissed):

azevaykin (approved):

PR #5063, opened 2024-05-31
Add TTL to CREATE and ALTER TABLE

Pseudolukian:
### Changelog category 


ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/alter_table.md
```diff
@@ -220,6 +220,18 @@ For example, this command disables automatic partitioning of the table:
...
+ALTER TABLE table_name 
+SET (TTL = Interval("PT10S") ON column_name);
+```
+
+If you need to remove the `TTL` from a table, you can do so with the following command:
+```sql
+ALTER TABLE table_name
+RESET (TTL);
+```
+
```
blinkov, line 234: There's already the same example below on this page:
Screenshot 2024-06-03 at 07 26 05

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/alter_table.md
```diff
@@ -220,6 +220,18 @@ For example, this command disables automatic partitioning of the table:
 ALTER TABLE series SET (AUTO_PARTITIONING_BY_SIZE = DISABLED);
 ```
 
+Using the `ALTER` command, you can also modify the `TTL` parameters of the table. For example, to increase the delay interval for record deletion:
```
blinkov, line 223:
```suggestion
Using the `ALTER TABLE` command, you can also modify the table's [TTL](../../../../concepts/ttl.md) parameters. For example, to increase the delay interval for record deletion:
```

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/alter_table.md
```diff
@@ -220,6 +220,18 @@ For example, this command disables automatic partitioning of the table:
...
+```sql
+ALTER TABLE table_name 
+SET (TTL = Interval("PT10S") ON column_name);
```
blinkov, line 226: This example doesn't make sense as TTL works asynchronously and won't clean up data that fast. Usually, it's in like days, not seconds.

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/create_table.md
```diff
@@ -143,9 +143,9 @@ Here, key is the name of the parameter and value is its value.
 
 For a list of valid parameter names and values, see the [{{ backend_name }} table description]({{ concept_table }}).
 
-For example, this code will create a table with enabled automatic partitioning by partition size and the preferred size of each partition is 512 MB:
+### Examples of Creating a String Table with Additional Parameters
```
blinkov, line 146:
```suggestion
### Examples of creating a table with additional parameters
```
This is a common way to write headers, but we're not currently using it in other places.

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/create_table.md
```diff
@@ -143,9 +143,9 @@ Here, key is the name of the parameter and value is its value.
...
 
-<small>Listing 4</small>
+This code will create a table with enabled automatic partitioning by partition size and the preferred size of each partition is 512 MB:
```
blinkov, line 148:
```suggestion
This code will create a table with enabled automatic partitioning by size and the preferred partition size of 512 MB:
```

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/create_table.md
```diff
@@ -159,6 +159,28 @@ WITH (
 );
 ```
 
+Creating a table with enabled [TTL](../../../../concepts/ttl.md) for automatically removing outdated records from the table. A record is deleted when the value of one of the columns (configurable by the user) of type: `Date`, `Datetime`, or `Timestamp` becomes greater than the current time (Epoch time). The lifespan of records can be extended (globally for the entire table) by the value specified in the `Interval` function:
```
blinkov, line 162:
```suggestion
Creating a table with enabled [TTL](../../../../concepts/ttl.md) to asynchronously remove outdated records from the table automatically. {{ ydb-short-name }} periodically checks if each record needs to be deleted by checking if the value of a specified column plus the provided interval is greater than the current time:
```

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/create_table.md
```diff
@@ -159,6 +159,28 @@ WITH (
...
+    title Utf8,
+    expire_at Timestamp
+    PRIMARY KEY (id)
+)
+WITH (
+    TTL = Interval("PT0S") ON expire_at
+);
+```
+
+In the provided example, a record will be deleted as soon as the value in the `expire_at` column matches the Epoch time, since the Interval function specifies the value "PT0S" is 0 seconds.
```
blinkov, line 175:
```suggestion
In the provided example, the specified interval is `"PT0S"`, which according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), means zero seconds. Thus, records will be deleted when the value in the `expire_at` column exceeds the current Epoch time.
```

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/create_table.md
```diff
@@ -159,6 +159,28 @@ WITH (
...
+Creating a table with enabled [TTL](../../../../concepts/ttl.md) for automatically removing outdated records from the table. A record is deleted when the value of one of the columns (configurable by the user) of type: `Date`, `Datetime`, or `Timestamp` becomes greater than the current time (Epoch time). The lifespan of records can be extended (globally for the entire table) by the value specified in the `Interval` function:
+```sql
+CREATE TABLE my_table (
+    id Uint64,
+    title Utf8,
+    expire_at Timestamp
+    PRIMARY KEY (id)
+)
+WITH (
+    TTL = Interval("PT0S") ON expire_at
```
blinkov, line 171: This is a corner case example. Let's add a more typical one first. For example, there's a column with an event time (timestamp at the moment of insert), and we want to discard events older than one week.

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/create_table.md
```diff
@@ -159,6 +159,28 @@ WITH (
...
+);
+```
+
+In the provided example, a record will be deleted as soon as the value in the `expire_at` column matches the Epoch time, since the Interval function specifies the value "PT0S" is 0 seconds.
+
+{% note info %}
+
+In the table, there can be only one TTL column. If in the TTL column has a `NULL` value, that row will never be deleted.
```
blinkov, line 179:
```suggestion
There can be only one TTL column in a table. If the TTL column has a `NULL` value, that row will never be deleted.
```

blinkov, line 179: In this patch, it looks like "rows" and "records" are used as synonyms. Let's stick to one of them to minimize the risk of confusion.

blinkov (changes requested):

ydb/docs/en/core/yql/reference/yql-core/syntax/_includes/create_table.md
blinkov, line 171: According to https://ydb.tech/docs/en/concepts/ttl, there's also an option to use an integer column and specify in which time unit to interpret values:

> The value in the TTL column with a numeric type (Uint32, Uint64, or DyNumber) is interpreted as a [Unix time](https://en.wikipedia.org/wiki/Unix_time) value. The following units are supported (set in the TTL settings):
> 
> - Seconds.
> - Milliseconds.
> - Microseconds.
> - Nanoseconds.

It would be helpful to cover this case, too.

blinkov:
Fixed in other PRs

Issue #4611, opened 2024-05-16
[viewer api] quota and consumption for hdd storage types must be returned


antonkovalenko:
As a follow-up for [#3430](https://github.com/ydb-platform/ydb/issues/3430)

1. Info about unused storage types must not be returned in `tenantInfo/storageUsage` response.
2. Currently database `cold-visits` with HDD storage without storage database quotas doesn't have HDD info in `tenantInfo/storageUsage` response. But must have info about quotas and consumption of attached HDD storage

![Image](https://github.com/ydb-platform/ydb/assets/692649/29118753-e15c-463f-a4bf-c9bf02de305f)

StekPerepolnen:
"StorageUsage": [
     {
        "Type": "SSD",
        "Size": "12345", // SSD Blob storage
        "Limit": "12345", // SSD Blob storage limit
        "SoftQuota": "12345", // SSD Blob storage limit
        "HardQuota": "12345", // SSD Blob storage limit
    },
    {
        "Type": "HDD",
        "Size": "12345", // SSD Blob storage
        "Limit": "12345", // SSD Blob storage limit
        "SoftQuota": "12345", // SSD Blob storage limit
        "HardQuota": "12345", // SSD Blob storage limit
    },
   {
        "Type": "None",
        "Size": "12345", // SSD Blob storage
        "Limit": "12345", // SSD Blob storage limit
        "SoftQuota": "12345", // SSD Blob storage limit
        "HardQuota": "12345", // SSD Blob storage limit
    }
]

StekPerepolnen:
I would like all the logic for calculating values for the `Size` and `Limit` to remain within the handler. Currently, HardQuota is always placed there. 
@artemmufazalov Could you clarify in which cases other values might be used?

artemmufazalov:
@StekPerepolnen 

> I'm going to remove `QuotaUsage` field like before and combine info from it with `StorageUsage` instead I would suggest that `Size` and `Limit` will go to the main Info card, and quotas will go to the details.

`StorageUsage` and `QuotaUsage` not only have different quotas, but have different current size. We need both sizes and both limits for extended info. Don't merge them, please

> @artemmufazalov Could you clarify in which cases other values might be used?

Recently, we agreed to use soft quota, so our diagnostics will match with data that ydb provides to monitoring services (soft quota as limit). So we need soft quota as limit as well.

[...]

PR #5069, opened 2024-05-31
quota and consumption for hdd storage types

StekPerepolnen:
### Changelog entry 

https://github.com/ydb-platform/ydb/issues/4611#issuecomment-2141624326
I'm going to remove QuotaUsage field like before and combine info from it with StorageUsage instead
I would suggest that Size and Limit will go to the main Info card, and quotas will go to the details.

If there is information about size, return storage information.
If there are quotas, return limit and quotas.

The pool type is specified in the config. It's a String that don't have to be like SSD or HDD. So I added the block None in case the name is not parsed correctly - this usage and quotas will go to None.

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/core/viewer/json_tenantinfo.h
+++ b/ydb/core/viewer/json_tenantinfo.h
@@ -55,6 +55,11 @@ class TJsonTenantInfo : public TViewerPipeClient<TJsonTenantInfo> {
     TString RootId; // id of root domain (tenant)
     NKikimrViewer::TTenantInfo Result;
 
+    struct TStorageQuota {
+        uint64 SoftQuota = 0;
+        uint64 HardQuota = 0;
+    };
+
 public:
     static constexpr NKikimrServices::TActivity::EType ActorActivityType() {
         return NKikimrServices::TActivity::VIEWER_HANDLER;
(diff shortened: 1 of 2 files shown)


ydb/core/viewer/json_tenantinfo.h
```diff
@@ -628,24 +645,32 @@ class TJsonTenantInfo : public TViewerPipeClient<TJsonTenantInfo> {
...
+                        auto& usage = storageUsageByType[type];
+                        usage.SoftQuota += quota.data_size_soft_quota();
+                        usage.HardQuota += quota.data_size_hard_quota();
+                    }
+
+                    for (const auto& [type, usage] : storageUsageByType) {
+                        auto& storageUsage = *tenant.AddStorageUsage();
+                        storageUsage.SetType(type);
+                        storageUsage.SetSize(usage.Size);
+                        storageUsage.SetLimit(usage.HardQuota);
```
adameat, line 670: what will happen here if we don't have storage quotas at all?

adameat (changes requested):

adameat (approved):

PR #5191, opened 2024-06-05
require admin access for force operations

adameat:
require admin access for force operations

* Improvement

### DIFF
--- a/ydb/core/viewer/viewer.cpp
+++ b/ydb/core/viewer/viewer.cpp
@@ -198,6 +198,92 @@ public:
     TString GetHTTPFORBIDDEN(const NMon::TEvHttpInfo* request) override;
     TString GetHTTPNOTFOUND(const NMon::TEvHttpInfo* request) override;
 
+    bool CheckAccessAdministration(const NMon::TEvHttpInfo* request) override {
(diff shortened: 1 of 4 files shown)


ydb/core/viewer/viewer.cpp
```diff
@@ -198,6 +198,92 @@ class TViewer : public TActorBootstrapped<TViewer>, public IViewer {
     TString GetHTTPFORBIDDEN(const NMon::TEvHttpInfo* request) override;
     TString GetHTTPNOTFOUND(const NMon::TEvHttpInfo* request) override;
 
+    bool CheckAccessAdministration(const NMon::TEvHttpInfo* request) override {
+        if (!KikimrRunConfig.AppConfig.GetDomainsConfig().GetSecurityConfig().GetEnforceUserTokenRequirement()) {
```
StekPerepolnen, line 202: the conditions are duplicated here. It looks strange to me `!KikimrRunConfig.AppConfig.GetDomainsConfig().GetSecurityConfig().GetEnforceUserTokenCheckRequirement()`
if the token should not be verified, is it so important to look at its presence? `request->UserToken.empty()`

adameat: GetEnforceUserTokenRequirement() and GetEnforceUserTokenCheckRequirement() are two different options

StekPerepolnen (approved):

PR #5197, opened 2024-06-05
YQL-18516 increase number of buckets in GraceJoin 32->128

lll-phill-lll:
### Changelog entry 

...

### Changelog category 
* Performance improvement

### Additional information

...


yumkam:
There are a number of `O(nbuckets^2)` code out there; it is fine (or even more efficient) with small nbuckets, but will affect things with larger nbuckets (but that probably follow.

PR #5266, opened 2024-06-06
Do not write an error if ydbd not present

maximyurchuk:
### Changelog entry 
Such error in logs may be confusing

### Changelog category 


.github/scripts/send_build_stats.py
```diff
@@ -59,7 +59,7 @@ def main():
     if not os.path.exists(YDBD_PATH):
         # can be possible due to incremental builds and ydbd itself is not affected by changes
         print("{} not exists, skipping".format(YDBD_PATH))
```
uzhastik, line 61: правильнее так: does not exist

uzhastik (approved):

PR #5336, opened 2024-06-07
Fix non-ASCIIZ input for pcreposix

igormunkin:
StringRef data might not be a NTBS, though the function `<TRegExMatch::Match>` expects ASCIIZ string. As a result of the patch the given argument string is explicitly copied and NUL terminator is appended to it.

Unfortunately, several existing tests reveals the error, but the canon results were invalid.

### Changelog category

* Bugfix

### DIFF
--- a/ydb/library/yql/udfs/common/hyperscan/hyperscan_udf.cpp
+++ b/ydb/library/yql/udfs/common/hyperscan/hyperscan_udf.cpp
@@ -201,7 +201,10 @@ namespace {
             }
 
             if (args[0]) {
-                const std::string_view input(args[0].AsStringRef());
+                // XXX: StringRef data might not be a NTBS, though the function
+                // <TRegExMatch::Match> expects ASCIIZ string. Explicitly copy
+                // the given argument string and append the NUL terminator to it.
+                const TString input(args[0].AsStringRef());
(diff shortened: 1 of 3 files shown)


ydb/library/yql/udfs/common/hyperscan/hyperscan_udf.cpp
```diff
@@ -201,7 +201,13 @@ namespace {
             }
 
             if (args[0]) {
-                const std::string_view input(args[0].AsStringRef());
+                // XXX: StringRef data might not be a NTBS, though the function
+                // <TRegExMatch::Match> expects ASCIIZ string. Explicitly copy
+                // the given argument string and append the NUL terminator to it.
+                // See YQL-15189 for more info.
+                const TStringRef inputRef = args[0].AsStringRef();
+                TString input(inputRef.Data(), inputRef.Size() + 1);
```
vitstn, line 209: TString input(inputRef.Data(), inputRef.Size()) is enough

igormunkin: Fixed, thanks.

vitstn (approved):

PR #5690, opened 2024-06-18
Use NO_EXPORT_DYNAMIC_SYMBOLS() for ydbd in order to reduce binary size

maximyurchuk:
### Changelog entry 

See info in 
https://wiki.yandex-team.ru/kikimr/ydb-qa/build-time-size-ideas/#inprogressnoexportdynamicsymbols
https://clubs.at.yandex-team.ru/arcadia/26226 

This change removes exporting of all symbols in binary 

### Changelog category 

### Additional information

Stack trace capturing is still works, see check in https://github.com/ydb-platform/ydb/pull/5706 and ydb/tests/functional/scheme_shard/test_alter_ops.py test


Enjection (approved):
Tried this branch locally, seem's that everything works. The only considerations I have is about some potential weird usages inside arcadia, I would try to apply this commit there first and see whether something will be broken or not.

PR #5721, opened 2024-06-19
grace_join: fix mixup ("","foo") == ("foo", "") in string key matches

yumkam:
YQL-18530

tpc ds 1/10 q87 produced incorrect results because ("foo","",42) incorrectly matched ("","foo",42) key tuples

### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
+++ b/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
@@ -33,6 +33,9 @@ void TTable::AddTuple(  ui64 * intColumns, char ** stringColumns, ui32 * strings
     // Processing variable length string columns
     if ( NumberOfKeyStringColumns != 0 || NumberOfKeyIColumns != 0) {
 
+        totalBytesForStrings += sizeof(ui32)*NumberOfKeyStringColumns;
+        totalBytesForStrings += sizeof(ui32)*NumberOfKeyIColumns;
+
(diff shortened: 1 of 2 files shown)


ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
```diff
@@ -55,11 +58,15 @@ void TTable::AddTuple(  ui64 * intColumns, char ** stringColumns, ui32 * strings
         char * currStrPtr = reinterpret_cast< char* > (startPtr);
 
         for( ui64 i = 0; i < NumberOfKeyStringColumns; i++) {
+            WriteUnaligned<ui32>(currStrPtr, stringsSizes[i] );
```
lll-phill-lll, line 61: nit: It would be nice to change `sizeof(ui32)` to `sizeof(<variable name>)` here and in other places. In the future, changing the type of a variable may lead to errors that are difficult to detect

lll-phill-lll (approved):

ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
yumkam: Hmm. Given that this is serialization that must match on pack/unpack side, that can be some `TStringOffset`. Using type of variable feels equally fragile (e.g. `auto variable = foo.size()` would be sleeping disaster).

PR #5738, opened 2024-06-19
YQ-2068 cache lookup results

zverevgeny:


yumkam:
This likely belongs to some follow-up patch: `TUnboxedKeyValueLruCacheWithTtl` relies on garbage collection by periodic calls to `.Tick()` method, it looks like not called anywhere currently.

PR #5743, opened 2024-06-19
Add allow unknown fields to CMS

pixcc:
### Changelog category 

### Additional information

This change allows CMS to ignore unknown fields when reading them from the local database. Overwriting the data will cause the unknown fields to be lost.

### DIFF
--- a/ydb/core/cms/cms_tx_load_state.cpp
+++ b/ydb/core/cms/cms_tx_load_state.cpp
@@ -91,7 +91,10 @@ public:
             request.Owner = owner;
             request.Order = order;
             request.Priority = priority;
-            google::protobuf::TextFormat::ParseFromString(requestStr, &request.Request);
+
+            google::protobuf::TextFormat::Parser parser;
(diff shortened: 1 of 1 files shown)


ydb/core/cms/cms_tx_load_state.cpp
```diff
@@ -9,6 +9,21 @@
...
+
+namespace {
+
+template<typename T>
+T ParseFromString(const TString& str) {
+    google::protobuf::TextFormat::Parser parser;
+    parser.AllowUnknownField(true);
+    T result;
+    parser.ParseFromString(str, &result);
+    return result;
```
CyberROFL, line 22: Я бы сохранил семантику `ParseFromString`, принимая `T` по ссылке/указателю:
```suggestion
bool ParseFromStringSafe(const TString& input, T* output) {
    google::protobuf::TextFormat::Parser parser;
    parser.AllowUnknownField(true);
    return parser.ParseFromString(str, output);
```

CyberROFL (approved):

PR #5796, opened 2024-06-20
grace join: reduce hash collisions

yumkam:
Low bits of hash are already used as bucket index, so all low bits of hashes in single buckets is same. This results in higher collision rate and 6x to 7x number of comparision per access (simulated)

### Changelog entry 

...

### Changelog category 

* Performance improvement

### Additional information

...

### DIFF
--- a/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
+++ b/ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
@@ -244,7 +244,7 @@ void ResizeHashTable(KeysHashTable &t, ui64 newSlots){
         if ( *it == 0)
             continue;
         ui64 hash = *it;
-        ui64 newSlotNum = (hash / NumberOfBuckets) % (newSlots);
+        ui64 newSlotNum = hash % (newSlots);
         auto newIt = newTable.begin() + t.SlotSize * newSlotNum;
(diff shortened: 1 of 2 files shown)


ydb/library/yql/minikql/comp_nodes/mkql_grace_join_imp.cpp
```diff
@@ -359,7 +359,7 @@ void TTable::Join( TTable & t1, TTable & t2, EJoinKind joinKind, bool hasMoreLef
             if (!HasBitSet(nullsPtr, 1))
             {
 
-                ui64 slotNum = hash % nSlots;
+                ui64 slotNum = (hash / NumberOfBuckets) % nSlots;
```
lll-phill-lll, line 362: Maybe `hash << BitsForNumberOfBuckets % nSlots`?

yumkam: On one hand, compiler optimizes divide-by-constant with proper combination of shift/muls/etc.
On other, if we make `BitsForNumberOfBuckets` runtime-configurable, compiler won't be able to optimize it.

In the end, I decided it is better to solve the problem differently, and choose `nSlots` to avoid collisions instead (this way, we need to spend more time once-per-resize instead of once-per-access, even if it is as trivial as bit-shift)

PR #5805, opened 2024-06-20
Make error for impossible TzDate casts more friendly

igormunkin:
There are several TzDatetime or TzTimestamp values, that cannot be casted to TzDate. Consider the example below:
```
select cast(TzDatetime("1970-01-01T23:59:59,Europe/Moscow") as TzDate)
```

There is no valid value, starting from the Unix epoch, to represent the 01/01/1970 midnight for Europe/Moscow timezone. Hence, this conversion is impossible and will fail at runtime.

As a result of the patch, the exception raised while building the the TzDate value is handled in UDF and turned into the user-friendly error with some diagnostics.

### Changelog category

* Bugfix


vitstn (approved):

ydb/library/yql/udfs/common/datetime2/datetime_udf.cpp
```diff
@@ -694,9 +694,16 @@ NUdf::TUnboxedValuePod DoAddYears(const NUdf::TUnboxedValuePod& date, i64 years,
     SIMPLE_STRICT_UDF(TMakeTzDate, TTzDate(TAutoMap<TResource<TMResourceName>>)) {
```
rvu1024, line 694: Она уже не strict получается

PR #5826, opened 2024-06-21
Keep track of dynamic proxies too and broadcast messages to them

alexvru:
### Changelog entry 

Keep track of dynamic proxies too and broadcast messages to them

### Changelog category 

* Bugfix 

### Additional information

This allows correct stopping of all proxies, including ones attached to dynamic nodes or to static nodes that were added online.

### DIFF
--- a/ydb/library/actors/core/actorsystem.cpp
+++ b/ydb/library/actors/core/actorsystem.cpp
@@ -125,6 +125,7 @@ namespace NActors {
                     if (!target) {
                         target = actorId;
                         ServiceMap->RegisterLocalService(recipient, target);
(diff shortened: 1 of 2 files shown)


ydb/library/actors/core/actorsystem.cpp
```diff
@@ -125,6 +125,7 @@ namespace NActors {
                     if (!target) {
                         target = actorId;
                         ServiceMap->RegisterLocalService(recipient, target);
+                        DynamicProxies.push_back(target);
```
snaury, line 128: Not thread safe!

snaury (approved):

PR #5845, opened 2024-06-21
correctly handle hitting MaxTabletsScheduled in drain

vporyadke:
### Changelog entry 

fix a bug where tablet drain could finish prematurely

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/core/mind/hive/hive_impl.cpp
+++ b/ydb/core/mind/hive/hive_impl.cpp
@@ -239,16 +239,16 @@ void THive::ExecuteProcessBootQueue(NIceDb::TNiceDb& db, TSideEffects& sideEffec
         }
         if (tablet->IsReadyToStart(now)) {
             TBestNodeResult bestNodeResult = FindBestNode(*tablet, record.SuggestedNodeId);
-            if (bestNodeResult.BestNode != nullptr) {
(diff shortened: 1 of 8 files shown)


ydb/core/mind/hive/hive_impl.cpp
```diff
@@ -810,14 +810,7 @@ void THive::Handle(TEvPrivate::TEvKickTablet::TPtr &ev) {
...
 
     BLOG_D("THive::Handle::TEvKickTablet TabletId=" << tabletId);
-    TBestNodeResult result = FindBestNode(*tablet);
-    if (result.BestNode == nullptr) {
-        Execute(CreateRestartTablet(tabletId));
-    } else if (result.BestNode != tablet->Node) {
-        if (IsTabletMoveExpedient(*tablet, *result.BestNode)) {
-            Execute(CreateRestartTablet(tabletId));
-        }
-    }
```
vporyadke, line 820: The checks here are pointless - we kill the tablet if there is no node for it, but leave it if there is a different node with the move not being expedient. Also the checks did not guarantee  that the tablet would be restarted on a different node - because things might change between this moment and the restart tablet transaction. Since I have to touch this code, removing them seems best - unconditional tablet restart is reasonable for Kick, that's what people use it for.

adameat (approved):

PR #6031, opened 2024-06-27
Tablet pipe client. Sending messages only for local tablets.

andrewstalin:
### Changelog entry

Added the flag "SendForLocalOnly" for tablet pipe client.  Used for sending messages only for local tablets. If a tablet is not local, a message "TEvNonLocalTablet" is returned.

### Changelog category

### DIFF
--- a/ydb/core/base/tablet_pipe.h
+++ b/ydb/core/base/tablet_pipe.h
@@ -32,6 +32,7 @@ namespace NKikimr {
             EvClientCheckDelay,
             EvClientShuttingDown,
             EvMessage, // replacement for EvSend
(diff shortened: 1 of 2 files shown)


ydb/core/base/tablet_pipe.h
```diff
@@ -365,6 +378,7 @@ namespace NKikimr {
             bool PreferLocal = false;
             bool CheckAliveness = false;
             bool ExpectShutdown = false;
+            bool SendForLocalOnly = false;
```
snaury, line 381: Config already has `ForceLocal`, in which case pipe will not even try to connect to remote nodes.

PR #6039, opened 2024-06-27
Fix Aws::InitAPI memory leak

jepett0:
Put InitAPI into the main function like is advised in the docs (it seems static is bad)


jepett0:
Continued in another [PR](https://github.com/ydb-platform/ydb/pull/6160)

PR #6072, opened 2024-06-28
IGNIETFERRO-2059: Fix a couple of string-related use-after-free errors

ubyte:
### Changelog entry 

Fixed access to freed memory when the original string is destroyed earlier than the TStringBuf constructed from it.

Also, the construction of empty TStringBuf from an empty string has been replaced with direct construction, since, firstly, it is faster, and secondly, it does not lead to invalid pointers, the comparison/copying of which may be problematic in hardened environments: 
> Some implementations might define that copying an invalid pointer value causes a system-generated runtime fault[.](https://timsong-cpp.github.io/cppwp/n4868/basic.stc#footnote-35.sentence-1) [⮥](https://timsong-cpp.github.io/cppwp/n4868/basic.stc#footnoteref-35)

### Changelog category 

* Bugfix

### DIFF
--- a/ydb/core/blobstorage/pdisk/blobstorage_pdisk_ut_http_request.h
+++ b/ydb/core/blobstorage/pdisk/blobstorage_pdisk_ut_http_request.h
@@ -27,7 +27,7 @@ struct THttpRequestMock : NMonitoring::IHttpRequest {
     }
 
     TStringBuf GetPostContent() const override {
-        return TString();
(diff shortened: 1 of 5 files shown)


ubyte:
CI check: https://nda.ya.ru/t/6TXcB25k76fWK5

vitstn (approved):

gridnevvvit (approved):

PR #6075, opened 2024-06-28
Do not ignore configure errors

maximyurchuk:
### Changelog entry 

Configure errors are ignored if `-k` specified.

Additional call without any compilation was added to handle that situation.

### Changelog category 

### Additional information

Check is here: https://github.com/ydb-platform/ydb/pull/6076


maximyurchuk:
It was fixed in 
https://github.com/ydb-platform/ydb/pull/6080

Issue #6122, opened 2024-07-01
PG Port listening


rekby:
1. Listen postgres port only if set LocalPgWireConfig.ListeningPort option
2. Start listening same way as YDB port - try to relisten on fails (now pg port often doesn't started to listen without error on ydbd restarts).

PR #6124, opened 2024-07-01
set pg port in config

rekby:
### Changelog entry 

...

### Changelog category 

### Additional information

...


rekby:
for compatible with #6122

shnikd (approved):

Issue #5737, opened 2024-06-19
Memory leak when exporting / importing from S3


jepett0:
LeakSanitizer [report](https://paste.yandex-team.ru/800d201d-be15-4980-b90c-41ba618a8f73)

steps to reproduce:
1. build `ydbd` with `--sanitize=leak` flag
2. start `ydbd` (you can use `start.sh` script from the [quickstart](https://ydb.tech/docs/en/quickstart) with a custom `ydbd` binary)
3. create `table_to_backup` table in your database
4. run `ydb export s3` to backup the database to some S3 storage (you can use [Yandex Cloud](https://yandex.cloud/en-ru/services/storage?utm_referrer=https%3A%2F%2Fwww.google.com%2F), it is free up to 1GB a month)
5. delete `table_to_backup` table in the database
6. run `ydb import s3` to restore the database from the backup created at step 4
7. run `stop.sh` script to shutdown the cluster
8. observe stderr for the LeakSanitizer report

PR #6160, opened 2024-07-02
Fix memory leak issue with AWS API usage

jepett0:
https://github.com/ydb-platform/ydb/issues/5737

A memory leak can be observed in the production and reproduced with a `ydbd` binary compiled with the `--sanitize=leak` flag and a `ydb import s3` call. Detailed reproduction steps are described in the [issue](https://github.com/ydb-platform/ydb/issues/5737).

The exact reason for the [leak](https://gist.github.com/jepett0/689eecc83996c5e5fd57d170b0567db9) is unknown to me. However, it seems to be caused by the indeterministic order of static variable destruction. Here is a relevant line from the [docs](https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/basic-use.html#sdk-initializing) of AWS C++ SDK:

> The SDK for C++ and its dependencies use C++ static objects, and the order of static object destruction is not determined by the C++ standard. To avoid memory issues caused by the nondeterministic order of static variable destruction, do not wrap the calls to Aws::InitAPI and Aws::ShutdownAPI into another static object.

At first glance, it does not seem like `TApiInitializer` objects have static storage duration in [the current code](https://github.com/ydb-platform/ydb/blob/8bbd471a54b7aa610b4eff1dd6df3158f5d58daf/ydb/core/wrappers/s3_storage_config.cpp#L126). However, if you change the `TApiInitializer` object to be a Singleton itself and run the same script as described in the issue, then you will get the exact same leak sanitizer [report](https://gist.github.com/jepett0/78d56973c5f8843c9e510b20193ff446). It gave me an idea that the underlying issue might be fixed in the same way as a static `TApiInitializer` would be.

I have done some research on the topic of memory issues in AWS SDK and that is what I have found:
- A [question](https://github.com/aws/aws-sdk-cpp/issues/456) asking if multiple InitAPI / ShutdownAPI calls are ok. (The answer is "no".)
- A [comment](https://github.com/aws/aws-sdk-cpp/issues/1550#issuecomment-1412601061) elaborating a bit more on the possible problems of a static RAII wrapper of InitAPI / ShutdownAPI calls.
- A [blame](https://github.com/apache/arrow/blame/fd6bf5a532245ae80908d52b90d5462f9b2f9882/cpp/src/arrow/filesystem/s3fs.cc#L801) of Apache Arrow code that contains changes that have happened between version 12 and 13. It is important to note that [the exact same leak](https://github.com/apache/arrow/issues/37375) was fixed by some change in the Arrow code which happened between versions 12 and 13 and I believe [this PR](https://github.com/apache/arrow/pull/33858) did it. To quote the fix:
> When statically linking error with AWS it was possible to have a crash on shutdown/exit. Now that should no longer be possible. S3 can only be initialized and finalized once. S3 (the AWS SDK) will not be finalized until after all CPU & I/O threads are finished.

## Proposed fix

So the fix that I propose in this PR is simple: call InitAPI somewhere in the beginning of the `main` function and call `ShutdownAPI` somewhere close to the end of the `main` function. Specifically, make `TApiInitializer` a data member of `TKikimrRunner`.

Unit tests have to call `InitAPI` and `ShutdownAPI` explicitly before each test run. It must be done once per process, so a unit test hook is a good choice because individual test cases are run sequentially in a [for loop](https://github.com/ydb-platform/ydb/blob/b286ca162f348d897fe141d87aacf262715f3492/library/cpp/testing/unittest/simple.h#L23) in the same process by default.


ydb/core/wrappers/s3_storage_config.cpp
```diff
@@ -1,81 +1,15 @@
 #include "s3_storage.h"
 #include "s3_storage_config.h"
 
-#include <contrib/libs/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/internal/AWSHttpResourceClient.h>
```
jepett0, line 4: These includes were not used, I believe

ydb/core/wrappers/s3_storage_config.cpp
```diff
@@ -1,81 +1,15 @@
...
-    }
-
-    ~TCurlInitializer() {
-        curl_global_cleanup();
-    }
-};
-
-struct TApiInitializer {
-    TApiInitializer() {
-        Options.httpOptions.initAndCleanupCurl = false;
```
jepett0, line 39: I suppose there is no need to have a separate TCurlInitializer (which calls `curl_global_init` and `cleanup`), because if you enable the `httpOptions.initAndCleanupCurl` option, then `curl_global_init` and `cleanup` [would be called](https://github.com/ydb-platform/ydb/blob/732076c926d51aae388debdb74b7457964e5e6ca/contrib/libs/aws-sdk-cpp/aws-cpp-sdk-core/source/http/HttpClientFactory.cpp#L126) from AWS SDK

ydb/core/wrappers/s3_storage_config.cpp
```diff
@@ -1,81 +1,15 @@
...
-        InitAPI(Options);
-
-        Internal::CleanupEC2MetadataClient(); // speeds up config construction
```
jepett0, line 42: This function is from `Internal` namespace. I don't think we should call it. Moreover, `TApiInitializer` would be initialized only once per `ydbd` application start, so its performance should not be an issue.

ydb/core/wrappers/s3_wrapper_ut.cpp
```diff
@@ -48,8 +65,8 @@ class TS3MockTest: public NUnitTest::TTestBase, private NExternalStorage::TS3Use
     }
 
     void TearDown() override {
-        S3Mock.Reset();
         Runtime.Reset();
+        S3Mock.Reset();
```
jepett0, line 69: Reset in reverse order of initialization. It does not really matter

ydb/services/ydb/backup_ut/ya.make
jepett0, line 1: I decided to move BackupRestore tests from `ydb_import_ut.cpp`, because its `ya.make` builds a huge binary with a lot of tests. None of these tests need `Y_TEST_HOOK_BEFORE_RUN`, except mine, so I decided to separate them.

ijon (dismissed):

jepett0:
@CyberROFL has pointed out the following potential problems:
- [KIKIMR-11055](https://st.yandex-team.ru/KIKIMR-11055). Data race in Aws::S3::S3Client. See old thread sanitizer warning [here](https://gist.github.com/jepett0/c6d529ed176c6cc750dc0f141ef31bd4).
- [KIKIMR-12129](https://st.yandex-team.ru/KIKIMR-12129). Potential data race in setenv, getenv fixed by replacing them with a call to [CleanupEC2MetadataClient](https://github.com/ydb-platform/ydb/blob/448b74bbf09747c0397424a9a50736bff38ef330/ydb/core/wrappers/s3_storage_config.cpp#L42).

I believe that the fix proposed in this PR should avoid both problems.

### Reasons why data races in Aws::S3::S3Client (KIKIMR-11055) are most probably avoided
[There are](https://gist.github.com/jepett0/c6d529ed176c6cc750dc0f141ef31bd4) 6 data races caught in KIKIMR-11055:
- 2 in `curl_slist_append`
- in `curl_easy_init`
- in `Curl_open`
- in `Curl_initinfo`
- in `Curl_saferealloc`
As of this PR `curl` [is managed](https://github.com/ydb-platform/ydb/pull/6160#discussion_r1662891415) by AWS SDK and I believe they should have written it properly, so the data races mentioned above should have been eliminated.

### Reasons why potential data races in setenv, getenv calls (KIKIMR-12129) are avoided
This PR does not introduce any new calls to them.

### Thread sanitizer tests
I have run unit tests in `ydb/services/ydb/backup_ut` multiple times under TSAN and got the following [results](https://wiki.yandex-team.ru/users/deminds/testing-backup-restore-under-thread-sanitizer):

ydb/core/driver_lib/run/run.cpp
```diff
@@ -1653,6 +1653,8 @@ TIntrusivePtr<TServiceInitializersList> TKikimrRunner::CreateServiceInitializers
         sil->AddServiceInitializer(new TGraphServiceInitializer(runConfig));
     }
 
+    sil->AddServiceInitializer(new TAwsApiInitializer(*this));
```
jepett0, line 1656: No service mask, always enabled.

ijon (approved):

PR #6235, opened 2024-07-03
Refactor event pb

maximyurchuk:
### Changelog entry 

- remove unused args
- remove unused funtions

### Changelog category 


ydb/library/actors/core/event_pb.cpp
```diff
@@ -221,4 +224,212 @@ namespace NActors {
...
+            fetchRope(len);
+        }
+    }
+
+    ui32 CalculateSerializedSizeImpl(const TVector<TRope> &payload, ssize_t recordSize) {
+        ssize_t result = recordSize;
+        if (result >= 0 && payload) {
+            ++result; // marker
+            char buf[MaxNumberBytes];
+            result += SerializeNumber(payload.size(), buf);
```
MBkkt, line 382: We can compute count of needed bytes faster:
https://github.com/protocolbuffers/protobuf/commit/7315f6d398d2d18443b26c513cbdcdbffaeebaa3

MBkkt:
```cpp
#if (defined(__x86__) || defined(__x86_64__) || defined(_M_IX86) || \
     defined(_M_X64)) &&                                            \
    !(defined(__LZCNT__) || defined(__AVX2__))
// X86 CPUs lacking the lzcnt instruction are faster with the bsr-based
// implementation. MSVC does not define __LZCNT__, the nearest option that
// it interprets as lzcnt availability is __AVX2__.
#define EVENT_PB_PREFER_BSR 1
#else
#define EVENT_PB_PREFER_BSR 0
#endif

template<typename T>
inline size_t VarintSize(T value) {
#if EVENT_PB_PREFER_BSR
  // Explicit OR 0x1 to avoid calling std::countl_zero(0), which
  // requires a branch to check for on platforms without a clz instruction.
  uint32_t log2value = (std::numeric_limits<T>::digits - 1) -
                       std::countl_zero(value | 0x1);
  return static_cast<size_t>((log2value * 9 + (64 + 9)) / 64);
#else
  uint32_t clz = std::countl_zero(value);
  return static_cast<size_t>(
      ((std::numeric_limits<T>::digits * 9 + 64) - (clz * 9)) / 64);
#endif
}
```

alexvru (approved):

PR #6247, opened 2024-07-03
Support fo big datetime in Cli: Missed break

azevaykin:
### Changelog entry 

Support fo big datetime in Cli

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/public/lib/ydb_cli/common/csv_parser.cpp
+++ b/ydb/public/lib/ydb_cli/common/csv_parser.cpp
@@ -135,10 +135,13 @@ public:
             break;
         case EPrimitiveType::Date32:
             Builder.Date32(GetArithmetic<i32>(token));
+            break;
         case EPrimitiveType::Datetime64:
             Builder.Datetime64(GetArithmetic<i64>(token));
+            break;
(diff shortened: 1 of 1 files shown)


uzhastik (approved):

ydb/public/lib/ydb_cli/common/csv_parser.cpp
```diff
@@ -152,7 +155,7 @@ class TCsvToYdbConverter {
             Builder.TzTimestamp(token);
             break;
         default:
-            TMisuseException() << "Unsupported primitive type: " << Parser.GetPrimitive();
+            throw TMisuseException() << "Unsupported primitive type: " << Parser.GetPrimitive();
```
MBkkt, line 158:
```suggestion
            ythrow TMisuseException() << "Unsupported primitive type: " << Parser.GetPrimitive();
```

Maybe we want to have file + line? text of error message is not unique

azevaykin: There are only `throw` statements everywhere in the file

PR #6277, opened 2024-07-04
Update LLVM part of combiner with spilling.

### LINKED ISSUE #5416: [[YQL]] Adapt WideLastCombinerWrapper to work with spilling in LLVM
Currently, spilling in WideLastCombiner works just when LLVM is disabled.
It will be great to be able to use spilling in WideLastCombiner when LLVM is turned on.

Tony-Romanov:


ydb/library/yql/minikql/comp_nodes/mkql_wide_combine.cpp
```diff
@@ -343,8 +343,8 @@ class TSpillingSupportState : public TComputationValue<TSpillingSupportState> {
     };
 
 public:
-    enum class ETasteResult: ui8 {
-        Init,
+    enum class ETasteResult: i8 {
+        Init = -1,
```
Darych, line 347: What is the reason to init from -1?

Darych (approved):

Issue #4955, opened 2024-05-28
support setting implTable config parameters on alter table add index


antonkovalenko:
YDB Backup files store information about index table partitioning config and other parameters.
Internal implementation of `Alter table add index` operation must provide a way to set implTable configuration, e.g. number of partitions and their boundaries

jepett0:
Some thoughts on the problem:

We need to decide on the optimal protocol for backups and restores of indices. The usual way to restore a scheme object is to execute the appropriate CreateSomeObjectRequest from the public API from the backup. However, index table partitioning settings cannot be specified in the CreateTableRequest currently.

We need to choose one of the solutions:
- add index table partitioning settings description in the CreateTableRequest from public API
- recreate a table with indices using CreateTableRequest + a AlterTableRequests for each indexImplTable

It seems that @CyberROFL is ok with the former solution, but would like to be able to execute AlterTableRequest not on the indexImplTable (because it is a part of the implementation), but on the index object itself.

AlterTableRequest on indexImplTable vs. AlterTableRequest on the index object itself:

Cons:
1) We don't need to know implementation details of index table.

Pros:
1) Such a backup / restore protocol could be implemented without any changes to the YDB code.
2) Maybe it is ok for backup / restore to know index tables implementation details, because this procedures will not be a part of the user's code, but a part of backup / restore system.
3) It is natural for a table to be a target of AlterTableRequest. One could argue that it is not obvious that it is possible to alter indexImplTable partitioning settings by issuing AlterTableRequest on its parent (the index object).

jepett0:
PR:
https://github.com/ydb-platform/ydb/pull/5119

PR #5119, opened 2024-06-03
Extend index tables' partitioning settings to be able to restore them from a backup with the same partitioning as before

jepett0:
https://github.com/ydb-platform/ydb/issues/4955

There are 3 major commits in this PR, whose purposes are the following:
- 90162f55be34706f7b08797aa3a0fa1703d29eca Extend `ydb_table.proto` to be able to specify partitioning settings of the `indexImplTable` and explicit partition boundaries (or uniform partition count) of the index table at the moment of its creation. There is also a functional test using GRPC calls that shows that it becomes possible to create a table with extended index partitioning settings.
- c96c24f11babd322026f7f5e9860ba7cdf8c6142 Changes in cpp SDK to be able to describe extended partitioning settings and explicit partition boundaries (if any) of secondary indexes of the table. This is necessary to be able to dump and restore the table with the extended index partitioning info.
- 49321ece51d08509414e40ff7ba089a0691d9078 Unit tests and functional tests of the feature.

Note: each major commit is preceded by a minor refactoring change. For the best reviewing experience, please review these changes separately.

PR #6280, opened 2024-07-04
Use the whole TTableDescription to describe an indexImplTable in TIndexDescription

jepett0:
It would be better (more future proof) to keep all the necessary settings of the indexImplTable in one message. I would not like to reinvent the bicycle, so the best option I see here is to reuse TTableDescription.

Additionally, I have added options `fillConfig` and `fillBoundaries` to control what parts of the whole TTableDescription would be return on DescribeTableIndex request. It should prevent the unnecessary use of CPU for parsing big protobufs (TableBoundaries can be big). We would need to enable `fillBoundaries` explicitly for backups, but this would be done in scope of https://github.com/ydb-platform/ydb/issues/4955.

And the last change is to make IndexImplTableDescription field a repeated field. The motivation is to prepare `flat_scheme_op.proto` for multiple indexImplTables which are said to be coming soon to implement vector indices. It is a breaking change, but there are no instances of YDB that have been using the new TIndexDescriptions with PartitionConfig and PartitionBoundaries, so it should be ok.

### DIFF
--- a/ydb/core/grpc_services/rpc_describe_table.cpp
+++ b/ydb/core/grpc_services/rpc_describe_table.cpp
@@ -92,7 +92,14 @@ private:
                     return Reply(Ydb::StatusIds::INTERNAL_ERROR, ctx);
                 }
 
-                FillIndexDescription(describeTableResult, tableDescription, splitKeyType);
+                try {
+                    FillIndexDescription(describeTableResult, tableDescription);
+                } catch (const std::exception& ex) {
+                    LOG_ERROR(ctx, NKikimrServices::GRPC_SERVER, "Unable to fill index description: %s", ex.what());
+                    Request_->RaiseIssue(NYql::ExceptionToIssue(ex));
+                    return Reply(Ydb::StatusIds::INTERNAL_ERROR, ctx);
+                }
+
                 FillChangefeedDescription(describeTableResult, tableDescription);
 
                 if (GetProtoRequest()->include_table_stats()) {
--- a/ydb/core/protos/flat_scheme_op.proto
+++ b/ydb/core/protos/flat_scheme_op.proto
@@ -999,7 +999,7 @@ message TIndexDescription {
 
     // DataSize + IndexSize of indexImplTable
     optional uint64 DataSize = 9;
-    repeated TTableDescription IndexImplTableDescription = 10;
+    repeated TTableDescription IndexImplTableDescriptions = 10;
 }
 
 message TIndexCreationConfig {
--- a/ydb/core/protos/flat_scheme_op.proto
+++ b/ydb/core/protos/flat_scheme_op.proto
@@ -982,10 +982,6 @@ enum EIndexState {
     EIndexStateWriteOnly = 3;
 }
 
-message TExplicitPartitions {
-    repeated TSplitBoundary SplitBoundary = 1;
-}
-
 message TIndexDescription {
     optional string Name = 1;
     optional uint64 LocalPathId = 2;
@@ -1000,15 +996,10 @@ message TIndexDescription {
     optional uint64 PathOwnerId = 7;
 
     repeated string DataColumnNames = 8;
+
     // DataSize + IndexSize of indexImplTable
     optional uint64 DataSize = 9;
-
-    // indexImplTable settings
-    oneof Partitions {
-        uint32 UniformPartitions = 10;
-        TExplicitPartitions ExplicitPartitions = 11;
-    }
-    optional TPartitioningPolicy PartitioningPolicy = 12;
+    repeated TTableDescription IndexImplTableDescription = 10;
 }
 
 message TIndexCreationConfig {
--- a/ydb/core/tx/datashard/export_common.cpp
+++ b/ydb/core/tx/datashard/export_common.cpp
@@ -55,11 +55,6 @@ TMaybe<Ydb::Table::CreateTableRequest> GenYdbScheme(
 
     try {
         FillTableBoundary(scheme, tableDesc, mkqlKeyType);
-    } catch (const yexception&) {
-        return Nothing();
-    }
-
-    try {
         FillIndexDescription(scheme, tableDesc);
     } catch (const yexception&) {
         return Nothing();
--- a/ydb/core/tx/datashard/export_common.cpp
+++ b/ydb/core/tx/datashard/export_common.cpp
@@ -59,7 +59,12 @@ TMaybe<Ydb::Table::CreateTableRequest> GenYdbScheme(
         return Nothing();
     }
 
-    FillIndexDescription(scheme, tableDesc, mkqlKeyType);
+    try {
+        FillIndexDescription(scheme, tableDesc);
+    } catch (const yexception&) {
+        return Nothing();
+    }
+
     FillStorageSettings(scheme, tableDesc);
     FillColumnFamilies(scheme, tableDesc);
     FillAttributes(scheme, pathDesc);
--- a/ydb/core/tx/schemeshard/schemeshard_impl.h
+++ b/ydb/core/tx/schemeshard/schemeshard_impl.h
@@ -1016,8 +1016,12 @@ public:
 
     void DescribeTable(const TTableInfo::TPtr tableInfo, const NScheme::TTypeRegistry* typeRegistry,
                        bool fillConfig, bool fillBoundaries, NKikimrSchemeOp::TTableDescription* entry) const;
-    void DescribeTableIndex(const TPathId& pathId, const TString& name, NKikimrSchemeOp::TIndexDescription& entry);
-    void DescribeTableIndex(const TPathId& pathId, const TString& name, TTableIndexInfo::TPtr indexInfo, NKikimrSchemeOp::TIndexDescription& entry);
+    void DescribeTableIndex(const TPathId& pathId, const TString& name,
+        bool fillConfig, bool fillBoundaries, NKikimrSchemeOp::TIndexDescription& entry
+    ) const;
+    void DescribeTableIndex(const TPathId& pathId, const TString& name, TTableIndexInfo::TPtr indexInfo,
+        bool fillConfig, bool fillBoundaries, NKikimrSchemeOp::TIndexDescription& entry
+    ) const;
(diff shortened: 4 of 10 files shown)


ydb/core/grpc_services/rpc_describe_table.cpp
```diff
@@ -92,7 +92,14 @@ class TDescribeTableRPC : public TRpcSchemeRequestActor<TDescribeTableRPC, TEvDe
                     return Reply(Ydb::StatusIds::INTERNAL_ERROR, ctx);
                 }
 
-                FillIndexDescription(describeTableResult, tableDescription, splitKeyType);
+                try {
+                    FillIndexDescription(describeTableResult, tableDescription);
```
jepett0, line 96: FillIndexDescription now calls FillColumnDescription which might throw

ydb/core/tx/schemeshard/schemeshard_path_describer.cpp
```diff
@@ -1238,12 +1238,7 @@ void TSchemeShard::DescribeTableIndex(const TPathId& pathId, const TString& name
     const auto& tableStats = indexImplTable->GetStats().Aggregated;
     entry.SetDataSize(tableStats.DataSize + tableStats.IndexSize);
 
-    *entry.MutablePartitioningPolicy() = indexImplTable->PartitionConfig().GetPartitioningPolicy();
-    if (const auto& explicitPartitions = indexImplTable->TableDescription.GetSplitBoundary();
-        !explicitPartitions.empty()
-    ) {
-        *entry.MutableExplicitPartitions()->MutableSplitBoundary() = explicitPartitions;
-    }
+    *entry.AddIndexImplTableDescription() = indexImplTable->TableDescription;
```
jepett0, line 1241: The entire TableDescription is copied here in the TIndexDescription. However, it does not contain a lot of info. Mainly, it does not contain TTablePartition which is heavy for multishard tables.

Screenshot 2024-07-04 at 12 12 30

ydb/core/ydb_convert/table_description.cpp
```diff
@@ -719,10 +719,9 @@ bool FillColumnDescription(NKikimrSchemeOp::TColumnTableDescription& out,
 
 template <typename TYdbProto>
 void FillTableBoundaryImpl(TYdbProto& out,
-    const google::protobuf::RepeatedPtrField<NKikimrSchemeOp::TSplitBoundary>& boundaries,
-    const NKikimrMiniKQL::TType& splitKeyType
-) {
-    for (const auto& boundary : boundaries) {
+        const NKikimrSchemeOp::TTableDescription& in, const NKikimrMiniKQL::TType& splitKeyType) {
```
jepett0, line 722: These changes restore these lines to the original state in which they were before the https://github.com/ydb-platform/ydb/pull/5119

ydb/core/ydb_convert/table_description.cpp
```diff
@@ -802,38 +801,53 @@ void FillPartitioningSettings(TYdbProto& out, const NKikimrSchemeOp::TPartitioni
...
-    case NKikimrSchemeOp::TIndexDescription::kUniformPartitions:
-        settings.set_uniform_partitions(tableIndex.GetUniformPartitions());
-        break;
-    case NKikimrSchemeOp::TIndexDescription::kExplicitPartitions:
-        FillTableBoundaryImpl(*settings.mutable_partition_at_keys(),
-            tableIndex.GetExplicitPartitions().GetSplitBoundary(),
+    if (indexImplTableDescription.HasUniformPartitionsCount()) {
+        settings.set_uniform_partitions(indexImplTableDescription.GetUniformPartitionsCount());
+    }
+    if (indexImplTableDescription.SplitBoundarySize()) {
```
jepett0, line 835: You could ask yourself: "What if __BOTH__ UniformPartitionsCount and SplitBoundary are specified?" My answer is: SplitBoundary has higher priority. The other option I see here is to throw an error. (FillIndexDescription already throws if FillColumnDescription cannot decide how to fill the columns.)

ydb/core/protos/flat_scheme_op.proto
```diff
@@ -1000,15 +996,10 @@ message TIndexDescription {
...
     // DataSize + IndexSize of indexImplTable
     optional uint64 DataSize = 9;
-
-    // indexImplTable settings
-    oneof Partitions {
-        uint32 UniformPartitions = 10;
-        TExplicitPartitions ExplicitPartitions = 11;
-    }
-    optional TPartitioningPolicy PartitioningPolicy = 12;
+    repeated TTableDescription IndexImplTableDescription = 10;
```
azevaykin, line 1002: According to the [Language Guide (proto 3)](https://protobuf.dev/programming-guides/proto3/) , repeated fields are used in the plural.
```suggestion
    repeated TTableDescription IndexImplTableDescriptions = 10;
```

jepett0: Ok. I agree. More on the issue:
- https://protobuf.dev/programming-guides/style/#repeated-fields
- https://google.aip.dev/144#:~:text=Repeated%20fields%20must%20use%20a,coin%20a%20new%20plural%20form.
- statistically speaking, only 12 out of 94 repeated protobuf field names are singular in flat_scheme_op.proto

azevaykin (dismissed):

ydb/core/tx/datashard/export_common.cpp
```diff
@@ -59,7 +59,12 @@ TMaybe<Ydb::Table::CreateTableRequest> GenYdbScheme(
         return Nothing();
     }
 
-    FillIndexDescription(scheme, tableDesc, mkqlKeyType);
+    try {
+        FillIndexDescription(scheme, tableDesc);
+    } catch (const yexception&) {
+        return Nothing();
+    }
```
MBkkt, line 66: NIT: There is no need to separate try catch for FillTableBoundary and FillIndexDescription

ydb/core/ydb_convert/table_description.cpp
MBkkt: > The other option I see here is to throw an error. (FillIndexDescription already throws if FillColumnDescription cannot decide how to fill the columns.)

Exception sounds better for me, because 
1. looks like error case
2. can be changed to something else in future

ydb/core/protos/flat_scheme_op.proto
```diff
@@ -1000,15 +996,10 @@ message TIndexDescription {
     optional uint64 PathOwnerId = 7;
 
     repeated string DataColumnNames = 8;
+
     // DataSize + IndexSize of indexImplTable
```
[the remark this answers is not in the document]
jepett0: IndexSize is the size of the index of the indexImplTable. It follows the general DataSize + IndexSize = TotalSize [convention](https://github.com/ydb-platform/ydb/blob/6f872bdf6fdd00795d70a46fb00121bb84d808e4/ydb/core/protos/table_stats.proto#L32) for tables.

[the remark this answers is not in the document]
jepett0: I would much rather leave this comment for you to deal with 😜

ydb/core/ydb_convert/table_description.cpp
jepett0: I checked how DescribeTable deals with this problem and found out that it just ignores UniformPartitionsCount, i.e. it does not even check if it exists in the TTableDescription. See the code around [here](https://github.com/ydb-platform/ydb/blob/29e1e4546446acf94c2a350d97b3dbfdf68f766f/ydb/core/tx/schemeshard/schemeshard_path_describer.cpp#L1183). The motivation is: there is no such thing as UniformPartitionsCount for an __existing__ table. If a table already exists, then it has __explicit__ partition boundaries which are represented by SplitBoundary field of the TTableDescription protobuf.

ydb/core/tx/schemeshard/schemeshard_path_describer.cpp
jepett0: I wonder why there is no column info in the IndexImplTableDescription... I will solve this problem in scope of https://github.com/ydb-platform/ydb/issues/4955

ydb/core/tx/schemeshard/schemeshard_impl.h
```diff
@@ -1016,8 +1016,12 @@ class TSchemeShard
 
     void DescribeTable(const TTableInfo::TPtr tableInfo, const NScheme::TTypeRegistry* typeRegistry,
                        bool fillConfig, bool fillBoundaries, NKikimrSchemeOp::TTableDescription* entry) const;
-    void DescribeTableIndex(const TPathId& pathId, const TString& name, NKikimrSchemeOp::TIndexDescription& entry);
-    void DescribeTableIndex(const TPathId& pathId, const TString& name, TTableIndexInfo::TPtr indexInfo, NKikimrSchemeOp::TIndexDescription& entry);
+    void DescribeTableIndex(const TPathId& pathId, const TString& name,
+        bool fillConfig, bool fillBoundaries, NKikimrSchemeOp::TIndexDescription& entry
```
jepett0, line 1020: I have added explicit `fillConfig` and `fillBoundaries` options to be able to control the behavior of `DescribeTableIndex`. Moreover, these options are supported by `DescribeTable` and it would be inconsistent to ignore them in `DescribeTableIndex`

ydb/core/tx/schemeshard/schemeshard_path_describer.cpp
```diff
@@ -8,6 +8,16 @@
 
 #include <util/stream/format.h>
 
+namespace {
+
+void FillPartitionConfig(const NKikimrSchemeOp::TPartitionConfig& in, NKikimrSchemeOp::TPartitionConfig& out) {
+    out.CopyFrom(in);
+    NKikimr::NSchemeShard::TPartitionConfigMerger::DeduplicateColumnFamiliesById(out);
+    out.MutableStorageRooms()->Clear();
```
jepett0, line 16: It turns out that PartitionConfig is not just copied, but some unnecessary info is erased before giving it out to users

azevaykin (dismissed):

MBkkt (dismissed):

azevaykin (approved):

PR #6355, opened 2024-07-05
Change TEvLocal actorsystem metrics

kruall:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/core/mind/local.cpp
+++ b/ydb/core/mind/local.cpp
@@ -278,13 +278,12 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
         auto& record = eventStatus->Record;
         record.SetStartTime(StartTime.GetValue());
         record.MutableResourceMaximum()->CopyFrom(ResourceLimit);
+        NActors::TExecutorPoolState userPoolState;
+        ctx.ExecutorThread.ActorSystem->GetExecutorPoolState(AppData()->UserPoolId, userPoolState);
+
         if (!record.GetResourceMaximum().HasCPU()) {
-            TExecutorPoolStats poolStats;
-            TVector<TExecutorThreadStats> statsCopy;
-            TVector<TExecutorThreadStats> sharedStatsCopy;
-            ctx.ExecutorThread.ActorSystem->GetPoolStats(AppData()->UserPoolId, poolStats, statsCopy, sharedStatsCopy);
-            if (!statsCopy.empty()) {
-                record.MutableResourceMaximum()->SetCPU(poolStats.CurrentThreadCount * 1000000);
+            if (userPoolState.PossibleMaxLimit) {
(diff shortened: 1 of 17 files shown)


alexvru (dismissed):

ydb/core/protos/local.proto
```diff
@@ -53,6 +63,7 @@ message TEvStatus {
     optional uint64 AvailableWeight = 5;
     optional NKikimrTabletBase.TMetrics ResourceMaximum = 8;
     optional uint64 StartTime = 7;
+    optional TActorSystemInfo ActorSystemInfo = 9;
```
vporyadke, line 66: TEvStatus is sent when Local first registers in Hive, and when the tenant is altered. If we want this info to be sent regularly, I would put in TabletMetrics message.

ydb/core/mind/local.cpp
```diff
@@ -111,6 +111,9 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
     ui64 MemLimit = 0;
     double NodeUsage = 0;
 
+    TInstant LastUpdate;
```
vporyadke, line 114: The name is confusing, last update of what? Also If we change it to use TabletMetrics message, there already is a SendTabletMetricsTime member

ydb/core/mind/local.cpp
```diff
@@ -278,15 +281,35 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
...
+        ctx.ExecutorThread.ActorSystem->GetExecutorPoolStates(poolStates);
+
+        TDuration passedTime = ctx.Now() - LastUpdate;
+        LastUpdate = ctx.Now();
+
+        auto *actorSystemInfo = record.MutableActorSystemInfo();
+        double cores = 0;
+        for (ui8 poolId = 0; poolId < poolStates.size(); ++poolId) {
+            auto &poolState = poolStates[poolId];
+            if (poolId != AppData()->IOPoolId) {
```
vporyadke, line 294: We already have a setting PoolsToMonitorForUsage. I believe it should be used here. But it also excludes Batch pool by default - is there a reason you're including it here?

vporyadke (changes requested):

ydb/core/protos/local.proto
```diff
@@ -32,6 +32,16 @@ message TLocalConfig {
     repeated NKikimrSchemeOp.TResourceProfile ResourceProfiles = 1;
 }
 
+message TActorSystemInfo {
```
vporyadke, line 35: Currently we send a single number TotalNodeUsage, that is the maximum over pool usages (and mem usage).
What do we expect to gain from sending this detailed info? Just seems like we will simply do the calculate usage + take maximum logic on Hive side instead.

vporyadke (approved):

alexvru (approved):

PR #6409, opened 2024-07-08
fix neighbours count

### LINKED ISSUE #6424: Corrctly count neighbours in Hive
When counting the number of "neighbours" (tablets of the same object, e.g. of the same table) of a tablet on a node, the tablet itself is counted. This is incorrect and makes the current node of the tablet less favorable, leading to more tablet moves than necessary.

vporyadke:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

...


adameat (approved):
also, dont forget about 24-3 (and 24-2)

PR #6456, opened 2024-07-09
set predicate-extract config default

ssmike:
### Changelog category


gridnevvvit (approved):

gridnevvvit:
the feature was enabled successfully in major installation so I think that is ok to make it enabled by default in 24-3

spuchin (approved):

fomichev3000 (approved):

PR #6568, opened 2024-07-11
fix wrong isolation level

### LINKED ISSUE #5887: Fix wrong isolation level for commit in per statement execution
If the last statement in query is ddl, the all query failed with error 
`[[bad]]assertion failed at ydb/core/kqp/ut/service/kqp_qs_queries_ut.cpp:2301, virtual void NKikimr::NKqp::NTestSuiteKqpQueryService::TTestCaseDdl_Dml::Execute_(NUnitTest::TTestContext &): (result.GetStatus() == EStatus::SUCCESS) failed: (INTERNAL_ERROR != SUCCESS) <main>: Error: ydb/core/kqp/executer_actor/kqp_data_executer.cpp:144  TKqpDataExecuter(): requirement Request.IsolationLevel != NKikimrKqp::ISOLATION_LEVEL_UNDEFINED failed`

Unittest example:
```
            auto result = db.ExecuteQuery(R"(
                CREATE TABLE TestDdlDml2 (
                    Key Uint64,
                    Value1 String,
                    Value2 String,
                    PRIMARY KEY (Key)
                );
                UPSERT INTO TestDdlDml2 (Key, Value1, Value2) VALUES (1, "1", "1");
                SELECT * FROM TestDdlDml2;
                ALTER TABLE TestDdlDml2 DROP COLUMN Value2;
                UPSERT INTO TestDdlDml2 (Key, Value1) VALUES (2, "2");
                SELECT * FROM TestDdlDml2;
                CREATE TABLE TestDdlDml33 (
                    Key Uint64,
                    PRIMARY KEY (Key)
                );
            )", TTxControl::NoTx()).ExtractValueSync();
            UNIT_ASSERT_VALUES_EQUAL_C(result.GetStatus(), EStatus::SUCCESS, result.GetIssues().ToString());
```

VPolka:
### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/core/kqp/session_actor/kqp_session_actor.cpp
+++ b/ydb/core/kqp/session_actor/kqp_session_actor.cpp
@@ -1099,7 +1099,7 @@ public:
             switch (tx->GetType()) {
                 case NKqpProto::TKqpPhyTx::TYPE_SCHEME:
                     YQL_ENSURE(tx->StagesSize() == 0);
-                    if (QueryState->HasTxControl() && QueryState->TxCtx->EffectiveIsolationLevel != NKikimrKqp::ISOLATION_LEVEL_UNDEFINED) {
+                    if (QueryState->HasTxControl() && !QueryState->HasImplicitTx() && QueryState->TxCtx->EffectiveIsolationLevel != NKikimrKqp::ISOLATION_LEVEL_UNDEFINED) {
                         ReplyQueryError(Ydb::StatusIds::PRECONDITION_FAILED,
                             "Scheme operations cannot be executed inside transaction");
                         return true;
--- a/ydb/core/kqp/session_actor/kqp_session_actor.cpp
+++ b/ydb/core/kqp/session_actor/kqp_session_actor.cpp
@@ -291,7 +291,6 @@ public:
     void CommitTx() {
         YQL_ENSURE(QueryState->HasTxControl());
         const auto& txControl = QueryState->GetTxControl();
-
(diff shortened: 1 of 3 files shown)


ydb/core/kqp/session_actor/kqp_session_actor.cpp
UgnineSirdis, line 1: I'm not sure that this will fix the problem. I've debugged this test and found out, that the error occurs when we are executing the last PhyTx: `ExecutePhyTx(tx==nullptr, commit=true)`. In this case we don't enter the first `if (tx) {}` block and so don't execute `QueryState->PrepareStatementTransaction(tx->GetType());`. So IsolationLevel is leaved as it was during previous PhyTx. I tried to add PrepareStatementTransaction also to the call with commit=true variant.
I think that the following code would be more safe:
```
bool ExecutePhyTx(const TKqpPhyTxHolder::TConstPtr& tx, bool commit) {
    if (tx) {
        QueryState->PrepareStatementTransaction(tx->GetType());
        switch (tx->GetType()) {
            ...
        }
    } else {
        YQL_ENSURE(commit);
        QueryState->PrepareStatementTransaction(NKqpProto::TKqpPhyTx::TYPE_GENERIC); // not sure that this is the most suitable type, but in this call the main thing is that it is not equal to TYPE_SCHEME
    }
    ...
}
```

UgnineSirdis (changes requested):

UgnineSirdis (approved):

PR #6609, opened 2024-07-12
add status processing

ildar-khisambeev:


ydb/core/protos/statistics.proto
```diff
@@ -147,6 +147,7 @@ message TEvStatisticsResponse {
         STATUS_SUCCESS = 1;
         STATUS_ABORTED = 2;
         STATUS_ERROR = 3;
+        STATUS_PROCESSING = 4;
```
azevaykin, line 150: Let it be `= 2`, before `STATUS_ABORTED`

azevaykin (approved):

PR #6642, opened 2024-07-12
fix(kqp): return one result row per key from stream join actor for left semi strategy

### LINKED ISSUE #6559: Stream join returns duplicates for left semi join

ulya-sidorina:


spuchin (approved):

spuchin:
Don't forget to merge into 24-3.

PR #6817, opened 2024-07-18
[check] retry tests

maximyurchuk:
### Changelog entry 

...

### Changelog category 

* New feature
* Experimental feature
* Improvement
* Performance improvement
* Bugfix 

### Additional information

...


ydb/library/actors/core/ut/event_pb_payload_ut.cpp
CyberROFL, line 1: Это тут к чему?

Issue #6394, opened 2024-07-08
VIEW: COUNT and other aggregate functions do not work inside a view


jepett0:
To reproduce:
1. Create a view:
```sql
CREATE VIEW user_group_view WITH (security_invoker = TRUE) AS
    SELECT COUNT(user_id) AS members_count
    FROM `/Root/dedicated/user_group_member`;
```

2. Select from the view:
```sql
SELECT * FROM user_group_view;
```

3. Observe an error:
```
Error
Expected callable type, but got: Unit
Error
Type annotation
Code: 1030
Error
At function: RemovePrefixMembers, At function: RemoveSystemMembers
Error
At function: Take
Error
At function: PersistableRepr, At function: SqlProject, At function: PersistableRepr, At function: SqlProject
Error
At function: AssumeColumnOrderPartial
Error
At function: Aggregate, At function: Apply
Error
Expected callable type, but got: Unit
```

jepett0:
PRs in stable:
https://github.com/ydb-platform/ydb/pull/7501
https://github.com/ydb-platform/ydb/pull/7523

PR #6820, opened 2024-07-18
Enable COUNT in view queries

jepett0:
https://github.com/ydb-platform/ydb/issues/6394
KIKIMR-21731

- Add ModuleResolver from KQP host.
    
    It seems that `COUNT` and other aggregate functions are implemented as imported modules. See, for example, [AST](https://gist.github.com/jepett0/1429d788390b52a618804e7c1955c42d) of the following query:
    ```sql
    SELECT COUNT(*) FROM `/Root/episodes`
    ```
    Note the following line:
    > (import aggregate_module '"/lib/yql/aggregate.yql")

    Aggregate module `count_traits_factory` is implemented [here](https://github.com/ydb-platform/ydb/blob/31ea815dbf71fb690ebea594e04097d0f221e56d/ydb/library/yql/mount/lib/yql/aggregate.yql#L670).

- Add `ExpandApplyForLambdas` transformation step after `RewriteReadFromView`.

    The same is [done](https://github.com/ydb-platform/ydb/blob/31ea815dbf71fb690ebea594e04097d0f221e56d/ydb/library/yql/providers/yt/provider/yql_yt_datasource.cpp#L859) to rewrite reads from views in YT. The necessity could be well understood by studying the differences in:
    - [transformations](https://gist.github.com/jepett0/1328e860f5b061a3c01b79ceb3ea831a) of
        ```sql
        SELECT COUNT(*) FROM `/Root/episodes`
        ```
    
        (Note how `Apply` of a lambda disappears due to some graph transformation between the first and the second debug printout of the TExprNode graph of the pure `SELECT`.)
    - [transformations](https://gist.github.com/jepett0/50a585504490a0299df6785a274a5e94) of the same query written inside a view
    

Both (ModuleResolver addition and ExpandApplyForLambdas step repetition) are necessary to enable aggregate functions in views.

Copying default KQP host's translation settings is done to make view compilation more similar to a pure `SELECT` compilation. It is not a necessary part of the fix.

### DIFF
--- a/ydb/core/kqp/host/kqp_translate.h
+++ b/ydb/core/kqp/host/kqp_translate.h
@@ -1,7 +1,8 @@
 #pragma once
 
-#include <ydb/core/kqp/provider/yql_kikimr_results.h>
-#include <ydb/core/kqp/common/kqp.h>
+#include <ydb/core/kqp/common/simple/query_ast.h>
+#include <ydb/core/kqp/provider/yql_kikimr_provider.h>
+#include <ydb/core/protos/table_service_config.pb.h>
 #include <ydb/library/yql/core/pg_settings/guc_settings.h>
 
 namespace NKikimr {
--- a/ydb/core/kqp/provider/rewrite_io_utils.cpp
+++ b/ydb/core/kqp/provider/rewrite_io_utils.cpp
@@ -15,23 +15,17 @@ using namespace NNodes;
 
 constexpr const char* QueryGraphNodeSignature = "SavedQueryGraph";
 
-NSQLTranslation::TTranslationSettings CreateViewTranslationSettings(const TString& cluster) {
-    NSQLTranslation::TTranslationSettings settings;
-
-    settings.DefaultCluster = cluster;
-    settings.ClusterMapping[cluster] = TString(NYql::KikimrProviderName);
-    settings.Mode = NSQLTranslation::ESqlMode::LIMITED_VIEW;
-
-    return settings;
-}
-
 TExprNode::TPtr CompileViewQuery(
     const TString& query,
     TExprContext& ctx,
-    const TString& cluster
+    NKikimr::NKqp::TKqpTranslationSettingsBuilder& settingsBuilder,
+    IModuleResolver::TPtr moduleResolver
 ) {
+    auto translationSettings = settingsBuilder.Build(ctx);
+    translationSettings.Mode = NSQLTranslation::ESqlMode::LIMITED_VIEW;
+
     TAstParseResult queryAst;
-    queryAst = NSQLTranslation::SqlToYql(query, CreateViewTranslationSettings(cluster));
+    queryAst = NSQLTranslation::SqlToYql(query, translationSettings);
 
     ctx.IssueManager.AddIssues(queryAst.Issues);
     if (!queryAst.IsOk()) {
@@ -39,7 +33,7 @@ TExprNode::TPtr CompileViewQuery(
     }
 
     TExprNode::TPtr queryGraph;
-    if (!CompileExpr(*queryAst.Root, queryGraph, ctx, nullptr, nullptr)) {
+    if (!CompileExpr(*queryAst.Root, queryGraph, ctx, moduleResolver.get(), nullptr)) {
         return nullptr;
     }
 
@@ -123,7 +117,8 @@ TExprNode::TPtr RewriteReadFromView(
     const TExprNode::TPtr& node,
     TExprContext& ctx,
     const TString& query,
-    const TString& cluster
+    NKikimr::NKqp::TKqpTranslationSettingsBuilder& settingsBuilder,
+    IModuleResolver::TPtr moduleResolver
 ) {
     YQL_PROFILE_FUNC(DEBUG);
 
@@ -132,7 +127,7 @@ TExprNode::TPtr RewriteReadFromView(
 
     TExprNode::TPtr queryGraph = FindSavedQueryGraph(readNode.Ptr());
     if (!queryGraph) {
-        queryGraph = CompileViewQuery(query, ctx, cluster);
+        queryGraph = CompileViewQuery(query, ctx, settingsBuilder, moduleResolver);
         if (!queryGraph) {
             ctx.AddError(TIssue(ctx.GetPosition(readNode.Pos()),
                          "The query stored in the view cannot be compiled."));
--- a/ydb/core/kqp/provider/yql_kikimr_datasource.cpp
+++ b/ydb/core/kqp/provider/yql_kikimr_datasource.cpp
@@ -3,6 +3,7 @@
 #include "yql_kikimr_provider_impl.h"
 
 #include <ydb/core/kqp/common/simple/services.h>
+#include <ydb/core/kqp/host/kqp_translate.h>
 #include <ydb/library/yql/providers/common/provider/yql_data_provider_impl.h>
 #include <ydb/library/yql/providers/common/config/yql_configuration_transformer.h>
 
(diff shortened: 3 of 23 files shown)


ydb/core/kqp/provider/rewrite_io_utils.cpp
```diff
@@ -15,31 +15,54 @@ using namespace NNodes;
...
 
-NSQLTranslation::TTranslationSettings CreateViewTranslationSettings(const TString& cluster) {
+NSQLTranslation::TTranslationSettings CreateViewTranslationSettings(const TString& cluster, EKikimrQueryType queryType) {
     NSQLTranslation::TTranslationSettings settings;
 
     settings.DefaultCluster = cluster;
     settings.ClusterMapping[cluster] = TString(NYql::KikimrProviderName);
     settings.Mode = NSQLTranslation::ESqlMode::LIMITED_VIEW;
 
+    settings.Flags.insert("FlexibleTypes");
```
jepett0, line 25: These flags are copied from the default [settings](https://github.com/ydb-platform/ydb/blob/2d8ae71907f4da2c07c941ea8efa277d4680d216/ydb/core/kqp/host/kqp_translate.cpp#L107) that KQP host is using to compile queries.

The following settings are deliberately left out (i.e. not copied from KQP host):
- PG (postgres) settings (PG syntax is not yet officially supported in combination with views)
- SQL version settings (I imagine, views don't have to support v0 syntax)
- external data source settings (they haven't been supported inside views yet)
- EndOfQueryCommit setting, because I believe views should not manage commits by themselves
- QueryParameters (views don't support parameters)
- ApplicationName (to be frank, I have no idea, what use does it have)
- GUCSettings (they are used in PG syntax mode that is not yet officially supported in combination with views)

One important thing to note here is that the flags (i.e. pragmas) depend on the query type used to compile the `SELECT` from a view. This query type might differ from the query type that was used to `CREATE VIEW`. This might lead to the results of internal view queries being dependent on the particular query type that is being used at the time of a `SELECT` from the view. This might be considered a bug, because a view should produce the same result whenever it is selected from. It should act almost like a real table that has its rows saved and persisted on disk to never change unless these rows change in the underlying tables from which the view selects from.

This "bug" will be fixed later, when we will decide which parts of the `CREATE VIEW` context to save inside the view.

dorooleg: 1. We can reuse the basic settings of KQP and remove unsuitable fields. Do we really need to keep two independent versions of TTranslationSettings?
2. What do we need to do to support external data sources? What is the difference between data shards and external data sources in case of views?

ydb/core/kqp/ut/view/input/cases/count_episodes/create_view.sql
```diff
@@ -0,0 +1,9 @@
+CREATE VIEW `/Root/count_episodes` WITH (security_invoker = TRUE) AS
```
dorooleg, line 1: What do these tests check? Do you have a validation for the results of these queries?

jepett0: This particular test (`count_episodes`) is a way to test `COUNT` inside a view, but with a bit more complex query than just `SELECT COUNT(*) FROM some_table`

dorooleg: It's an intresting approach. Thank you for exlanation :)

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: 1. We could reuse `TKqpTranslationSettingsBuilder` to compile the view query if that is what you suggest. However, the ideal way to get `TTranslationSettings` for the view query compilation is to fetch them from the parsing context of the `CREATE VIEW` statement and persist them in the body of the view. We should not use effective `TTranslationSettings` at the moment of a `SELECT` from a view, because they might differ from those that were in effect during `CREATE VIEW` statement. (For example, `query type` might differ leading to default pragmas being different.)

2. We haven't tested this scenario at all. I'm sure there will be some hidden differences that will become a problem for views in combination with external data sources. We will probably think about this scenario later, after the first release of views.

dorooleg: 1. My proposal is to try to reuse the part of settings between TKqpTranslationSettingsBuilder and CreateViewTranslationSettings. Also, sometimes we want to override the persisted view translations settings. For example, we want to display a warning if the user uses some expression inside the view. We have already used this approach for bindings. `EBindingsMode BindingsMode;` prefix. 
2. We think about using the view inside YQ service and there we use the external sources. And it was interesting to know the status in this part. I got the answer here, thank you)

jepett0: 1. In some sense, I have done exactly that: reused (copy-pasted) some parts of the settings that TKqpTranslationSettingsBuilder builds. I plan to persist translation settings inside a view at the moment of the view creation for later use during `SELECT` from the view. It would eliminate the need to build the settings here, during `RewriteReadFromView`. We would just take the persisted ones and use them.

    As for overriding of the settings to display a warning if some forbidden expression is used inside a view, we can do this at the moment of persisting of the settings during `CREATE VIEW`. Or are you trying to say that we might someday need to override the settings of some old view that was created years ago to forbid some discontinued expression inside it? Well, in that case it seems suitable to override it inside the dedicated, view-specific `CreateViewTranslationSettings` function and don't mix it with the general `TKqpTranslationSettingsBuilder`.

    And lastly, all I am trying to say is that I will rewrite `CreateViewTranslationSettings` function later. The changes in this function that are introduced in this PR are temporary. It is better to think all of this through when we start implementing `TTranslationSettings` persisting during `CREATE VIEW`.

dorooleg (dismissed):

ydb/core/kqp/provider/rewrite_io_utils.cpp
jepett0: Used TKqpTranslationSettingsBuilder in the latest [revision](https://github.com/ydb-platform/ydb/blob/fee0d0ddce1cbf573356840741486942e08e697c/ydb/core/kqp/provider/rewrite_io_utils.cpp#L24)

ydb/core/kqp/provider/yql_kikimr_datasource.cpp
```diff
@@ -768,7 +772,15 @@ class TKikimrDataSource : public TDataProviderBase {
...
 
                 const auto& query = tableDesc.Metadata->ViewPersistedData.QueryText;
-                return RewriteReadFromView(node, ctx, query, cluster);
+                NKqp::TKqpTranslationSettingsBuilder settingsBuilder(
+                    SessionCtx->Query().Type,
+                    SessionCtx->Config()._KqpYqlSyntaxVersion.Get().GetRef(),
+                    cluster,
+                    query,
+                    SessionCtx->Config().BindingsMode,
+                    GUCSettings
```
jepett0, line 781: [GUCSettings](https://pgpedia.info/g/guc.html) are needed here. That is why we pass them in CreateKikimrDataSource.

We cannot, unfortunately,  pass the KQP translation settings builder itself from the KQP host to the Kikimr DataSource, because it needs `query` to be built.

ydb/core/kqp/host/kqp_translate.h
```diff
@@ -1,7 +1,8 @@
 #pragma once
 
-#include <ydb/core/kqp/provider/yql_kikimr_results.h>
```
jepett0, line 3: Here I remove unnecessary includes and add only the explicitly needed ones

ydb/core/kqp/ut/view/input/cases/aggregates_and_window/create_view.sql
```diff
@@ -0,0 +1,43 @@
+CREATE VIEW `/Root/aggregates_and_window` WITH (security_invoker = TRUE) AS
+    SELECT
```
jepett0, line 2: This test might be an overkill, but it uses SUM, AVG and COUNT aggregate functions as well as a WINDOW clause

gridnevvvit (approved):

PR #6840, opened 2024-07-18
deprioritise system tablets in balancer

vporyadke:
### Changelog entry 

make less coordinator&mediator moves

### Changelog category 

* Performance improvement

### Additional information

...

### DIFF
--- a/ydb/core/mind/hive/balancer.cpp
+++ b/ydb/core/mind/hive/balancer.cpp
@@ -62,17 +62,17 @@ void BalanceNodes<NKikimrConfig::THiveConfig::HIVE_NODE_BALANCE_STRATEGY_RANDOM>
 }
 
 template<>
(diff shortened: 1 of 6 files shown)


ydb/core/mind/hive/balancer.cpp
```diff
@@ -62,17 +62,17 @@ void BalanceNodes<NKikimrConfig::THiveConfig::HIVE_NODE_BALANCE_STRATEGY_RANDOM>
 }
 
 template<>
-void BalanceTablets<NKikimrConfig::THiveConfig::HIVE_TABLET_BALANCE_STRATEGY_OLD_WEIGHTED_RANDOM>(std::vector<TTabletInfo*>& tablets, EResourceToBalance resourceToBalance) {
+void BalanceTablets<NKikimrConfig::THiveConfig::HIVE_TABLET_BALANCE_STRATEGY_OLD_WEIGHTED_RANDOM>(std::vector<TTabletInfo*>::iterator first, std::vector<TTabletInfo*>::iterator last, EResourceToBalance resourceToBalance) {
```
adameat, line 65: I would advise of not using first-last semantics, where it's clear that it's a begin-end semantics.
the key difference is that "last" usually points to existing element, while "end" points to the position, after the last element

adameat (approved):

Issue #6699, opened 2024-07-15
[QueryService] Upserting and replacing rows with RETURNING raise INTERNAL ERROR


alex268:
If I have a simple table like this
```
CREATE TABLE test1 (id Int32, v Text, PRIMARY KEY(id))
```
and execute next request via QueryService
```
UPSERT INTO test1 (id, v) VALUES (1, '321') RETURNING id, v;
REPLACE INTO test1 (id, v) VALUES (1, '111') RETURNING v, id;
```
I get an error
```
code = INTERNAL_ERROR(code=400030), issues = [
#1060 Execution (S_FATAL) 14:2 - 14:2: #1 contrib/ydb/core/kqp/opt/kqp_opt_kql.cpp:982  operator()(): requirement effectsMap[effect.Raw()] failed (S_FATAL),
#1060 Execution (S_FATAL) 14:2 - 14:2: #1 contrib/ydb/core/kqp/opt/kqp_opt_kql.cpp:982  operator()(): requirement effectsMap[effect.Raw()] failed (S_FATAL)
]
```
The same query via TableService works without problems.

ssmike:
Works in master, but the columns order is wrong. Fix attached.

ssmike:
@alex268 Which branch are you using? Could you test against fresh master? Your query just works in ut

alex268:
I reproduce this error right now on the fresh YDB docker image. As I see this image was created July 19, but I am not sure that it was built from master branch. Original error was raised in this [test](https://github.com/ydb-platform/ydb-jdbc-driver/blob/master/jdbc/src/test/java/tech/ydb/jdbc/impl/YdbQueryConnectionImplTest.java#L689-L691) and now it is still not working.
Information about docker image I used for tests:
```
        "Id": "sha256:be9c2e54403ef2c3f87de230b9c05a40c39c165490ef8f7b44678b1d09fcbfb0",
        "RepoTags": [
            "cr.yandex/yc/yandex-docker-local-ydb:trunk"
        ],
        "RepoDigests": [
            "cr.yandex/yc/yandex-docker-local-ydb@sha256:047286eec93a5204b30bf6cebe9e7b1338889bee8efc6b38113e66d270e49e42"
        ],
        "Parent": "",
        "Comment": "buildkit.dockerfile.v0",
        "Created": "2024-07-19T01:38:22.013840011Z",
```

ssmike:
AST ut
(
[...]

PR #6861, opened 2024-07-19
fix columns order in returning list

ssmike:
### Changelog category 
* Bugfix 


ssmike:
#6699

gridnevvvit (approved):

PR #6882, opened 2024-07-19
redirect stdin, stdout, stderr to the standard stream or files as des…

Nikita-Levuskin:
…ired

### Changelog entry 

redirect stdin,stdout,stderr to the standard stream or files as desired

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/tests/library/harness/daemon.py
+++ b/ydb/tests/library/harness/daemon.py
@@ -59,9 +59,9 @@ class Daemon(object):
         command,
         cwd,
         timeout,
-        stdin_file=yatest_common.work_path('stdin'),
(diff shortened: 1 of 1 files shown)


ydb/tests/library/harness/daemon.py
```diff
@@ -59,9 +59,9 @@ def __init__(
         command,
         cwd,
         timeout,
-        stdin_file=yatest_common.work_path('stdin'),
-        stdout_file=yatest_common.work_path('stdout'),
-        stderr_file=yatest_common.work_path('stderr'),
+        stdin_file='stdin',
```
rekby, line 62: Why do you need remove `yatest_common.work_path`?

PR #7103, opened 2024-07-25
fix viewer redirects

adameat:
### Changelog entry 

fix viewer redirects when we make a call to newer node via older one

### Changelog category 

* Bugfix


the-ancient-1 (approved):
LGTM

PR #7111, opened 2024-07-25
Remove cpu reqs from tests in order to speed up checks

maximyurchuk:
### Changelog entry 

Due to poor parallelism remove cpu requirements

Some of requirements was set in 
https://a.yandex-team.ru/review/1740558/details 

Looks like they are not needed

But that change not actually helps, because RAM implies CPU requirements (that a bug! see DEVTOOLSSUPPORT-48411)

### Changelog category 

### Additional in


ydb/library/yql/parser/pg_wrapper/test/ya.make
```diff
@@ -15,7 +15,6 @@ ELSE()
 ENDIF()
 
 REQUIREMENTS(
-    cpu:4
```
vitstn, line 18: please don't

gridnevvvit (dismissed):

ydb/library/yql/parser/pg_wrapper/test/ya.make
maximyurchuk: reverted (it was stated that pg_wrapper tests do work in 16 threads)

gridnevvvit (approved):

PR #7121, opened 2024-07-26
Use EvRead-based table reading in datashard_ut_write test

lex007in:
### Changelog entry 

Use EvRead in datashard_ut_write test instead of old TReadTableState.

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/core/tx/datashard/datashard_ut_write.cpp
+++ b/ydb/core/tx/datashard/datashard_ut_write.cpp
@@ -1,5 +1,4 @@
 #include "datashard_active_transaction.h"
-#include "datashard_ut_read_table.h"
 #include <ydb/core/tx/data_events/payload_helper.h>
 #include <ydb/core/tx/datashard/ut_common/datashard_ut_common.h>
 #include <ydb/core/base/tablet_pipecache.h>
@@ -11,7 +10,6 @@ using namespace NKikimr::NDataShard;
 using namespace NKikimr::NDataShard::NKqpHelpers;
 using namespace NSchemeShard;
 using namespace Tests;
-using namespace NDataShardReadTableTest;
 
 Y_UNIT_TEST_SUITE(DataShardWrite) {
     const TString expectedTableState = "key = 0, value = 1\nkey = 2, value = 3\nkey = 4, value = 5\n";
@@ -52,7 +50,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 1);
             UNIT_ASSERT_VALUES_EQUAL(tableState, expectedTableState);
         }
     }
@@ -78,8 +76,8 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read tables =========\n";
         {
-            auto tableState1 = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
-            auto tableState2 = TReadTableState(server, MakeReadTableSettings("/Root/table-2")).All();
+            auto tableState1 = ReadTable(server, shards1.at(0), "table-1", tableId1, 1);
+            auto tableState2 = ReadTable(server, shards2.at(0), "table-2", tableId2, 2);
             UNIT_ASSERT_VALUES_EQUAL(tableState1, "key = 0, value = 1\n");
             UNIT_ASSERT_VALUES_EQUAL(tableState2, "key = 2, value = 3\n");
         }
@@ -111,7 +109,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 1);
             UNIT_ASSERT_VALUES_EQUAL(tableState, expectedTableState);
         }
     }
@@ -134,7 +132,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 1);
             UNIT_ASSERT_VALUES_EQUAL(tableState, "key64 = 0, key32 = 1, value64 = 2, value32 = 3, valueUtf8 = String_4\n"
                                                  "key64 = 5, key32 = 6, value64 = 7, value32 = 8, valueUtf8 = String_9\n"
                                                  "key64 = 10, key32 = 11, value64 = 12, value32 = 13, valueUtf8 = String_14\n");
@@ -147,7 +145,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table with 1th row deleted =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 2);
             UNIT_ASSERT_VALUES_EQUAL(tableState, "key64 = 5, key32 = 6, value64 = 7, value32 = 8, valueUtf8 = String_9\n"
                                                  "key64 = 10, key32 = 11, value64 = 12, value32 = 13, valueUtf8 = String_14\n");
         }
@@ -231,7 +229,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 1);
             UNIT_ASSERT_VALUES_EQUAL(tableState, expectedTableState);
         }
     }    
@@ -252,7 +250,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 1);
             UNIT_ASSERT_VALUES_EQUAL(tableState, expectedTableState);
         }
 
@@ -272,7 +270,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table with 1th row deleted =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 2);
             UNIT_ASSERT_VALUES_EQUAL(tableState, "key = 2, value = 3\nkey = 4, value = 5\n");
         }
     }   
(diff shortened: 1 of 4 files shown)


ydb/core/tx/datashard/datashard_ut_write.cpp
```diff
@@ -52,7 +50,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards.at(0), "table-1", tableId, 1);
```
MBkkt, line 53: maybe unit_assert_values(shards.size(), values)
or cycle in readtable over shards?

azevaykin: I like the second option: ReadTable iterates over all the shards.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2459,6 +2459,73 @@ namespace {
         TDeque<TPendingRequest> QuotaRequests;
     };
 
+    std::unique_ptr<TEvDataShard::TEvReadResult> ReadAllTable(
```
azevaykin, line 2462:
```suggestion
    std::unique_ptr<TEvDataShard::TEvReadResult> ReadEntireTable(
```

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2459,6 +2459,73 @@ namespace {
...
+        Tests::TServer::TPtr server,
+        ui64 tabletId,
```
azevaykin, line 2464: If we read the entire table than there should be array of tabletIds.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2459,6 +2459,73 @@ namespace {
...
+        auto sender = runtime.AllocateEdgeActor();
+
+        runtime.SendToPipe(
+            tabletId,
+            sender,
+            request.release()
+        );
+
+        TAutoPtr<IEventHandle> handle;
+        runtime.GrabEdgeEventRethrow<TEvDataShard::TEvReadResult>(handle, TDuration::Seconds(5));
```
azevaykin, line 2498: EvRead is iterator based reading.
There could be several results.

Example:
https://github.com/ydb-platform/ydb/blob/main/ydb/core/tx/datashard/datashard_ut_read_iterator.cpp#L726

MBkkt, line 2498: We can unify this code with `ydb/core/tx/datashard/datashard_ut_read_iterator.cpp` code:
Lets make its private functions: AddRangeQuery, AddKeyQuery, SendRead, SendReadAsync, WaitReadResult publicly available for all datashard tests.
And we can use it to implement this function: AddRangeQuery+SendRead

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2459,6 +2459,73 @@ namespace {
...
+        UNIT_ASSERT(handle);
+
+        return std::unique_ptr<TEvDataShard::TEvReadResult>(handle->Release<TEvDataShard::TEvReadResult>().Release());
+    }
+
+    TString PrintTableFromResult(const TEvDataShard::TEvReadResult& event, const NKikimrTxDataShard::TEvGetInfoResponse::TUserTable& userTable) {
```
azevaykin, line 2504: Result should be made from one event, but from several events.

azevaykin:
There are several `SELECT key, value FROM /Root/table` left.
For example [here](https://github.com/ydb-platform/ydb/pull/7121/files/2e6b7ad86027c665aa7281fab8b9761d532304c3#diff-5d6e9e26d5539b1edf6d6936803a156968d292386ce0cf440c1cb65f2cf65423R980)
And it's a good example of multi-shard reading.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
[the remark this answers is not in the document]
azevaykin: There should be something like that
https://github.com/ydb-platform/ydb/blob/main/ydb/core/tx/datashard/datashard_ut_read_iterator.cpp#L560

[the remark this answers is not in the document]
MBkkt: Yes, I suggested to move this function to common code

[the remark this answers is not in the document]
azevaykin: By the way, you can move functions like `WaitReadResult` from `datashard_ut_read_iterator.cpp` to `datashard_ut_common.cpp`

azevaykin: This is exactly I have said https://github.com/ydb-platform/ydb/pull/7121#discussion_r1692998532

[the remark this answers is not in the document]
lex007in: I set a limited timeout to prevent slowing down the test run in case of problems affecting all tests. However, this timeout would not be enough when running tests with profilers with significant performance overhead, such as Valgrind. So, I'll remove timeout.

ydb/core/tx/datashard/datashard_ut_write.cpp
```diff
@@ -147,7 +145,7 @@ Y_UNIT_TEST_SUITE(DataShardWrite) {
 
         Cout << "========= Read table with 1th row deleted =========\n";
         {
-            auto tableState = TReadTableState(server, MakeReadTableSettings("/Root/table-1")).All();
+            auto tableState = ReadTable(server, shards, "table-1", tableId, 2000);
```
azevaykin, line 148: Please, hide the magic number 2000. It has no semantic meaning for tests like this.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2515,4 +2556,123 @@ void SendViaPipeCache(
...
+
+    return WaitReadResult(server, timeout);
+}
+
+TString ReadTable(
+    Tests::TServer::TPtr server,
+    std::span<const ui64> tabletIds,
+    const TString& tableName,
+    const TTableId& tableId,
+    ui64 startReadId)
```
azevaykin, line 2655:
```suggestion
    ui64 startReadId = 1000)
```

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
MBkkt: @lex007in just rename it to SendReadPart

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
azevaykin: Or even remove it.  
But leave it as it is in GetBaseReadRequest.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2515,4 +2556,123 @@ void SendViaPipeCache(
...
+    ui64 readId,
+    NKikimrDataEvents::EDataFormat format,
+    const TRowVersion& readVersion)
+{
+    auto request = std::make_unique<TEvDataShard::TEvRead>();
+    auto& record = request->Record;
+
+    record.SetReadId(readId);
+    record.MutableTableId()->SetOwnerId(tableId.PathId.OwnerId);
+    record.MutableTableId()->SetTableId(userTable.GetPathId());
```
azevaykin, line 2586:
```suggestion
    record.MutableTableId()->SetTableId(tableId.PathId.LocalPathId);
```

`userTable` is not necessary.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2515,4 +2556,123 @@ void SendViaPipeCache(
...
+{
+    SendReadAsync(server, tabletId, request, sender, node, clientConfig, clientId);
+
+    return WaitReadResult(server, timeout);
+}
+
+TString ReadTable(
+    Tests::TServer::TPtr server,
+    std::span<const ui64> tabletIds,
+    const TString& tableName,
```
azevaykin, line 2653: https://github.com/ydb-platform/ydb/pull/7121/files#r1693924255

azevaykin, line 2653:
```suggestion
```

`tableName` should be removed due to
https://github.com/ydb-platform/ydb/pull/7121/files#r1693924255

ydb/core/tx/datashard/datashard_ut_write.cpp
lex007in: We need to distinguish `ReadTable` calls in one test to generate a unique `readId`. Maybe make parameter like`readOrder` in `ReadTable` and multiply by 1000 and increment inside `ReadTable`? Alternatively, make a test helper class that holds and tracks `readId`?

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
lex007in: We still need `userTable` for column Ids (for making requests) and for column names and column types (for printing).

azevaykin: > We still need `userTable` for column Ids (for making requests) and for column names and column types (for printing).

Yes, you are right.

azevaykin: Nevertheless, let's try remove `tableName`. You can enumerate `tablesMap` and find the `userTable` by `tableId.PathId`. 

If you can do it then the result function call would trivial like that:
```
auto tableState = ReadTable(server, shards, tableId);
```

ydb/core/tx/datashard/datashard_ut_write.cpp
azevaykin: Let's do `ui64 startReadId = 1000` as suggestion here https://github.com/ydb-platform/ydb/pull/7121#discussion_r1693921476

lex007in: So, I'll add a default value for the `startReadId` parameter and leave this non-default 2000 because there are multiple `ReadTable` calls in this particular test (and do so for other simialr), is that correct?

azevaykin: Yes, let's do it.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -34,6 +34,67 @@ using namespace Tests;
...
+
+        auto request = GetBaseReadRequest(tableId, userTable, readId, format);
+
+        AddFullRangeQuery(*request);
+
+        SendReadAsync(server, tabletId, request.release(), sender);
+    }
+
+    void PrintTableFromResult(TStringBuilder& out, const TEvDataShard::TEvReadResult& event, const NKikimrTxDataShard::TEvGetInfoResponse::TUserTable& userTable) {
+        const auto& description = userTable.GetDescription();
```
azevaykin, line 74: Only `description` is needed in this function. So, it can passed to the function and `userTable` could be deleted.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
```diff
@@ -2515,4 +2587,122 @@ void SendViaPipeCache(
...
+    const TRowVersion& readVersion)
+{
+    auto request = std::make_unique<TEvDataShard::TEvRead>();
+    auto& record = request->Record;
+
+    record.SetReadId(readId);
+    record.MutableTableId()->SetOwnerId(tableId.PathId.OwnerId);
+    record.MutableTableId()->SetTableId(userTable.GetPathId());
+
+    const auto& description = userTable.GetDescription();
```
azevaykin, line 2619: Only `description` is needed in this function. So, it can passed to the function and `userTable` could be deleted.

ydb/core/tx/datashard/ut_common/datashard_ut_common.cpp
lex007in: From any shard?

azevaykin: You should call `userTable.GetDescription()` outside `PrintTableFromResult` and pass only `description` inside.

azevaykin (approved):

PR #7246, opened 2024-07-30
renaming mvp config parameters

StekPerepolnen:
### Changelog entry 
renamed oidc `endpoint` parameters to `url-path`

### Changelog category 

* Improvement

### DIFF
--- a/ydb/mvp/oidc_proxy/mvp.cpp
+++ b/ydb/mvp/oidc_proxy/mvp.cpp
@@ -229,9 +229,9 @@ void TMVP::TryGetOidcOptionsFromConfig(const YAML::Node& config) {
     OpenIdConnectSettings.SessionServiceEndpoint = oidc["session_service_endpoint"].as<std::string>("");
     OpenIdConnectSettings.SessionServiceTokenName = oidc["session_service_token_name"].as<std::string>("");
     OpenIdConnectSettings.AuthorizationServerAddress = oidc["authorization_server_address"].as<std::string>("");
-    OpenIdConnectSettings.AuthEndpoint = oidc["auth_endpoint"].as<std::string>(OpenIdConnectSettings.DEFAULT_AUTH_ENDPOINT);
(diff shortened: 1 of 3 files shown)


UgnineSirdis (dismissed):

ydb/mvp/oidc_proxy/mvp.cpp
```diff
@@ -229,9 +229,9 @@ void TMVP::TryGetOidcOptionsFromConfig(const YAML::Node& config) {
     OpenIdConnectSettings.SessionServiceEndpoint = oidc["session_service_endpoint"].as<std::string>("");
     OpenIdConnectSettings.SessionServiceTokenName = oidc["session_service_token_name"].as<std::string>("");
     OpenIdConnectSettings.AuthorizationServerAddress = oidc["authorization_server_address"].as<std::string>("");
-    OpenIdConnectSettings.AuthEndpoint = oidc["auth_endpoint"].as<std::string>(OpenIdConnectSettings.DEFAULT_AUTH_ENDPOINT);
-    OpenIdConnectSettings.TokenEndpoint = oidc["token_endpoint"].as<std::string>(OpenIdConnectSettings.DEFAULT_TOKEN_ENDPOINT);
-    OpenIdConnectSettings.ExchangeEndpoint = oidc["exchange_endpoint"].as<std::string>(OpenIdConnectSettings.DEFAULT_EXCHANGE_ENDPOINT);
+    OpenIdConnectSettings.AuthUrlPath = oidc["auth_url_path"].as<std::string>(OpenIdConnectSettings.DEFAULT_AUTH_URL_PATH);
```
UgnineSirdis, line 232: Why do we need such detailed parameters? Why they are not simply `https://<HOST>:<PORT>/<PATH>`?

StekPerepolnen: Initially, the PATH config was not stored in the config on the right side, and this behavior remains. This was done so that the right side could keep the existing config and everything would continue to work as before. The left side was made similarly because I thought it would be more complicated to deal with if Yandex and Nebius were implemented differently.

UgnineSirdis (approved):

PR #7324, opened 2024-07-31
Limit batches in PDisk, add lwprobes to detect batch bursts

va-kuznecov:
### Changelog entry 

...

### Changelog category 

* New feature
* Experimental feature
* Improvement
* Performance improvement
* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
+++ b/ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
@@ -13,6 +13,8 @@
 
 #include <util/system/unaligned_mem.h>
 
+constexpr size_t MAX_REQS_PER_CYCLE = 200; // 200 requests take ~0.2ms in EnqueueAll function
+
 namespace NKikimr {
 namespace NPDisk {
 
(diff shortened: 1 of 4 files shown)


ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
```diff
@@ -13,6 +13,8 @@
 
 #include <util/system/unaligned_mem.h>
 
+constexpr size_t MAX_REQS_PER_CYCLE = 200; // 200 requests take ~0.2ms in EnqueueAll function
```
robdrynkin, line 16: Maybe MAX_REQS_PER_CYCLE -> MAX_US_PER_CYCLE=200 and check timer below?

ydb/tools/oss_slice/__init__.py
robdrynkin, line 1: ?

ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
va-kuznecov: Yes, it would be better if TInstant::Now() were very fast. But, unfortunately, it's very expensive to call Now() on every loop iteration, so we need to use workarounds

ydb/tools/oss_slice/__init__.py
va-kuznecov: my bad

va-kuznecov:
moved to https://github.com/ydb-platform/ydb/pull/9099

PR #7384, opened 2024-08-01
grace join: fix FROM ANY vs bloomfilter

yumkam:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

Regression by #6951/774401feac5989c0a56eec46682e9d3827bdf3d1


yumkam:
foo

|A|B|
|-|-|
| a | b |
| a | c |

bar
|A|
|-|
empty

```sql
SELECT * FROM ANY foo LEFT JOIN bar USING(A)
```
should produce one row, but produces two rows.

aakulaga-ydb (approved):

PR #7448, opened 2024-08-05
Add YQL syntax highlighting to YDB CLI

vityaman:
### Changelog entry 

Add YQL syntax highlighting to YDB CLI

### Changelog category 

* New feature

### Additional information

- Use YQL ANTLR4 grammar from the backend
- Generate ANTLR recognizers from the grammar
- Highlight YQL syntax using generated lexer
- Mimic `Monaco` theme like in the Web UI


vityaman:
There are no issue for this feature currently. Maybe it exists, but I haven't found it.  @asmyasnikov, should I create an issue to `Add YQL syntax highlighting to YDB CLI`. Also, there are no issue for suggestions (autocompletion).

vityaman:
The presented solution is naive, but simple. It would also be good to run a parser to get more information to highlight different types of names in different colors.

vityaman:
I have some kind of internal concern about the CPU load, because parsing is triggered on every input buffer change event.

I conducted such an experiment. 

1. Copied the `SELECT CustomerName, City FROM Customers request;`. 

2. Opened the test application with `replxx` and `htop -p ...` to watch the `CPU` time.

3. For ~10 seconds, I copied this request and fixed the `CPU` indicator at the end.

Results. Without the feature, CPU was `4.3`, with the feature is `31.7`.
Conclusion. When we add such features to our REPL, it starts to be more cpu intensive.

vityaman:
Currently, there are no tests, because I don't know, what is the best way to make them. 

I propose to add a unit test (`ut`) for `YQLHighlight::Apply` at `ydb/public/lib/ydb_cli/commands/interactive`.

And also I think that it would be nice to have some flag as CLI argument to disable syntax highlighting feature in order not to interfere users if we made a mistake somewhere. Maybe some such flag is already existed (like `--disable-experimental-features`)?

@asmyasnikov, what do you think about it?

vityaman:
Maybe we should check how it works on terminals without color support.

vityaman:
This MR is also useful to evaluate the impact of adding `ANTLR Runtime` on the executable binary size.

ydb/public/lib/ydb_cli/commands/interactive/yql_highlight.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+        namespace {
+
+            YQLHighlight::Color ColorOf(const YQLHighlight::ColorSchema& schema,
+                                        const antlr4::Token* token) {
+                const auto type = token->getType();
+                if (type == YQLLexer::STRING_VALUE) {
+                    return schema.string;
+                }
+                if (YQLLexer::ID_PLAIN <= type && type <= YQLLexer::ID_QUOTED) {
```
pnv1, line 28: Why <= and not ==? Looks weird

vityaman: I agree. Wanted to be consistent with this range checks. Also, I have an idea to add static assertions here to make sure that we have covered all token types.

pnv1: Not a big deal if we won't cover some new type. It will just be white, right?
But with this code we rely on enum's actual values. Not sure if its good at all. It is generated, right? So if somebody changes the order of type declarations in grammar file (which is not restricted), values will change and this code may be broken?

vityaman: Yes, if someone changes the order of definitions in the grammar, the code will stop working. To avoid this, I can replace this with a large switch case that uses all tokens.

pnv1: > To avoid this, I can replace this with a large switch-case that uses all tokens.

Looks like there is no other choice. We show explicitly what token types we want to color. And do not rely on the order of declaration appearance in the grammar.

A default option should be everything unknown, with a default color. So that we will notice it if we miss a type or a new type will appear.

vityaman: @pnv1, I rewrote `ColorOf` function to use switch-case. 

There is no reason to make it in a single `switch-case`, because we can't check it for an exhaustiveness, as ANTLR4-generated Lexer stores token indexes in an anon enum. So I decided to split it into functions as in future there can be added more logic like highlighting table and column names differently.

It should be optimized into huge RO table lookup (https://godbolt.org/z/x3avjEdaG).

The other option is to use `std::unordered_map` that can be less performant.

pnv1: Yeah, enum is huge now. But since we have no other choice, I think its good

pnv1 (dismissed):

vityaman:
There are also a few things to do:

- [ ] Add [monaco](https://github.com/ydb-platform/monaco-yql-languages/blob/main/src/themes/themes.contribution.ts#L27) color schema to be consistent with embedded UI.

- [ ] After a merge of a new ANTLR4 grammar for YQL, rebase and use that grammar instead of copy-pasted.

vityaman:
I am also thinking about doing fork from this branch and add autosuggestions there just to make a prototype and get early feedback from you.

vityaman:
Hm, not sure that the cause of tests failure is CLI highlighting. Maybe I should rebase onto trunk. I am going to ignore it for now, focusing only on code related to highlighting and suggestions. When we agree on the changes and ANTLR4 YQL grammar from the backend will be used, then I'll start to prepare MRs for a merging.

@asmyasnikov, I think we do not really need to mark `ok-to-test` before that ANTLR4 YQL grammar will be merged, because there are no tests related to my area yet. It will save test runner resources and also bot will not send these messages that clutter up the dialogue a bit.

vityaman:
# Monaco theme

## Sample query

```sql
SELECT 123467, "Hello, {name}!", (1 + (5 * 1 / 0)), MIN(identifier)
FROM `local/test/space/table` JOIN test;
```

## YDB Embedded UI

I used a code editor at [Yandex Query](https://yq.yandex.cloud/folders/) to get an actual YQL code highlighting. 

image

image

## Monaco colors

I asked DeepSeek to give me hex colors of the Monaco theme and prepared the table.

image

image

RGB 0-5 is used because of terminal colors limitation (at least at Replxx API).

Application of these colors produces the following result.

image

image

image

And here I understood that these colors are for a dark theme and hard to read on a bright background. Theoretically, we can check for a terminal background and apply a different theme based on it, but it seems an overkill (but nice) for me.

## Simple VSCode-like

Then I tried that colors...

image

...and got

image

image

image

It seems quite readable for me and It looks familiar. 

Identifiers can be made darker. I don't like it.

image

Also braces and operators can be made darker to highlight them a little, while keeping identifiers with a default color. I like it.

image

image

image

## Missed features

1. Now quoted identifiers are highlighted differently in Embedded UI and the same can easily be  done in CLI. **Do we need it?**

2. Function, column, table and etc names are highlighted with the same color. This is because we use only lexical analysis stage for highlighting, where there are all only identifiers. To highlight differently, we should run parser and inspect a `ParseTree` (with error nodes, but something will be valid). This will provide us with a nice colors, but a little bit more complicated (current implementation is so simple). **Do we need it?**

pnv1:
@vityaman
1. The colours are OK by itself. But the whole idea was to be more consistent with WEB UI colors, isn't it? Strings in UI are red/orange while in your last picture its yellow. Path to table in UI was dark green vs light grey in last picture.
If the problem is that the colors are too pale for CLI, maybe we should find colors that looks bright in CLI (on black/dark background in the first place), but also corresponds to UI colors? I.e. Strings will remain red/orange, path to table wil remain (dark) green.
2. Can you give an example? What is the difference?
3. The more consistent with UI - the better. I hope the difference in performance shouldn't be noticeable for user. The only question I guess is how hard it is to implement.

vityaman:
@pnv1

> Path to table in UI was dark green vs light grey in last picture.

This is what I meant in the second question. Tried to choice another color.

> Strings in UI are red/orange while in your last picture its yellow

Tried to choice another color.

New version.

```rust
return {
  .keyword = Color::BLUE,
  .operation = rgb666(3, 3, 3),
  .simple_identifier = Color::DEFAULT,
  .quoted_identifier = rgb666(1, 3, 3),
  .string = rgb666(3, 0, 0),
  .number = Color::BRIGHTGREEN,
  .unknown = Color::DEFAULT,
};
```

image

image

While Web UI looks like this.

image

image


>  The more consistent with UI - the better. I hope the difference in performance shouldn't be noticeable for user. The only question I guess is how hard it is to implement.

I will return with an answer a later, will figure out how to do this. Maybe this can be done using parse listeners. For example, on each `TerminalNode` examine its `ParseStack` to make a color decision, or switching some `currentColor` on interested rule entrance highlighting each terminal node with this `currentColor`.

pnv1:
> image
> 
> image
> 

I like new version. Colors are readable on both light and dark themes and looks pretty similar to those in WEB UI

vityaman:
@pnv1

I explored syntax highlighting through parser a little. It was easy to rewrite existing solution to use listener, but highlighting functions is harder, than I thought, as it requires some non-trivial `ParseTree` analysis. With such complex logic unit tests are essential.

But I found myself wanting to essentially find out if the current token is followed by a bracket to recognize the function what can be done without a parser with just a little hack at token list traversal.

Actually we can't highlight function names even without following brace after an identifier, because we don't have enough semantic information. To do this, we need the predicate on a names space. 

Questions

1. Web UI is able to do this, because they has predefined list of [builtinFunctions](https://github.com/ydb-platform/monaco-yql-languages/blob/v1.0.6/src/yql/yql.keywords.ts#L10), also they have [such rule](https://github.com/ydb-platform/monaco-yql-languages/blob/v1.0.6/src/yql/yql.ts#L93) for it. I can port these rules to YDB CLI. **Do we need it?**

2. While Monaco Editor does highlighting using regex, we still can consider to run parser in CLI to highlight different identifiers differently. For example, highlight column names with color different from table. Web UI is not capable of it, so this is an extra feature (seems is not for this MR). **Do we need it?**

pnv1:
> Questions
My opinion: 
1. Predefined list of built-in functions is better than current solution where everything is white. Also it is consistent with UI.
2. Its good that we have that opportunity. But until it is realised, the version with predefined list like in UI is the best option

ydb/public/lib/ydb_cli/commands/interactive/yql_highlight.h
```diff
@@ -0,0 +1,68 @@
...
+                    Color type;
+                    Color variable;
+                    Color quoted;
+                } identifier;
+                Color string;
+                Color number;
+                Color unknown;
+
+                static ColorSchema Monaco();
+                static ColorSchema Debug();
```
vityaman, line 35: This color schema has better contrast and contains unique colors, so it is applicable for unit and manual testing.

ydb/public/lib/ydb_cli/commands/interactive/yql_highlight_ut.cpp
```diff
@@ -0,0 +1,172 @@
...
+            "kkkk qqqqqqqqqqqqqqqqqqqqqqqq kkkk vvvvo");
+        Check(highlight, "SELECT Bool(phone) FROM customer",
+              "kkkkkk ttttovvvvvo kkkk vvvvvvvv");
+    }
+
+    Y_UNIT_TEST(Emoji) {
+        YQLHighlight highlight(Coloring);
+        Check(highlight, "☺", "uuu");
+        Check(highlight, "\"☺\"", "sssuu");
+        Check(highlight, "`☺`", "qqquu");
```
vityaman, line 155: I don't know how it works, but it is highlighted normally:

image

ydb/public/lib/ydb_cli/commands/interactive/yql_highlight.h
```diff
@@ -0,0 +1,68 @@
...
+        public:
+            using Color = replxx::Replxx::Color;
+            using Colors = replxx::Replxx::colors_t;
+
+            struct ColorSchema {
+                Color keyword;
+                Color operation;
+                struct {
+                    Color function;
+                    Color type;
```
vityaman, line 26: Added type highlighting. 

Web UI

image

image

T UI

image

image

ydb/public/lib/ydb_cli/commands/interactive/yql_highlight_ut.cpp
```diff
@@ -0,0 +1,186 @@
+#include "yql_highlight.h"
+
+#include <library/cpp/testing/unittest/registar.h>
+#include <util/string/split.h>
+#include <util/system/compiler.h>
+
+using namespace NYdb::NConsoleClient;
+
+Y_UNIT_TEST_SUITE(YqlHighlightTests) {
```
vityaman, line 9: Decided to add tests because logic became complex.

vityaman:
Also I thought that that predicates with huge switch cases can be rewritten to use `std::initializer_list`. On `-O3` the effect of `std::find` will be the same as switch-case: https://godbolt.org/z/KobG69e9r

vityaman:
Merged a fresh `main` and adopted an existing solution. 

Despite currently I use only lexer which does not have protobuf as an include dependency, I don't really like, that I will depend on protobuf after adding autocompletion as then I will need the AST.

I can't generate a basic CPP parser, because there is a `NULL` token in grammar which is not a problem at the backend as of usage of custom `.stg` that adds `TOKEN_` prefix for each token.

pnv1 (approved):

asmyasnikov (approved):

PR #7511, opened 2024-08-06
Check client certificate/token when option EnforceUserTokenCheckRequirement is on

UgnineSirdis:
### Changelog entry 

Support EnforceUserTokenCheckRequirement option for the case of mTLS

### Changelog category 

* Improvement

### Additional information

...


UgnineSirdis:
If the cluster is in state when no token is required (and no client certificate is checked), this option helps to debug problems that can occur after configuration update. So if current state is dynamic node registration without using token/client certificate, we can setup client cert settings, node registration settings, EnforceUserTokenRequirement=false, EnforceUserTokenCheckRequirement=true in static nodes. After this dynamic nodes that still don't use client certificate will continue register. Then we can try mTLS on dynamic nodes: if client cert is specified, server starts checking it and returns related errors if something is set up wrong. But during this other nodes will continue work.

dcherednik (approved):

PR #7645, opened 2024-08-12
per-tenant balancing in hive

vporyadke:
### Changelog entry 

...

### Changelog category 

* Improvement

### Additional information

...


vporyadke:
This PR makes Hive balancer trigger on a by-subdomain basis, rather than on the whole cluster. The consequences are

* Not triggering balancer on the difference in usage between nodes in different domains
* Being able to run multiple balancers on different domains concurrently (still on the same mailbox) instead of just one balancer on the whole cluster

This is only relevant for
 
* Root Hive in a multi-tenant cluster - which should only have tenant Hives in it... in the future.
* Exclusive nodes in a shared database

Illustrations of the changes in behavior (only trigger logic is changed, not the behavior of the balancer itself):
image
image

PR #7679, opened 2024-08-12
system user access to system statistics table

alexd65536:
### Changelog category 

### DIFF
--- a/ydb/core/statistics/database/ut/ut_database.cpp
+++ b/ydb/core/statistics/database/ut/ut_database.cpp
@@ -3,6 +3,8 @@
 #include <ydb/core/statistics/events.h>
 #include <ydb/core/statistics/database/database.h>
 
+#include <ydb/public/sdk/cpp/client/ydb_table/table.h>
+
 #include <thread>
 
 namespace NKikimr::NStat {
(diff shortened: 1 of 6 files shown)


ydb/core/statistics/database/ut/ut_database.cpp
```diff
@@ -87,6 +89,41 @@ Y_UNIT_TEST_SUITE(StatisticsSaveLoad) {
...
+                SELECT * FROM `/Root/Database/.metadata/_statistics`;
+            )", NYdb::NTable::TTxControl::BeginTx().CommitTx()).ExtractValueSync();
+            status = result.GetStatus();
+        };
+        std::thread testThread(test);
+
+        runtime.SimulateSleep(TDuration::Seconds(10));
+        testThread.join();
+
+        UNIT_ASSERT_VALUES_EQUAL(status, NYdb::EStatus::SCHEME_ERROR);
```
azevaykin, line 124: May be `UNAUTHORIZED` ?

adameat (approved):
I personally suggest to avoid bool's with default values at the end of functions.
and I suggest to replace isSystemUser with some kind of ID / name of actual user.

ydb/core/statistics/database/ut/ut_database.cpp
alexd65536: у нас исторически возвращается в этом месте SCHEME_ERROR

PR #7697, opened 2024-08-13
Fix unexpected read iterator stream reset

### LINKED ISSUE #7674: DataShard: read iterators quota bug and unexpected stream resets
Datashard doesn't handle decreasing quotas very well. When client specifies lower quotas in TEvReadAck while TEvReadContinue or TTxReadContinue are inflight, the read could fail by the `ShouldStop()` condition too early, before `LastProcessedKey` is initialized, which causes query to unexpectedly restart later. It's unclear whether this specific problem causes duplicate and out-of-order rows for some clients.

snaury:
### Changelog entry 

Fixed an issue where a SELECT query could produce unexpected duplicate rows.

### Changelog category 

* Bugfix 

### Additional information

Fixes #7674.

### DIFF
--- a/ydb/core/tx/datashard/datashard__read_iterator.cpp
+++ b/ydb/core/tx/datashard/datashard__read_iterator.cpp
@@ -5,6 +5,7 @@
 #include "datashard_locks_db.h"
 #include "probes.h"
 
+#include <ydb/core/base/counters.h>
 #include <ydb/core/formats/arrow/arrow_batch_builder.h>
 
 #include <ydb/library/actors/core/monotonic_provider.h>
@@ -315,6 +316,8 @@ public:
         , Self(self)
         , TableId(state.PathId.OwnerId, state.PathId.LocalPathId, state.SchemaVersion)
         , FirstUnprocessedQuery(State.FirstUnprocessedQuery)
+        , LastProcessedKey(State.LastProcessedKey)
+        , LastProcessedKeyErasedOrMissing(State.LastProcessedKeyErasedOrMissing)
     {
         GetTimeFast(&StartTime);
         EndTime = StartTime;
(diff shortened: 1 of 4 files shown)


gridnevvvit (approved):

ydb/core/tx/datashard/datashard__read_iterator.cpp
```diff
@@ -1683,6 +1712,7 @@ class TDataShard::TReadOperation : public TOperation, public IReadOperation {
         if (Reader->HasUnreadQueries()) {
             Reader->UpdateState(state, ResultSent);
             if (!state.IsExhausted()) {
```
MBkkt, line 1714: Why don't we need `!wasExhausted` check for else branch?

ydb/core/tx/datashard/datashard__read_iterator.cpp
```diff
@@ -567,6 +573,7 @@ class TReader {
                FirstUnprocessedQuery++;
             else
                FirstUnprocessedQuery--;
+            LastProcessedKey.clear();
```
MBkkt, line 576: Can you explain please when do we need to clear `LastProcessedKey`?
For an example I don't understand why doesn't it needed for `case EReadStatus::StoppedByLimit:`

ydb/core/tx/datashard/datashard__read_iterator.cpp
snaury: Because iterator hasn't produced any results at this time, they cannot be acked and cannot be changed. Since this is the first result ever the exhausted counter hasn't been incremented yet, even if state was somehow exhausted initially.

ydb/core/tx/datashard/datashard__read_iterator.cpp
snaury: LastProcessedKey is used to resume iteration in a separate LocalDB transaction. When it's not set we start iteration as specified in the query. Since we move to the next query I clear LastProcessedKey to guarantee the next query doesn't use some unexpected last processed key from the previous query. Query iteration is supposed to set this key to the correct value before exiting, but conditions have become way too complicated over time, and it's better to be safe than sorry the next time.

When subquery returns StoppedByLimit the key to resume from has already been set to the correct value.

MBkkt (approved):

PR #7743, opened 2024-08-13
Add UT for many source paths

qrort:
### Changelog category 

### Additional information

...


qrort:
https://github.com/ydb-platform/ydb/blob/3eb317d4502e4b84cefabe55e614c32fae88cf43/ydb/public/api/client/yc_private/ydb/v1/backup.proto#L80 to check <= 256 constraint here

PR #7754, opened 2024-08-13
Config option for node registration token

UgnineSirdis:
### Changelog entry 

Config option for dynamic node registration token.
Like config option `StaffApiUserToken` for CMS request.

### Changelog category 

* New feature

### Additional information

...


ydb/core/protos/auth.proto
```diff
@@ -55,6 +55,7 @@ message TAuthConfig {
     optional string AccessServiceType = 79 [default = "Yandex_v2"]; // For now the following values are supported: "Yandex_v2", "Nebius_v1"
     optional string CertificateAuthenticationDomain = 80 [default = "cert"];
     optional bool EnableLoginAuthentication = 81 [default = true];
+    optional string NodeRegistrationToken = 82 [default = "root@builtin", (Ydb.sensitive) = true];
```
ijon, line 58: I would have made a comment about how it is better to not shove default values into the protobuf specs (for a number of reason), but... having seen the entire message spec I think its late.

ijon (approved):

PR #7829, opened 2024-08-15
Wait for all CAs inside Executer before shutdown

abyss7:
### Changelog entry 

Implement additional state for TKqpDataExecuter in which it will wait for all children Compute Actors for graceful shutdown.

Also Executer will respect node disconnects and timeouts - to not hang on waiting forever.

### Changelog category 

* New feature

### DIFF
--- a/ydb/core/kqp/executer_actor/kqp_executer_impl.h
+++ b/ydb/core/kqp/executer_actor/kqp_executer_impl.h
@@ -1713,7 +1713,7 @@ protected:
         LOG_E("Sending timeout response to: " << Target);
 
         Request.Transactions.crop(0);
-        this->PassAway();
+        this->Shutdown();
     }
 
     void FillResponseStats(Ydb::StatusIds::StatusCode status) {
@@ -1778,7 +1778,7 @@ protected:
         ExecuterStateSpan.EndError(response.DebugString());
 
         Request.Transactions.crop(0);
-        this->PassAway();
+        this->Shutdown();
     }
 
 protected:
@@ -1846,6 +1846,12 @@ protected:
     }
 
 protected:
+    // Introduced separate method from `PassAway()` - to not get confused with expectations from other actors,
+    // that `PassAway()` should kill actor immediately.
+    virtual void Shutdown() {
+        PassAway();
+    }
+
     void PassAway() override {
         YQL_ENSURE(AlreadyReplied && ResponseEv);
         this->Send(Target, ResponseEv.release());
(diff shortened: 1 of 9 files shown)


ydb/core/kqp/executer_actor/kqp_executer_impl.h
```diff
@@ -1624,14 +1624,14 @@ class TKqpExecuterBase : public TActorBootstrapped<TDerived> {
...
-            if (task.ComputeActorId) {
+            if (task.ComputeActorId && !task.Meta.Completed) {
                 LOG_I("aborting compute actor execution, message: " << issues.ToOneLineString()
                     << ", compute actor: " << task.ComputeActorId << ", task: " << task.Id);
 
                 auto ev = MakeHolder<TEvKqp::TEvAbortExecution>(NYql::NDq::YdbStatusToDqStatus(code), issues);
                 this->Send(task.ComputeActorId, ev.Release());
             } else {
-                LOG_I("task: " << task.Id << ", does not have Compute ActorId yet");
+                LOG_I("task: " << task.Id << ", does not have the CA id yet or is already complete");
```
Darych, line 1634: Maybe we can add checks here also and write more explicit message, is it already completed or CA id is not known.

ydb/core/kqp/ut/query/kqp_limits_ut.cpp
```diff
@@ -830,6 +833,120 @@ Y_UNIT_TEST_SUITE(KqpLimits) {
...
+        bool firstEvState = false;
+        ui32 totalEvState = 0;
+        TActorId executerId;
+        ui32 actorCount = 3; // TODO: get number of actors properly.
+
+        auto& runtime = *kikimr.GetTestServer().GetRuntime();
+        runtime.SetObserverFunc([&](TAutoPtr<IEventHandle>& ev) {
+            if (ev->GetTypeRewrite() == NYql::NDq::TEvDqCompute::TEvState::EventType) {
+                ++totalEvState;
+                if (!firstEvState) {
```
Darych, line 864: Based on test description I would expect that `firstEvState` is true and after this event it becomes false.

Darych (approved):

ydb/core/kqp/ut/query/kqp_limits_ut.cpp
abyss7: Probably the variable name is too vague - `firstEvState` indicates if we already encountered the first event `State`, so we replace only the first one.

gridnevvvit (approved):

PR #8037, opened 2024-08-20
Use 'mock_cloud' if empty string is passed as cloud id to TFolderServiceAdapterMock

siarheivesialou:


siarheivesialou:
The recent change has broken test_client: previously passing empty string caused 'mock_cloud' to be used.
This PR returns this behavior back.

qyryq (approved):

FloatingCrowbar (approved):

PR #8219, opened 2024-08-23
Remove unnecessary message

MBkkt:
ACCEPTED is unnecessary, because 
1. if schemeshard doesn't restart, it doesn't change anything from schemeshard/datashard point of view
2. if schemeshard restart, it's same because we anyway restart scan for this shard from scratch, so there won't be any difference


ijon:
@CyberROFL take a look, please

ijon (dismissed):

CyberROFL (approved):

PR #8266, opened 2024-08-26
describe handler added default `family` and `familyname` values

StekPerepolnen:
### Changelog entry 

Added default values for `Family` and `FamilyName` fields when use describe handler

### Changelog category 

* Improvement

### Additional information

...


adameat (approved):
I'm very suspicious about this change, but I don't see any other option at the moment.

PR #8388, opened 2024-08-28
Fix PDisk config is being overwritten

SammyVimes:
### Changelog category 

* Bugfix


va-kuznecov (approved):
LGTM

also got verbal OK from cthulhu

PR #8536, opened 2024-08-30
Stopped pushing null as arguments to TCoIf in column shards

pavelvelikhov:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/core/kqp/opt/physical/predicate_collector.cpp
+++ b/ydb/core/kqp/opt/physical/predicate_collector.cpp
@@ -143,12 +143,13 @@ bool AbstractTreeCanBePushed(const TExprBase& expr, const TExprNode* ) {
 
 bool CheckExpressionNodeForPushdown(const TExprBase& node, const TExprNode* lambdaArg) {
     if constexpr (NKikimr::NSsa::RuntimeVersion >= 5U) {
-        if (node.Maybe<TCoIf>() || node.Maybe<TCoJust>() || node.Maybe<TCoCoalesce>()) {
-            // Temporary fix for https://github.com/ydb-platform/ydb/issues/7967
-            if (auto ifPred = node.Maybe<TCoIf>()) {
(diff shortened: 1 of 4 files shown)


ydb/core/kqp/ut/join/data/queries/tpcds34.sql
```diff
@@ -0,0 +1,38 @@
+pragma TablePathPrefix = "/Root/test/ds/";
```
zverevgeny, line 1: I'd suggest to add a minimal query that reproduces the bug being fixed and give a reference to this tpc-h query

ydb/core/kqp/opt/physical/predicate_collector.cpp
```diff
@@ -144,6 +144,12 @@ bool AbstractTreeCanBePushed(const TExprBase& expr, const TExprNode* ) {
 bool CheckExpressionNodeForPushdown(const TExprBase& node, const TExprNode* lambdaArg) {
     if constexpr (NKikimr::NSsa::RuntimeVersion >= 5U) {
         if (node.Maybe<TCoIf>() || node.Maybe<TCoJust>() || node.Maybe<TCoCoalesce>()) {
+            // Temporary fix for https://github.com/ydb-platform/ydb/issues/7967
+            if (auto ifPred = node.Maybe<TCoIf>()) {
```
zverevgeny, line 148: Double check for Maybe<CoIf>. For me, extracting CoIf to a separate code branch would make this logic more readable

ydb/core/kqp/ut/join/data/queries/tpcds34.sql
pavelvelikhov: Added a query to KqpOlap unit test.
Keeping this query though - let's make sure we check all benchmark queries that have failed so far

zverevgeny (approved):

PR #8712, opened 2024-09-04
Fix use-after-free in CommittingOps tracking

snaury:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

There was a problem with CommittingOps tracking, where a distributed transaction could have failed without adding to this set, but would try to remove itself on completion. When other immediate transactions attached to the same version it could cause refcount underflow and remove the version record with iterators still pointing there, causing use-after-free. Reproducing this issue is very tricky, since distributed tx must finish execution before another immediate tx (non read-only and attached to the same version) executes, which usually cannot happen. However, new EvWrite api uses local mvcc snapshots for uncommitted writes, which makes it possible to attach to a version first and execute out-of-order later.

This patch adds strict validation to CommittingOps tracking, and makes sure Remove (with the same version) is only called once after a corresponding Add.

Fixes KIKIMR-21932.

### DIFF
--- a/ydb/core/tx/datashard/datashard_pipeline.h
+++ b/ydb/core/tx/datashard/datashard_pipeline.h
@@ -424,11 +424,13 @@ private:
             ui64 Step;
             ui64 TxId;
             mutable ui32 Counter;
+            mutable ui32 TxCounter;
 
             TItem(const TRowVersion& from)
                 : Step(from.Step)
                 , TxId(from.TxId)
                 , Counter(1u)
+                , TxCounter(0u)
             {}
 
             friend constexpr bool operator<(const TItem& a, const TItem& b) {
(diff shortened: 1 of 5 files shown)


kungasc (dismissed):

kungasc (approved):

ydb/core/tx/datashard/datashard_pipeline.h
```diff
@@ -458,17 +462,29 @@ class TPipeline : TNonCopyable {
...
-            if (auto it = TxIdMap.find(txId); it != TxIdMap.end()) {
-                if (--it->second->Counter == 0)
-                    ItemsSet.erase(it->second);
-                TxIdMap.erase(it);
-            }
+        inline void Remove(ui64 txId, TRowVersion version) {
+            auto it = TxIdMap.find(txId);
+            Y_VERIFY_S(it != TxIdMap.end(), "Removing immediate tx " << txId << " " << version
+                    << " does not match a previous Add");
+            Y_VERIFY_S(TRowVersion(it->second->Step, it->second->TxId) == version, "Removing immediate tx " << txId << " " << version
```
azevaykin, line 469: it->second 8 раз встречается. Можно вынести в отдельную переменную для читаемости.

PR #8738, opened 2024-09-04
fix query multiple rowsets

### LINKED ISSUE #1023: Two result sets are displayed for 1 table row
First result set in correct 
image

But the Second one is incorrect (no rows at all)

image

I expect to see only first result set.

adameat:
### Changelog entry 

closes https://github.com/ydb-platform/ydb-embedded-ui/issues/1023

### Changelog category 

* Bugfix 


alexvru (approved):

antonkovalenko (approved):
nice

PR #8749, opened 2024-09-04
Change grow factor according to the memory in rh hash

lll-phill-lll:
### Changelog entry 

RobinHood hash will grow less on low memory. 
...

### Changelog category 

* New feature

### Additional information

...


yumkam:
I think RH code assumes capacity is power-of-2

Issue #8465, opened 2024-08-29
INTERNAL_ERROR on cancel proposal with followers


spuchin:
YDB version 24-3-7

Aug 29 14:24:21 HOST PROCESS[*]: 2024-08-29T11:24:21.393815Z :KQP_COMPUTE ERROR: SelfId: [7903:7408520626554243129:24368], TxId: 878451059008582, task: 2. Ctx: { SessionId : ydb://session/3?node_id=&id=ZDg5ZTIzZGEtMzhkNDBhZTMtYTA4Zjk1NGUtNzljZGE4ZDU=.  TraceId :TRACE.  CustomerSuppliedId : .  CurrentExecutionId : .  Database : DB.  PoolId : . }. InternalError: INTERNAL_ERROR DEFAULT_ERROR: { <main>: Error: contrib/ydb/core/kqp/executer_actor/kqp_data_executer.cpp:653  CancelProposal(): requirement !state.DatashardState->Follower failed }.

PR #8761, opened 2024-09-04
Do not invalidate request with INTERNAL_ERROR in case of cancelation tx on the follower using depended reads

dcherednik:
The data query on the table with READ_REPLICA_SETTINGS set and STALE_RO one shard tx in case of depended reads (for example joint table with itself) can perform via legacy propose transaction routine on the follower. It can cause YQL_ENSURE if the query was cancelled.

Actually no need to send TEvCancelTransactionProposal event on the folower datashard, we can just skip such shards.

### Changelog category
* Bugfix


dcherednik:
https://github.com/ydb-platform/ydb/issues/8465

gridnevvvit (approved):

PR #8804, opened 2024-09-05
Split WideCombiners state into buckets earlier

lll-phill-lll:
### Changelog entry 
Split wideCombiner's state into buckets earlier - when the memory limit is set. It should prevent OOM after hash table x2 growth
...

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/library/yql/minikql/aligned_page_pool.h
+++ b/ydb/library/yql/minikql/aligned_page_pool.h
@@ -218,7 +218,7 @@ public:
         IsMaximumLimitValueReached = isReached;
     }
 
-    bool GetMaximumLimitValueReached() noexcept {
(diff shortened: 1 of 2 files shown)


ydb/library/yql/minikql/aligned_page_pool.h
```diff
@@ -218,6 +218,10 @@ class TAlignedPagePoolImpl {
         IsMaximumLimitValueReached = isReached;
     }
 
+    bool GetMaximumLimitValueReached() noexcept {
```
zverevgeny, line 221: const

lll-phill-lll: fixed, thanks!

vitstn (approved):

PR #8934, opened 2024-09-09
move redirects to base class

adameat:
### Changelog entry 

move redirects to base class, solve problem with redirects in serverless databases

### Changelog category 

* Improvement
* Bugfix 

### DIFF
--- a/ydb/core/viewer/json_pipe_req.cpp
+++ b/ydb/core/viewer/json_pipe_req.cpp
@@ -93,24 +93,32 @@ void TViewerPipeClient::SendDelayedRequests() {
     }
 }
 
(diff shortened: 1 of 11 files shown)


alexvru (dismissed):

alexvru (approved):

antonkovalenko (commented):
>solve problem with redirects in serverless databases

could you please describe the problem, how was it reproduced?

adameat:
> > solve problem with redirects in serverless databases
> 
> could you please describe the problem, how was it reproduced?

serverless databases generate unnecessary redirects. they can't be seen, but they only affect performance.

antonkovalenko (approved):

PR #8984, opened 2024-09-09
Execute LogWrite and ChunkWrite from GetEvents thread in PDisk

va-kuznecov:
### Changelog entry 

...

### Changelog category 

* Experimental feature

### Additional information

...

### DIFF
--- a/ydb/core/blobstorage/pdisk/blobstorage_pdisk_completion.h
+++ b/ydb/core/blobstorage/pdisk/blobstorage_pdisk_completion.h
@@ -20,7 +20,7 @@ struct TCompletionAction {
     TString ErrorReason;
     // Only reads should be executed in a separate thread(s) sinse their complitions consist of
     // time-consuming decyphering of read data
-    bool ShouldBeExecutedInCompletionThread = false;
+    bool ShouldBeExecutedInCompletionThread = true;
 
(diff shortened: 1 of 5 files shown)


ydb/core/blobstorage/pdisk/blobstorage_pdisk_completion.h
```diff
@@ -18,6 +18,11 @@ struct TCompletionAction {
     NWilson::TTraceId TraceId;
     EIoResult Result = EIoResult::Unknown;
     TString ErrorReason;
+    // Only reads should be executed in a separate thread since their complitions consist of
+    // time-consuming decyphering of read data. But currently some completion actions can write
+    // to BlockDevice from Exec() and it's more safe to use WhiteList to allow only
+    // LogWrite and ChunkWrite to be executed from GetThread
+    bool ShouldBeExecutedInCompletionThread = true;
```
SammyVimes, line 25:
```suggestion
    // Only reads should be executed in a separate thread since their completions consist of
    // time-consuming deciphering of read data. But currently some completion actions can write
    // to BlockDevice from Exec() and it's more safe to use WhiteList to allow only
    // LogWrite and ChunkWrite to be executed from GetThread
    bool ShouldBeExecutedInCompletionThread = true;
```

alexvru (approved):

PR #9036, opened 2024-09-10
Implement excess key columns drop for BlockMapJoinCore computation node

igormunkin:
This patchset implements drop of the specified key columns, to reduce excess data in the result, produced by the `BlockMapJoinCore` computation node. As a result of the patch, program builder handler for `BlockMapJoinCore` accepts the vector with the specified key columns, that match indices of the corresponding "left" wide flow fields.

It's worth mentioning why only **left key** columns are processed:
* Right key columns are dropped by default, since they are the same as the left ones;
* Right payload columns respect the following contract: input type provides only those fields, required by output wide type;
* Left non-key columns also respect the aforementioned contract.

All in all, only the key columns used to process the join, but not required in the output type, should be processed by this computation node.

**NB**: MKQL version is not bumped, since there is no way to emit `BlockMapJoinCore` at the moment.

### Changelog category

* Improvement

### DIFF
--- a/ydb/library/yql/minikql/comp_nodes/mkql_block_map_join.cpp
+++ b/ydb/library/yql/minikql/comp_nodes/mkql_block_map_join.cpp
@@ -27,7 +27,7 @@ template <bool RightRequired>
 class TBlockJoinState : public TBlockState {
 public:
     TBlockJoinState(TMemoryUsageInfo* memInfo, TComputationContext& ctx,
-                    const TVector<TType*>& inputItems, const TSet<ui32>& keyDrops,
+                    const TVector<TType*>& inputItems, const THashSet<ui32>& keyDrops,
                     const TVector<TType*> outputItems,
                     NUdf::TUnboxedValue**const fields)
         : TBlockState(memInfo, outputItems.size())
@@ -175,7 +175,7 @@ private:
     size_t InputWidth_;
     size_t OutputWidth_;
     TUnboxedValueVector Inputs_;
-    const TSet<ui32> KeyDrops_;
+    const THashSet<ui32> KeyDrops_;
     const std::vector<arrow::ValueDescr> InputsDescr_;
     TVector<std::unique_ptr<IBlockReader>> Readers_;
     TVector<std::unique_ptr<IBlockItemConverter>> Converters_;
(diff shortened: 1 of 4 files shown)


ydb/library/yql/minikql/comp_nodes/mkql_block_map_join.cpp
```diff
@@ -27,13 +39,14 @@ template <bool RightRequired>
 class TBlockJoinState : public TBlockState {
 public:
     TBlockJoinState(TMemoryUsageInfo* memInfo, TComputationContext& ctx,
-                    const TVector<TType*>& inputItems,
+                    const TVector<TType*>& inputItems, const TSet<ui32>& keyDrops,
```
vitstn, line 42: если порядок неважен, по умолчанию лучше использовать HashSet

ydb/library/yql/minikql/comp_nodes/mkql_block_map_join.cpp
```diff
@@ -54,30 +67,40 @@ class TBlockJoinState : public TBlockState {
 
     void CopyRow() {
         // Copy items from the "left" flow.
+        size_t builderIndex = 0;
         for (size_t i = 0; i < InputWidth_; i++) {
-            AddItem(GetItem(i), i);
+            if (KeyDrops_.contains(i)) {
```
vitstn, line 72: кажется тут вообще удобнее всего иметь vector<ui8> шириной Width для максимально быстрой проверки

vitstn: может быть вообще сделаем Vector<ui32> с реальными входными индексами (выходные лежат плотно)? тогда цикл будет идти только по реально используемым

vitstn (changes requested):
!

ydb/library/yql/minikql/comp_nodes/mkql_block_map_join.cpp
igormunkin: [Fixed](https://github.com/ydb-platform/ydb/pull/9036/commits/36ab859e3b7f34e9d79ecc0550309ab0ccab1aa4).

ydb/library/yql/minikql/comp_nodes/mkql_block_map_join.cpp
igormunkin: Neat! [Done](https://github.com/ydb-platform/ydb/pull/9036/commits/0cb5832f0a8192e820932f7a377343b69d0cc909).

vitstn (approved):

Issue #8942, opened 2024-09-09
NodeBroker: delta protocol with incremental updates


snaury:
- [ ] Write RFC
- [x] Implement blue-green workaround
- [x] Merge blue-green workaround to 24-3
- [x] Merge blue-green workaround to 24-2
- [ ] NodeBroker refactoring?
- [ ] DynamicNameserver refactoring?
- [ ] Implement delta protocol in NodeBroker
- [ ] Implement delta protocol in DynamicNameserver

snaury:
Closing in favor of #11064.

PR #9122, opened 2024-09-12
NodeBroker: use deltas when returning recently added nodes

snaury:
### Changelog entry 

NodeBroker now sends deltas when listing recently added nodes.

### Changelog category 

* Bugfix 

### Additional information

This fixes one of the issues where a blue-green deploy is stalled in big clusters due to cached node lists becoming invalidated with each new added node and excessive traffic when it is repeatedly resent to all nodes in the cluster. With a minor backwards compatible change to TEvListNodes it becomes possible to only send recently added nodes.

Related to #8942.

### DIFF
--- a/ydb/core/mind/node_broker.cpp
+++ b/ydb/core/mind/node_broker.cpp
@@ -290,14 +290,45 @@ void TNodeBroker::AddDelayedListNodesRequest(ui64 epoch,
 
 void TNodeBroker::ProcessListNodesRequest(TEvNodeBroker::TEvListNodes::TPtr &ev)
 {
-    ui64 version = ev->Get()->Record.GetCachedVersion();
+    auto *msg = ev->Get();
 
     NKikimrNodeBroker::TNodesInfo info;
     Epoch.Serialize(*info.MutableEpoch());
     info.SetDomain(AppData()->DomainsInfo->GetDomain()->DomainUid);
     TAutoPtr<TEvNodeBroker::TEvNodesInfo> resp = new TEvNodeBroker::TEvNodesInfo(info);
-    if (version != Epoch.Version)
+
+    bool optimized = false;
+
+    if (msg->Record.HasCachedVersion()) {
+        if (msg->Record.GetCachedVersion() == Epoch.Version) {
+            // Client has an up-to-date list already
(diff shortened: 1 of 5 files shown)


CyberROFL (dismissed):

CyberROFL (approved):

ydb/core/mind/node_broker_impl.h
```diff
@@ -341,6 +341,10 @@ class TNodeBroker : public TActor<TNodeBroker>
     TSchedulerCookieHolder EpochTimerCookieHolder;
     TString EpochCache;
 
+    TString EpochDeltasCache;
+    TVector<ui64> EpochDeltasVersions;
+    TVector<ui64> EpochDeltasEndOffsets;
```
pixcc, line 346: Should we put it all in one class? With interface something like:

```
AddDelta(delta, version);
Clear();
ContainsVersion(version);
GetDelta(version);
```

ydb/core/mind/node_broker.cpp
```diff
@@ -290,14 +290,45 @@ void TNodeBroker::AddDelayedListNodesRequest(ui64 epoch,
...
+        if (msg->Record.GetCachedVersion() == Epoch.Version) {
+            // Client has an up-to-date list already
+            optimized = true;
+        } else {
+            // We may be able to only send added nodes in the same epoch when
+            // all deltas are cached up to the current epoch inclusive.
+            ui64 neededFirstVersion = msg->Record.GetCachedVersion() + 1;
+            if (!EpochDeltasVersions.empty() &&
+                EpochDeltasVersions.front() <= neededFirstVersion &&
+                EpochDeltasVersions.back() == Epoch.Version &&
```
pixcc, line 312: Is this an additional defence or is there really a case where this is false?

pixcc (approved):

ydb/core/mind/node_broker.cpp
snaury: Yes, this protects against version incrementing without a corresponding delta. This should currently only happen when epoch increments, and we clear deltas in that case, but it's better to be defensive than sending incorrect replies erroneously.

PR #9239, opened 2024-09-13
Balancing: add new limits, timeouts and sensors, fix blob on main by ingress, but not realy

robdrynkin:

### DIFF
--- a/ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
+++ b/ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
@@ -51,7 +51,7 @@ namespace NBalancing {
             } else if (ev->Sender == DeleterId) {
                 IsDeleteCompleted = true;
             } else {
-                STLOG(PRI_WARN, BS_VDISK_BALANCING, BSVB05, "Unexpected id", (Id, ev->Sender));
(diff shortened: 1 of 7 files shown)


ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
```diff
@@ -160,6 +164,11 @@ namespace NBalancing {
                 }
 
                 merger.Clear();
+
+                if (SendOnMainParts.Size() >= MAX_TO_SEND_PER_EPOCH && TryDeleteParts.Size() >= MAX_TO_DELETE_PER_EPOCH) {
```
va-kuznecov, line 180: maybe || ?

if any of limits is exceeded actor should stop.

va-kuznecov (changes requested):

ydb/core/blobstorage/vdisk/balance/balancing_actor.cpp
robdrynkin: No, thats not a mistake, I check `TryDeleteParts.Size() < MAX_TO_DELETE_PER_EPOCH` and `SendOnMainParts.Size() < MAX_TO_SEND_PER_EPOCH` above before pushing to queues. If I will check `||`, there could be a situation when I will only send parts on main for a long time, without deletes

va-kuznecov (approved):

PR #9333, opened 2024-09-16
CBO cost function for YT

alephonea:
* Cost of a join is calculated as the sum of children costs, input byte sizes, and output byte size.
* Output byte size is set proportional to output cardinality.
* The cardinality of output is estimated using number of unique values.
* In CollectCboStats(), fetch information necessary for estimating table's in-memory size.
* Change CalcInMemorySize() to calculate lower bound in memory when the second leaf label is unknown.
* Cost is adjusted for Map join and Lookup join.
  - In case of lookup join, only fraction of table size for the large table contributes to cost
  - map join ops can be fused, avoiding the cost of reading the large table repeatedly.


ydb/library/yql/providers/yt/provider/yql_yt_provider_context.cpp
```diff
@@ -0,0 +1,214 @@
...
+    const TVector<TString>& /*rightJoinKeys*/,
+    EJoinAlgoType joinAlgo,
+    EJoinKind /*joinKind*/) {
+    if (joinAlgo == EJoinAlgoType::LookupJoin) {
+        return !leftJoinKeys.empty() && IsLookupJoinApplicable(*left->Stats, *right->Stats, leftJoinKeys);
+    }
+    if (joinAlgo == EJoinAlgoType::MapJoin) {
+        return IsMapJoinApplicable(*right->Stats);
+    }
+    return joinAlgo == EJoinAlgoType::MergeJoin;
```
pavelvelikhov, line 143: This is somewhat wierd from a clarity point of view
Why not just return true at this point?

ydb/library/yql/providers/yt/provider/yql_yt_provider_context.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    if (joinAlgo == EJoinAlgoType::MapJoin &&
+        (leftSpecific->JoinAlgo == EJoinAlgoType::MapJoin || leftSpecific->JoinAlgo == EJoinAlgoType::LookupJoin) &&
+        rightStats.Type == EStatisticsType::BaseTable) {
+        // Optimistically assume that consecutive MapJoins are fused together.
+        leftReadBytes = 0;
+    }
+
+    double outputCost =
```
pavelvelikhov, line 198: So this might be ok for the first iteration, but really you need separate cost functions for different join algorithms

pavelvelikhov (dismissed):

ydb/library/yql/providers/yt/provider/yql_yt_provider_context.cpp
alephonea: Ack
Right now the cost function is simple enough now to include all join algorithms.

ydb/library/yql/providers/yt/provider/yql_yt_provider_context.cpp
alephonea: MergeJoin is used as a generic join algorithm
If we simply return true here, CBO will enumerate many more plans for different values of join algo enum.

rvu1024 (dismissed):

rvu1024 (approved):

PR #9485, opened 2024-09-19
Describe view using legacy msg bus service

jepett0:
We want YDB CLI to return view query text on `ydb scheme describe view_to_describe` command.

There is no generic GRPC service that can return SchemeShard's private protobuf info to YDB CLI. We have to choose from the following options:
1. Create a dedicated GRPC service for views and add only the `DescribeView` call there. This approach seems to be the currently adopted one. See the recently added [replication](https://github.com/ydb-platform/ydb/pull/4904/files#diff-bf2791eba66b7e5364bbd71a7eec8828509fce5456bb00685df8efa8238c7d1b) service and [object storage](https://github.com/ydb-platform/ydb/pull/4434/files#diff-05f77927ddda0bfc47eff950d8e1bb74edd298e1ed3a575aacbbb4dc7ca24cae) service which both have a single RPC in them.
2. Add the `DescribeView` call to the [TableService](https://github.com/ydb-platform/ydb/blob/main/ydb/public/api/grpc/ydb_table_v1.proto). I'm not sure if adding the `DescribeView` call here is a good option since this service is concerned with tables only.
3. Use the legacy GRPC service (a.k.a. Message Bus) to issue a generic [SchemeDescribe](https://github.com/ydb-platform/ydb/blame/d8ee31dd2a8eb84c5487dc662917b2a0099251cf/ydb/public/lib/deprecated/client/grpc_client.h#L57) request that returns a private SchemeShard's protobuf. It would differ significantly from what YDB CLI is currently doing to describe a scheme object, but it would provide a good solution for a generic describe scheme object problem. Whenever we decide that we are ready to remove the legacy `TGrpcClient` we would have to rewrite this code too. However, I don't think it would take too much time to rewrite it.

The current PR implements the third approach.

### DIFF
--- a/ydb/public/lib/ydb_cli/commands/ydb_service_scheme.cpp
+++ b/ydb/public/lib/ydb_cli/commands/ydb_service_scheme.cpp
@@ -1,5 +1,10 @@
 #include "ydb_service_scheme.h"
 
+#include <ydb/core/protos/flat_scheme_op.pb.h>
+#include <ydb/library/grpc/client/grpc_common.h>
+#include <ydb/library/yql/public/issue/yql_issue_message.h>
+#include <ydb/public/lib/base/msgbus_status.h>
+#include <ydb/public/lib/deprecated/client/grpc_client.h>
 #include <ydb/public/lib/json_value/ydb_json_value.h>
 #include <ydb/public/lib/ydb_cli/common/pretty_table.h>
 #include <ydb/public/lib/ydb_cli/common/scheme_printers.h>
@@ -247,10 +252,10 @@ int TCommandDescribe::Run(TConfig& config) {
         return TryTopicConsumerDescribeOrFail(driver, result);
     }
     ThrowOnError(result);
-    return PrintPathResponse(driver, result);
+    return PrintPathResponse(driver, result, config);
 }
 
-int TCommandDescribe::PrintPathResponse(TDriver& driver, const NScheme::TDescribePathResult& result) {
+int TCommandDescribe::PrintPathResponse(TDriver& driver, const NScheme::TDescribePathResult& result, TConfig& config) {
     NScheme::TSchemeEntry entry = result.GetEntry();
     Cout << "<" << EntryTypeToString(entry.Type) << "> " << entry.Name << Endl;
     switch (entry.Type) {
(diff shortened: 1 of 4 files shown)


ydb/public/lib/ydb_cli/commands/ydb_service_scheme.cpp
```diff
@@ -543,6 +550,50 @@ int TCommandDescribe::DescribeReplication(const TDriver& driver) {
...
+        );
+    }
+    if (transportError) {
+        throw yexception() << "grpc transport error message: " << transportError->first
+            << ", code: " << transportError->second;
+    }
+    if (response.status() != NKikimr::NMsgBusProxy::MSTATUS_OK) {
+        throw yexception() << "scheme describe error status: " << response.status()
+            << ", issues: " << NYql::IssuesFromMessageAsString(response.issues());
+    }
```
jepett0, line 592: I haven't written unit tests yet, because I'm not sure this approach would be the accepted one, but if you are interested in the output of the `ydb scheme describe` command, here is how it looks like:

```
...$ ydb scheme describe v
<view> v

Query text: select * from t

...$ ydb scheme describe vvv
Status: SCHEME_ERROR
Issues: 
<main>: Error: Path not found
```

I'm not sure where the error message is generated. You can notice that its format is different from what I have written for the `response.status() != MSTATUS_OK` case.

However, error handling is covered 🤷

ydb/public/lib/ydb_cli/commands/ydb_service_scheme.cpp
```diff
@@ -543,6 +550,50 @@ int TCommandDescribe::DescribeReplication(const TDriver& driver) {
...
+}
+
+int TCommandDescribe::DescribeView(TConfig& config) {
+    const NKikimr::NGRpcProxy::TGrpcError* transportError = nullptr;
+    NKikimrClient::TResponse response;
+    {
+        auto clientConfig = NYdbGrpc::TGRpcClientConfig(config.Address);
+        if (config.EnableSsl) {
+            clientConfig.EnableSsl = config.EnableSsl;
+            clientConfig.SslCredentials.pem_root_certs = config.CaCerts;
```
jepett0, line 565: I have tested:
- no authentication
- YDB token authentication

I imagine I would need to test all of the other methods to be sure that the GRPC client is configured correctly 😢

ydb/public/lib/ydb_cli/commands/ydb_service_scheme.cpp
```diff
@@ -1,5 +1,10 @@
 #include "ydb_service_scheme.h"
 
+#include <ydb/core/protos/flat_scheme_op.pb.h>
```
CyberROFL, line 3: No way.

blinkov:
@jepett0, the legacy MessageBus API is in the middle of being removed from the codebase; you need to go with other options.

jepett0:
Will implement using the approach 1. See [PR](https://github.com/ydb-platform/ydb/pull/9513)

PR #8991, opened 2024-09-09
Capture TablePathPrefix (and other parts of the parser context) in CREATE VIEW

jepett0:
https://github.com/ydb-platform/ydb/issues/8992

We need to capture the parser context in order to use it later during a `SELECT` from the view to ensure that the results of the compilation of the view query text are the same as they were during `CREATE VIEW`. Postgres achieves the same by persisting AST of the view query, but our parser cannot handle this. More details on the problem can be found in the [RFC](https://github.com/ydb-platform/ydb-rfc/blob/main/view_context_capture.md).

PR #9513, opened 2024-09-19
Describe VIEW for YDB CLI

jepett0:
- https://github.com/ydb-platform/ydb/issues/9618

We want YDB CLI to return view query text on `ydb scheme describe view_to_describe` command.

There is no generic GRPC service that can return SchemeShard's private protobuf info to YDB CLI. Ideally, we would like to implement informational tables that contain up-to-date information about the objects in the scheme. However, this is a great undertaking and it will delay the first release of views significantly.

We have decided to add a dedicated GRPC service for views and add (for now) only the `DescribeView` call [there](https://github.com/ydb-platform/ydb/pull/9513/files#diff-3ec42035d720eec1f55d44e44b66c32b675dc34d905bcec845d972d3ddadec95R9). (Eventually, we would also add `CreateView`, `AlterView` and `DropView` calls if users asked.) This approach seems to be the currently adopted one across YDB. See the recently added [replication](https://github.com/ydb-platform/ydb/pull/4904/files#diff-bf2791eba66b7e5364bbd71a7eec8828509fce5456bb00685df8efa8238c7d1b) service and [object storage](https://github.com/ydb-platform/ydb/pull/4434/files#diff-05f77927ddda0bfc47eff950d8e1bb74edd298e1ed3a575aacbbb4dc7ca24cae) service which both have a single RPC in them.


ydb/public/lib/ydb_cli/commands/ydb_service_scheme.cpp
```diff
@@ -543,6 +545,19 @@ int TCommandDescribe::DescribeReplication(const TDriver& driver) {
     return PrintDescription(this, OutputFormat, result, &TCommandDescribe::PrintReplicationResponsePretty);
 }
 
+int TCommandDescribe::PrintViewResponsePretty(const NYdb::NView::TDescribeViewResult& result) const {
+    Cout << "\nQuery text: " << result.GetViewDescription().GetQueryText() << Endl;
```
jepett0, line 549: The output looks like this:
- basic example, the view exists
```
...$ ydb scheme describe v
<view> v

Query text: select * from t
```
- the view does not exist
```
...$ ydb scheme describe vvv
Status: SCHEME_ERROR
Issues: 
<main>: Error: Path not found
```
- the view has [captured](https://github.com/ydb-platform/ydb/pull/8991) the `TablePathPrefix` pragma
```
...$ ydb scheme describe path/to/a/folder/v
<view> v

Query text: PRAGMA TablePathPrefix = "/Root/dedicated/path/to/a/folder";
select * from `/Root/dedicated/v`
```

CyberROFL (commented):
lgtm

ydb/core/grpc_services/rpc_view.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+private:
+    void DescribeScheme() {
+        auto ev = std::make_unique<TEvTxUserProxy::TEvNavigate>();
+        SetAuthToken(ev, *Request_);
+        SetDatabase(ev.get(), *Request_);
+        ev->Record.MutableDescribePath()->SetPath(GetProtoRequest()->path());
+
+        Send(MakeTxProxyID(), ev.release());
```
jepett0, line 38: If you are wondering, all describe requests sent to tx proxy [are going](https://github.com/ydb-platform/ydb/blob/56fbcfc12d0e9b60d405a93e782ffcb084376446/ydb/core/tx/tx_proxy/proxy_impl.cpp#L407) through both:
- [scheme cache](https://github.com/ydb-platform/ydb/blob/56fbcfc12d0e9b60d405a93e782ffcb084376446/ydb/core/tx/tx_proxy/describe.cpp#L293) ([with](https://github.com/ydb-platform/ydb/blob/56fbcfc12d0e9b60d405a93e782ffcb084376446/ydb/core/tx/tx_proxy/describe.cpp#L282) sync version flag) to retrieve the path's SchemeShard id
- [SchemeShard](https://github.com/ydb-platform/ydb/blob/56fbcfc12d0e9b60d405a93e782ffcb084376446/ydb/core/tx/tx_proxy/describe.cpp#L391) to retrieve the most up-to-date info about the path

ydb/public/lib/ydb_cli/commands/ydb_service_scheme.cpp
ijon, line 549: I would add `\n` after the `Query text:` -- to make query text more separate and easy for the user to select.

CyberROFL (dismissed):

ijon (dismissed):

ijon (approved):

PR #9552, opened 2024-09-20
Remove clang14 from PRs and other workflows

maximyurchuk:
### Changelog entry 

Arcadia doesn't need clang14 
Cmake builds are not working at current moment, ignoring them

### Changelog category 


maximyurchuk:
Merged in 
https://github.com/ydb-platform/ydb/pull/9798

Issue #9227, opened 2024-09-13
Syntax highlighting in YDB CLI in interactive mode


vityaman:
We "could-have" YQL syntax highlighting in [YDB CLI](https://github.com/ydb-platform/ydb/tree/main/ydb/apps/ydb) in interactive mode with color schema like we have in [Embedded UI](https://github.com/ydb-platform/ydb-embedded-ui/tree/main/src/utils/monaco/yql) [Monaco](https://github.com/ydb-platform/monaco-yql-languages) editor. This can be done using [YQL ANTLR4 grammar](https://github.com/ydb-platform/ydb/blob/main/ydb/library/yql/sql/v1/SQLv1Antlr4.g.in).

vityaman:
Merged the MR [ Add YQL syntax highlighting to YDB CLI #7448](https://github.com/ydb-platform/ydb/pull/7448).

vityaman:
We noticed, that current implementation is not so good, as it does not distinguish ANSI or extended SQL syntax types and parses text as extended SQL, so ANSI query can be highlighted incorrectly. 

We should add an abstraction and support both SQL dialects. 

Maybe we should use [yql/sql/sql.h interface](https://github.com/ydb-platform/ydb/blob/main/ydb/library/yql/sql/sql.h).

vityaman:
We noticed, that current solution highlight some identifiers incorrectly as in the YQL keywords can be identifiers. 

Examples

```sql
select 1 as window;
select group from (SELECT 1 as group);

select group from (select 1 as group) (group = column)
select min(x) from T group by y; (group = keyword)
```

vityaman:
Connected with 
- https://github.com/ydb-platform/ydb/issues/15129

PR #9687, opened 2024-09-24
#9227 Fix bug with UTF-8 highlighting in YDB CLI

vityaman:
### Changelog entry 

Fix bug with UTF-8 highlighting in YDB CLI

### Changelog category 

* Bugfix 

### Additional information

- Added tests with queries containing UTF-8 characters with length 2, 3, 4
- Fixed a bug in the color filling procedure
- Outcome: ANTLR `Token::(Start|End)Index` returns an index of character in a `TVector<int>`, not `TVector<char>`.
- Connected with #9227 
- Was found during MR #9404, thanks to @nepal


vityaman:
I figured out, that there are no bugs actually in a current implementation. The tests are incorrect, as `replxx` gives us an array of colors for a utf8 string, but I construct an array of colors for ascii string in tests. Will fix this in MR #9404.

PR #9835, opened 2024-09-27
[docs_release] add on/create/branches

blinkov:
### Changelog entry 

...

### Changelog category 

### Additional information

...


blinkov:
This naive approach doesn't work as in this case GitHub sends a different kind of event to the workflow.

PR #9933, opened 2024-10-01
Add condition on DYNAMODB_STREAM_JSON mode

Pseudolukian:
### Changelog category


blinkov:
A similar change has already been merged as part of #9956

PR #9938, opened 2024-10-01
Add common libs for yql plugin. Ydb git part.

### LINKED ISSUE #9398: Global directories for UDFs in YQL agent config
Reported by @gritukan from YTsaurus.

YT uses a YQL agent with a config. There must be an option in this config to specify directories with UDFs available for all queries of all users of the cluster, for instance a Python UDF or an s-expression UDF.

For example, if //home/yql is such a directory and I create file //home/yql/my_sum.yql with the content

```
$script = @@#py
def my_sum(a, b):
    return a + b
@@;
$concat = SystemPython3_8::concat(Callable<(Optional<Int64>, Optional<Int64>)->Int64>, $script);

```

Then my_sum function should be available to all queries.

Tony-Romanov:


gritukan (approved):

Krock21:
Do you have any description or github issue for it?

Could you please link it?

Tony-Romanov:
https://github.com/ydb-platform/ydb/issues/9398

PR #10128, opened 2024-10-05
NodeWarden header optimization

SammyVimes:
### Changelog category 

* Improvement


va-kuznecov (approved):

ydb/core/blobstorage/nodewarden/node_warden_impl.h
```diff
@@ -180,41 +191,7 @@ namespace NKikimr::NStorage {
             return NKikimrServices::TActivity::NODE_WARDEN;
         }
```
va-kuznecov, line 193: This review is totally OK for me, but TNodeWarden is an actor and should be moved to .cpp file

PR #10166, opened 2024-10-07
Disable merges for indexImplTables partitions when build is in progress

jepett0:
Ticket:
- https://github.com/ydb-platform/ydb/issues/8583

### Description of the change

[Do not merge](https://github.com/ydb-platform/ydb/pull/10166/commits/eb8cb41312279104d815f9497a31233717d9d149#diff-3489c7d6c6ac1a283c26f0ebf1056af0bfea000001fcb2303d141016793db947) partitions of indexImplTables if they are not finished being built (i.e. the state of the index object is not equal to [EIndexStateReady](https://github.com/ydb-platform/ydb/blob/f6e0d968f423f1c82824c21783a2b508e2ab6baa/ydb/core/protos/flat_scheme_op.proto#L1038)).

### Context

We have implemented saving split boundaries of indexImplTables to the backup recently:
- https://github.com/ydb-platform/ydb/pull/7589

However, the indexImplTable is created empty during the process of an import from S3. The data is being [added](https://github.com/ydb-platform/ydb/blob/63805c5c7a7ac9bba1e2bc5187d3c14d432db4d2/ydb/core/tx/schemeshard/schemeshard_build_index__progress.cpp#L593) to it at the [filling](https://github.com/ydb-platform/ydb/blob/63805c5c7a7ac9bba1e2bc5187d3c14d432db4d2/ydb/core/tx/schemeshard/schemeshard_info_types.h#L2912) stage of the [BuildIndexes](https://github.com/ydb-platform/ydb/blob/63805c5c7a7ac9bba1e2bc5187d3c14d432db4d2/ydb/core/tx/schemeshard/schemeshard_info_types.h#L2776) phase of the import process. If there are a lot of partitions in the index, it would take a lot of time to fill them. During the filling stage, SchemeShard might decide to merge the partitions of this indexImplTable. We would like to prevent it, because it will slow down the import process significantly. DataShards cannot build index if its partitions are being merged and would have to wait for the merge to end.

### Notes

Commits are separated for your convince:
- preliminary refactoring: c92a1d73fd0fdaf1d538310a8a1398f1cefb77f4
- the main change: eb8cb41312279104d815f9497a31233717d9d149
- unit tests: ccb3ff40aecab1dc99ec1f0a8ca1d75e927257fa

### DIFF
--- a/ydb/core/tx/schemeshard/ut_helpers/helpers.h
+++ b/ydb/core/tx/schemeshard/ut_helpers/helpers.h
@@ -62,6 +62,10 @@
     template<bool OPT1, bool OPT2>                                                                                 \
     void N(NUnitTest::TTestContext&)
 
+namespace NYdb::NTable {
+    struct TGlobalIndexSettings;
+}
+
 namespace NSchemeShardUT_Private {
     using namespace NKikimr;
 
@@ -361,6 +365,7 @@ namespace NSchemeShardUT_Private {
         NKikimrSchemeOp::EIndexType IndexType = NKikimrSchemeOp::EIndexTypeGlobal;
         TVector<TString> IndexColumns;
         TVector<TString> DataColumns;
+        TVector<NYdb::NTable::TGlobalIndexSettings> GlobalIndexSettings = {};
     };
 
     std::unique_ptr<TEvIndexBuilder::TEvCreateRequest> CreateBuildColumnRequest(ui64 id, const TString& dbName, const TString& src, const TString& columnName, const Ydb::TypedValue& literal);
--- a/ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__table_stats.cpp
@@ -264,10 +264,11 @@ bool TTxStoreTableStats::PersistSingleStats(const TPathId& pathId,
                                                                subDomainInfo->EffectiveStoragePools(),
                                                                shardInfo->BindedChannels);
 
+    const auto pathElement = Self->PathsById[pathId];
     LOG_DEBUG_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
                 "TTxStoreTableStats.PersistSingleStats: main stats from"
                     << " datashardId(TabletID)=" << datashardId << " maps to shardIdx: " << shardIdx
-                    << ", pathId: " << pathId << ", pathId map=" << Self->PathsById[pathId]->Name
+                    << ", pathId: " << pathId << ", pathId map=" << pathElement->Name
                     << ", is column=" << isColumnTable << ", is olap=" << isOlapStore);
 
     const TPartitionStats newStats = PrepareStats(ctx, rec, channelsMapping);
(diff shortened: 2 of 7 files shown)


ydb/core/tx/schemeshard/ut_helpers/helpers.h
```diff
@@ -361,6 +365,7 @@ namespace NSchemeShardUT_Private {
         NKikimrSchemeOp::EIndexType IndexType = NKikimrSchemeOp::EIndexTypeGlobal;
         TVector<TString> IndexColumns;
         TVector<TString> DataColumns;
+        TVector<NYdb::NTable::TGlobalIndexSettings> GlobalIndexSettings = {};
```
jepett0, line 368: Please, evaluate the following design choice.

I have decided to use a structure defined in the public API to describe a TBuildIndexConfig in SchemeShard's unit tests.

Pros:
- We don't have to repeat the same code in the unit test helpers.

Cons:
- Using public API for unit test helpers might limit us greatly. We would have to change the public description of an index every time we decide to test some new functionality in the SchemeShard tests.
  - Counterargument: ut helpers already [use](https://github.com/ydb-platform/ydb/blob/33cb4333c3daf2ab4a78a3964e595b7e24623d5b/ydb/core/tx/schemeshard/ut_helpers/helpers.cpp#L1677) a protobuf from the public API to describe an index to be created. It is only natural to use the SDK to work conveniently with the public protobuf. The dependency on the public API and the limitations that come with it would not be lifted, if we didn't use SDK here.

MBkkt (dismissed):

ydb/core/tx/schemeshard/schemeshard_impl.cpp
```diff
@@ -1464,6 +1464,12 @@ const TTableInfo* TSchemeShard::GetMainTableForIndex(TPathId indexTableId) const
     if (!parentEl || !(*parentEl)->IsTableIndex())
         return nullptr;
 
+    if (index) {
+        if (const auto* indexInfo = Indexes.FindPtr(parentId)) {
```
MBkkt, line 1468: Why is it possible that we don't find `indexInfo`?
Maybe it should be an assert?

ydb/core/tx/schemeshard/schemeshard_impl.h
```diff
@@ -559,7 +559,7 @@ class TSchemeShard
     NKikimrSchemeOp::TPathVersion  GetPathVersion(const TPath& pathEl) const;
     ui64 GetAliveChildren(TPathElement::TPtr pathEl, const std::optional<TPathElement::EPathType>& type = std::nullopt) const;
 
-    const TTableInfo* GetMainTableForIndex(TPathId indexTableId) const;
+    const TTableInfo* GetMainTableForIndex(TPathId indexTableId, const TTableIndexInfo** index = nullptr) const;
```
CyberROFL, line 562: `std::optional<TTableIndexInfo*>`?

CyberROFL (approved):

ydb/core/tx/schemeshard/schemeshard_info_types.cpp
```diff
@@ -1932,6 +1933,10 @@ bool TTableInfo::CheckCanMergePartitions(const TSplitSettings& splitSettings,
         return false;
     }
 
+    if (index && index->State != NKikimrSchemeOp::EIndexStateReady) {
+        return false;
+    }
```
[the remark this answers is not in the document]
jepett0: Reworked this completely. Now the check is a separate [clause](https://github.com/ydb-platform/ydb/blob/ccb3ff40aecab1dc99ec1f0a8ca1d75e927257fa/ydb/core/tx/schemeshard/schemeshard__table_stats.cpp#L413) in the if statement that defines if we need to merge partitions.

ydb/core/tx/schemeshard/schemeshard_impl.cpp
jepett0: Changed the code, now this comment is irrelevant.

Now we don't do the double check:
- path element IsTableIndex()
- Indexes contain the path ID
to define that the path ID represents an index.

ydb/core/tx/schemeshard/schemeshard_impl.h
jepett0: Changed the code, now the comment is irrelevant.

Do not change the signature of the GetMainTableForIndex function.

ijon (approved):

PR #10244, opened 2024-10-09
Add AES128 implementation for blobstorage crypto module 

SammyVimes:
### Changelog category 

### Additional information

Not used as of now, only for unit tests

### DIFF
--- a/ydb/core/blobstorage/crypto/aes128.h
+++ b/ydb/core/blobstorage/crypto/aes128.h
@@ -1,0 +1,66 @@
+#include <memory>
+#include <openssl/evp.h>
+#include <openssl/err.h>
+#include <sstream>
+#include <string>
+#include <vector>
+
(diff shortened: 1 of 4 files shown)


ydb/core/blobstorage/crypto/aes128.h
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+class AES128 {
+public:
+    AES128();
+
+    void SetKeyAndIV(const std::vector<ui8>& key, const std::vector<ui8>& iv);
+
+    size_t Encrypt(const ui8* plaintext, ui8* ciphertext, size_t len);
```
va-kuznecov, line 40: I would advice using the usual order of arguments, as in all our code. We use Intel asm notation, which is similar to memcpy notation
`void *memcpy(void dest, const void src, size_t n);`

PR #10333, opened 2024-10-11
KIKIMR-22049: fix signed headers for S3 multipart upload

pixcc:
### Changelog entry 

Fix `There were headers present in the request which were not signed` error returned from MinIO during multipart uploading in S3 exports

### Changelog category 

* Bugfix 

### Additional information

MinIO returns an error when the signed headers contain `x-amz-storage-class` that is not present in the request's http headers

### DIFF
--- a/ydb/core/wrappers/s3_storage.cpp
+++ b/ydb/core/wrappers/s3_storage.cpp
@@ -251,10 +251,10 @@ private:
 
 }; // TInputStreamContext
 
-template <typename TEvRequest, typename TEvResponse, template<typename, typename> typename TBaseContext = TContextBase>
-class TContextWithStorageClass : public TBaseContext<TEvRequest, TEvResponse> {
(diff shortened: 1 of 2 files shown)


ydb/core/wrappers/s3_storage.cpp
```diff
@@ -251,31 +251,41 @@ class TInputStreamContext: public TContextBase<TEvRequest, TEvResponse> {
 
 }; // TInputStreamContext
 
-template <typename TEvRequest, typename TEvResponse>
-class TPutInputStreamContext: public TInputStreamContext<TEvRequest, TEvResponse> {
+template <typename TEvRequest, typename TEvResponse, template<typename, typename> typename TBaseContext = TContextBase>
```
CyberROFL, line 254:
```suggestion
template <typename TEvRequest, typename TEvResponse, template <typename, typename> typename TBaseContext = TContextBase>
```

CyberROFL (approved):

PR #10334, opened 2024-10-11
Plan2svg converter viewer handler

Hor911:


adameat (approved):
don't feel happy about it :disappointed:

PR #10422, opened 2024-10-15
Fix UB in right shift

yumkam:
If the value of rhs is negative or is not less than the number of bits in lhs, the behavior is undefined (quote from cppref, not authoritative, feel free to correct).

### Changelog entry 

...

### Changelog category 

### Additional information

No, overwriting won't help. Compiler can deduce that when prefix == 0, above line leads to UB, and just throw out conditional assignment (and leave value undefined).
(Most likely, this is not a problem with current version of clang, but it is not predictable if/when it won't be)


yumkam:
I have not realized, `ui128::operator >>` is not buiiltin

yumkam:
(ui128 is not builtin, but there are UB inside `ui128::operator>>` when rhs is `>=128`)

avevad (dismissed):

avevad:
Nice catch.
Indeed, ISO C++ [expr.shift 7.6.7.1] says:
> The behavior is undefined if the right operand is
> negative, or greater than or equal to the width of the promoted left operand.

avevad (approved):

Issue #6743, opened 2024-07-16
index build: persist IndexImplTables' descriptions


jepett0:
Index building is a long-running operation and it needs to be able to recover in the same state as before after SchemeShard's reboots. The user-requested parameters of indexImplTables should __not__ be lost.

After a little investigation, it turns out that indexImplTable descriptions have not been persisted in the [PersistCreateBuildIndex](https://github.com/ydb-platform/ydb/blob/bc2bf72c522858fca1678f58f3aeff3e3ef78ce7/ydb/core/tx/schemeshard/schemeshard_build_index.cpp#L59) function.

PR #10579, opened 2024-10-17
Index build: do not lose the requested partitioning info of indexImplTables in case of SchemeShard reboots

jepett0:
Ticket:
- https://github.com/ydb-platform/ydb/issues/6743

### Context

Index building is a long-running operation that is not a single transaction, but many. YDB does persist the state of the individual transactions (i.e. SchemeShard's proposals) and is able to continue them from the last state even if the SchemeShard reboots during the execution of the transaction. However, index build operations require their own persistent [description](https://github.com/ydb-platform/ydb/blob/cebe29d29c824b0985aa74db4567b3ae5fafb564/ydb/core/tx/schemeshard/schemeshard_schema.h#L1281) for the SchemeShard to [continue](https://github.com/ydb-platform/ydb/blob/cebe29d29c824b0985aa74db4567b3ae5fafb564/ydb/core/tx/schemeshard/schemeshard__init.cpp#L4458) them if it reboots.

### Description

[PersistCreateBuildIndex ](https://github.com/ydb-platform/ydb/blob/cebe29d29c824b0985aa74db4567b3ae5fafb564/ydb/core/tx/schemeshard/schemeshard_build_index.cpp#L47) is a function that is responsible for the index build operation ability to recover from the SchemeShard's reboots. As you can see, it didn't persist the ImplTableDescriptions (they store the partitioning info about the indexImplTables to be created) as well as the Specialized Index Description (i.e. specific vector index description). Previously, if the SchemeShard had rebooted before the indexImplTable was created, but after the index build operation had started, then the partitioning info of the indexImplTables would have been lost. A [unit test](https://github.com/ydb-platform/ydb/pull/10579/files#diff-1176b341bd2e55345be061401e1688178cb65a4ce494d0eb5c5b827f3a4ac44d) Is added to reproduce the scenario.

### Note

The main commit is:
- https://github.com/ydb-platform/ydb/pull/10579/commits/a0b430662021c5d2060a2882a9853dd7cca75b1a

The others are supplementary (they add unit tests).

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard_build_index.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_build_index.cpp
@@ -46,7 +46,8 @@ void TSchemeShard::Handle(TEvPrivate::TEvIndexBuildingMakeABill::TPtr& ev, const
 
 void TSchemeShard::PersistCreateBuildIndex(NIceDb::TNiceDb& db, const TIndexBuildInfo& info) {
     Y_ABORT_UNLESS(info.BuildKind != TIndexBuildInfo::EBuildKind::BuildKindUnspecified);
-    db.Table<Schema::IndexBuild>().Key(info.Id).Update(
+    auto persistedBuildIndex = db.Table<Schema::IndexBuild>().Key(info.Id);
+    persistedBuildIndex.Update(
         NIceDb::TUpdate<Schema::IndexBuild::Uid>(info.Uid),
         NIceDb::TUpdate<Schema::IndexBuild::DomainOwnerId>(info.DomainPathId.OwnerId),
         NIceDb::TUpdate<Schema::IndexBuild::DomainLocalId>(info.DomainPathId.LocalPathId),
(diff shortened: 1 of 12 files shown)


ydb/public/sdk/cpp/client/ydb_table/table.cpp
```diff
@@ -2384,6 +2384,9 @@ TGlobalIndexSettings TGlobalIndexSettings::FromProto(const TProto& proto) {
     };
 }
 
+// explicit instantiations to avoid defining the template in the .h file
+template TGlobalIndexSettings TGlobalIndexSettings::FromProto(const Ydb::Table::GlobalIndexSettings& proto);
```
MBkkt, line 2388: Move this function to header sounds better to me

ydb/public/sdk/cpp/client/ydb_table/table.cpp
```diff
@@ -2458,6 +2461,9 @@ TVectorIndexSettings TVectorIndexSettings::FromProto(const TProto& proto) {
     };
 }
 
+// explicit instantiations to avoid defining the template in the .h file
+template TVectorIndexSettings TVectorIndexSettings::FromProto(const Ydb::Table::VectorIndexSettings& proto);
```
MBkkt, line 2465: Move this function to header sounds better to me

ydb/public/sdk/cpp/client/ydb_table/table.cpp
jepett0: Maybe even make it non-template, because the only type it currently uses is Ydb::Table::GlobalIndexSettings. Moreover, SerializeTo can only serialize it to the aforementioned protobuf.

MBkkt (dismissed):

ijon (dismissed):

azevaykin (approved):

PR #10713, opened 2024-10-22
Test math module

Pseudolukian:
### Changelog category


blinkov:
It's easier to do this again than to fix conflicts.

PR #10741, opened 2024-10-22
Fix a bug in YQL node creation for indexImplTable table settings

### LINKED ISSUE #7644: broken `CREATE TABLE` and `ALTER TABLE ADD INDEX` YQL node creation with index table settings
[This code](https://github.com/ydb-platform/ydb/blob/5ff39baa13338aa0310607be94962a86af5d7953/ydb/library/yql/sql/v1/query.cpp#L291) is never called which was verified by @azevaykin by adding `Y_ABORT_UNLESS` in the `if` clause. However, if you make it work by copying the implementation to some other place (for example to [create vector index](https://github.com/ydb-platform/ydb/blob/38223418e2311742bf61d53c0f390abe3f8365fb/ydb/library/yql/sql/v1/query.cpp#L321) node), it will fail with an exception.

We need to fix the bug to enable `CREATE TABLE` and `ALTER TABLE ADD INDEX` clauses with index table settings.

The reason seems to be a missing `node.Q` wrapper.

jepett0:
Ticket:
- https://github.com/ydb-platform/ydb/issues/7644

Forgot to quote the tableSettings node that is created while parsing `CREATE TABLE` with an index definition or `ALTER TABLE ADD INDEX` with indexImplTable settings. It is not possible to issue this commands via YQL, because the relevant grammar is not implemented yet, so it is not possible for users to catch this bug.

For those, who are curios if the test does in fact catch the bug, here is the [log](https://gist.github.com/jepett0/1ddecdbe929ffb914497adf7f584019f) of the test after I have reverted the fix:
<details>
<summary>test failure</summary>

```text
(CompileExpr(*alterTableAst, alterTableExpr, exprContext, nullptr, nullptr)) <main>: Error: First item in list is not an atom, did you forget quote?
```

</details>

### DIFF
--- a/ydb/core/kqp/provider/yql_kikimr_provider_ut.cpp
+++ b/ydb/core/kqp/provider/yql_kikimr_provider_ut.cpp
@@ -1,8 +1,60 @@
 #include <ydb/core/kqp/provider/yql_kikimr_provider_impl.h>
+#include <ydb/library/yql/ast/yql_expr.h>
+#include <ydb/library/yql/sql/v1/source.h>
+#include <ydb/library/yql/sql/v1/context.h>
 
 #include <library/cpp/testing/unittest/registar.h>
 
 namespace NYql {
+
+namespace {
+
+using namespace NSQLTranslationV1;
+
+TContext CreateDefaultParserContext(NYql::TIssues& issues) {
+    NSQLTranslation::TTranslationSettings settings;
+    settings.DefaultCluster = "/Cluster";
+    settings.AssumeYdbOnClusterWithSlash = true;
+
+    return TContext(settings, {}, issues);
+}
+
+TAstNode* CreateAlterTable(TContext& parserContext, const TString& tableName, const TAlterTableParameters& params) {
+    TTableRef tableRef(tableName, parserContext.Scoped->CurrService, parserContext.Scoped->CurrCluster, {});
+    {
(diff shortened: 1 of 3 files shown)


ydb/core/kqp/provider/yql_kikimr_provider_ut.cpp
jepett0, line 1: In my opinion the main purpose of this test is to check all the steps between (and including) the AST node creation and the dedicated TKi node creation. So I have put the test in this file, since the `TKiAlterTable` class is defined in the `<path prefix of generated files>/ydb/core/kqp/provider/yql_kikimr_expr_nodes.decl.inl.h`.

The bug can actually be caught earlier than the TKi node is created. (You can see in the failed test [log](https://gist.github.com/jepett0/1ddecdbe929ffb914497adf7f584019f) that it fails on the `CompileExpr` function.) However, the parser work is finished only when the AlterTableRequest [is built](https://github.com/ydb-platform/ydb/blob/80744192966ba11b73922f86c4a21a018c977dfc/ydb/core/kqp/provider/yql_kikimr_exec.cpp#L1342) from the TKiAlterTable node in `yql_kikimr_exec.cpp`. So in my opinion to check the correctness of the parsing pipeline we need to build the dedicated TKi node from the input.

vitstn (approved):

PR #10831, opened 2024-10-24
Views: if exists / if not exists for DDL

### LINKED ISSUE #10818: views: IF EXISTS / IF NOT EXISTS to help with scripting view dropping / creation
KIKIMR-22087

jepett0:
### Description

```sql
CREATE VIEW IF NOT EXISTS ...;
DROP VIEW IF EXISTS ...;
```

Introduce `IF ...` modifiers for the VIEW DDL to not produce errors if the object already exists / does not exist.

### Context

We would like VIEW users to be able to batch create views with scripts (which should not produce errors during the execution) like the following one:

```sql
CREATE OR REPLACE VIEW first_view ...;
CREATE OR REPLACE VIEW second_view ...;
CREATE OR REPLACE VIEW third_view ...;
...
```

However, CREATE OR REPLACE VIEW is a bit harder to implement (it needs to drop the view and recreate it in one transaction + update the server side query cache ☆) than the simple alternative:
```sql
DROP VIEW IF EXISTS first_view;
CREATE VIEW first_view ...;

DROP VIEW IF EXISTS second_view;
CREATE VIEW second_view ...;
...
```

<details>
<summary>☆</summary>

Ideally the server side query cache entry for the queries referencing the view-to-be-dropped should be deleted in the same transaction as the view. However, this feature is not easy to implement (see this [PR](https://github.com/ydb-platform/ydb/pull/1960) for more info). I suppose the query cache update lag would be less of an issue for the scripts using `DROP VIEW` + `CREATE VIEW` rather than a single `CREATE OR REPLACE VIEW`, because the later is expected to be atomic.
</details>

### DIFF
--- a/ydb/core/kqp/ut/view/view_ut.cpp
+++ b/ydb/core/kqp/ut/view/view_ut.cpp
@@ -344,7 +344,38 @@ Y_UNIT_TEST_SUITE(TCreateAndDropViewTest) {
         {
             const auto creationResult = session.ExecuteSchemeQuery(creationQuery).GetValueSync();
             UNIT_ASSERT(!creationResult.IsSuccess());
-            UNIT_ASSERT(creationResult.GetIssues().ToString().Contains("error: path exist, request accepts it"));
+            UNIT_ASSERT_STRING_CONTAINS(creationResult.GetIssues().ToString(), "error: path exist, request accepts it");
+        }
+    }
+
+    Y_UNIT_TEST(CreateViewOccupiedName) {
+        TKikimrRunner kikimr(TKikimrSettings().SetWithSampleTables(false));
+        EnableViewsFeatureFlag(kikimr);
+        auto session = kikimr.GetQueryClient().GetSession().ExtractValueSync().GetSession();
+
+        constexpr const char* path = "table";
+
+        const TString createTable = std::format(R"(
+                CREATE TABLE {} (key Int32, value Utf8, PRIMARY KEY (key));
+            )", path
+        );
+        ExecuteQuery(session, createTable);
+
+        auto checkError = [&session](const TString& query, const TString& expectedError) {
+            const auto result = session.ExecuteQuery(query, NQuery::TTxControl::NoTx()).ExtractValueSync();
+            UNIT_ASSERT(!result.IsSuccess());
+            UNIT_ASSERT_STRING_CONTAINS(result.GetIssues().ToString(), expectedError);
+        };
+
+        const TString queryTemplate = std::format(R"(
+                CREATE VIEW {{}}{} WITH (security_invoker = true) AS SELECT 1;
+            )", path
+        );
+        const TString expectedError = std::format("path: '/Root/{}', error: unexpected path type", path);
+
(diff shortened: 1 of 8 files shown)


ydb/library/yql/sql/v1/format/sql_format_ut.h
```diff
@@ -1539,8 +1539,8 @@ Y_UNIT_TEST(ObfuscatePragma) {
 
 Y_UNIT_TEST(CreateView) {
     TCases cases = {
-        {"creAte vIEw TheView wiTh (security_invoker = trUE) As SELect 1",
-            "CREATE VIEW TheView WITH (security_invoker = TRUE) AS\nSELECT\n\t1;\n"},
+        {"creAte vIEw iF nOt EXistS TheView wiTh (security_invoker = trUE) As SELect 1",
```
vitstn, line 1542: exists часть опциональная, и должен быть тест без нее

ydb/library/yql/sql/v1/format/sql_format_ut.h
```diff
@@ -1549,8 +1549,8 @@ Y_UNIT_TEST(CreateView) {
 
 Y_UNIT_TEST(DropView) {
     TCases cases = {
-        {"dRop viEW theVIEW",
-            "DROP VIEW theVIEW;\n"},
+        {"dRop viEW iF EXistS theVIEW",
```
vitstn, line 1552: тоже самое

ydb/core/kqp/ut/view/view_ut.cpp
```diff
@@ -428,6 +447,32 @@ Y_UNIT_TEST_SUITE(TCreateAndDropViewTest) {
...
+        ExecuteQuery(session, creationQuery);
+
+        const TString dropQuery = std::format(R"(
+                DROP VIEW IF EXISTS `{}`;
+            )",
+            path
+        );
+        ExecuteQuery(session, dropQuery);
+        // an attempt to drop an already deleted view does not produce an error
+        ExecuteQuery(session, dropQuery);
```
UgnineSirdis, line 473: We also need more tests to the following cases:
- DROP VIEW query must fail if there is no such view
- DROP VIEW query must fail if we are attempting to delete view, but it is not a view (topic, table etc)
- DROP VIEW IF EXISTS query must fail if we are attempting to delete view, but it is not a view (topic, table etc)

All these cases should work also for ExecuteQuery calls (not only for ExecuteSchemeQuery) - they are newer API

Also the same cases, but for CREATE type

UgnineSirdis (changes requested):

ydb/core/kqp/ut/view/view_ut.cpp
jepett0: Added all 3 `DROP VIEW` tests you mentioned.

As for the `CREATE VIEW`, the test that checks that the `CREATE VIEW` statement does in fact create a view has already existed, so I haven't added it. However, I have added the other two tests for the `CREATE VIEW` statement.

vitstn (approved):

UgnineSirdis (approved):

Issue #10340, opened 2024-10-11
Add clickbench performance tests on PostgreSQL syntax over YDB


SloNN:
We want to track PostgreSQL performance results over clickbench dataset while making performance testing.

Please consider to add the tests to pipeline.

PR #10861, opened 2024-10-24
Add clickbench performance tests on PostgreSQL syntax over YDB

iddqdex:
https://github.com/ydb-platform/ydb/issues/10340

### DIFF
--- a/ydb/library/workload/clickbench/clickbench.cpp
+++ b/ydb/library/workload/clickbench/clickbench.cpp
@@ -1,5 +1,6 @@
 #include "clickbench.h"
 #include "data_generator.h"
+#include <ydb/library/workload/benchmark_base/workload.h_serialized.h>
 
 #include <library/cpp/resource/resource.h>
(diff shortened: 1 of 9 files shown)


ydb/library/workload/clickbench/clickbench.cpp
```diff
@@ -164,6 +173,8 @@ void TClickbenchWorkloadParams::ConfigureOpts(NLastGetopt::TOpts& opts, const EC
             .StoreResult(&ExternalQueries);
         opts.AddLongOption('c', "check-canonical", "Use deterministic queries and check results with canonical ones.")
             .NoArgument().StoreTrue(&CheckCanonicalFlag);
+        opts.AddLongOption( "pg-syntax", "Use postgree syntax in queries instead YQL one.")
```
pnv1, line 176: We should probably use "--syntax pg" option with default value "yql". We already have such option in ydb sql command, would be consistent with it. Also would be easier to support new syntaxes, if they will appear

pnv1 (approved):

PR #10867, opened 2024-10-24
correct and speed up compaction

ivanmorozov333:
* Improvement
* Performance improvement
* Bugfix


fexolm (approved):

ydb/core/tx/columnshard/engines/changes/general_compaction.cpp
```diff
@@ -237,26 +237,35 @@ std::shared_ptr<TGeneralCompactColumnEngineChanges::IMemoryPredictor> TGeneralCo
 }
 
 ui64 TGeneralCompactColumnEngineChanges::TMemoryPredictorChunkedPolicy::AddPortion(const TPortionInfo& portionInfo) {
-    SumMemoryFix += portionInfo.GetRecordsCount() * (2 * sizeof(ui64) + sizeof(ui32) + sizeof(ui16));
+    SumMemoryFix += portionInfo.GetRecordsCount() * (2 * sizeof(ui64) + sizeof(ui32) + sizeof(ui16)) + portionInfo.GetTotalBlobBytes();
```
fexolm, line 240: Maybe we could use sizeof(SomeStruct) here? Just wondering...

PR #10930, opened 2024-10-25
Leader could be empty at the very beginning in partitions stats

### LINKED ISSUE #10846: KqpSystemView.PartitionStatsFollower is flaky under ASAN
An example of the failure:
- https://nebius.testmo.net/automation/runs/results/73095?group_by=automation_run_tests:status_id&group_id=3&test_id=1226553975

You can see in the context that it fails about 10% of the time with the same error:
```
Test crashed (return code: -6)
```

azevaykin:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Additional information

...


jepett0 (approved):
Looks good to me, but I'm not an expert in this code

ijon (approved):

PR #10983, opened 2024-10-28
Preparation for clang-18 migration

Nechda:
This PR is a preparation for the clang-18 migration.

### DIFF
--- a/ydb/core/blob_depot/data_uncertain.cpp
+++ b/ydb/core/blob_depot/data_uncertain.cpp
@@ -242,3 +242,15 @@ namespace NKikimr::NBlobDepot {
     }
 
 } // NKikimr::NBlobDepot
+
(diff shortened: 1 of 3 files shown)


abyss7:
Would like to see some comments about proposed changes to `enum class` and corresponding `Out` function

Nechda:
Moving the `enum class` out of the class scope has been required due to a new diagnostic that was introduced in clang-18.0.1 in the [commit](https://github.com/llvm/llvm-project/commit/29bd78b2f61e638f6c26bc417ae476754b91e985)

Therefore, in clang 16, we are unable to define the `void Out()` function for the hidden type `EKeyBlobState` directly within the class. Instead, I have moved `EKeyBlobState` to another structure and specialized `void Out() out` of the class' scope.

abyss7 (approved):

vitstn (approved):

rvu1024 (approved):

PR #10989, opened 2024-10-28
Library import 241028-1143

alexv-smirnov:


maximyurchuk:
Import with fix: https://github.com/ydb-platform/ydb/pull/11021

PR #11023, opened 2024-10-29
Support PG types in schemeshard

lex007in:
### Changelog entry 

* Add test for shard splitting by PG key
* Support PG types for PartitionKeySchema in PQ

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__table_stats_histogram.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__table_stats_histogram.cpp
@@ -89,6 +89,7 @@ TSerializedCellVec ChooseSplitKeyByHistogram(const NKikimrTableStats::THistogram
             splitKey[i] = keyMed.GetCells()[i];
         } else {
             // med == lo and med != hi, so we want to find a value that is > med and <= hi
+            // TODO: support this optimization for integer pg types
             if (IsIntegerType(columnType) && !keyMed.GetCells()[i].IsNull()) {
                 // For integer types we can add 1 to med
                 ui64 val = 0;
--- a/ydb/core/tx/schemeshard/schemeshard__table_stats_histogram.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__table_stats_histogram.cpp
@@ -6,7 +6,6 @@ namespace NKikimr {
 namespace NSchemeShard {
 
 static bool IsIntegerType(NScheme::TTypeInfo type) {
-    // TODO: support pg types
     switch (type.GetTypeId()) {
     case NScheme::NTypeIds::Bool:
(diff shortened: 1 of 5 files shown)


ydb/core/tx/schemeshard/schemeshard__table_stats_histogram.cpp
```diff
@@ -6,7 +6,6 @@ namespace NKikimr {
 namespace NSchemeShard {
 
 static bool IsIntegerType(NScheme::TTypeInfo type) {
-    // TODO: support pg types
```
lex007in, line 9: Generic not-only-integer realization is enough. Tested in SplitShardsWhithPgKey test.

azevaykin (dismissed):

ydb/core/tx/schemeshard/ut_split_merge/ut_split_merge.cpp
```diff
@@ -158,6 +158,84 @@ Y_UNIT_TEST_SUITE(TSchemeShardSplitBySizeTest) {
...
+                           {NLs::PartitionCount(1)});
+
+        TestAlterTable(runtime, ++txId, "/MyRoot", R"(
+                        Name: "Table"
+                        PartitionConfig {
+                            PartitioningPolicy {
+                                MinPartitionsCount: 100
+                                MaxPartitionsCount: 100
+                                SizeToSplit: 1
+                                FastSplitSettings {
```
ijon, line 203: Use of `FastSplitSettings` was removed from split-merge code quite some time ago. So it is meaningless to specify them in tests.

There are still some not updated tests (and they surprisingly continue to work ok), but new tests should not employ `FastSplitSettings`.

ydb/core/tx/schemeshard/schemeshard__table_stats_histogram.cpp
ijon: Its enough to work but not enough to work well.
Below in the same file `IsIntegerType()` is used to choose between fast and slow branches for split prefix search. And failing to support integer pg types as integer forces `ChooseSplitKeyByHistogram()` to choose the slow pass.

ijon (changes requested):

ydb/core/tx/schemeshard/schemeshard__table_stats_histogram.cpp
lex007in: Yes, support for this optimization requires the ability to increment a value of integer pg type which is not  currently supported. I'll do it in separate issue: https://github.com/ydb-platform/ydb/issues/11168

ijon (approved):

PR #11059, opened 2024-10-29
Fix excessive read latency during and after shard splits

### LINKED ISSUE #11036: DataShard: read actors retry TEvRead to shards long after they finished splitting
Looks like DataShard currently replies with a very generic error:

```
    if (!IsStateNewReadAllowed()) {
        replyWithError(
            Ydb::StatusIds::OVERLOADED,
            TStringBuilder() << "Shard " << TabletID() << " is splitting/merging"
                << " (node# " << SelfId().NodeId() << " state# " << DatashardStateName(State) << ")");
        return;
    }
```

Which causes ReadActor (LookupActor?) to keep retrying with an expotential backoff without resolving new partitioning. I've seen this causing 3+ seconds latency (due to expotential backoff) until it finally requests the new shard.

snaury:
### Changelog entry 

Fixed excessive read latency during and after some shard splits.

### Changelog category 

* Bugfix 

### Additional information

It was observed that reads sometimes take seconds during frequent shard splits. Turns out shards replied with an `OVERLOADED` status even after split has already finished, which caused KQP to retry reads repeatedly with an exponential backoff, until eventually a guard condition (after multiple seconds) would cause read actor to finally re-resolve the table again. A correct `NOT_FOUND` status (which indicates the table no longer exists) fixes this problem.

Fixes #11036.

### DIFF
--- a/ydb/core/tx/datashard/datashard__read_iterator.cpp
+++ b/ydb/core/tx/datashard/datashard__read_iterator.cpp
@@ -2657,6 +2657,16 @@ void TDataShard::Handle(TEvDataShard::TEvRead::TPtr& ev, const TActorContext& ct
         return;
     }
 
+    if (State == TShardState::PreOffline ||
+        State == TShardState::Offline)
+    {
+        replyWithError(
+            Ydb::StatusIds::NOT_FOUND,
+            TStringBuilder() << "Shard " << TabletID() << " finished splitting/merging"
+                << " (node# " << SelfId().NodeId() << " state# " << DatashardStateName(State) << ")");
+        return;
+    }
+
     if (!IsStateNewReadAllowed()) {
         replyWithError(
             Ydb::StatusIds::OVERLOADED,
--- a/ydb/core/tx/datashard/datashard_ut_read_iterator.cpp
+++ b/ydb/core/tx/datashard/datashard_ut_read_iterator.cpp
@@ -3934,7 +3934,7 @@ Y_UNIT_TEST_SUITE(DataShardReadIterator) {
     Y_UNIT_TEST(HandleMvccGoneInContinue) {
         // TODO
     }
-};
+}
 
 Y_UNIT_TEST_SUITE(DataShardReadIteratorSysTables) {
     Y_UNIT_TEST(ShouldRead) {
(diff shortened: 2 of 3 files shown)


ydb/core/tx/datashard/datashard_ut_read_iterator.cpp
```diff
@@ -4719,4 +4719,70 @@ Y_UNIT_TEST_SUITE(DataShardReadIteratorConsistency) {
...
+        runtime.SimulateSleep(TDuration::MilliSeconds(1));
+
+        auto readStartTs = runtime.GetCurrentTime();
+        blockedReads.Unblock();
+        blockedReads.Stop();
+        auto readResponse = runtime.WaitFuture(std::move(readFuture));
+        UNIT_ASSERT_VALUES_EQUAL(readResponse.operation().status(), Ydb::StatusIds::SUCCESS);
+        auto readLatency = runtime.GetCurrentTime() - readStartTs;
+        Cerr << "... read latency was " << readLatency << Endl;
+        UNIT_ASSERT_C(readLatency < TDuration::MilliSeconds(100),
```
azevaykin, line 4782: Is that all right under sanitizer?

ydb/core/tx/datashard/datashard__read_iterator.cpp
```diff
@@ -2657,6 +2657,16 @@ void TDataShard::Handle(TEvDataShard::TEvRead::TPtr& ev, const TActorContext& ct
         return;
     }
 
+    if (State == TShardState::PreOffline ||
+        State == TShardState::Offline)
+    {
+        replyWithError(
+            Ydb::StatusIds::NOT_FOUND,
```
azevaykin, line 2664: Don't you want to catch is the test  `EvReadResult` and check status 'NOT_FOUND' ?

ydb/core/tx/datashard/datashard_ut_read_iterator.cpp
snaury: This is simulated time, it shouldn't change regardless of how slow this runs.

ydb/core/tx/datashard/datashard__read_iterator.cpp
snaury: Why? I want to observe and test the effect (high latency), not some implementation detail.

azevaykin (approved):

PR #11061, opened 2024-10-29
24-3: Fix excessive read latency during and after shard splits

### LINKED ISSUE #11036: DataShard: read actors retry TEvRead to shards long after they finished splitting

snaury:
### Changelog entry 

Fixed excessive read latency during and after some shard splits.

### Changelog category 

* Bugfix 

### Additional information

It was observed that reads sometimes take seconds during frequent shard splits. Turns out shards replied with an `OVERLOADED` status even after split has already finished, which caused KQP to retry reads repeatedly with an exponential backoff, until eventually a guard condition (after multiple seconds) would cause read actor to finally re-resolve the table again. A correct `NOT_FOUND` status (which indicates the table no longer exists) fixes this problem.

Fixes #11036.

### DIFF
--- a/ydb/core/tx/datashard/datashard__read_iterator.cpp
+++ b/ydb/core/tx/datashard/datashard__read_iterator.cpp
@@ -2645,6 +2645,16 @@ void TDataShard::Handle(TEvDataShard::TEvRead::TPtr& ev, const TActorContext& ct
         return;
     }
 
+    if (State == TShardState::PreOffline ||
+        State == TShardState::Offline)
+    {
+        replyWithError(
+            Ydb::StatusIds::NOT_FOUND,
+            TStringBuilder() << "Shard " << TabletID() << " finished splitting/merging"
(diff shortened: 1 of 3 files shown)


fomichev3000 (approved):

snaury:
По зафейлившимся тестам, посмотрел более ранние PR в 24-3, там те же тесты фейлятся.

PR #11393, opened 2024-11-07
Try to lsof, when fail to lock file

eivanov89:
### Changelog entry 

Try to lsof, when fail to lock file

### Changelog category 

* Improvement


the-ancient-1 (approved):
LGTM

Issue #11478, opened 2024-11-11
Memory leak during periodic task run on the query service session


dcherednik:
`==1727294==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 128 byte(s) in 1 object(s) allocated from:
warning: address range table at offset 0x0 has a premature terminator entry at offset 0x10
warning: address range table at offset 0x30 has a premature terminator entry at offset 0x40
warning: address range table at offset 0x60 has a premature terminator entry at offset 0x70
warning: address range table at offset 0x90 has a premature terminator entry at offset 0xa0
warning: address range table at offset 0xc0 has a premature terminator entry at offset 0xd0
warning: address range table at offset 0xa20 has a premature terminator entry at offset 0xa30
warning: address range table at offset 0xa50 has a premature terminator entry at offset 0xa60
warning: address range table at offset 0xa80 has a premature terminator entry at offset 0xa90
warning: address range table at offset 0xab0 has a premature terminator entry at offset 0xac0
warning: address range table at offset 0xae0 has a premature terminator entry at offset 0xaf0
warning: address range table at offset 0xb10 has a premature terminator entry at offset 0xb20
warning: address range table at offset 0xb40 has a premature terminator entry at offset 0xb50
warning: address range table at offset 0xb70 has a premature terminator entry at offset 0xb80
warning: address range table at offset 0xba0 has a premature terminator entry at offset 0xbb0
warning: address range table at offset 0xbd0 has a premature terminator entry at offset 0xbe0
warning: address range table at offset 0xc00 has a premature terminator entry at offset 0xc10
warning: address range table at offset 0xc30 has a premature terminator entry at offset 0xc40
warning: address range table at offset 0xc60 has a premature terminator entry at offset 0xc70
warning: address range table at offset 0xc90 has a premature terminator entry at offset 0xca0
warning: address range table at offset 0xcc0 has a premature terminator entry at offset 0xcd0
warning: address range table at offset 0xcf0 has a premature terminator entry at offset 0xd00
warning: address range table at offset 0xd20 has a premature terminator entry at offset 0xd30
warning: address range table at offset 0xd50 has a premature terminator entry at offset 0xd60
warning: address range table at offset 0x1300 has a premature terminator entry at offset 0x1310
warning: address range table at offset 0x1360 has a premature terminator entry at offset 0x1370
warning: address range table at offset 0x1390 has a premature terminator entry at offset 0x13a0
[...]

PR #11489, opened 2024-11-11
Fix memory leak in case of launch periodic keep-alive task for query service session

dcherednik:


dcherednik:
https://github.com/ydb-platform/ydb/issues/11478

Gazizonoki (approved):

pnv1 (approved):

Issue #11520, opened 2024-11-12
Migrate to STYLE_CPP from CPP_STYLE_TEST_16


andrei-levitskii:
We've recently introduced a new CPP linting macro - `STYLE_CPP`. It's targeted to become a replacement for the old `CPP_STYLE_TEST_16`.

Unlike `CPP_STYLE_TEST_16` `STYLE_CPP` collects its targets from `SRCS` macro. It's recommended to list header files located outside of the module directory in `SRCS` macro. If you happen not to list header files in `SRCS` you can use [`HEADERS`](https://github.com/ydb-platform/ydb/blob/main/build/ymake.core.conf#L5857) macro

If you need to lint multiple nested directories consider using autoincludes. Consult YaTeam for more.

PR #11549, opened 2024-11-13
Add cpp style default configs map

andrei-levitskii:
### Changelog entry 

...

### Changelog category 

### Additional information

https://github.com/ydb-platform/ydb/issues/11520


maximyurchuk (changes requested):
This file is exported from arcadia, you have to add that file into sync instead of changing it in current repo

PR #11721, opened 2024-11-18
show down and frozen nodes in ui even when they are dead

vporyadke:
[KIKIMR-25191](http://st/KIKIMR-25191)

### Changelog entry 

show down and frozen nodes in ui even when they are dead

### Changelog category 

### Additional information

Useful to be able to remove the Down/Freeze flag

### DIFF
--- a/ydb/core/mind/hive/monitoring.cpp
+++ b/ydb/core/mind/hive/monitoring.cpp
@@ -2427,7 +2427,10 @@ public:
                 TNodeInfo& node = *nodeInfo;
                 TNodeId id = node.Id;
 
-                if (!node.IsAlive() && TInstant::MilliSeconds(node.Statistics.GetLastAliveTimestamp()) < aliveLine) {
+                if (!node.IsAlive()
+                    && TInstant::MilliSeconds(node.Statistics.GetLastAliveTimestamp()) < aliveLine
+                    && !node.Down
+                    && !node.Freeze) {
                     continue;
                 }
 


ydb/core/mind/hive/monitoring.cpp
```diff
@@ -2427,7 +2427,10 @@ class TTxMonEvent_LandingData : public TTransactionBase<THive> {
                 TNodeInfo& node = *nodeInfo;
                 TNodeId id = node.Id;
 
-                if (!node.IsAlive() && TInstant::MilliSeconds(node.Statistics.GetLastAliveTimestamp()) < aliveLine) {
+                if (!node.IsAlive()
```
adameat, line 2430: it means that node will disappear on changing any of these flags. 
it would look strange... maybe consider removing this check at all?

ydb/core/mind/hive/monitoring.cpp
```diff
@@ -2427,7 +2427,10 @@ class TTxMonEvent_LandingData : public TTransactionBase<THive> {
...
+                    && TInstant::MilliSeconds(node.Statistics.GetLastAliveTimestamp()) < aliveLine
+                    && !node.Down
+                    && !node.Freeze) {
```
CyberROFL, line 2433:
```suggestion
                    && !node.Freeze)
                {
```
Иначе сливается. Но можно (лучше) отдельно поправить.

CyberROFL (approved):

PR #11744, opened 2024-11-19
Fix "Refs" field name.

Tony-Romanov:
The actual name of this field is "Ref". 

https://github.com/ydb-platform/ydb/blob/main/ydb/library/yql/providers/yt/gateway/native/yql_yt_native.cpp#L2981


nepal:
Committed to upstream at r15384683

PR #11782, opened 2024-11-20
Fix tpch pg queries, make they as yql ones

iddqdex:
YQL-19298


igormunkin (approved):
@iddqdex, thanks for the patch! Tests are fine now, hence LGTM.

PR #11841, opened 2024-11-21
Fix UI error "Failed to resolve database"

### LINKED ISSUE #11153: error resolving database nodes for NBS tenant

StekPerepolnen:
### Changelog entry 

In 24-3, support for redirecting requests to dynodes was added. The fix adjusts the scenario where a redirect is not possible: instead of throwing a "Failed to resolve database" error, the request is processed locally on the node that received it.

### Changelog category 

* Improvement

### DIFF
--- a/ydb/core/viewer/json_pipe_req.cpp
+++ b/ydb/core/viewer/json_pipe_req.cpp
@@ -748,6 +748,12 @@ void TViewerPipeClient::RequestDone(ui32 requests) {
     }
 }
 
+void TViewerPipeClient::ResetAndBootstrap() {
+    Requests = 0;
+    DelayedRequests.clear();
+    Bootstrap();
+}
+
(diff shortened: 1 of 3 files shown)


ydb/core/viewer/json_pipe_req.cpp
```diff
@@ -748,6 +748,12 @@ void TViewerPipeClient::RequestDone(ui32 requests) {
     }
 }
 
+void TViewerPipeClient::ResetAndBootstrap() {
```
adameat, line 751: why?!

StekPerepolnen: Simply calling Bootstrap is not an option, as it resets the Request counter. This manifests in some requests waiting for a timeout (10 seconds) before completing.

Implementing a clean solution for accurate Request counting quickly is challenging. Resetting the counter is the easiest approach.

adameat (approved):

PR #11935, opened 2024-11-25
add kafka port to quick start

rekby:
merge after https://github.com/ydb-platform/ydb/pull/11892

### Changelog entry 

...

### Changelog category 

### Additional information

...


ydb/docs/en/core/quickstart.md
```diff
@@ -86,9 +86,11 @@ Normally, {{ ydb-short-name }} stores data on multiple SSD/NVMe or HDD raw disk
       ```bash
       docker run -d --rm --name ydb-local -h localhost \
         --platform linux/amd64 \
-        -p 2135:2135 -p 2136:2136 -p 8765:8765 \
+        -p 2135:2135 -p 2136:2136 -p 8765:8765 -p 9092:9092 \
         -v $(pwd)/ydb_certs:/ydb_certs -v $(pwd)/ydb_data:/ydb_data \
         -e GRPC_TLS_PORT=2135 -e GRPC_PORT=2136 -e MON_PORT=8765 \
+        -e YDB_KAFKA_API_ENABLED=1 \
```
blinkov, line 92: Why it's `=1` instead of the port number like in the neighboring settings?
And to avoid doing it twice we could add the postgresql port, too.

rekby: You are right - it will be consistent with other params and better understand numbers if enable kafka with dedicated port, I can change it.

What about: `YDB_KAFKA_PROXY_PORT=9092` ?

postgres has different flags for enable and set custom port:
`YDB_EXPERIMENTAL_PG` and `PGWIRE_LISTENING_PORT`

I think it will be difficult to read a lot of params for customization in quick start.

May be create dedicated page for describe all env variable for container?

rekby: @blinkov `YDB_KAFKA_API_ENABLED=1` replaced by `YDB_KAFKA_PROXY_PORT=9092`

fomichev3000 (approved):

PR #11969, opened 2024-11-25
Fix: vector index description wasn't persisted

MBkkt:


ydb/core/tx/schemeshard/schemeshard_info_types.h
```diff
@@ -2371,11 +2371,16 @@ struct TTableIndexInfo : public TSimpleRefCount<TTableIndexInfo> {
     using EType = NKikimrSchemeOp::EIndexType;
     using EState = NKikimrSchemeOp::EIndexState;
 
-    TTableIndexInfo(ui64 version, EType type, EState state)
+    TTableIndexInfo(ui64 version, EType type, EState state, std::string_view description)
```
ijon, line 2374: Is there a reason to use `std::string_view` here?

ydb/core/tx/schemeshard/schemeshard_info_types.h
```diff
@@ -2391,8 +2396,20 @@ struct TTableIndexInfo : public TSimpleRefCount<TTableIndexInfo> {
         return result;
     }
 
+    TString DescriptionToStr() const {
+        return std::visit([](const auto& v) {
+            if constexpr (requires { v.SerializeAsString(); }) {
```
ijon, line 2401: Can you explain why this check is necessary? I guess, we expect all variants of `SpecializedIndexDescription` to be of protobuf message type.

ydb/core/tx/schemeshard/schemeshard_info_types.h
```diff
@@ -2391,8 +2396,20 @@ struct TTableIndexInfo : public TSimpleRefCount<TTableIndexInfo> {
         return result;
     }
 
+    TString DescriptionToStr() const {
```
ijon, line 2399: `ToStr` has a misleading connotation. According to our codebase it doesn't have a meaning of "encode value to a sequence of bytes" but rather "make something printable to be viewed by a person".
I would recommend using something more accurate: `SerializedDescription` or something.

ydb/core/tx/schemeshard/schemeshard_info_types.h
MBkkt: In my head it's commonly best way to pass something like const string reference.
Also `ParseFromString` uses `string_view` (abseil but it's kind of same):
```cpp
bool ParseFromString(y_absl::string_view data);
```

Do you have some other preference? What and why?

ydb/core/tx/schemeshard/schemeshard_info_types.h
MBkkt: We expect monostate for global secondary indexes, but I agree that is maybe better to check this instead

ydb/core/tx/schemeshard/schemeshard_info_types.h
MBkkt: Ok, changed to `SerializeDescription`, not `SerializedDescription` because I want it to be clear that it's not just getter

ydb/core/tx/schemeshard/schemeshard_schema.h
```diff
@@ -1043,9 +1044,10 @@ struct Schema : NIceDb::Schema {
         struct AlterVersion :   Column<3, NScheme::NTypeIds::Uint64> {};
         struct IndexType :      Column<4, NScheme::NTypeIds::Uint32> { using Type = NKikimrSchemeOp::EIndexType; static constexpr Type Default = NKikimrSchemeOp::EIndexTypeInvalid; };
         struct State :          Column<5, NScheme::NTypeIds::Uint32> { using Type = NKikimrSchemeOp::EIndexState; static constexpr Type Default = NKikimrSchemeOp::EIndexStateInvalid; };
+        struct Description :    Column<6, NScheme::NTypeIds::String> {};
```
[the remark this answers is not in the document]
MBkkt: Ok

ydb/core/tx/schemeshard/schemeshard__operation_consistent_copy_tables.cpp
```diff
@@ -35,6 +35,7 @@ NKikimrSchemeOp::TModifyScheme CreateIndexTask(NKikimr::NSchemeShard::TTableInde
     operation->SetName(dst.LeafName());
 
     operation->SetType(indexInfo->Type);
+    Y_ENSURE(indexInfo->Type != NKikimrSchemeOp::EIndexType::EIndexTypeGlobalVectorKmeansTree);
```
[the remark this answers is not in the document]
MBkkt: There is a problem, I don't like the idea of ​​schemeshard aborting if something is written incorrectly and we pass the wrong index type in some schemeshard operation, I prefer abort this operation instead.

Do you have any suggestions on what I can replace this "ensure" code with?

ydb/core/tx/schemeshard/schemeshard_info_types.h
ijon: `const TString&`.
We don't store original argument value in the object and don't construct TTableIndexInfo from anything but TString value, so string_view doesn't bring any benefits here.

ydb/core/tx/schemeshard/schemeshard__operation_consistent_copy_tables.cpp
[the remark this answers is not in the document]
MBkkt: Fixed

ijon (approved):

PR #12067, opened 2024-11-27
Refactor tracing configurator to be available from actor system threads

UgnineSirdis:
### Changelog entry 

There are situations when we need to generate a new trace id (or throttle existing one) not only from grpc proxy actor, but in several other places (long living topic session). For this case I refactored tracing configurator so that it can be available from every actor system thread.

### Changelog category 

* Improvement

### Additional information

...

### DIFF
--- a/ydb/core/base/wilson_tracing_control.cpp
+++ b/ydb/core/base/wilson_tracing_control.cpp
@@ -22,7 +22,14 @@ public:
     {}
 
     TSamplingThrottlingControl* GetTracingControlPtr() {
+        if (Y_UNLIKELY(!Control)) {
+            Control = CreateNewTracingControl();
+        }
         return Control.Get();
+    }
+
+    void ResetTracingControl() {
+        Control = nullptr;
     }
 
 private:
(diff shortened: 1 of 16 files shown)


ydb/core/base/wilson_tracing_control.cpp
```diff
@@ -0,0 +1,35 @@
...
+    Y_ASSERT(HasAppData()); // In general we must call this from actor thread
+    if (!HasAppData()) {
+        return nullptr;
+    }
+
+    return AppData()->TracingConfigurator->GetControl();
+}
+
+TSamplingThrottlingControl* GetTracingControlTls() {
+    static thread_local TIntrusivePtr<TSamplingThrottlingControl> Control = CreateNewTracingControl();
```
snaury, line 23: Поддержка сложных thread_local'ов (с деструкторами) очень плохая под разными операционками и компиляторами, я боюсь в таком виде оно может перестать собираться.

snaury (dismissed):

snaury (approved):

PR #12092, opened 2024-11-28
tests of ttl utility functions on SS

swalrus1:


ijon (commented):
I would suggest incorporating tier index into the error messages (for example: "tier 1: missing ApplyAfterSeconds").
The user gives a sequence of tiers, and error message without accurate addressing wouldn't be that helpful (even if sequence dize is supposed to be small).

ijon (dismissed):

ijon (approved):

PR #12135, opened 2024-11-29
YQ-3871 Restart session on query stopping (if query read historical data)

kardymonds:
### Changelog entry 

...

### Changelog category 

### Additional information

...


ydb/core/fq/libs/row_dispatcher/topic_session.cpp
```diff
@@ -762,6 +765,41 @@ void TTopicSession::Handle(NFq::TEvRowDispatcher::TEvStopSession::TPtr& ev) {
...
+    TMaybe<ui64> minMessageOffset;
+    for (auto& [readActorId, clientPtr] : Clients) {
+        if (info.ReadActorId == readActorId || !clientPtr->NextMessageOffset) {
+            continue;
+        }
+        if (!minMessageOffset) {
+            minMessageOffset = clientPtr->NextMessageOffset;
+            continue;
+        }
+        minMessageOffset = std::min(minMessageOffset, clientPtr->NextMessageOffset);
```
yumkam, line 783: nit: Given the way `TMaybe<>::operator<` works, it is possible to just use `std::min` here, without checks for `!clientPtr->...` and `!minMessage...`. On other hand, it would kinda obfuscate logic.
Yet another alternative -- return early once we've found `clientPtr->NMO <= info.NMO` instead of looking for min (... but there will be less info to be logged)

ydb/tests/fq/pq_async_io/mock_pq_gateway.h
```diff
@@ -0,0 +1,42 @@
...
+#include <ydb/library/yql/dq/actors/compute/dq_compute_actor_async_io.h>
+#include <ydb/library/yql/dq/actors/protos/dq_events.pb.h>
+#include <yql/essentials/minikql/mkql_alloc.h>
+
+#include <ydb/public/sdk/cpp/client/ydb_topic/topic.h>
+#include <ydb/core/testlib/basics/runtime.h>
+
+#include <library/cpp/testing/unittest/registar.h>
+
+#include <ydb/library/yql/providers/pq/gateway/dummy/yql_pq_blocking_queue.h>
```
yumkam, line 17: Won't this require some `PEERDIR` update?

ydb/core/fq/libs/row_dispatcher/topic_session.cpp
kardymonds: Тогда как есть оставлю)

yumkam (dismissed):

yumkam (approved):

PR #12208, opened 2024-12-02
Library import 241202-1134

alexv-smirnov:


nepal (changes requested):
obsolete, do not commit

PR #12216, opened 2024-12-02
`--upload-batch-bytes` when it is set to maximum returns `Too much data` in `ydb tools restore`

pixcc:
### Changelog entry 

Fixed a bug where `ydb tools restore` was failing with `Too much data` if `--upload-batch-bytes` option value was set exactly to it's maximum possible value (16MiB)

### Changelog category 

* Bugfix

### DIFF
--- a/ydb/apps/ydb/CHANGELOG.md
+++ b/ydb/apps/ydb/CHANGELOG.md
@@ -1,6 +1,6 @@
 * Fixed a bug in TPC-H tables schema where the `partsupp` table had incorrect list of key columns
 * Enhanced parallelism of data restoring in `ydb tools restore`
-* Fixed a bug that `--upload-batch-bytes` when it is set to maximum returns `Too much data` in `ydb tools restore`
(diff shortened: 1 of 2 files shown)


CyberROFL (dismissed):

ydb/apps/ydb/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 * Fixed a bug in TPC-H tables schema where the `partsupp` table had incorrect list of key columns
-
 * Enhanced parallelism of data restoring in `ydb tools restore`
+* Fixed a bug that `--upload-batch-bytes` when it is set to maximum returns `Too much data` in `ydb tools restore`
```
pnv1, line 3: Fixed a bug where `ydb tools restore` could fail with `Too much data` when a batch size was equal to a value in `--upload-batch-bytes` option

pnv1 (changes requested):

pnv1 (approved):

PR #12287, opened 2024-12-04
Audit log: No permission to connect to the database

azevaykin:
### Changelog entry 

When a user tries to connect without the necessary permissions, this error is returned:

```
Status: UNAUTHORIZED
Issues: 
<<main>: Error: User has no permission to perform query on this database, database: /Domain 0/testdb, user: bigpig1-no access@ldap, from ip: ipv4:10.131.0.31:34736, code: 200000
<main>: Error: Endpoint list is empty for database /Domain0/testdb, cluster endpoint bigpig1:2135.
```

In this case, nothing is written to the console log.

What has been fixed:
1. The user does not need to see such a detailed error, it is enough to say “Access Denied”
2. A detailed message should be recorded in the audit log

The result console message is:
```
Status: UNAUTHORIZED
Issues:
<main>: Error: No permission to connect to the databased, code: 200000
<main>: Error: Endpoint list is empty for database /local, cluster endpoint localhost:19140.
```

### Changelog category 

* Bugfix 

### Additional information

...

### DIFF
--- a/ydb/core/grpc_services/grpc_request_check_actor.h
+++ b/ydb/core/grpc_services/grpc_request_check_actor.h
@@ -170,9 +170,10 @@ public:
         }
 
         {
-            auto [error, issue] = CheckConnectRight();
+            bool error = CheckConnectRight();
             if (error) {
-                ReplyUnauthorizedAndDie(*issue);
+                AuditLogAccessDenied(GrpcRequestBaseCtx_, CheckedDatabaseName_, TBase::GetUserSID(), TBase::GetSanitizedToken());
+                ReplyUnauthorizedAndDie();
                 return;
             }
         }
@@ -430,8 +431,8 @@ private:
     }
 
 private:
-    void ReplyUnauthorizedAndDie(const NYql::TIssue& issue) {
-        GrpcRequestBaseCtx_->RaiseIssue(issue);
+    void ReplyUnauthorizedAndDie() {
+        GrpcRequestBaseCtx_->RaiseIssue(MakeIssue(NKikimrIssues::TIssuesIds::ACCESS_DENIED));
         GrpcRequestBaseCtx_->ReplyWithYdbStatus(Ydb::StatusIds::UNAUTHORIZED);
         GrpcRequestBaseCtx_->FinishSpan();
         PassAway();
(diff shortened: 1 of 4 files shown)


ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -545,30 +546,22 @@ class TGrpcRequestCheckActor
...
-            return {false, std::nullopt};
+            return false;
         }
 
-        const TString error = TStringBuilder()
-            << "User has no permission to perform query on this database"
-            << ", database: " << CheckedDatabaseName_
-            << ", user: " << TBase::GetUserSID()
-            << ", from ip: " << GrpcRequestBaseCtx_->GetPeerName();
-        LOG_INFO(*TlsActivationContext, NKikimrServices::GRPC_PROXY_NO_CONNECT_ACCESS, "%s", error.c_str());
```
ijon, line 562: I think there should be a message about connect-db right violation in the common log.
Common log and audit log are different and are intended for a different audiences, so audit log message is not a substitute for a message in the common log.

ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -545,30 +546,22 @@ class TGrpcRequestCheckActor
...
         Counters_->IncDatabaseAccessDenyCounter();
 
         if (!AppData()->FeatureFlags.GetCheckDatabaseAccessPermission()) {
-            return {false, std::nullopt};
+            return false;
         }
 
-        LOG_INFO(*TlsActivationContext, NKikimrServices::GRPC_SERVER, "%s", error.c_str());
-        return {true, MakeIssue(NKikimrIssues::TIssuesIds::ACCESS_DENIED, error)};
+        return true;
```
ijon, line 564: There we deny user the ability to differentiate between two causes of ACCESS_DENIED:
1. inability to work with the database as a whole (no connect-db right)
2. no sufficient rights to work with the specific path in the database (all other schema rights)

I think that ACCESS_DENIED for no connect-db right should come with _some_ message specific to that cause. Empty ACCESS_DENIED is not descriptive enough.

ijon (changes requested):

ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -546,22 +546,31 @@ class TGrpcRequestCheckActor
...
 
         if (!AppData()->FeatureFlags.GetCheckDatabaseAccessPermission()) {
-            return false;
+            return {false, std::nullopt};
         }
 
-        return true;
+        const TString error = "No permission to connect to the database";
+        LOG_INFO_S(TlsActivationContext->AsActorContext(), NKikimrServices::GRPC_SERVER, 
+            "AUDIT: "
```
ijon, line 566: This is not an audit message, so `AUDIT:` mark should not be used.

`AUDIT:` mark is only present in common log messages that duplicate (some) audit log messages and that are left in the code for temporary backward compatibility with the audit log consumers that lagging behind in adopting separate audit log stream.
As a rule, `AUDIT:` should not be added to the newly introduced common log messages.

azevaykin: a1852a1af71dfd3b7b036f990283450aef94db6a

ijon (approved):

PR #12631, opened 2024-12-16
Enable automatic box management with distconf by default

alexvru:
### Changelog entry 

Enable automatic box management with distconf by default

### Changelog category 

### Additional information

With unspecified automatic\_box\_management field, it is enabled automatically.


adameat (approved):
I would suggest to add default to proto

PR #12900, opened 2024-12-24
Make authentication optional in kafka proxy of YDB Topics

a-serebryanskiy:
### Changelog entry 

Now if you specify `pqconfig.require_credentials_in_new_protocol: false` in cluster config, it will not require credentials to read/write/administrate topics through Kafka proxy. 

### Changelog category 

* Improvement

### Additional information

...


blinkov (commented):
no changes in documentation

ydb/core/kafka_proxy/kafka_connection.cpp
```diff
@@ -711,6 +712,15 @@ class TKafkaConnection: public TActorBootstrapped<TKafkaConnection>, public TNet
         }
     }
 
+    bool RequireAuthentication(EApiKey apiKey) {
```
niksaveliev, line 715: I think this check is necessary to ensure that the client has gone through the standard flow: API_VERSIONS -> SASL_HANDSHAKE -> SASL_AUTHENTICATE, rather than just starting to send us requests. SASL_AUTHENTICATE, in addition to authorization, sets up the database, cloud, folder, and other things in the context. You can see this in the TEvAuthResult event. Not sure everything will work properly if this part is skipped.

niksaveliev (approved):

ydb/core/kafka_proxy/kafka_connection.cpp
a-serebryanskiy: I've checked ydbd behaviour on local database in read/write/create topics scenarios and everything works well. May be there are some tests I can add to ensure correct behaviour?

PR #12928, opened 2024-12-24
Use sized operator delete for TOwnedCellVec

snaury:
### Changelog entry 

...

### Changelog category 

### Additional information

It was pointed out that `TOwnedCellVec` could use sized operator delete, since we use c++20 anyway, and this could be a bit more efficient. To make that work use a custom destroying delete operator.

### DIFF
--- a/ydb/core/scheme/scheme_tablecell.cpp
+++ b/ydb/core/scheme/scheme_tablecell.cpp
@@ -112,8 +112,11 @@ namespace {
 
 } // namespace
 
-void TOwnedCellVec::TData::operator delete(void* mem) noexcept {
-    ::free(mem);
+void TOwnedCellVec::TData::operator delete(TData* data, std::destroying_delete_t) noexcept {
+    void* ptr = data;
+    size_t size = data->DataSize;
+    data->~TData();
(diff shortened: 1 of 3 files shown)


ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -210,7 +209,7 @@ TOwnedCellVec::TInit TOwnedCellVec::AllocateFromSerialized(std::string_view data
         }
     }
 
-    void* mem = ::malloc(size);
+    void* mem = ::operator new(size);
     if (Y_UNLIKELY(!mem)) {
         throw std::bad_alloc();
     }
```
MBkkt, line 215:
```suggestion
```

ydb/core/scheme/scheme_tablecell.cpp
```diff
@@ -136,7 +138,7 @@ TOwnedCellVec::TInit TOwnedCellVec::Allocate(TOwnedCellVec::TCellVec cells) {
...
```
MBkkt, line 144:
```suggestion
```

MBkkt (approved):

PR #13174, opened 2025-01-04
Enable arena-based allocations for QueryService output rows

zinal:
### Changelog entry 

...

### Changelog category 

* Performance improvement
* Not for changelog (changelog entry is not required)

### Additional information

...

PR #13180, opened 2025-01-05
stream lookup: GetRangePartitioning() optimization

zinal:
### Changelog entry 

...

### Changelog category 

* Performance improvement

### Additional information

...


zinal:
According to measurements, this gives extra 20% of saved execution time (after applying fix from #13174) for large index-based range scan queries.

gridnevvvit (approved):

PR #13296, opened 2025-01-11
IGNIETFERRO-2114: correction of a debug check in the sorting predicate, a false positive of which leads to an abort

ubyte:
### Changelog entry 

Depending on the version of STL used, its configuration, and the distribution of sortable elements, the std::sort algorithm may compare some element with itself (for example, libcxx can do this).
In that case, the sorting predicate is obliged to return the value `false`, and not lead to an abortion.

### Changelog category 

* Bugfix 

### Additional information

* crash backtrace: https://nda.ya.ru/t/TQUdfPbp7B62M3

### DIFF
--- a/ydb/core/tx/scheme_board/cache.cpp
+++ b/ydb/core/tx/scheme_board/cache.cpp
@@ -1030,19 +1030,17 @@ class TSchemeCache: public TMonitorableActor<TSchemeCache> {
                 }
             }
 
+            const size_t emptyEndKeyPrefixCount = CountIf(partitioning, [](const auto& partition) {
+                Y_ABORT_UNLESS(partition.Range);
+                return !partition.Range->EndKeyPrefix;
+            });
+            Y_ABORT_UNLESS(emptyEndKeyPrefixCount <= 1);
             Sort(partitioning.begin(), partitioning.end(), [&schema](const auto& lhs, const auto& rhs) {
-                Y_ABORT_UNLESS(lhs.Range && rhs.Range);
-                Y_ABORT_UNLESS(lhs.Range->EndKeyPrefix || rhs.Range->EndKeyPrefix);
-
-                if (!lhs.Range->EndKeyPrefix) {
-                    return false;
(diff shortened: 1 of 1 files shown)


ydb/core/tx/scheme_board/cache.cpp
```diff
@@ -1030,20 +1030,18 @@ class TSchemeCache: public TMonitorableActor<TSchemeCache> {
...
             }
 
+            const size_t emptyEndKeyPrefixCount = CountIf(partitioning, [](const auto& partition) {
+                Y_ABORT_UNLESS(partition.Range);
+                return !partition.Range->EndKeyPrefix;
+            });
+            Y_ABORT_UNLESS(emptyEndKeyPrefixCount <= 1);
             Sort(partitioning.begin(), partitioning.end(), [&schema](const auto& lhs, const auto& rhs) {
-                Y_ABORT_UNLESS(lhs.Range && rhs.Range);
-                Y_ABORT_UNLESS(lhs.Range->EndKeyPrefix || rhs.Range->EndKeyPrefix);
```
ubyte, line 1035: This line in some conditions causes an abort, even if the array contains no more than one element that does not have the EndKeyPrefix.

ydb/core/tx/scheme_board/cache.cpp
```diff
@@ -1030,20 +1030,18 @@ class TSchemeCache: public TMonitorableActor<TSchemeCache> {
                 }
             }
 
+            const size_t emptyEndKeyPrefixCount = CountIf(partitioning, [](const auto& partition) {
+                Y_ABORT_UNLESS(partition.Range);
+                return !partition.Range->EndKeyPrefix;
+            });
+            Y_ABORT_UNLESS(emptyEndKeyPrefixCount <= 1);
```
ubyte, line 1037: It is also much faster to check the preconditions once outside the sorting routine, rather than rechecking them multiple times during pairwise element comparison.

CyberROFL (approved):

PR #13352, opened 2025-01-14
Inline up to 14 bytes in TCell

snaury:
### Changelog entry 

...

### Changelog category 

### Additional information

This will remove edge cases where TCell and TUnboxedValuePod behave differently for data between 9 and 14 bytes long.

### DIFF
--- a/ydb/core/scheme/scheme_tablecell.h
+++ b/ydb/core/scheme/scheme_tablecell.h
@@ -10,30 +10,75 @@
 #include <util/system/unaligned_mem.h>
 #include <util/memory/pool.h>
 
+#include <bit>
 #include <deque>
 #include <type_traits>
 
 namespace NKikimr {
 
-#pragma pack(push,4)
+static_assert(std::endian::native == std::endian::little, "TCell expects little endian architecture for data packing");
+
 // Represents one element in a tuple
 // Doesn't own the memory buffer that stores the actual value
-// Small values (<= 8 bytes) are stored inline
+// Small values (<= 14 bytes) are stored inline
 struct TCell {
     template<typename T>
     using TStdLayout = std::enable_if_t<std::is_standard_layout<T>::value, T>;
 
+public:
+    // 14 bytes ensures parity with TUnboxedValuePod
+    static constexpr size_t MaxInlineSize() { return 14; }
+    static constexpr bool CanInline(size_t size) { return size <= MaxInlineSize(); }
+
 private:
-    ui32 DataSize_ : 30;
-    ui32 IsInline_ : 1;
-    ui32 IsNull_   : 1;
-    union {
-        i64 IntVal;
-        const char* Ptr;
-        double DoubleVal;
-        float FloatVal;
-        char Bytes[8];
+    // NotInline (low bit):
+    //   0: IsInline() returns true
+    //   1: IsInline() returns false
+    // NotRefValue (high bit):
+    //   0: data is TRefValue
+    //   1: data is TInlineValue
+    // Null values are IsInline() with a nullptr data pointer and zero size
+    // TCell filled with zeroes is always a null value
+    // This bit arrangment allows methods to check a single flag
+    enum EKind : ui8 {
+        // Both bits 0: null value
+        KindNull = 0,
+        // Low bit 1: ref value
+        KindRefValue = 1,
+        // High bit 1: inline value
+        KindInlineValue = 2,
     };
+
+    struct TRefValue {
+        const char* Ptr;
+        ui64 Size : 62;
+        ui64 Kind : 2;
+    };
+
+    struct TInlineValue {
+        char Data[15];
+        ui8 Size : 6;
+        ui8 Kind : 2;
+    };
+
+private:
+    union {
+        char Raw[16];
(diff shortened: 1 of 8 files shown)


ydb/core/scheme/scheme_tablecell.h
```diff
@@ -10,79 +10,148 @@
...
 namespace NKikimr {
 
-#pragma pack(push,4)
+static_assert(std::endian::native == std::endian::little, "TCell expects little endian architecture for data packing");
+
 // Represents one element in a tuple
 // Doesn't own the memory buffer that stores the actual value
-// Small values (<= 8 bytes) are stored inline
+// Small values (<= 14 bytes) are stored inline
 struct TCell {
```
MBkkt, line 24: Do we need some alignas?
Before your changes it was 8?
But maybe it's ok if we never read with assumption that this data is aligned

ydb/core/scheme/scheme_tablecell.h
```diff
@@ -10,79 +10,148 @@
...
             switch (size) {
-                case 8: memcpy(&IntVal, ptr, 8); break;
-                case 7: memcpy(&IntVal, ptr, 7); break;
-                case 6: memcpy(&IntVal, ptr, 6); break;
-                case 5: memcpy(&IntVal, ptr, 5); break;
-                case 4: memcpy(&IntVal, ptr, 4); break;
-                case 3: memcpy(&IntVal, ptr, 3); break;
-                case 2: memcpy(&IntVal, ptr, 2); break;
-                case 1: memcpy(&IntVal, ptr, 1); break;
+                case 15: ::memcpy(Inline.Data, ptr, 15); break;
```
MBkkt, line 100: CanInline: size <= 14 ... switch(size) case 15:

Why does MaxInlineSize 14 but not 15?

ydb/core/scheme/scheme_tablecell.h
```diff
@@ -127,46 +196,39 @@ struct TCell {
...
+                case 8: ::memcpy(dst, Inline.Data, 8); break;
+                case 7: ::memcpy(dst, Inline.Data, 7); break;
+                case 6: ::memcpy(dst, Inline.Data, 6); break;
+                case 5: ::memcpy(dst, Inline.Data, 5); break;
+                case 4: ::memcpy(dst, Inline.Data, 4); break;
+                case 3: ::memcpy(dst, Inline.Data, 3); break;
+                case 2: ::memcpy(dst, Inline.Data, 2); break;
+                case 1: ::memcpy(dst, Inline.Data, 1); break;
+                case 0: break;
+            }
```
MBkkt, line 226:
```suggestion
                case 0: break;
                default: std::unreachable();
            }
```
Or any alternative, I think it should improve codegen

ydb/core/scheme/scheme_tablecell.h
```diff
@@ -10,79 +10,148 @@
...
+                case 9: ::memcpy(Inline.Data, ptr, 9); break;
+                case 8: ::memcpy(Inline.Data, ptr, 8); break;
+                case 7: ::memcpy(Inline.Data, ptr, 7); break;
+                case 6: ::memcpy(Inline.Data, ptr, 6); break;
+                case 5: ::memcpy(Inline.Data, ptr, 5); break;
+                case 4: ::memcpy(Inline.Data, ptr, 4); break;
+                case 3: ::memcpy(Inline.Data, ptr, 3); break;
+                case 2: ::memcpy(Inline.Data, ptr, 2); break;
+                case 1: ::memcpy(Inline.Data, ptr, 1); break;
+                case 0: break;
```
MBkkt, line 115:
```suggestion
                case 0: break;
                default: std::unreachable();
```
```suggestion
                case 0: break;
                default: std::unreachable();
            }
```
Or any alternative, I think it should improve codegen

ydb/core/scheme/scheme_tablecell.h
```diff
@@ -127,46 +196,39 @@ struct TCell {
...
+            }
         }
     }
 };
 
-#pragma pack(pop)
-
-static_assert(sizeof(TCell) == 12, "TCell must be 12 bytes");
+static_assert(sizeof(TCell) == 16, "TCell must be 16 bytes");
```
MBkkt, line 231: btw I think such comments is legacy of old standards when static_assert without macro was impossible

ydb/core/scheme/scheme_tablecell.h
snaury: We have enough space for 15, but the goal here is parity with TUnboxedValuePod, which inlines up to 14. It's better to cover all physically possible cases, in case MaxInlineSize is bumped to 15 later.

ydb/core/scheme/scheme_tablecell.h
snaury: It already covers all possible cases, no need for unreachable.

ydb/core/scheme/scheme_tablecell.h
snaury: True, but I don't see a point in removing the comment.

MBkkt (approved):

PR #13385, opened 2025-01-15
avoid integer overflow

uzhastik:
### Changelog entry 

...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/core/fq/libs/compute/common/utils.cpp
+++ b/ydb/core/fq/libs/compute/common/utils.cpp
@@ -159,12 +159,19 @@ struct TDurationParser {
         return res;
     }
 
-    constexpr ui32 ConsumeNumberPortion() noexcept {
-        ui32 dec = 1;
-        ui32 res = 0;
+    constexpr ui64 ConsumeNumberPortion() noexcept {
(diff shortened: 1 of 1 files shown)


ydb/core/fq/libs/compute/common/utils.cpp
```diff
@@ -159,12 +159,19 @@ struct TDurationParser {
...
+    constexpr ui64 ConsumeNumberPortion() noexcept {
+        ui64 dec = 1;
+        ui64 res = 0;
+        // max ui64 = 18 446 744 073 709 551 615
+        constexpr ui64 MAX_NUMBER = 999'999'999'999'999'999ull;
         while (!Src.empty() && IsDigit(Src.back())) {
-            res += (Src.back() - '0') * dec;
-            dec *= 10;
+            if (res < MAX_NUMBER) {
+                res += (Src.back() - '0') * dec;
```
yumkam, line 169: With `10000000000000000000000000001`,  until first `1` processed `res` remains 1 (no overflow detected), but there are overflow in `dec` (*unsigned* overflow, no UB; but there were no UB in original code either)

PR #13401, opened 2025-01-15
Remove YQL from CODEOWNERS

maximyurchuk:
Because YQL in Arcadia


spuchin (dismissed):

rvu1024:
ydb/library/yql/dq
ydb/library/yql/providers/dq
ydb/library/yql/providers/yt
ydb/library/yql/tests
ydb/library/yql/tools

Issue #13427, opened 2025-01-16
Add counter for total file handles in KQP spilling


spuchin:
We need to reduce our usage of file handles in spilling, as at the moment we can reach system limit, which cannot be configured inside YDB:
https://github.com/ydb-platform/ydb/issues/11637

As a first step, we need to actually monitor our file handles usage in spilling, so I propose we add sensor for total usage per node, similar to other spilling counters:
https://github.com/ydb-platform/ydb/blob/27eb5606466f5fe75662cc3e820820943cf55609/ydb/library/yql/dq/actors/spilling/spilling_counters.h#L13

PR #13611, opened 2025-01-21
#13427 Add a monitoring counter for fds kept open by the spilling system

uniquelogin:
Monitor file descriptor usage from the moment a new spill file is registered with the spilling system until it's closed and released. Seems to work but might be a bit raw.

TODO: Does not account for additional opens in TReadFileOp/TWriteFileOp; do we need that? These additional fds are short-lived (only held open for the duration of read/write call) and the total amount is bounded by IoThreadPoolWorkersCount.

TODO: The spilling system actor shutdown sequence should be extracted into a function in the _ut.cpp.

Why are these long-lived fds being held open anyway? TReadFileOp and TWriteFileOp only need the file name. To keep files alive during GC, maybe?


lll-phill-lll:
Thank you for the Pull Request!

Regarding your questions:
1. These file descriptors can be ignored. Counting them will complicate the code, as it will require passing the counters into internal structures, which will make the already complicated code even more complicated. While you correctly noted that the number of simultaneously open temporary file descriptors does not exceed the number of threads, which in our case is 2. This number may change, but most likely not much.
2. Regarding the second question, I do not have an answer to it. This is probably a place for improvement. And file descriptors can indeed not be kept open. But this question requires additional investigation.

lll-phill-lll (approved):

spuchin (approved):

PR #13732, opened 2025-01-22
added max-age

StekPerepolnen:
### Changelog entry
https://nebius.atlassian.net/browse/NBYDB-770
add max-age to session cookie depending on iam response
specification said that set-cookie max-age attribute is unsigned 31 bit

### Changelog category
- Improvement

### DIFF
--- a/ydb/mvp/oidc_proxy/oidc_impersonate_start_page_nebius.cpp
+++ b/ydb/mvp/oidc_proxy/oidc_impersonate_start_page_nebius.cpp
@@ -83,9 +83,7 @@ void THandlerImpersonateStart::ProcessImpersonatedToken(const NJson::TJsonValue&
     if (!jsonExpiresIn->GetUInteger(&expiresIn)) {
         return ReplyBadRequestAndPassAway("Wrong OIDC provider response: failed to extract `expires_in`");
     }
-    if (expiresIn > std::numeric_limits<i32>::max()) {
-        expiresIn = std::numeric_limits<i32>::max();
-    }
+    expiresIn = std::min(expiresIn, 604800ULL); // clean cookies no less than once a week.
     TString impersonatedCookieName = CreateNameImpersonatedCookie(Settings.ClientId);
     TString impersonatedCookieValue = Base64Encode(impersonatedToken);
     BLOG_D("Set impersonated cookie: (" << impersonatedCookieName << ": " << NKikimr::MaskTicket(impersonatedCookieValue) << ")");
--- a/ydb/mvp/oidc_proxy/oidc_impersonate_start_page_nebius.cpp
+++ b/ydb/mvp/oidc_proxy/oidc_impersonate_start_page_nebius.cpp
@@ -66,13 +66,23 @@ void THandlerImpersonateStart::RequestImpersonatedToken(TString& sessionToken, T
     Become(&THandlerImpersonateStart::StateWork);
 }
 
-void THandlerImpersonateStart::ProcessImpersonatedToken(const TString& impersonatedToken) {
+void THandlerImpersonateStart::ProcessImpersonatedToken(const NJson::TJsonValue& jsonValue) {
+    const NJson::TJsonValue* jsonImpersonatedToken;
+    const NJson::TJsonValue* jsonExpiresIn;
+    if (!jsonValue.GetValuePointer("impersonation", &jsonImpersonatedToken)) {
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: impersonation not found");
+    }
+    if (!jsonValue.GetValuePointer("expires_in", &jsonExpiresIn)) {
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: expires_in not found");
+    }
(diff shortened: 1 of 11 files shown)


ydb/mvp/oidc_proxy/oidc_impersonate_start_page_nebius.cpp
```diff
@@ -66,13 +66,23 @@ void THandlerImpersonateStart::RequestImpersonatedToken(TString& sessionToken, T
...
+void THandlerImpersonateStart::ProcessImpersonatedToken(const NJson::TJsonValue& jsonValue) {
+    const NJson::TJsonValue* jsonImpersonatedToken;
+    const NJson::TJsonValue* jsonExpiresIn;
+    if (!jsonValue.GetValuePointer("impersonation", &jsonImpersonatedToken)) {
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: impersonation not found");
+    }
+    if (!jsonValue.GetValuePointer("expires_in", &jsonExpiresIn)) {
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: expires_in not found");
+    }
+    TString impersonatedToken = jsonImpersonatedToken->GetStringRobust();
```
UgnineSirdis, line 78: `GetStringRobust`? Will it be OK to get value like `["my_token"]` ? I think that token must be a string and other cases will not be processed correctly, event if we get variable of type string here.

ydb/mvp/oidc_proxy/oidc_session_create_nebius.cpp
```diff
@@ -33,14 +33,24 @@ void THandlerSessionCreateNebius::RequestSessionToken(const TString& code) {
...
+void THandlerSessionCreateNebius::ProcessSessionToken(const NJson::TJsonValue& jsonValue) {
+    const NJson::TJsonValue* jsonAccessToken;
+    const NJson::TJsonValue* jsonExpiresIn;
+    if (!jsonValue.GetValuePointer("access_token", &jsonAccessToken)) {
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: access_token not found");
+    }
+    if (!jsonValue.GetValuePointer("expires_in", &jsonExpiresIn)) {
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: expires_in not found");
+    }
+    TString sessionToken = jsonAccessToken->GetStringRobust();
```
UgnineSirdis, line 45: The same comments about the same code here

ydb/mvp/oidc_proxy/oidc_session_create_yandex.cpp
```diff
@@ -45,8 +52,8 @@ void THandlerSessionCreateYandex::ProcessSessionToken(const TString& sessionToke
     SetHeader(meta, "authorization", token);
     meta.Timeout = TDuration::Seconds(10);
 
-    NActors::TActorSystem* actorSystem = ctx.ActorSystem();
-    NActors::TActorId actorId = ctx.SelfID;
+    NActors::TActorSystem* actorSystem = TlsActivationContext->ActorSystem();
```
UgnineSirdis, line 55: `TActivationContext::ActorSystem()` is maybe more beautiful for my opinion. But I don't insist :)

UgnineSirdis (changes requested):

ydb/mvp/oidc_proxy/oidc_impersonate_start_page_nebius.cpp
```diff
@@ -69,14 +69,23 @@ void THandlerImpersonateStart::RequestImpersonatedToken(TString& sessionToken, T
...
     }
     if (!jsonValue.GetValuePointer("expires_in", &jsonExpiresIn)) {
-        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: expires_in not found");
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: `expires_in` not found");
+    }
+    if (!jsonExpiresIn->GetUInteger(&expiresIn)) {
+        return ReplyBadRequestAndPassAway("Wrong OIDC provider response: failed to extract `expires_in`");
+    }
+    if (expiresIn > std::numeric_limits<i32>::max()) {
+        expiresIn = std::numeric_limits<i32>::max();
```
UgnineSirdis, line 87: I think this is too much for default limit. 2147483647 seconds are about 24855 days, 68 years. It will be a problem if your cookie will not expire such a long time. What if it is being set for some kind of error?

UgnineSirdis (changes requested):

ydb/mvp/oidc_proxy/oidc_impersonate_start_page_nebius.cpp
StekPerepolnen: According to the specification, Set-Cookie should be an unsigned 31-bit value.

However, I agree that extending the lifespan is unnecessary to ensure cookies are periodically cleared. I would suggest setting the limit to one day (which, in practice, will almost always be 12 hours unless IAM changes it).

UgnineSirdis (approved):

PR #13775, opened 2025-01-23
Extract IsAdministrator

kungasc:
### Changelog entry 

...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/core/base/auth.cpp
+++ b/ydb/core/base/auth.cpp
@@ -1,0 +1,43 @@
+#include "auth.h"
+
+namespace NKikimr {
+
+namespace {
+
+bool HasToken(const TAppData* appData, const NACLib::TUserToken& userToken) {
+    for (const auto& sid : appData->AdministrationAllowedSIDs) {
+        if (userToken.IsExist(sid)) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+}
+
+bool IsAdministrator(const TAppData* appData, const TString& userToken) {
+    if (appData->AdministrationAllowedSIDs.empty()) {
+        return true;
+    }
+
+    if (!userToken) {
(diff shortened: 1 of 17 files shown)


ydb/core/base/auth.cpp
```diff
@@ -0,0 +1,43 @@
...
+bool IsAdministrator(const TAppData* appData, const TString& userToken) {
+    if (appData->AdministrationAllowedSIDs.empty()) {
+        return true;
+    }
+
+    if (!userToken) {
+        return false;
+    }
+
+    return HasToken(appData, NACLib::TUserToken(userToken));
```
ijon, line 28: Minor note: `NACLib::TUserToken()` will cause process crash if `userToken` contains invalid (or empty) protobuf. It was ok when token string were getting retrieved from the known context where token string guaranteed to be valid. But utility method lacks that guarantee.

ydb/core/grpc_services/rpc_rate_limiter_api.cpp
```diff
@@ -66,19 +67,8 @@ class TRateLimiterRequest : public TRpcOperationRequestActor<TDerived, TRequest>
...
-                for (auto &sid : AppData()->AdministrationAllowedSIDs) {
-                    if (userToken.IsExist(sid)) {
-                        allowed = true;
-                        break;
-                    }
-                }
-            }
 
-            if (!allowed) {
+            if (!IsAdministrator(AppData(), self->Request_->GetSerializedToken())) {
```
ijon, line 71: I believe that `GetSerializedToken()` could be replaced with `GetInternalToken()` that returns `TUserToken` that is already exists in the request context.

ijon (approved):

PR #13904, opened 2025-01-27
SystemView Auth Sort results

kungasc:
### Changelog entry 

...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/core/sys_view/auth/groups.cpp
+++ b/ydb/core/sys_view/auth/groups.cpp
@@ -30,14 +30,25 @@ protected:
     void FillBatch(NKqp::TEvKqpCompute::TEvScanData& batch, const TNavigate::TEntry& entry) override {
         Y_ABORT_UNLESS(entry.Status == TNavigate::EStatus::Ok);
         Y_ABORT_UNLESS(CanonizePath(entry.Path) == TBase::TenantName);
+
+        TVector<const TDomainInfo::TGroup*> groups(::Reserve(entry.DomainInfo->Groups.size()));
+        for (const auto& group : entry.DomainInfo->Groups) {
+            groups.push_back(&group);
+        }
+        std::sort(groups.begin(), groups.end(), [](const auto* left, const auto* right) {
+            return left->Sid < right->Sid;
+        });
+        Y_DEBUG_ABORT_UNLESS(groups.end() == std::unique(groups.begin(), groups.end(), [](const auto* left, const auto* right) {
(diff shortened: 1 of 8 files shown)


ydb/core/sys_view/auth/groups.cpp
```diff
@@ -30,14 +30,25 @@ class TGroupsScan : public TAuthScanBase<TGroupsScan> {
     void FillBatch(NKqp::TEvKqpCompute::TEvScanData& batch, const TNavigate::TEntry& entry) override {
         Y_ABORT_UNLESS(entry.Status == TNavigate::EStatus::Ok);
         Y_ABORT_UNLESS(CanonizePath(entry.Path) == TBase::TenantName);
+
+        TVector<const TDomainInfo::TGroup*> groups(::Reserve(entry.DomainInfo->Groups.size()));
```
azevaykin, line 34: Minor: these lines of code: "push back, sort, assert" are practically the same in all FillBatch methods.

ydb/core/sys_view/auth/permissions.cpp
```diff
@@ -49,32 +46,48 @@ class TPermissionsScan : public TAuthScanBase<TPermissionsScan> {
...
-                    case Schema::AuthPermissions::Permission::ColumnId:
-                        cells.push_back(TCell(permission.data(), permission.size()));
-                        break;
-                    default:
+            auto acePermissions = ConvertACLMaskToYdbPermissionNames(ace.GetAccessRight());
+            for (const auto& permission : acePermissions) {
+                permissions.emplace_back(ace.HasSID() ? ace.GetSID() : TString{}, std::move(permission));
+            }
+        }
+        std::sort(permissions.begin(), permissions.end(), [](const auto& left, const auto& right) {
```
azevaykin, line 54: Does this affect performance on a large test? Could you add new times to the description of this PR ?

kunga: In large test we measure only scheme shard time as it's the bottleneck of the database

Time of a scan's execution doesn't matter a lot as it isn't block anything, and of course, if we enumerate millions of permissions it takes long

But sorting doesn't seem to affect that time much as we sort quite a short list for a specific path

kunga: Also we expect these method to be called for only some paths as paging or some range filters should be applied

azevaykin (approved):

PR #13969, opened 2025-01-29
Use the root domain as a database for all operations with users and groups

andrewstalin:
### Changelog entry

Now, with the "DomainLoginOnly" flag set, authorization occurs on the root SchemeShard. "SQL Script" creates users and groups in the same way on the root SchemeShard. But "QueryService" creates them in a domain. It is  proposed to unify this behavior and, in all cases where the "DomainLoginOnly" flag is set, use the root SchemeShard to store users and authorize them. Otherwise, either the domain or the root SchemeShard will be used, depending on the value specified in the "Database" parameter.

### Changelog category

### DIFF
--- a/ydb/core/kqp/executer_actor/kqp_scheme_executer.cpp
+++ b/ydb/core/kqp/executer_actor/kqp_scheme_executer.cpp
@@ -123,7 +123,7 @@ public:
         Become(&TKqpSchemeExecuter::ExecuteState);
     }
 
-    TString GetDatabaseName() const {
+    TString GetDatabaseForLoginOperation() const {
         const auto domainLoginOnly = AppData()->AuthConfig.GetDomainLoginOnly();
         const auto domain = AppData()->DomainsInfo ? AppData()->DomainsInfo->GetDomain() : nullptr;
         const auto domainName = domain ? domain->Name : "";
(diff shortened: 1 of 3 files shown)


ydb/tests/functional/tenants/test_users_groups_with_acl.py
```diff
@@ -0,0 +1,129 @@
...
+    def setup_class(cls):
+        config_generator = KikimrConfigGenerator()
+        config_generator.yaml_config["auth_config"] = {
+            "domain_login_only": cls.DomainLoginOnly,
+        }
+        cls.cluster = KiKiMR(config_generator)
+        cls.cluster.start()
+
+    @classmethod
+    def teardown_class(cls):
```
azevaykin, line 23: It's better to use `harness ydb_fixtures`

ydb/core/kqp/executer_actor/kqp_scheme_executer.cpp
```diff
@@ -120,6 +122,15 @@ class TKqpSchemeExecuter : public TActorBootstrapped<TKqpSchemeExecuter> {
         Become(&TKqpSchemeExecuter::ExecuteState);
     }
 
+    TString GetDatabaseName() const {
+        const auto appData = AppData();
+        if (appData && appData->AuthConfig.GetDomainLoginOnly()) {
```
ijon, line 127: AppData() result cannot be nullptr if called from within an actor system (which it is).

ydb/core/kqp/executer_actor/kqp_scheme_executer.cpp
```diff
@@ -120,6 +122,15 @@ class TKqpSchemeExecuter : public TActorBootstrapped<TKqpSchemeExecuter> {
         Become(&TKqpSchemeExecuter::ExecuteState);
     }
 
+    TString GetDatabaseName() const {
```
ijon, line 125: There is already common (for kqp) method for that -- `NKikimr::NKqp::NSchemeHelpers::SetDatabaseForLoginOperation`, see [ydb/core/kqp/gateway/utils/scheme_helpers.cpp#L59](https://github.com/ydb-platform/ydb/blob/main/ydb/core/kqp/gateway/utils/scheme_helpers.cpp#L59).

Please consider using it instead.

ydb/tests/functional/tenants/test_users_groups_with_acl.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_query_create_group_by_tenant_admin(self):
+        self.query_create_group(self.tenant_admin_driver_config)
+
+
+class TestCreateSubjectsWithDomainLoginOnly(BaseCreateSubjects):
+    DomainLoginOnly = True
+
+
+class TestCreateSubjectsWithoutDomainLoginOnly(BaseCreateSubjects):
+    DomainLoginOnly = False
```
ijon, line 129: Also pytest has plenty of ways to conveniently [parametrize test functions and classes](https://docs.pytest.org/en/stable/how-to/parametrize.html#parametrize).

ydb/core/kqp/executer_actor/kqp_scheme_executer.cpp
```diff
@@ -120,6 +123,15 @@ class TKqpSchemeExecuter : public TActorBootstrapped<TKqpSchemeExecuter> {
...
```
adameat, line 126: I would suggest to change naming to specify that it's not just a database name... it's a database-owner of the security state... otherwise someone could reuse it not looking into the implementation.

andrewstalin: Is that okay?

ijon (approved):

Issue #12724, opened 2024-12-18
S3 backup & restore VIEW


jepett0:
**В рамках релиза 25-3 только включение ФФ**

- [x] `export s3`: https://github.com/ydb-platform/ydb/pull/13677
- [x] `import s3`: https://github.com/ydb-platform/ydb/pull/13361

### Changelog entry 

Backup-restore VIEWs (flag `EnableViewExport`).

### Changelog category 

* New feature

PR #13974, opened 2025-01-29
backup restore test suite refactoring

jepett0:
Refactoring so the future code review would be easier.

Previously the backup and restore lambdas received an object to be backed-up and restored. However, this parameter was unused in the local backup (a.k.a. dump) and restore tests. To eliminate this awkwardness, I decided to make the lambdas "whole database", i.e. by default they backup and restore the whole database. Unfortunately, the TS3Mock class lacks the list objects request processing, so it seems impossible to recursively list the bucket on import. Therefore, I have added the `items` parameter to the BackupRestoreS3::CreateRestoreLambda function in order to imitate the "whole database restore" behavior of the created lambda. You have to specify the items to be imported yourself, unfortunately.

### Changelog category 

### Additional information

- https://github.com/ydb-platform/ydb/issues/12724

### DIFF
--- a/ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
+++ b/ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
@@ -102,28 +102,10 @@ TDataQueryResult ExecuteDataModificationQuery(TSession& session,
     return result;
 }
 
-NQuery::TExecuteQueryResult ExecuteQuery(NQuery::TSession& session, const TString& script, bool isDDL = false) {
-    const auto result = session.ExecuteQuery(
-        script,
-        isDDL ? NQuery::TTxControl::NoTx() : NQuery::TTxControl::BeginTx().CommitTx()
-    ).ExtractValueSync();
-    UNIT_ASSERT_C(result.IsSuccess(), "query:\n" << script << "\nissues:\n" << result.GetIssues().ToString());
-    return result;
-}
-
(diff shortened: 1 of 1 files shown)


ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
```diff
@@ -472,24 +487,19 @@ void TestRestoreDirectory(const char* directory, TSchemeClient& client, TBackupF
...
-        UNIT_ASSERT_C(result.IsSuccess(), result.GetIssues().ToString());
-    }
-
-    auto CreateBackupLambda(const TDriver& driver, const TFsPath& pathToBackup, bool schemaOnly = false) {
-        return [&driver, &pathToBackup, schemaOnly](const char* table) {
-            Y_UNUSED(table);
-            // TO DO: implement NDump::TClient::Dump and call it instead of BackupFolder
-            NBackup::BackupFolder(driver, "/Root", ".", pathToBackup, {}, schemaOnly, false);
+    auto CreateBackupLambda(const TDriver& driver, const TFsPath& fsPath, const TString& dbPath = "/Root") {
+        return [&]{
```
CyberROFL, line 491: Здесь и далее:
```suggestion
        return [&] {
```

Или:
```cpp
        return [&]() {
```

CyberROFL (approved):

PR #14153, opened 2025-02-03
TOPIC: setting retention_storage_mb via YQL does not work

jepett0:
PR to highlight the bug

### Changelog category 


jepett0:
This was a draft PR to illustrate an issue by adding a test that breaks in CI

PR #14238, opened 2025-02-05
push ExtractMembers into TDqLookupSourceWrap

yumkam:
### Changelog entry 

...

### Changelog category 

### Additional information

When right-side table contains more columns than used, we can just drop them.
Extend this optimization for TDqLookupSource (and refactor a bit to reduce copy-paste)


yumkam:
Re-run large tests from `ydb/tests/fq/generic/streaming` with current main merged, seems works (and manually verified ast, matches expected, unused fields removed)

uzhastik (approved):

PR #14281, opened 2025-02-06
dq streamlookup join: treat as wide-channel-capable connection

yumkam:
### Changelog entry 

...

### Changelog category 

### Additional information

...


yumkam:
Triggers exception in ydb streaming
```
ydb/library/yql/dq/runtime/dq_columns_resolve.cpp:15
        YQL_ENSURE(TryFromString(columnName, idx), "Expecting number as column name");
```

yumkam:
Does not worth continued effort

yumkam:
Needs fixes/rework in `RebuildStageInputsAsWide` for case ShuffleMode hash (column names needs renaming; probably, need separate list)

PR #14510, opened 2025-02-12
Test with backup & restore rate limiter's properties

CyberROFL:
### Changelog entry 

...

### Changelog category 

### Additional information

...

### DIFF
--- a/ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
+++ b/ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
@@ -69,10 +69,26 @@ bool operator==(
     return left == right;
 }
 
+bool operator==(
+    const Ydb::RateLimiter::MeteringConfig& lhs,
+    const Ydb::RateLimiter::MeteringConfig& rhs
+) {
+    return google::protobuf::util::MessageDifferencer::Equals(lhs, rhs);
+}
+
+bool operator==(const TMeteringConfig& lhs, const TMeteringConfig& rhs) {
+    Ydb::RateLimiter::MeteringConfig left;
+    lhs.SerializeTo(left);
+    Ydb::RateLimiter::MeteringConfig right;
+    rhs.SerializeTo(right);
+    return left == right;
+}
+
 bool operator==(const TDescribeResourceResult& lhs, const TDescribeResourceResult& rhs) {
     UNIT_ASSERT_C(lhs.IsSuccess(), lhs.GetIssues().ToString());
     UNIT_ASSERT_C(rhs.IsSuccess(), rhs.GetIssues().ToString());
-    return lhs.GetHierarchicalDrrProps() == rhs.GetHierarchicalDrrProps();
+    return lhs.GetHierarchicalDrrProps() == rhs.GetHierarchicalDrrProps()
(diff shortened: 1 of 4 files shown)


Gazizonoki (dismissed):

pnv1 (dismissed):

ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
```diff
@@ -857,39 +873,74 @@ void DropRateLimiter(
...
+    std::vector<TDescribeResourceResult> original;
+    for (const auto& [resource, _] : rateLimiters) {
+        original.emplace_back(DescribeRateLimiter(rateLimiterClient, path, resource));
     }
 
     backup();
 
-    for (int i = 2; i >= 0; --i) {
-        DropRateLimiter(rateLimiterClient, coordinationNodePath, rateLimiters[i]);
+    for (size_t i = rateLimiters.size() - 1; i >= 0; --i) {
```
jepett0, line 936: Wouldn't it be an infinite loop? By the way, since we are talking about the preferred type for the loop variable, why do you think `size_t` is better? I have found the following in the [Google style guide](https://google.github.io/styleguide/cppguide.html#Integer_Types) for C++:

> We use int very often, for integers we know are not going to be too big, e.g., loop counters. Use plain old int for such things. You should assume that an int is at least 32 bits, but don't assume that it has more than 32 bits. If you need a 64-bit integer type, use int64_t or uint64_t.
...
You should not use the unsigned integer types such as uint32_t, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, do not use unsigned types to say a number will never be negative. Instead, use assertions for this.

ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
```diff
@@ -857,39 +873,74 @@ void DropRateLimiter(
...
 
     restore();
-    for (int i = 0; i < 3; ++i) {
-        UNIT_ASSERT_EQUAL_C(
-            DescribeRateLimiter(rateLimiterClient, coordinationNodePath, rateLimiters[i]),
-            originalDescriptions[i],
-            "i: " << i
-        );
+    for (size_t i = 0; i < rateLimiters.size(); ++i) {
+        UNIT_ASSERT_EQUAL(DescribeRateLimiter(rateLimiterClient, path, rateLimiters[i].first), original[i]);
```
jepett0, line 943: I have added `_C` to the `UNIT_ASSERT_EQUAL` so that any potential error would be easier to debug. This assertion macro does not print the descriptions, so I'm guessing it would be impossible to know exactly which descriptions differ. I recommend adding the `"i: " << i` message back

jepett0 (changes requested):

pnv1 (approved):

ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
jepett0: What is really needed here is the printout of the mismatched descriptions. I guess, it the index by itself won't help much, so it is ok to remove it here

jepett0 (approved):

Gazizonoki (approved):

PR #14530, opened 2025-02-13
Create prefix impl table and ability to create prefixed vector index

MBkkt:
### Changelog category 

### Additional information

It adds additional layer for vector index: prefix impl table (prefix columns and id column)

### DIFF
--- a/ydb/core/base/table_index.cpp
+++ b/ydb/core/base/table_index.cpp
@@ -37,8 +37,13 @@ bool ContainsSystemColumn(const auto& columns) {
     return false;
 }
 
-constexpr std::string_view ImplTables[] = {
-    ImplTable, NTableVectorKmeansTreeIndex::LevelTable, NTableVectorKmeansTreeIndex::PostingTable,
+const TString ImplTables[] = {
+    ImplTable,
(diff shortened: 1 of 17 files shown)


ydb/core/base/table_index.cpp
```diff
@@ -28,8 +28,22 @@ bool Contains(const auto& names, std::string_view str) {
...
+    return false;
+}
+
+const TString ImplTables[] = {
+    ImplTable,
+    NTableVectorKmeansTreeIndex::LevelTable,
+    NTableVectorKmeansTreeIndex::PostingTable,
+    NTableVectorKmeansTreeIndex::PrefixTable,
+    TString{NTableVectorKmeansTreeIndex::PostingTable} + NTableVectorKmeansTreeIndex::BuildSuffix0,
+    TString{NTableVectorKmeansTreeIndex::PostingTable} + NTableVectorKmeansTreeIndex::BuildSuffix1,
```
MBkkt, line 46: I'm not really sure is it needed or not, but I find place where we check that impl table one of this or crash.
So it's my paranoia

CyberROFL (approved):
schemeshard - ok

gridnevvvit (approved):

PR #14554, opened 2025-02-13
rework node priorities to make dc preference overridable

### LINKED ISSUE #14323: ensure system tablets are alive when an entire dc is in a crash-loop
System tablets (specifically, coordinators and mediators) are kept in one datacenter when possible. Assume we are deploying a new problematic version, and at first rolled it out on one datacenter, causing all nodes there to become stuck in a crash-loop. Then, system tablets remain in this DC, are not working, and the entire database is effectively down, despite only one location having problems. This issue is for improving the behavior in this case.

- [x] Run simulations to see whether the "stick system tablets together" functionality is the cause of the problems, or if it would be the same with it turned off.
- [ ] Make a fix based on the simulations

vporyadke:
### Changelog entry 

Improved node priority logic in Hive.

### Changelog category 

* Improvement

### Additional information

Updated node priorities logic to avoid issues when system tablets remain in a datacenter even when all nodes of the datacenter are in a crash loop.

### DIFF
--- a/ydb/core/mind/hive/hive_impl.cpp
+++ b/ydb/core/mind/hive/hive_impl.cpp
@@ -1176,15 +1176,15 @@ TNodeInfo* THive::SelectNode<NKikimrConfig::THiveConfig::HIVE_NODE_SELECT_STRATE
     return itNode->Node;
 }
 
-TVector<THive::TSelectedNode> THive::SelectMaxPriorityNodes(TVector<TSelectedNode> selectedNodes, const TTabletInfo& tablet) const
+TVector<THive::TSelectedNode> THive::SelectMaxPriorityNodes(TVector<TSelectedNode> selectedNodes, const TTabletInfo& tablet, TDataCenterPriority& dcPriority) const
 {
     i32 priority = std::numeric_limits<i32>::min();
     for (const TSelectedNode& selectedNode : selectedNodes) {
-        priority = std::max(priority, selectedNode.Node->GetPriorityForTablet(tablet));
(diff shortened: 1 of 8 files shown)


ydb/core/mind/hive/hive_impl.cpp
```diff
@@ -1274,53 +1274,21 @@ THive::TBestNodeResult THive::FindBestNode(const TTabletInfo& tablet, TNodeId su
...
     for (size_t numGroup = 0; numGroup < dataCentersGroups.size(); ++numGroup) {
         const NKikimrHive::TDataCentersGroup* dcGroup = dataCentersGroups[numGroup];
-        if (dcGroup->DataCenterSize()) {
-            for (TDataCenterId dc : dcGroup->GetDataCenter()) {
-                indexDC2Group[dc] = candidateGroups.data() + numGroup;
-            }
-        } else {
-            for (const ui64 dcId : dcGroup->GetDataCenterNum()) {
-                indexDC2Group[DataCenterToString(dcId)] = candidateGroups.data() + numGroup;
-            }
```
CyberROFL, line 1289: Это точно deprecated? Как можно в этом удостовериться?

CyberROFL (dismissed):

ydb/core/mind/hive/hive_impl.cpp
vporyadke: It was introduced in April 2021, made deprecated in September 2021, internal tracker shows no signs of it ever being set anywhere, and it is not set on any of our main prod clusters.

adameat (dismissed):

CyberROFL (approved):

PR #14704, opened 2025-02-17
Avoid persistent and in-memory tx status getting out-of-sync

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

It was discovered that in certain situation tx status parts and in-memory state may diverge, allowing transaction ids to be reused, which would unexpectedly become committed or removed after a restart. Make sure in-memory state doesn't diverge from the state that could be restored from disk on restart.

Fixes KIKIMR-22538.

### DIFF
--- a/ydb/core/tablet_flat/flat_database.h
+++ b/ydb/core/tablet_flat/flat_database.h
@@ -226,7 +226,8 @@ public:
     void UpdateApproximateFreeSharesByChannel(const THashMap<ui32, float>& approximateFreeSpaceShareByChannel);
     TString SnapshotToLog(ui32 table, TTxStamp);
 
-    TAutoPtr<TSubset> Subset(ui32 table, TArrayRef<const TLogoBlobID> bundle, TEpoch before) const;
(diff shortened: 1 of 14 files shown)


ydb/core/tablet_flat/flat_database.h
```diff
@@ -235,11 +236,12 @@ class TDatabase {
     TBundleSlicesMap LookupSlices(ui32 table, TArrayRef<const TLogoBlobID> bundles) const;
     void ReplaceSlices(ui32 table, TBundleSlicesMap slices);
 
-    void Replace(ui32 table, TArrayRef<const TPartView>, const TSubset&);
-    void ReplaceTxStatus(ui32 table, TArrayRef<const TIntrusiveConstPtr<TTxStatusPart>>, const TSubset&);
+    void Replace(ui32 table, const TSubset&,
+            TArrayRef<const TPartView>, TArrayRef<const TIntrusiveConstPtr<TTxStatusPart>>);
```
azevaykin, line 240: unnecessary spaces

snaury: Поправил

azevaykin (approved):

Issue #14709, opened 2025-02-17
view: fix a bug if the query contains 'LIKE'


jepett0:
https://st.yandex-team.ru/KIKIMR-22563

```sql
CREATE VIEW demo_view WITH (security_invoker = TRUE) AS
  SELECT
    CASE WHEN datt.v LIKE Unwrap(patt.v) THEN 1 ELSE 0 END AS attr
  FROM (SELECT "aaa"u AS v) AS patt
  CROSS JOIN (SELECT "bbb"u AS v) AS datt;
```

jepett0:
Reproducible on main:

Status: GENERIC_ERROR
Issues: 
<main>:3:20: Error: Name not found: Udf1

jepett0:
PR in main:
https://a.yandex-team.ru/arc/review/8166949/details

PR #14788, opened 2025-02-19
YQL parsing object creation: init feature nodes

jepett0:
### Changelog category 

### Description for reviewers 

While working on the followig bug in views:
- https://github.com/ydb-platform/ydb/issues/14709

we have discovered that the object feature nodes haven't been initialized by the translator. It seems to matter only for views, because views have a non-trivial feature, `query_ast` which might potentially need initialization for the proper view query validation to happen.


jepett0:
This change will be merged to Arcadia

PR #14804, opened 2025-02-19
Return virtual timestamp for execute query | execute scan api

dcherednik:
Untitled Diagram drawio-2

- YDB has CDC feature which allows to get async stream of changes to process it somewhere outside.
- Although we speak about async changes it is important to know whether particular changes already present in data we read from CDC or not.
- CDC already has feature called virtual timestamp which allows to restore transactions order on the receiver side (used in Aardappel)

The simplest way to achieve it - return virtual timestamp as side information in the execute query result. This feature already implemented for read table result.

Execution modes which can return virtual timestamp in current code base:
ExecuteScanQuery (TableService)
ExecuteQuery RO with MVCC (QueryService)
ExecuteQuery RW non interactive

Execution modes which unable to return virtual timestamp
ExecuteQuery RW interactive

### DIFF
--- a/ydb/docs/en/core/concepts/cdc.md
+++ b/ydb/docs/en/core/concepts/cdc.md
@@ -52,6 +52,12 @@ Returning these timestamps for each query may cause noticeable overhead in case 
 
 For explicitly created transactions (transactions created via BeginTransactionRequest rpc call), these timestamps are always returned and do not require the flag.
 
+{% note info %}
+
+StreamReadTable and StreamExecuteScanQuery calls also have snapshot_timestamp fields in the response.
+
+{% endnote %}
+
 ## Initial table scan {#initial-scan}
 
 By default, a changefeed only includes records about those table rows that changed after the changefeed was created. Initial table scan enables you to export, to the changefeed, the values of all the rows that existed at the time of changefeed creation.
--- a/ydb/docs/en/core/concepts/cdc.md
+++ b/ydb/docs/en/core/concepts/cdc.md
@@ -38,6 +38,20 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
 
 {% endnote %}
 
+## Getting virtual timstamps during query execution
+
+As was writen above virtual timestamp represents coordination time. There is a possibility to return a virtual timestamp during query execution. It can be used to check whether data read or written in a transaction is present in the CDC async replica.
+
+The public API has two field to represent transaction reletaed virtual timestamp.
+
+1. 'snapshot_timestamp' – Represents the moment in time when YDB takes an MVCC or persistent snapshot. It is empty if the request is executed without a snapshot. This is mostly relevant for read-only transactions and is returned during the first part of the result set. The practical purpose is to implement synchronization between readers in transactions and readers from replicas.
+
+1. 'commit_timestamp' – Represents the moment in time when YDB commits data into persistent storage. This is mostly relevant for read-write and write-only transactions and is returned in the last part of the result set. The practical purpose is to implement synchronization between transactional writers and readers from replicas.
+
+Returning these timestamps for each query may cause noticeable overhead in case of tiny result. To use them, the 'with_tx_timestamp' flag must be set in the 'TransactionControl' message.
+
+For explicitly created transactions, these timestamps are always returned and do not require the flag.
+
 ## Initial table scan {#initial-scan}
 
 By default, a changefeed only includes records about those table rows that changed after the changefeed was created. Initial table scan enables you to export, to the changefeed, the values of all the rows that existed at the time of changefeed creation.
--- a/ydb/docs/en/core/concepts/cdc.md
+++ b/ydb/docs/en/core/concepts/cdc.md
@@ -40,17 +40,17 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
 
 ## Getting virtual timstamps during query execution
 
-As was writen above virtual timestamp represents coordination time. There is a possibility to return a virtual timestamp during query execution. It can be used to check whether data read or written in a transaction is present in the CDC async replica.
+As was written above virtual timestamp represents Global coordinator time. There is a possibility to return a virtual timestamp during query execution. It can be used to check whether data read or written in a transaction is present in the CDC async replica.
 
-The public API has two field to represent transaction reletaed virtual timestamp.
+The YDB API has two fields to get virtual timestamp from transaction.
 
-1. 'snapshot_timestamp' – Represents the moment in time when YDB takes an MVCC or persistent snapshot. It is empty if the request is executed without a snapshot. This is mostly relevant for read-only transactions and is returned during the first part of the result set. The practical purpose is to implement synchronization between readers in transactions and readers from replicas.
+1. 'snapshot_timestamp' – Represents the moment in time when YDB takes an MVCC or persistent snapshot. It is empty if the request is executed without a snapshot. This is mostly relevant for read-only transactions and is returned with the first part of the stream result. The practical purpose is to implement synchronization between readers in transactions and readers from replicas.
 
-1. 'commit_timestamp' – Represents the moment in time when YDB commits data into persistent storage. This is mostly relevant for read-write and write-only transactions and is returned in the last part of the result set. The practical purpose is to implement synchronization between transactional writers and readers from replicas.
+1. 'commit_timestamp' – Represents the moment in time when YDB commits data into persistent storage. This is mostly relevant for read-write and write-only transactions and is returned with the last part of the stream result. The practical purpose is to implement synchronization between transactional writers and readers from replicas.
 
 Returning these timestamps for each query may cause noticeable overhead in case of tiny result. To use them, the 'with_tx_timestamp' flag must be set in the 'TransactionControl' message.
 
-For explicitly created transactions, these timestamps are always returned and do not require the flag.
+For explicitly created transactions (transactions created via BeginTransactionRequest rpc call), these timestamps are always returned and do not require the flag.
 
 ## Initial table scan {#initial-scan}
 
--- a/ydb/public/api/protos/ydb_query.proto
+++ b/ydb/public/api/protos/ydb_query.proto
@@ -83,6 +83,9 @@ message TransactionControl {
     }
 
     bool commit_tx = 10;
+
+    // If set the timestamp will be returned along with ExecuteQueryResponsePart
+    bool return_virtual_timestamp = 11;
 }
 
 message BeginTransactionRequest {
@@ -94,6 +97,8 @@ message BeginTransactionRequest {
 message TransactionMeta {
     // Transaction identifier
     string id = 1 [(Ydb.length).le = 1024];
+    // Optional timestamp that represents snapshot acquired to start transaction
+    VirtualTimestamp snapshot_timestamp = 2;
 }
 
 message BeginTransactionResponse {
@@ -113,6 +118,9 @@ message CommitTransactionRequest {
 message CommitTransactionResponse {
     StatusIds.StatusCode status = 1;
     repeated Ydb.Issue.IssueMessage issues = 2;
+
+    // Optional timestamp that corresponds to the commit transaction
+    VirtualTimestamp commit_timestamp = 3;
 }
 
 message RollbackTransactionRequest {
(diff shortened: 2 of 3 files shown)


dcherednik:
There are possibility to return virtual timestamp during query execution. This timestamp can be compared with CDC virtual timestamp to determinate whether data written via SQL (or read) already present in async replica or not.

Case 1 - non interactive tx with commit_tx flag.
In this case we need to set with_tx_timestamp flag in the TransactionControl message. After tx execution we got commit_timestamp in the ExecuteQueryResponsePart message. This timestamp corresponds to the moment of commit (write) in the database. If this tx is read only than snapshot_timestamp in the TransactionMeta sets to the MVCC snapshot timestamp. 

Case 2 - interactive tx. snapshot_timestamp is set to MVCC snapshot timestamp always. CommitTransactionResponse contains commit_timestamp.

ydb/public/api/protos/ydb_query.proto
```diff
@@ -82,6 +83,9 @@ message TransactionControl {
     }
 
     bool commit_tx = 10;
+
+    // If set the timestamp will be returned along with ExecuteQueryResponsePart
+    bool with_tx_timestamp = 11;
```
spuchin, line 88: with_tx_timestamp(s)?

ydb/public/api/protos/ydb_query.proto
```diff
@@ -200,6 +209,9 @@ message ExecuteQueryResponsePart {
     Ydb.TableStats.QueryStats exec_stats = 5;
 
     TransactionMeta tx_meta = 6;
+
+    // Optional timestamp that corresponds to the commit transaction
+    VirtualTimestamp commit_timestamp = 7;
```
spuchin, line 214: На ролбеке же нет коммита, зачем тут таймстамп?

dcherednik: Это ExecuteQueryResponsePart. ui GitHub такой ui)

ydb/public/api/protos/ydb_query.proto
```diff
@@ -112,6 +118,9 @@ message CommitTransactionRequest {
 message CommitTransactionResponse {
     StatusIds.StatusCode status = 1;
     repeated Ydb.Issue.IssueMessage issues = 2;
+
+    // Optional timestamp that corresponds to the commit transaction
+    VirtualTimestamp commit_timestamp = 3;
```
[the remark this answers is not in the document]
dcherednik: Там есть. 
image

dcherednik:
There is a possibility to return a virtual timestamp during query execution. This timestamp represents YDB coordination time and can be compared with the CDC virtual timestamp to check whether data read or written in a transaction is present in the CDC async replica.

There are two types of virtual timestamps in the public API:

    snapshot_timestamp – Represents the moment in time when YDB takes an MVCC or persistent snapshot. It is empty if the request is executed without a snapshot. This is mostly relevant for read-only transactions and is returned during the first part of the result set. The practical purpose is to implement synchronization between readers in transactions and readers from replicas.

    commit_timestamp – Represents the moment in time when YDB commits data into persistent storage. This is mostly relevant for read-write and write-only transactions and is returned in the last part of the result set. The practical purpose is to implement synchronization between transactional writers and readers from replicas.

Returning these timestamps for each query may cause noticeable overhead in case of tiny result. To use them, the with_tx_timestamp flag must be set in the TransactionControl message.

For explicitly created transactions, these timestamps are always returned and do not require the flag.

ydb/docs/en/core/concepts/cdc.md
```diff
@@ -38,6 +38,20 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
 
 {% endnote %}
 
+## Getting virtual timstamps during query execution
+
+As was writen above virtual timestamp represents coordination time. There is a possibility to return a virtual timestamp during query execution. It can be used to check whether data read or written in a transaction is present in the CDC async replica.
```
UgnineSirdis, line 43: writen  -> written

UgnineSirdis, line 43: In the previous section coordination time was named a bit differently: `Global coordinator time`.

ydb/docs/en/core/concepts/cdc.md
```diff
@@ -38,6 +38,20 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
...
+
+The public API has two field to represent transaction reletaed virtual timestamp.
```
UgnineSirdis, line 45: The public API has two field to represent transaction reletaed virtual timestamp.

I would use another state:
`The YDB API has two fields to represent transaction related to virtual timestamp.`

ydb/docs/en/core/concepts/cdc.md
```diff
@@ -38,6 +38,20 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
...
+
+1. 'snapshot_timestamp' – Represents the moment in time when YDB takes an MVCC or persistent snapshot. It is empty if the request is executed without a snapshot. This is mostly relevant for read-only transactions and is returned during the first part of the result set. The practical purpose is to implement synchronization between readers in transactions and readers from replicas.
+
+1. 'commit_timestamp' – Represents the moment in time when YDB commits data into persistent storage. This is mostly relevant for read-write and write-only transactions and is returned in the last part of the result set. The practical purpose is to implement synchronization between transactional writers and readers from replicas.
```
UgnineSirdis, line 49: `is returned in the last part` - maybe `is returned with the last part` is better?

ydb/docs/en/core/concepts/cdc.md
```diff
@@ -38,6 +38,20 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
 
 {% endnote %}
 
+## Getting virtual timstamps during query execution
+
+As was writen above virtual timestamp represents coordination time. There is a possibility to return a virtual timestamp during query execution. It can be used to check whether data read or written in a transaction is present in the CDC async replica.
+
+The public API has two field to represent transaction reletaed virtual timestamp.
+
+1. 'snapshot_timestamp' – Represents the moment in time when YDB takes an MVCC or persistent snapshot. It is empty if the request is executed without a snapshot. This is mostly relevant for read-only transactions and is returned during the first part of the result set. The practical purpose is to implement synchronization between readers in transactions and readers from replicas.
```
UgnineSirdis, line 47: `is returned during the first part` - maybe `is returned with the first part` is better?

ydb/public/api/protos/ydb_query.proto
dcherednik: In case of write only non interactive TX only commit_timestamp will be set. So in any case the naming will be a bit unclear from plural/singular perspective

ydb/docs/en/core/concepts/cdc.md
```diff
@@ -38,6 +38,26 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
 
 {% endnote %}
 
+## Getting virtual timstamps during query execution
+
+As was written above virtual timestamp represents Global coordinator time. There is a possibility to return a virtual timestamp during query execution. It can be used to check whether data read or written in a transaction is present in the CDC async replica.
```
blinkov, line 43: This kind of content should go to reference documentation about SDK/CLI/etc., not concepts. Make sure to include bidirectional links with the `#virtual-timestamps` section above though.

alexv-smirnov: Looks for me like it describes server behaviour, not specific to the interface (SDK/CLI/etc), so Concepts is a proper place for it?

ydb/docs/en/core/concepts/cdc.md
```diff
@@ -38,6 +38,26 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
...
+
+The YDB API has two fields to get virtual timestamp from transaction.
+
+1. 'snapshot_timestamp' – Represents the moment in time when YDB takes an MVCC or persistent snapshot. It is empty if the request is executed without a snapshot. This is mostly relevant for read-only transactions and is returned with the first part of the stream result. The practical purpose is to implement synchronization between readers in transactions and readers from replicas.
+
+1. 'commit_timestamp' – Represents the moment in time when YDB commits data into persistent storage. This is mostly relevant for read-write and write-only transactions and is returned with the last part of the stream result. The practical purpose is to implement synchronization between transactional writers and readers from replicas.
+
+Returning these timestamps for each query may cause noticeable overhead in case of tiny result. To use them, the 'with_tx_timestamp' flag must be set in the 'TransactionControl' message.
+
+For explicitly created transactions (transactions created via BeginTransactionRequest rpc call), these timestamps are always returned and do not require the flag.
```
fomichev3000, line 53: Too complicated logic AFAIK. Let's make it simple -- set with_tx_timestamp to obtain snapshot_timestamp/commit_timestamp

ydb/docs/en/core/concepts/cdc.md
```diff
@@ -38,6 +38,26 @@ By default, virtual timestamps are not uploaded to the changefeed. To enable the
 
 {% endnote %}
 
+## Getting virtual timstamps during query execution
```
fomichev3000, line 41: The whole idea of returning virtual timestamps is not about CDC, sorry. You can use it without CDC. CDC is just a good example of this technology.
I can't find a good proper place for this information... But https://ydb.tech/docs/en/concepts/transactions (actually, YDB transactions and _queries_) looks like a better place to introduce virtual timestamps and to return them on query response

fomichev3000 (approved):

fomichev3000 (changes requested):
Sorry, I wanted to request changes.

ydb/public/api/protos/ydb_query.proto
```diff
@@ -93,6 +97,8 @@ message BeginTransactionRequest {
 message TransactionMeta {
     // Transaction identifier
     string id = 1 [(Ydb.length).le = 1024];
+    // Optional timestamp that represents snapshot acquired to start transaction
```
snaury, line 100:
```suggestion
    // Optional timestamp that represents snapshot acquired at the start of the transaction
```

ydb/public/api/protos/ydb_query.proto
```diff
@@ -112,6 +118,9 @@ message CommitTransactionRequest {
 message CommitTransactionResponse {
     StatusIds.StatusCode status = 1;
     repeated Ydb.Issue.IssueMessage issues = 2;
+
+    // Optional timestamp that corresponds to the commit transaction
```
snaury, line 122:
```suggestion
    // Optional timestamp that corresponds to the commit of the transaction
```

ydb/public/api/protos/ydb_query.proto
```diff
@@ -200,6 +209,9 @@ message ExecuteQueryResponsePart {
     Ydb.TableStats.QueryStats exec_stats = 5;
 
     TransactionMeta tx_meta = 6;
+
+    // Optional timestamp that corresponds to the commit transaction
```
snaury, line 213:
```suggestion
    // Optional timestamp that corresponds to the commit of the transaction
```

ydb/public/api/protos/ydb_query.proto
snaury: Я не помню мы это обсуждали или нет? Не очень понятно зачем отдельная опция на уровне отдельного запроса, почему не возвращать эти timestamp'ы всегда?

dcherednik:
Так было опасение что возврат таймстампа на каждый ответ в случае запросов вида SELECT `int_type_field` FROM table WHERE key=x; приведет к заметному оверхеду.

snaury:
> Так было опасение что возврат таймстампа на каждый ответ в случае запросов вида SELECT `int_type_field` FROM table WHERE key=x; приведет к заметному оверхеду.

К оверхеду на что конкретно? Типа два числа в ответе это будет заметный оверхед? На сервере мне кажется if'чик проверять будет дороже чем просто эти числа всегда передавать. Тут главное конечно не брать снепшот, если это не нужно, и в случае immediate чтений прокидывать этот timestamp с шардов. Шарды знают из какой версии они читали. Или мы это где-то обсудили и решили что это типа дорого/сложно, и вы хотите с опцией `with_tx_timestamp` всегда брать снепшот?

ydb/docs/en/core/concepts/cdc.md
alexv-smirnov: Actually, it's what is written here. There is one flag to return timestamps for a request. Depending on a request, there can be first, second, or both returned.

PR #14813, opened 2025-02-19
Ingore basic scheme limits for internal operations

### LINKED ISSUE #14772: Operations issued internally on 'ydb import' should be able to complete despite breaking scheme limits
Imports are long-running operations and the state of the database:
- the total number of paths in the database
- the total number of data shards in the database

might change due to the concurrent usage of the database that is being restored.

We don't want to forbid the concurrent usage of the database that is being restored, because an import can take days to finish. However, the usage of the database:
- creating new scheme objects
- increasing the load on some tables or indices

might lead to a breakage of scheme limits of the database:
- path limit
- shard limit

Once the scheme limits are broken, the import operation itself might be stopped. We want the import operation to be able to finish even if the scheme limits are exhausted.

### Example

The total number of data shards in the database A before the backup was 19'800:
- big table A: 10'000 shards
- big table B: 9'000 shards
- small table C: 800 shards

The database A is imported into a clean database B.

The limit on the number of shards in the database A and B is 20'000.

The import takes some time to finish. Let's imagine we have successfully imported only the table A and are importing the table B right now. While the import of the table B is running, the load on the table A increases and it splits, adding 300 shards to itself. The total number of shards becomes: 10'000 + 300 + 9'000 = 19'300.

Then the import operation starts importing the table C. However, it is unable to import the table, because the limit on the number of shards in the database B is 20'000.

jepett0:
### Changelog category 

### Description for reviewers 

Hacky solution for inability to import a database from a backup that was successfully created. Shard limit might be broken on import. Probably due to the index building. We decided to ignore the shard limits if the operation is a child of the import operation (i.e. is internal).

- https://github.com/ydb-platform/ydb/issues/14772

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard_build_index__create.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_build_index__create.cpp
@@ -119,7 +119,7 @@ public:
                     .PathsLimit(2) // index and impl-table
                     .DirChildrenLimit();
 
-                if (!request.GetOperationParams().labels().contains("internal")) {
+                if (bool isInternal = Self->TxIdToImport.contains(TTxId(request.GetTxId())); !isInternal) {
                     checks
                         .ShardsLimit(1); // impl-table
                 }
--- a/ydb/core/tx/schemeshard/schemeshard_build_index__create.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_build_index__create.cpp
@@ -115,10 +115,14 @@ public:
                 }
 
                 checks
-                    .IsValidLeafName()
-                    .PathsLimit(2) // index and impl-table
-                    .DirChildrenLimit()
-                    .ShardsLimit(1); // impl-table
+                    .IsValidLeafName();
+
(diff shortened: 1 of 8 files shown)


ydb/core/tx/schemeshard/schemeshard_build_index__create.cpp
```diff
@@ -117,8 +117,12 @@ class TSchemeShard::TIndexBuilder::TTxCreate: public TSchemeShard::TIndexBuilder
                 checks
                     .IsValidLeafName()
                     .PathsLimit(2) // index and impl-table
-                    .DirChildrenLimit()
-                    .ShardsLimit(1); // impl-table
+                    .DirChildrenLimit();
+
+                if (bool isInternal = Self->TxIdToImport.contains(TTxId(request.GetTxId())); !isInternal) {
```
CyberROFL, line 122: Мне кажется, иметь явный флажок в сообщении лучше, чем как-то косвенно пытаться понять.

CyberROFL (approved):

PR #15117, opened 2025-02-27
oidc add poller actor

StekPerepolnen:
### Changelog entry 

after the commit https://github.com/ydb-platform/ydb/pull/13316, a poller actor is needed in oidc_proxy and meta. Without this, the build hangs and does not work.

### Changelog category 

* Bugfix


UgnineSirdis:
We discussed that the task will be divided into two parts:
1. Fix itself. This PR.
2. Test whose absense caused the current problem that is fixed with this PR: in that test we setup real server that serves by http interface (not only actor one) and call simple command like ping. Setup of oidc proxy must be reusable and customizable in the future, the test will use this setup with several lines of code.

UgnineSirdis (approved):

PR #15250, opened 2025-03-03
Topic data handler in viewer

FloatingCrowbar:
### Changelog entry 

Get topic data handler for UI.

### Changelog category 

* New feature

### Description for reviewers 

...

### DIFF
--- a/ydb/core/viewer/json_pipe_req.cpp
+++ b/ydb/core/viewer/json_pipe_req.cpp
@@ -556,47 +556,28 @@ void TViewerPipeClient::RequestBSControllerPDiskUpdateStatus(const NKikimrBlobSt
 }
 
 THolder<NSchemeCache::TSchemeCacheNavigate> TViewerPipeClient::SchemeCacheNavigateRequestBuilder (
-        std::function<void (NSchemeCache::TSchemeCacheNavigate::TEntry&)> fillRequestEntryFunc
+        NSchemeCache::TSchemeCacheNavigate::TEntry&& entry
 ) {
     THolder<NSchemeCache::TSchemeCacheNavigate> request = MakeHolder<NSchemeCache::TSchemeCacheNavigate>();
-    NSchemeCache::TSchemeCacheNavigate::TEntry entry;
     entry.RedirectRequired = false;
     entry.Operation = NSchemeCache::TSchemeCacheNavigate::EOp::OpPath;
-    fillRequestEntryFunc(entry);
-    request->ResultSet.emplace_back(entry);
+    request->ResultSet.emplace_back(std::move(entry));
     return request;
 }
 
-void TViewerPipeClient::RequestSchemeCacheNavigateWithParams(const TString& path, ui32 access, bool showPrivate) {
-    auto request = SchemeCacheNavigateRequestBuilder(
-        [&](auto& entry) {
-            entry.Path = SplitPath(path);
-            entry.Access = access;
-            entry.ShowPrivatePath = showPrivate;
-        }
-    );
-    if (!Event->Get()->UserToken.empty())
-        request->UserToken = new NACLib::TUserToken(Event->Get()->UserToken);
-
-    SendRequest(MakeSchemeCacheID(), new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()));
-}
-
 void TViewerPipeClient::RequestSchemeCacheNavigate(const TString& path) {
-    auto request = SchemeCacheNavigateRequestBuilder(
-        [&](auto& entry) {
-            entry.Path = SplitPath(path);
-        }
-    );
+    NSchemeCache::TSchemeCacheNavigate::TEntry entry;
+    entry.Path = SplitPath(path);
+
+    auto request = SchemeCacheNavigateRequestBuilder(std::move(entry));
     SendRequest(MakeSchemeCacheID(), new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()));
 }
 
 void TViewerPipeClient::RequestSchemeCacheNavigate(const TPathId& pathId) {
-    auto request = SchemeCacheNavigateRequestBuilder(
-        [&](auto& entry) {
-            entry.TableId.PathId = pathId;
-            entry.RequestType = NSchemeCache::TSchemeCacheNavigate::TEntry::ERequestType::ByTableId;
-        }
-    );
+    NSchemeCache::TSchemeCacheNavigate::TEntry entry;
+    entry.TableId.PathId = pathId;
+    entry.RequestType = NSchemeCache::TSchemeCacheNavigate::TEntry::ERequestType::ByTableId;
+    auto request = SchemeCacheNavigateRequestBuilder(std::move(entry));
     SendRequest(MakeSchemeCacheID(), new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()));
 }
 
@@ -639,6 +620,25 @@ TViewerPipeClient::TRequestResponse<NSchemeShard::TEvSchemeShard::TEvDescribeSch
     if (response.Span) {
         response.Span.Attribute("path", path);
     }
+    return response;
+}
+
+TViewerPipeClient::TRequestResponse<TEvTxProxySchemeCache::TEvNavigateKeySetResult> TViewerPipeClient::MakeRequestSchemeCacheNavigateWithToken(
+        const TString& path, bool showPrivate, ui32 access, ui64 cookie
+) {
+    NSchemeCache::TSchemeCacheNavigate::TEntry entry;
+    entry.Path = SplitPath(path);
+    entry.ShowPrivatePath = showPrivate;
+    entry.Access = access;
+    auto request = SchemeCacheNavigateRequestBuilder(std::move(entry));
+
+    if (!Event->Get()->UserToken.empty())
+         request->UserToken = new NACLib::TUserToken(Event->Get()->UserToken);
+
+    auto response = MakeRequest<TEvTxProxySchemeCache::TEvNavigateKeySetResult>(
+            MakeSchemeCacheID(),
+            new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()), 0 /*flags*/, cookie
+    );
     return response;
 }
 
--- a/ydb/core/viewer/json_pipe_req.cpp
+++ b/ydb/core/viewer/json_pipe_req.cpp
@@ -567,7 +567,7 @@ THolder<NSchemeCache::TSchemeCacheNavigate> TViewerPipeClient::SchemeCacheNaviga
     return request;
 }
 
-void TViewerPipeClient::RequestSchemeCacheNavigateWtihParams(const TString& path, ui32 access, bool showPrivate) {
+void TViewerPipeClient::RequestSchemeCacheNavigateWithParams(const TString& path, ui32 access, bool showPrivate) {
     auto request = SchemeCacheNavigateRequestBuilder(
         [&](auto& entry) {
             entry.Path = SplitPath(path);
--- a/ydb/core/viewer/json_pipe_req.cpp
+++ b/ydb/core/viewer/json_pipe_req.cpp
@@ -555,24 +555,47 @@ void TViewerPipeClient::RequestBSControllerPDiskUpdateStatus(const NKikimrBlobSt
     SendRequestToPipe(pipeClient, request.Release());
 }
 
-void TViewerPipeClient::RequestSchemeCacheNavigate(const TString& path) {
+THolder<NSchemeCache::TSchemeCacheNavigate> TViewerPipeClient::SchemeCacheNavigateRequestBuilder (
+        std::function<void (NSchemeCache::TSchemeCacheNavigate::TEntry&)> fillRequestEntryFunc
+) {
     THolder<NSchemeCache::TSchemeCacheNavigate> request = MakeHolder<NSchemeCache::TSchemeCacheNavigate>();
     NSchemeCache::TSchemeCacheNavigate::TEntry entry;
-    entry.Path = SplitPath(path);
     entry.RedirectRequired = false;
     entry.Operation = NSchemeCache::TSchemeCacheNavigate::EOp::OpPath;
+    fillRequestEntryFunc(entry);
     request->ResultSet.emplace_back(entry);
+    return request;
+}
+
+void TViewerPipeClient::RequestSchemeCacheNavigateWtihAclCheck(const TString& path, ui32 access) {
+    auto request = SchemeCacheNavigateRequestBuilder(
+        [&](auto& entry) {
+            entry.Path = SplitPath(path);
+            entry.Access = access;
+        }
+    );
+    if (!Event->Get()->UserToken.empty())
+        request->UserToken = new NACLib::TUserToken(Event->Get()->UserToken);
+
+    SendRequest(MakeSchemeCacheID(), new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()));
+}
+
+void TViewerPipeClient::RequestSchemeCacheNavigate(const TString& path) {
+    auto request = SchemeCacheNavigateRequestBuilder(
+        [&](auto& entry) {
+            entry.Path = SplitPath(path);
+        }
+    );
     SendRequest(MakeSchemeCacheID(), new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()));
 }
 
 void TViewerPipeClient::RequestSchemeCacheNavigate(const TPathId& pathId) {
-    THolder<NSchemeCache::TSchemeCacheNavigate> request = MakeHolder<NSchemeCache::TSchemeCacheNavigate>();
-    NSchemeCache::TSchemeCacheNavigate::TEntry entry;
-    entry.TableId.PathId = pathId;
-    entry.RequestType = NSchemeCache::TSchemeCacheNavigate::TEntry::ERequestType::ByTableId;
-    entry.RedirectRequired = false;
-    entry.Operation = NSchemeCache::TSchemeCacheNavigate::EOp::OpPath;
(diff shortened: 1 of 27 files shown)


ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+        auto setData = [&](NJson::TJsonValue& jsonRecord, TString&& data) {
+            jsonRecord.InsertValue("OriginalSize", data.size());
+            if (data.size() > MaxMessageSize) {
+                data.resize(MaxMessageSize);
+            }
+            totalSize += data.size();
+            jsonRecord.InsertValue("Message", std::move(data));
+        };
+        fullJsonResponse.InsertValue("StartOffset", cmdRead.GetStartOffset());
```
adameat, line 39: do you know that you can use jsonRecord["StartOffset"] = ..   ?

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+    void Bootstrap() {
+        DoWork();
+    }
+
+private:
+    void DoWork() {
+        ui64 totalSize = 0;
+        const auto& response = ReadResponse->Record.GetPartitionResponse();
+        Y_ABORT_UNLESS (response.HasCmdReadResult());
```
adameat, line 25: please no abort() here

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+        switch (codec) {
+            case NPersQueueCommon::GZIP: {
+                auto [iterator, ins] = Codecs.emplace(codecId, MakeHolder<NYdb::NTopic::TGzipCodec>());
+                Y_ABORT_UNLESS(ins);
+                return iterator->second.Get();
+                break;
+            }
+            case NPersQueueCommon::ZSTD: {
+                auto [iterator, ins] = Codecs.emplace(codecId, MakeHolder<NYdb::NTopic::TZstdCodec>());
+                Y_ABORT_UNLESS(ins);
```
adameat, line 107: abort()

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+    ui64 MaxMessageSize;
+    ui64 MaxTotalSize;
+};
+
+
+void TTopicData::HandleDescribe(TEvTxProxySchemeCache::TEvNavigateKeySetResult::TPtr& ev) {
+    auto ev_ = std::unique_ptr<TEvTxProxySchemeCache::TEvNavigateKeySetResult>(ev->Release().Release());
+    const auto& result = ev_->Request;
+
+    if (!TBase::IsSuccess(ev_)) {
```
adameat, line 127: you wouldn't need it if you used MakeRequests...() semantics

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+        TStringBuilder error;
+        if (result->ResultSet.size() != 0) {
+            switch (result->ResultSet[0].Status) {
```
adameat, line 130: the same about this switch

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+        error << "While trying to find topic: '" << TopicPath << "' got error '" << TBase::GetError(ev_) << "'";
+        return ReplyAndPassAway(GetHTTPINTERNALERROR("text/plain", error));
+    }
+    const auto response = result->ResultSet.front();
+    if (response.Self->Info.GetPathType() != NKikimrSchemeOp::EPathTypePersQueueGroup) {
+        auto error = TStringBuilder() << "No such topic '" << TopicPath << "";
+        return ReplyAndPassAway(GetHTTPBADREQUEST("text/plain", error));
+    }
+    if (Event->Get()->UserToken.empty()) {
+        if (AppData(ActorContext())->EnforceUserTokenRequirement || AppData(ActorContext())->PQConfig.GetRequireCredentialsInNewProtocol()) {
```
adameat, line 165: it has more complex logic than that... see EnforceUserTokenCheckRequirement for example.
I would suggest to find a way to avoid put that kind of logic here.

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+            return ReplyAndPassAway(GetHTTPFORBIDDEN("text/plain", "Unauthenticated access is forbidden, please provide credentials"));
+        }
+    } else {
+        NACLib::TUserToken token(Event->Get()->UserToken);
+        if (!response.SecurityObject->CheckAccess(NACLib::EAccessRights::SelectRow, token)) {
```
adameat, line 170: is there a way to not check access in viewer handler? I think, it should be done outside of viewer code.

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+        if (partitionId == PartitionId) {
+            TabletId = partition.GetTabletId();
+            return SendPQReadRequest();
+        }
+    }
+    ReplyAndPassAway(GetHTTPBADREQUEST("text/plain", "No such partition in topic"));
+}
+
+void TTopicData::SendPQReadRequest() {
+    const auto& ctx = ActorContext();
```
adameat, line 188: please, see how viewer works with tablets in json_pipe_req.cpp
it's also required to support tracing and error handling

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+}
+
+void TTopicData::HandlePQResponse(TEvPersQueue::TEvResponse::TPtr& ev) {
+    ReadResponse = ev->Release();
+    const auto& record = ReadResponse->Record;
+    if (record.GetStatus() ==  NMsgBusProxy::MSTATUS_ERROR) {
+        switch (record.GetErrorCode()) {
+            case ::NPersQueue::NErrorCode::READ_ERROR_TOO_SMALL_OFFSET:
+            case ::NPersQueue::NErrorCode::READ_ERROR_TOO_BIG_OFFSET:
+                return ReplyAndPassAway(GetHTTPBADREQUEST("text/plain", "Bad offset"), record.GetErrorReason());
```
adameat, line 215: is it actually an error? isn't it a reason to return an empty result? I'm not sure, but errors like that look suspicious.

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+                return ReplyAndPassAway(GetHTTPINTERNALERROR("text/plain", "Error trying to read messages"), record.GetErrorReason());
+        }
+        return;
+    }
+    const auto& response = record.GetPartitionResponse();
+
+    if (response.HasCmdReadResult()) {
+        const auto& readResult = response.GetCmdReadResult();
+        if (readResult.GetReadingFinished()) {
+            ReplyAndPassAway(GetHTTPBADREQUEST("text/plain", "Bad partition-id"));
```
adameat, line 227: how do you expect to handle all these errors in UI?

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+            return;
+        }
+    } else {
+        return ReplyAndPassAway(GetHTTPINTERNALERROR("text/plain", "No data received from topic"));
```
adameat, line 231: is it an error?

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+    Register(new TUnpackDataToJsonActor(std::move(ReadResponse), SelfId()),
+                TMailboxType::HTSwap, AppData()->BatchPoolId);
```
adameat, line 234: we don't always have BatchPool in our binary.
and why do you need a separate actor here? you already have an actor processing user request...

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+    if (!GetIntegerParam("offset", Offset))
+        return;
+
+    Limit = FromStringWithDefault<ui32>(params.Get("limit"), 10);
+    if (Limit > MAX_MESSAGES_LIMIT) {
+        return ReplyAndPassAway(Viewer->GetHTTPBADREQUEST(Event->Get(), "text/plain", "Too many messages requested"));
+    }
+
+    if (params.Has("topic_path")) {
+        TopicPath = params.Get("topic_path");
```
adameat, line 291: how about just "path" ?
and it's easier to just read it, without Has() - and check for empty string. you don't want empty path, am I right?

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+        return false;
+    }
+}
+
+void TTopicData::Bootstrap() {
+    if (NeedToRedirect()) {
+        return;
+    }
+    const auto& params(Event->Get()->Request.GetParams());
+    Timeout = FromStringWithDefault<ui32>(params.Get("timeout"), 10000);
```
adameat, line 276: this parameter has already been parsed in the base class

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+    Timeout = std::min(Timeout, 30000u);
+
+
+    if (!GetIntegerParam("partition", PartitionId))
+        return;
+    if (!GetIntegerParam("offset", Offset))
+        return;
+
+    Limit = FromStringWithDefault<ui32>(params.Get("limit"), 10);
```
adameat, line 285: usually we don't imply limit by default ...

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+    Become(&TThis::StateRequestedDescribe, TDuration::MilliSeconds(Timeout), new TEvents::TEvWakeup());
+
+}
+
+void TTopicData::ReplyAndPassAway() {
+    if (!Response.IsDefined()) {
+        return ReplyAndPassAway(GetHTTPINTERNALERROR("text/plain", "Could not get topic data"));
+    }
+    ReplyAndPassAway(GetHTTPOKJSON(std::move(Response)));
```
adameat, line 304: you don't need std::move() here

ydb/core/viewer/viewer_topic_data.h
```diff
@@ -0,0 +1,171 @@
...
+    static constexpr ui32 READ_TIMEOUT_MS = 1000;
+    static constexpr ui32 MAX_MESSAGES_LIMIT = 1000;
+
+public:
+    static YAML::Node GetSwagger() {
+        YAML::Node node = YAML::Load(R"___(
+        get:
+            tags:
+              - viewer
+            summary: ACL information
```
adameat, line 77: really?

ydb/core/viewer/viewer_topic_data.h
```diff
@@ -0,0 +1,171 @@
...
+                in: query
+                description: partition to read from
+                required: true
+                type: integer
+              - name: offset
+                in: query
+                description: start offset to read from
+                required: true
+                type: integer
+              - name: limit
```
adameat, line 100: if I asked "offset=10 limit=10". and then I asked "offset=20 limit=10" - would I start right after the previous chunk of data?

ydb/core/viewer/viewer_topic_data.h
```diff
@@ -0,0 +1,171 @@
...
+                                        items:
+                                            type: object
+                                            properties:
+                                                Offset:
+                                                    type: integer
+                                                CreateTimestamp:
+                                                    type: integer
+                                                WriteTimestamp:
+                                                    type: integer
+                                                Timestamp Diff:
```
adameat, line 133: really? Timestamp Diff ?

maybe it would be easier to describe response type as a proto message in viewer.proto and use it reflection to make swagger description? and use it to make actuall response with automatic proto-json conversion?
 as in /viewer/nodes for example...

adameat (changes requested):

ydb/core/viewer/viewer_topic_data.h
FloatingCrowbar: Yes

ydb/core/viewer/json_pipe_req.cpp
```diff
@@ -544,24 +544,48 @@ void TViewerPipeClient::RequestBSControllerPDiskUpdateStatus(const NKikimrBlobSt
...
+    );
     SendRequest(MakeSchemeCacheID(), new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()));
 }
 
 void TViewerPipeClient::RequestSchemeCacheNavigate(const TPathId& pathId) {
-    THolder<NSchemeCache::TSchemeCacheNavigate> request = MakeHolder<NSchemeCache::TSchemeCacheNavigate>();
-    NSchemeCache::TSchemeCacheNavigate::TEntry entry;
-    entry.TableId.PathId = pathId;
-    entry.RequestType = NSchemeCache::TSchemeCacheNavigate::TEntry::ERequestType::ByTableId;
-    entry.RedirectRequired = false;
```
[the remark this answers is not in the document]
FloatingCrowbar: Just moved it to RequestBuilder function (because it has the same value in every invocation anyway).

ydb/core/viewer/viewer_topic_data.cpp
FloatingCrowbar: We discussed that it may be reasonable for topic since requiring the client to always specify it explicitly is not good for UX and reading topic with no limit would mean large amount of data is transfered every time this is called.

ydb/core/viewer/viewer_topic_data.h
FloatingCrowbar: That's what was requested in ticket :)
I refactored everything to use a proto anyways.

alexnick88 (dismissed):

ydb/core/viewer/json_pipe_req.cpp
```diff
@@ -618,6 +623,25 @@ TViewerPipeClient::TRequestResponse<NSchemeShard::TEvSchemeShard::TEvDescribeSch
...
+    entry.Access = access;
+    auto request = SchemeCacheNavigateRequestBuilder(std::move(entry));
+
+    if (!Event->Get()->UserToken.empty())
+         request->UserToken = new NACLib::TUserToken(Event->Get()->UserToken);
+
+    auto response = MakeRequest<TEvTxProxySchemeCache::TEvNavigateKeySetResult>(
+            MakeSchemeCacheID(),
+            new TEvTxProxySchemeCache::TEvNavigateKeySet(request.Release()), 0 /*flags*/, cookie
+    );
```
adameat, line 641: it would be nice to add at least path to Span.Attributes - good for debugging

ydb/core/viewer/protos/viewer.proto
```diff
@@ -837,3 +837,28 @@ message TFeatureFlagsConfig {
     repeated TDatabase Databases = 2;
 }
 
+message TTopicDataResponse {
+    message TMessage {
+        uint64 Offset = 1;
+        uint64 CreateTimestamp = 2;
+        uint64 WriteTimestamp = 3;
+        uint64 TimestampDiff = 4;
+        string Message = 5;
```
adameat, line 846: string or bytes?
if you put string here you imply that it's a valid utf-8 text string.

adameat (changes requested):

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+        auto* messageProto = ProtoResponse.AddMessages();
+        messageProto->SetOffset(r.GetOffset());
+        messageProto->SetCreateTimestamp(r.GetCreateTimestampMS());
+        messageProto->SetWriteTimestamp(r.GetWriteTimestampMS());
+        i64 diff = r.GetWriteTimestampMS() - r.GetCreateTimestampMS();
+        if (diff < 0) {
+            diff = 0;
+        }
+        messageProto->SetTimestampDiff(diff);
+        messageProto->SetSize(dataChunk.GetData().size());
```
adameat, line 168: is it actual size for compressed data?

ydb/core/viewer/viewer_topic_data.cpp
adameat: then it should be described in swagger description (usually, it's not an expected behavior)
and it's expected to have a "truncated" flag if limit is implicitly applied.

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+    }
+    const auto& cmdRead = response.GetCmdReadResult();
+
+    auto setData = [&](NKikimrViewer::TTopicDataResponse::TMessage& protoMessage, TString&& data) {
+        protoMessage.SetOriginalSize(data.size());
+        if (data.size() > maxSingleMessageSize) {
+            data.resize(maxSingleMessageSize);
+        }
+        totalSize += data.size();
+        if (EncodeMessageData)
```
adameat, line 146: it's not like you have an option: you can't put an arbitrary (non-utf8 conformant) string - it breaks json specification

I suggest to just avoid this option and always encode it into base64. (can proto2json do it automatically?)

adameat (dismissed):

ydb/core/viewer/viewer_topic_data.cpp
FloatingCrowbar: Yes. OriginalSize is used for uncompressed data size.

adameat (approved):

alexnick88 (approved):

PR #15337, opened 2025-03-04
Trunсate configs

mregrock:
### Changelog entry 

Truncate unnecessary fields for configs

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/deploy/yaml_config_examples/block-4-2.yaml
+++ b/ydb/deploy/yaml_config_examples/block-4-2.yaml
@@ -77,8 +77,6 @@ config:
       data_center: 'zone-a'
       rack: '8'
   interconnect_config:
-      start_tcp: true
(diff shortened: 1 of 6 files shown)


blinkov (changes requested):
These configs can be trimmed down even more aggressively:

1.  `body` is not really necessary
2.  no need to specify drive type if it's equal to `default_disk_type`
3. `host_config_id` thing is not really necessary
4. TLS certificates can be specified only once
5. `SectorMap:1:64` can be replaced with `RAM` disk type alias
6. `self_management_config` shouldn't be necessary unless it is a migration, same for `yaml_config_enabled`

Afaik, this is the current smallest working config: https://github.com/ydb-platform/ydb/blob/main/ydb/library/yaml_config/ut_transform/simplified_configs/mirror-3dc-3-nodes-distconf.yaml

PR #15526, opened 2025-03-10
Remove formating in config fetch

mregrock:
### Changelog entry 

Remove formating in config fetch

### Changelog category 

### Description for reviewers 

...


ydb/public/lib/ydb_cli/commands/ydb_dynamic_config.cpp
```diff
@@ -129,7 +129,7 @@ int TCommandConfigFetch::Run(TConfig& config) {
         }
 
         if (!OutDir) {
-            Cout << WrapYaml(cfg);
+            Cout << cfg << "\n";
```
pnv1, line 132: Use `Endl` instead of `"\n"`

pnv1 (approved):

PR #15714, opened 2025-03-13
Replace for dual config and dry-run support

mregrock:
### Changelog entry 

Replace for dual config and dry-run support

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/public/lib/ydb_cli/commands/ydb_dynamic_config.cpp
+++ b/ydb/public/lib/ydb_cli/commands/ydb_dynamic_config.cpp
@@ -235,6 +235,15 @@ void TCommandConfigReplace::Config(TConfig& config) {
         .StoreTrue(&AllowUnknownFields);
     config.Opts->AddLongOption("force", "Ignore metadata on config replacement")
         .StoreTrue(&Force);
+    config.Opts->AddLongOption("dedicated-cluster-yaml", "Path to dedicated cluster section of configuration")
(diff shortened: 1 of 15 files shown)


Enjection (dismissed):

ydb/public/lib/ydb_cli/commands/ydb_dynamic_config.cpp
```diff
@@ -242,7 +270,17 @@ int TCommandConfigReplace::Run(TConfig& config) {
         settings.AllowUnknownFields();
     }
 
-    auto status = client.ReplaceConfig(DynamicConfig, settings).GetValueSync();
+    auto status = [&]() {
+        if (SwitchDedicatedStorageSection && !*SwitchDedicatedStorageSection) {
+            return client.ReplaceConfigDisableDedicatedStorageSection(ClusterYaml.value(), settings).GetValueSync();
+        } else if (SwitchDedicatedStorageSection && *SwitchDedicatedStorageSection) {
+            return client.ReplaceConfigEnableDedicatedStorageSection(ClusterYaml.value(), StorageYaml.value(), settings).GetValueSync();
```
alexvru, line 277: What if we don't have some of them set? We should generate some message better than "bad optional access".

alexvru (changes requested):

alexvru (dismissed):

PR #15728, opened 2025-03-14
YDB CLI options parsing test

UgnineSirdis:
### Changelog entry 

YDB CLI options parsing test

### Changelog category 

### Description for reviewers 

YDB CLI options parsing test
Test before options refactoring

### DIFF
--- a/ydb/apps/ydb/ut/parse_command_line.cpp
+++ b/ydb/apps/ydb/ut/parse_command_line.cpp
@@ -279,7 +279,7 @@ Y_UNIT_TEST_SUITE(ParseOptionsTest) {
                 "-v",
                 "-e", GetEndpoint(),
                 "-d", GetDatabase(),
-                "scheme", "ls", GetDatabase(),
+                "scheme", "ls",
             }
         );
     }
@@ -299,7 +299,7 @@ Y_UNIT_TEST_SUITE(ParseOptionsTest) {
         RunCli(
             {
                 "-v",
-                "scheme", "ls", GetDatabase(),
+                "scheme", "ls",
             },
             {
                 {"YDB_TOKEN", "42"}
@@ -323,7 +323,7 @@ Y_UNIT_TEST_SUITE(ParseOptionsTest) {
             {
                 "-v",
                 "-p", "test_profile",
-                "scheme", "ls", GetDatabase(),
+                "scheme", "ls",
             },
             {
                 {"YDB_TOKEN", "42"}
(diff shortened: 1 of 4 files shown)


ydb/apps/ydb/ut/parse_command_line.cpp
```diff
@@ -0,0 +1,748 @@
...
+        )yaml",
+        "endpoint"_a = GetEndpoint(),
+        "database"_a = GetDatabase()
+        );
+
+        TString tokenFile = EnvFile("iam_token", "token");
+        ExpectToken("iam_token");
+        RunCli({
+            "-v",
+            "-e", GetEndpoint(),
```
pnv1, line 360: You can even add `--profile other_test_profile` here to make sure it is not used anyway

ydb/apps/ydb/ut/parse_command_line.cpp
```diff
@@ -0,0 +1,748 @@
...
+};
+
+Y_UNIT_TEST_SUITE(ParseOptionsTest) {
+    Y_UNIT_TEST_F(EndpointAndDatabaseFromCommandLine, TCliTestFixture) {
+        RunCli(
+            {
+                "-v",
+                "-e", GetEndpoint(),
+                "-d", GetDatabase(),
+                "scheme", "ls", GetDatabase(),
```
pnv1, line 282: Database is actually not needed. It is a default path for `scheme ls`

ydb/apps/ydb/ut/parse_command_line.cpp
```diff
@@ -0,0 +1,748 @@
...
+            {"YDB_PASSWORD", "pwd"},
+        });
+
+        ExpectUserAndPassword("test-user", "test-password", "not-empty-pwd-token");
+        RunCli({
+            "-v",
+            "-e", GetEndpoint(),
+            "-d", GetDatabase(),
+            "--user", "test-user",
+            "--no-password",
```
pnv1, line 522: Seems like `--no-password` shouldn't be used with `--password-file`. They should be mutually exclusive

ydb/apps/ydb/ut/parse_command_line.cpp
```diff
@@ -0,0 +1,748 @@
...
+                        password-file: {password_file}
+            test_profile_with_both_passwords:
+                endpoint: {endpoint}
+                database: {database}
+                authentication:
+                    method: static-credentials
+                    data:
+                        user: users
+                        password: pwd
+                        password-file: {password_file}
```
pnv1, line 561: This situation is not OK too, better make such situation with two password options not possible...

ydb/apps/ydb/ut/parse_command_line.cpp
UgnineSirdis: Yes, it makes sense, thank you! I will fix it in the next PR (it is already done)

ydb/apps/ydb/ut/parse_command_line.cpp
UgnineSirdis: I agree with you. Normally ydb cli does not write both passwords in profile, so they can be easily validated in such a way. I can fix it in next PR (the code is already written). So now it is commented, but soon it will be uncommented :)

pnv1 (approved):

PR #15899, opened 2025-03-18
Inflight limit for ReadRows to prevent dynnodes OOM

SammyVimes:
The limit is required to prevent dynnodes crashes by OOM when the network capacity is not enough to transport requested data to the client via ReadRows gRPC API.

The limit is configurable, default value is 0 (unlimited) for backward compatibility.

### Changelog category 

* Bugfix


dcherednik (dismissed):
GRPC part looks fine for me

dcherednik (approved):

PR #15900, opened 2025-03-18
use sparsed instead of mono-plain

ivanmorozov333:
use different bitset storages

* Improvement
* Performance improvement


swalrus1 (dismissed):

ijon (approved):
`ydb/core/protos/flat_scheme_op.proto` ok

swalrus1 (approved):

PR #15948, opened 2025-03-19
Fix config delivery to quoter (#15847)

FloatingCrowbar:
### Changelog entry 

Fix the possible race on quoter config delivery

### Changelog category 

* Bugfix 

### Description for reviewers 

...


maximyurchuk (approved):
Ок, так как вливаем в 24-4 правку

PR #16024, opened 2025-03-20
Refactor import s3 state proto so that it can contain not only checksum state, but other info too

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


ydb/core/tx/datashard/datashard_impl.h
```diff
@@ -790,7 +790,7 @@ class TDataShard
             struct ProcessedBytes :  Column<4, NScheme::NTypeIds::Uint64> {};
             struct WrittenBytes :    Column<5, NScheme::NTypeIds::Uint64> {};
             struct WrittenRows :     Column<6, NScheme::NTypeIds::Uint64> {};
-            struct ChecksumState :   Column<7, NScheme::NTypeIds::String> { using Type = NKikimrBackup::TChecksumState; };
+            struct DownloadState :   Column<7, NScheme::NTypeIds::String> { using Type = NKikimrBackup::TS3DownloadState; };
```
CyberROFL, line 793: Кроме этого всё ок.

PR #16171, opened 2025-03-24
Allow tablets to reboot on event parsing failure

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Replace some Y_ABORTs with Y_ENSUREs in the actor system. This will allow tablets to reboot in very rare cases where event parsing fails (e.g. due to protocol mismatches).

### DIFF
--- a/ydb/library/actors/core/actorid.h
+++ b/ydb/library/actors/core/actorid.h
@@ -49,8 +49,8 @@ namespace NActors {
             Raw.N.NodeId = nodeId | (poolId << PoolIndexShift);
         }
 
-        explicit TActorId(ui32 nodeId, const TStringBuf& x) noexcept {
-            Y_ABORT_UNLESS(x.size() <= MaxServiceIDLength, "service id is too long");
+        explicit TActorId(ui32 nodeId, const TStringBuf& x) {
+            Y_ENSURE(x.size() <= MaxServiceIDLength, "service id is too long");
             Raw.N.LocalId = 0;
             Raw.N.Hint = 0;
             Raw.N.NodeId = nodeId | ServiceMask;
--- a/ydb/library/actors/core/actorsystem.cpp
+++ b/ydb/library/actors/core/actorsystem.cpp
@@ -204,10 +204,8 @@ namespace NActors {
             if (ev->HasBuffer()) {
                 CheckEventMemory(ev->GetChainBuffer());
             }
-            Y_ABORT_UNLESS(ev->Recipient == recipient,
-                "Event rewrite from %s to %s would be lost via interconnect",
-                ev->Recipient.ToString().c_str(),
-                recipient.ToString().c_str());
+            Y_ENSURE(ev->Recipient == recipient,
+                "Event rewrite from " << ev->Recipient << " to " << recipient << " would be lost via interconnect");
             recipient = InterconnectProxy(recpNodeId);
             ev->Rewrite(TEvInterconnect::EvForward, recipient);
         }
--- a/ydb/library/actors/core/event.h
+++ b/ydb/library/actors/core/event.h
@@ -88,8 +88,8 @@ namespace NActors {
 
         template <typename TEventType>
         TEventType* Get() {
-            if (Type != TEventType::EventType)
-                Y_ABORT("Event type %" PRIu32 " doesn't match the expected type %" PRIu32, Type, TEventType::EventType);
+            Y_ENSURE(Type == TEventType::EventType,
+                "Event type " << Type << " doesn't match the expected type " << TEventType::EventType);
 
             if (!Event) {
                 static TEventSerializedData empty;
@@ -100,7 +100,7 @@ namespace NActors {
                 return static_cast<TEventType*>(Event.Get());
             }
 
-            Y_ABORT("Failed to Load() event type %" PRIu32 " class %s", Type, TypeName<TEventType>().data());
+            Y_ENSURE(false, "Failed to Load() event type " << Type << " class " << TypeName<TEventType>());
         }
 
         template <typename T>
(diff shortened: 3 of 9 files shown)


ydb/library/actors/core/actorid.h
```diff
@@ -49,8 +49,8 @@ namespace NActors {
             Raw.N.NodeId = nodeId | (poolId << PoolIndexShift);
         }
 
-        explicit TActorId(ui32 nodeId, const TStringBuf& x) noexcept {
-            Y_ABORT_UNLESS(x.size() <= MaxServiceIDLength, "service id is too long");
+        explicit TActorId(ui32 nodeId, const TStringBuf& x) {
+            Y_ENSURE(x.size() <= MaxServiceIDLength, "service id is too long");
```
alexvru, line 53: When would this help?

ydb/library/actors/core/actorsystem.cpp
```diff
@@ -204,10 +204,8 @@ namespace NActors {
             if (ev->HasBuffer()) {
                 CheckEventMemory(ev->GetChainBuffer());
             }
-            Y_ABORT_UNLESS(ev->Recipient == recipient,
-                "Event rewrite from %s to %s would be lost via interconnect",
-                ev->Recipient.ToString().c_str(),
-                recipient.ToString().c_str());
+            Y_ENSURE(ev->Recipient == recipient,
```
alexvru, line 207: And this?

ydb/library/actors/core/event.h
```diff
@@ -88,8 +88,8 @@ namespace NActors {
 
         template <typename TEventType>
         TEventType* Get() {
-            if (Type != TEventType::EventType)
-                Y_ABORT("Event type %" PRIu32 " doesn't match the expected type %" PRIu32, Type, TEventType::EventType);
+            Y_ENSURE(Type == TEventType::EventType,
```
alexvru, line 91: Doubtful, would mask the problem of event type id shift.

ydb/library/actors/core/event.h
```diff
@@ -154,8 +154,8 @@ namespace NActors {
         }
 
         static ui32 MakeFlags(ui32 channel, TEventFlags flags) {
-            Y_ABORT_UNLESS(channel < (1 << ChannelBits));
-            Y_ABORT_UNLESS(flags < (1 << ChannelShift));
+            Y_ENSURE(channel < (1 << ChannelBits));
```
alexvru, line 157: This too.

ydb/library/actors/core/event_load.h
```diff
@@ -93,7 +93,7 @@ namespace NActors {
         }
 
         TRope EraseBack(size_t count) {
-            Y_ABORT_UNLESS(count <= Rope.GetSize());
+            Y_ENSURE(count <= Rope.GetSize());
```
alexvru, line 96: Not used in deserialization.

ydb/library/actors/core/event_pb.cpp
```diff
@@ -24,7 +24,7 @@ namespace NActors {
     }
 
     void TRopeStream::BackUp(int count) {
-        Y_ABORT_UNLESS(count <= TotalByteCount);
+        Y_ENSURE(count <= TotalByteCount);
```
alexvru, line 27: Not used for deserialization.

ydb/library/actors/core/event_pb.cpp
```diff
@@ -218,7 +218,7 @@ namespace NActors {
     }
 
     bool TAllocChunkSerializer::WriteAliasedRaw(const void*, int) {
-        Y_ABORT_UNLESS(false);
+        Y_ENSURE(false);
```
alexvru, line 221: Serializer, not deserializer.

ydb/library/actors/core/event_pb.cpp
```diff
@@ -404,8 +404,8 @@ namespace NActors {
                     total += section.Size;
                 }
                 size_t serialized = CalculateSerializedSizeImpl(payload, recordSize);
-                Y_ABORT_UNLESS(total == serialized, "total# %zu serialized# %zu byteSize# %zd payload.size# %zu", total,
-                    serialized, byteSize, payload.size());
+                Y_ENSURE(total == serialized, "total# " << total << " serialized# " << serialized
```
alexvru, line 407: Serialization path, not deserizaliation.

ydb/library/actors/core/event_pb.h
```diff
@@ -205,7 +205,8 @@ namespace NActors {
         static IEventBase* Load(TEventSerializedData *input) {
             THolder<TEventPBBase> ev(new TEv());
             if (!input->GetSize()) {
-                Y_PROTOBUF_SUPPRESS_NODISCARD ev->Record.ParseFromString(TString());
+                Y_ENSURE(ev->Record.ParseFromString(TString()),
```
alexvru, line 208: This would be strange -- failed deserialization of empty string.

alexvru (changes requested):

ydb/library/actors/core/actorsystem.cpp
snaury: That's why we're checking this invariant and won't sent anything unless it's true.

ydb/library/actors/core/event.h
snaury: It's not masking anything, this is checked in both release and debug builds, but instead of Y_ABORT this would throw an exception, which can be handled e.g. by logging a CRIT and restarting the offending tablet (instead of the process).

ydb/library/actors/core/event_load.h
snaury: But it is used in serialization, e.g. when using `TAllocChunkSerializer`.

ydb/library/actors/core/event_pb.cpp
snaury: Doesn't mean we want processes (as opposed to specific components when possible) to always crash.

ydb/library/actors/core/event_pb.cpp
snaury: It doesn't affect IC (will crash anyway, but in the future we could close the specific connection for example), but will help when serializing in tablets.

ydb/library/actors/core/event_pb.h
snaury: It's not strange when you accidentally set some fields to `required`. We currently ignore it completely, but it's an error.

alexvru (approved):

PR #16282, opened 2025-03-26
Do not seek on small prefixes in prefix kmeans

kungasc:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Also batches uploading rows

Tested on wikipedia with query

```
ALTER TABLE wikipedia
    ADD INDEX prefix_index
    GLOBAL USING vector_kmeans_tree
    ON (wiki_id, embedding)
    WITH (similarity=inner_product, vector_type=float, vector_dimension=768, levels=1, clusters=128);

```

Before: 2h 43m

image

After: 5s

```

2025-03-31T08:35:10.735229Z :BUILD_INDEX NOTICE: Starting TPrefixKMeansScan Id: 281474976915661 TabletId: 72075186224037953 PathId { OwnerId: 72075186224037897 LocalId: 25 } SeqNoGeneration: 10 SeqNoRound: 1 Settings { metric: SIMILARITY_INNER_PRODUCT vector_type: VECTOR_TYPE_FLOAT vector_dimension: 768 } Seed: 72075186224037953 Upload: UPLOAD_BUILD_TO_POSTING K: 128 NeedsRounds: 3 Child: 11660617 LevelName: "/Root/testdb/wikipedia/prefix_index2/indexImplLevelTable" PostingName: "/Root/testdb/wikipedia/prefix_index2/indexImplPostingTable" PrefixName: "/Root/testdb/wikipedia/prefix_index2/indexImplPrefixTable" EmbeddingColumn: "embedding" PrefixColumns: 1 row version v1743410110780/18446744073709551615

2025-03-31T08:35:15.634298Z :BUILD_INDEX NOTICE: FinishTPrefixKMeansScan Id: 281474976915661 Parent: 12036061 Child: 12036062 K: 128 Clusters: 0 State: SAMPLE Round: 0 / 3 LevelBuf size: 0 PostingBuf size: 0 PrefixBuf size: 0 UploadTable: /Root/testdb/wikipedia/prefix_index2/indexImplPrefixTable UploadBuf size: 0 Id: 281474976915661 TabletId: 72075186224037953 RequestSeqNoGeneration: 10 RequestSeqNoRound: 1 Status: DONE UploadRows: 861302 UploadBytes: 600810546 ReadRows: 485859 ReadBytes: 1498875015

```

### DIFF
--- a/ydb/core/tx/datashard/build_index.cpp
+++ b/ydb/core/tx/datashard/build_index.cpp
@@ -25,6 +25,7 @@ namespace NKikimr::NDataShard {
 
 #define LOG_N(stream) LOG_NOTICE_S(*TlsActivationContext, NKikimrServices::TX_DATASHARD, stream)
 #define LOG_T(stream) LOG_TRACE_S(*TlsActivationContext, NKikimrServices::TX_DATASHARD, stream)
+#define LOG_I(stream) LOG_INFO_S(*TlsActivationContext, NKikimrServices::TX_DATASHARD, stream)
 #define LOG_D(stream) LOG_DEBUG_S(*TlsActivationContext, NKikimrServices::TX_DATASHARD, stream)
 #define LOG_W(stream) LOG_WARN_S(*TlsActivationContext, NKikimrServices::TX_DATASHARD, stream)
 #define LOG_E(stream) LOG_ERROR_S(*TlsActivationContext, NKikimrServices::TX_DATASHARD, stream)
@@ -185,7 +186,7 @@ public:
     TInitialState Prepare(IDriver* driver, TIntrusiveConstPtr<TScheme>) override {
         TActivationContext::AsActorContext().RegisterWithSameMailbox(this);
 
-        LOG_D("Prepare " << Debug());
+        LOG_I("Prepare " << Debug());
 
         Driver = driver;
 
--- a/ydb/core/tx/datashard/build_index.cpp
+++ b/ydb/core/tx/datashard/build_index.cpp
@@ -255,7 +255,7 @@ public:
 
         UploadStatusToMessage(progress->Record);
 
-        LOG_N("Finished " << Debug() << " " << progress->Record.ShortDebugString());
+        LOG_N("Finish" << Debug() << " " << progress->Record.ShortDebugString());
         this->Send(ProgressActorId, progress.Release());
 
         Driver = nullptr;
--- a/ydb/core/tx/datashard/build_index.cpp
+++ b/ydb/core/tx/datashard/build_index.cpp
@@ -254,9 +254,8 @@ public:
 
         UploadStatusToMessage(progress->Record);
 
+        LOG_N("Finished " << Debug() << " " << progress->Record.ShortDebugString());
         this->Send(ProgressActorId, progress.Release());
-
-        LOG_D("Finish " << Debug());
 
         Driver = nullptr;
         this->PassAway();
@@ -517,6 +516,9 @@ void TDataShard::Handle(TEvDataShard::TEvBuildIndexCreateRequest::TPtr& ev, cons
 void TDataShard::HandleSafe(TEvDataShard::TEvBuildIndexCreateRequest::TPtr& ev, const TActorContext& ctx) {
     const auto& record = ev->Get()->Record;
     TRowVersion rowVersion(record.GetSnapshotStep(), record.GetSnapshotTxId());
+
+    LOG_N("Starting TBuildIndexScan " << record.ShortDebugString()
+        << " row version " << rowVersion);
 
     // Note: it's very unlikely that we have volatile txs before this snapshot
     if (VolatileTxManager.HasVolatileTxsAtSnapshot(rowVersion)) {
--- a/ydb/core/tx/datashard/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/local_kmeans.cpp
@@ -49,7 +49,7 @@ public:
         Size = 0;
         lock.unlock();
 
-        LOG_N("Finished TLocalKMeansScan " << Response->Record.ShortDebugString());
+        LOG_N("FinishTLocalKMeansScan " << Response->Record.ShortDebugString());
         ctx.Send(ResponseActorId, std::move(Response));
     }
 
--- a/ydb/core/tx/datashard/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/local_kmeans.cpp
@@ -48,6 +48,8 @@ public:
         }
         Size = 0;
         lock.unlock();
+
+        LOG_N("Finished TLocalKMeansScan " << Response->Record.ShortDebugString());
         ctx.Send(ResponseActorId, std::move(Response));
     }
 
--- a/ydb/core/tx/datashard/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/local_kmeans.cpp
@@ -469,6 +469,7 @@ public:
         }
     }
 
+private:
     bool InitAggregatedClusters()
     {
         if (Clusters.size() == 0) {
--- a/ydb/core/tx/datashard/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/local_kmeans.cpp
@@ -443,9 +443,17 @@ public:
 
     EScan Feed(TArrayRef<const TCell> key, const TRow& row) final
     {
-        LOG_T("Feed " << Debug());
         ++ReadRows;
         ReadBytes += CountBytes(key, row);
+        
+        return Feed(key, *row);
+    }
+
+private:
+    EScan Feed(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
+    {
+        LOG_T("Feed " << Debug());
+        
         switch (State) {
             case EState::SAMPLE:
                 return FeedSample(row);
@@ -464,7 +472,6 @@ public:
         }
     }
 
-private:
     bool InitAggregatedClusters()
     {
         if (Clusters.size() == 0) {
@@ -544,10 +551,10 @@ private:
         return true;
     }
 
-    EScan FeedSample(const TRow& row)
+    EScan FeedSample(TArrayRef<const TCell> row)
     {
-        Y_ASSERT(row.Size() == 1);
-        const auto embedding = row.Get(0).AsRef();
+        Y_ASSERT(row.size() == 1);
+        const auto embedding = row.at(0).AsRef();
         if (!this->IsExpectedSize(embedding)) {
             return EScan::Feed;
         }
@@ -571,15 +578,15 @@ private:
         return MaxProbability != 0 ? EScan::Feed : EScan::Reset;
     }
 
-    EScan FeedKMeans(const TRow& row)
+    EScan FeedKMeans(TArrayRef<const TCell> row)
     {
-        Y_ASSERT(row.Size() == 1);
+        Y_ASSERT(row.size() == 1);
         const ui32 pos = FeedEmbedding(*this, Clusters, row, 0);
-        AggregateToCluster(pos, row.Get(0).Data());
+        AggregateToCluster(pos, row.at(0).Data());
         return EScan::Feed;
     }
 
-    EScan FeedUploadMain2Build(TArrayRef<const TCell> key, const TRow& row)
+    EScan FeedUploadMain2Build(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
     {
         const ui32 pos = FeedEmbedding(*this, Clusters, row, EmbeddingPos);
         if (pos > K) {
@@ -589,7 +596,7 @@ private:
         return FeedUpload();
     }
 
-    EScan FeedUploadMain2Posting(TArrayRef<const TCell> key, const TRow& row)
+    EScan FeedUploadMain2Posting(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
     {
         const ui32 pos = FeedEmbedding(*this, Clusters, row, EmbeddingPos);
         if (pos > K) {
@@ -599,7 +606,7 @@ private:
         return FeedUpload();
     }
 
-    EScan FeedUploadBuild2Build(TArrayRef<const TCell> key, const TRow& row)
+    EScan FeedUploadBuild2Build(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
     {
         const ui32 pos = FeedEmbedding(*this, Clusters, row, EmbeddingPos);
         if (pos > K) {
@@ -609,7 +616,7 @@ private:
         return FeedUpload();
     }
 
-    EScan FeedUploadBuild2Posting(TArrayRef<const TCell> key, const TRow& row)
+    EScan FeedUploadBuild2Posting(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
     {
         const ui32 pos = FeedEmbedding(*this, Clusters, row, EmbeddingPos);
         if (pos > K) {
--- a/ydb/core/tx/datashard/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/local_kmeans.cpp
@@ -443,18 +443,13 @@ public:
         return EScan::Feed;
     }
 
-    EScan Feed(TArrayRef<const TCell> key, const TRow& row) final
-    {
-        ++ReadRows;
-        ReadBytes += CountBytes(key, row);
-        
-        return Feed(key, *row);
-    }
-
-private:
-    EScan Feed(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
+    EScan Feed(TArrayRef<const TCell> key, const TRow& row_) final
     {
         LOG_T("Feed " << Debug());
+
+        ++ReadRows;
+        ReadBytes += CountBytes(key, row_);
+        auto row = *row_;
         
         switch (State) {
             case EState::SAMPLE:
--- a/ydb/core/tx/datashard/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/local_kmeans.cpp
@@ -193,7 +193,7 @@ public:
     TInitialState Prepare(IDriver* driver, TIntrusiveConstPtr<TScheme>) final
     {
         TActivationContext::AsActorContext().RegisterWithSameMailbox(this);
-        LOG_D("Prepare " << Debug());
+        LOG_I("Prepare " << Debug());
 
         Driver = driver;
         return {EScan::Feed, {}};
@@ -270,7 +270,7 @@ protected:
 
     void HandleWakeup(const NActors::TActorContext& /*ctx*/)
     {
-        LOG_T("Retry upload " << Debug());
+        LOG_I("Retry upload " << Debug());
 
         if (!WriteBuf.IsEmpty()) {
             Upload(true);
@@ -388,7 +388,7 @@ public:
         : TLocalKMeansScanBase{buildId, table, std::move(lead), parent, child, request, std::move(result)}
     {
         this->Dimensions = request.GetSettings().vector_dimension();
-        LOG_D("Create " << Debug());
+        LOG_I("Create " << Debug());
     }
 
     EScan Seek(TLead& lead, ui64 seq) final
--- a/ydb/core/tx/datashard/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/local_kmeans.cpp
@@ -587,7 +587,7 @@ private:
     EScan FeedUploadMain2Build(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
     {
         const ui32 pos = FeedEmbedding(*this, Clusters, row, EmbeddingPos);
-        if (pos > K) {
+        if (pos >= K) {
             return EScan::Feed;
         }
         AddRowMain2Build(ReadBuf, Child + pos, key, row);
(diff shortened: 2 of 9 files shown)


ydb/core/tx/datashard/build_index.cpp
```diff
@@ -254,10 +255,9 @@ class TBuildScanUpload: public TActor<TBuildScanUpload<Activity>>, public NTable
 
         UploadStatusToMessage(progress->Record);
 
+        LOG_N("Finish" << Debug() << " " << progress->Record.ShortDebugString());
```
MBkkt, line 258: Maybe also `LOG_I`?

ydb/core/tx/datashard/local_kmeans.cpp
```diff
@@ -48,6 +48,8 @@ class TResult {
         }
         Size = 0;
         lock.unlock();
+
+        LOG_N("FinishTLocalKMeansScan " << Response->Record.ShortDebugString());
```
MBkkt, line 52: same here and in other places, maybe `LOG_I`?

ydb/core/tx/datashard/local_kmeans.cpp
```diff
@@ -544,10 +549,10 @@ class TLocalKMeansScan final: public TLocalKMeansScanBase, private TCalculation<
         return true;
     }
 
-    EScan FeedSample(const TRow& row)
+    EScan FeedSample(TArrayRef<const TCell> row)
```
MBkkt, line 552: FeedSample is duplicated between LocalKMeans and PrefixKMeans, I forgot about this when copied local_kmeans to write prefix_kmeans.

ydb/core/tx/datashard/local_kmeans.cpp
```diff
@@ -544,10 +549,10 @@ class TLocalKMeansScan final: public TLocalKMeansScanBase, private TCalculation<
...
     {
-        Y_ASSERT(row.Size() == 1);
-        const auto embedding = row.Get(0).AsRef();
+        Y_ASSERT(row.size() == 1);
+        const auto embedding = row.at(0).AsRef();
```
MBkkt, line 555: Get was opeator[] btw, but I don't think it's matter, just in case

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -298,26 +303,26 @@ class TPrefixKMeansScanBase: public TActor<TPrefixKMeansScanBase>, public NTable
         Driver->Touch(EScan::Final);
     }
 
-    EScan FeedUpload()
+    ui64 GetProbability()
+    {
+        return Rng.GenRand64();
+    }
+
+    bool ShouldWaitUpload()
```
MBkkt, line 311: All function can be written like (because how you implemented `TryUpload`)
```cpp
return  TryUpload(LevelBuf, LevelTable, LevelTypes, true)
     || TryUpload(PostingBuf, PostingTable, PostingTypes, true)
     || TryUpload(PrefixBuf, PrefixTable, PrefixTypes, true);
```

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -298,26 +303,26 @@ class TPrefixKMeansScanBase: public TActor<TPrefixKMeansScanBase>, public NTable
         Driver->Touch(EScan::Final);
     }
 
-    EScan FeedUpload()
+    ui64 GetProbability()
+    {
+        return Rng.GenRand64();
+    }
```
MBkkt, line 309: This code block unnecessary moved

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -409,82 +408,59 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
-                    || TryUpload(PrefixBuf, PrefixTable, PrefixTypes, false))
-                {
-                    return EScan::Sleep;
-                }
-
-                return EScan::Final;
+        LOG_D("Seek " << seq << " " << Debug());
+
+        if (IsExhausted) {
+            if (UploadInProgress()
```
MBkkt, line 435: `UploadInProgress` already checked inside `TryUpload` (IMO should be assert inside `TryUpload`, to make it more easy to reader)

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -619,35 +686,28 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
-        return FeedUpload();
     }
 
-    EScan FeedUploadBuild2Posting(TArrayRef<const TCell> key, const TRow& row)
+    void FeedUploadBuild2Posting(TArrayRef<const TCell> key, TArrayRef<const TCell> row)
     {
         const ui32 pos = FeedEmbedding(*this, Clusters, row, EmbeddingPos);
-        if (pos > K) {
-            return EScan::Feed;
+        if (pos < K) {
```
MBkkt, line 708: Good catch, can you also fix same condition in other places (should be `>= K` instead of `> K`)?
It's not a bug, because only `[0...K)` and max_uint is possible, but still incorrect code

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+            if (!InitAggregatedClusters()) {
+                // We don't need to do anything,
+                // because this datashard doesn't have valid embeddings for this parent
+                return true;
+            }
+            Round = 1;
+            return false; // do KMEANS
+        }
+
+        if (State == EState::KMEANS) {
```
MBkkt, line 548: NIT
```suggestion
        } else if (State == EState::KMEANS) {
```

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -102,15 +102,19 @@ class TPrefixKMeansScanBase: public TActor<TPrefixKMeansScanBase>, public NTable
 
     ui32 PrefixColumns;
     TSerializedCellVec Prefix;
-    bool HasNextPrefix = false;
+    TBufferData PrefixRows;
+    bool IsFirstPrefixFeed = true;
+    bool IsPrefixRowsValid = true;
```
MBkkt, line 107: I think you can easily extend current tests to test both behavior (now only small prefixes tested):
1. Lets add max rows count in prefix kmeans proto message
2. Run prefix_kmeans tests with 0 and 1000 max rows count

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
-            default:
-                Y_ASSERT(false);
-                return EScan::Final;
+
+        if (IsFirstPrefixFeed && IsPrefixRowsValid) {
+            PrefixRows.AddRow(TSerializedCellVec{key}, TSerializedCellVec::Serialize(*row));
+            if (PrefixRows.IsReachLimits(Limits)) {
+                PrefixRows.Clear();
+                IsPrefixRowsValid = false;
+            }
```
MBkkt, line 480: Doesn't really matter, but I think it's better
```suggestion
            if (PrefixRows.IsReachLimits(Limits)) {
                PrefixRows.Clear();
                IsPrefixRowsValid = false;
            } else {
                PrefixRows.AddRow(TSerializedCellVec{key}, TSerializedCellVec::Serialize(*row));
            }
```

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -102,15 +102,19 @@ class TPrefixKMeansScanBase: public TActor<TPrefixKMeansScanBase>, public NTable
 
     ui32 PrefixColumns;
     TSerializedCellVec Prefix;
-    bool HasNextPrefix = false;
+    TBufferData PrefixRows;
```
MBkkt, line 105: Maybe at least todo about better is storing something like `std::vector<std::pair<TSerializedCellVec, TSerializedCellVec>>` instead of `std::vector<std::pair<TSerializedCellVec, TString>>` to avoid reparsing values

Maybe just template parameter for TBufferData?

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+        LOG_D("Exhausted " << Debug());
+
+        if (!FinishPrefix()) {
+            return EScan::Reset;
         }
+            
+        IsExhausted = true;
+        
+        // call Seek to wait uploads
+        return EScan::Reset;
```
MBkkt, line 499: We probably can call this code (wait uploads) here?
And avoid additional seek overhead

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
-                return FeedKMeans(row);
-            case EState::UPLOAD_BUILD_TO_BUILD:
-                return FeedUploadBuild2Build(key, row);
-            case EState::UPLOAD_BUILD_TO_POSTING:
-                return FeedUploadBuild2Posting(key, row);
-            default:
-                Y_ASSERT(false);
-                return EScan::Final;
+
+        if (IsFirstPrefixFeed && IsPrefixRowsValid) {
```
MBkkt, line 475: We don't need `IsFirstPrefixFeed`, we have `Sample` state, you can just put this code inside FeedSample

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -409,82 +408,59 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+        ReadBytes += CountBytes(key, row);
+
+        if (Prefix && !TCellVectorsEquals{}(Prefix.GetCells(), key.subspan(0, PrefixColumns))) {
+            if (!FinishPrefix()) {
+                // scan current prefix rows with a new state again
+                return EScan::Reset;
+            }
+        }
+
         if (!Prefix) {
```
MBkkt, line 464: this code is needed only for FeedSample, we don't need it if it's not `IsFirstPrefixFeed`

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+    bool FinishPrefix()
+    {
+        if (FinishPrefixImpl()) {
+            StartNewPrefix();
+            LOG_D("FinishPrefix finished " << Debug());
+            return true;
+        } else {
+            IsFirstPrefixFeed = false;
+            
+            if (IsPrefixRowsValid) {
```
MBkkt, line 512: We don't need bool flag, just check PrefixRows.empty or not should be enough

ydb/core/tx/datashard/upload_stats.h
```diff
@@ -78,8 +78,8 @@ struct TUploadRetryLimits {
 };
 
 struct TUploadLimits: TUploadRetryLimits {
-    ui64 BatchRowsLimit = 500;
-    ui64 BatchBytesLimit = 1u << 23; // 8MB
+    ui64 BatchRowsLimit = 50000;
+    ui64 BatchBytesLimit = 8388608; // 8MB
```
MBkkt, line 82: maybe 2-4MB for prefix_kmeans? Because it commonly will use x2-3 memory, but in general up to x4 (level, prefix, posting, upload)

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+                    if (FinishPrefixImpl()) {
+                        StartNewPrefix();
+                        LOG_D("FinishPrefix finished in " << iteration << " iterations " << Debug());
+                        return true;
+                    } else {
+                        LOG_D("FinishPrefix not finished in " << iteration << " iterations " << Debug());
+                    }
+                }
+            } else {
+                LOG_D("FinishPrefix not finished, rescanning rows " << Debug());
```
MBkkt, line 528: Not sure, will we read smaller?
```suggestion
                LOG_D("FinishPrefix not finished, rescanning rows " << Debug());
                if (State == Upload) {
                   lead.SetTags(ScanTags);
                } else {
                   lead.SetTags({&EmbeddingTag, 1});
                   EmbeddingPosForKMeans = 0;
                }
```

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+            
+            if (IsPrefixRowsValid) {
+                LOG_D("FinishPrefix not finished, manually feeding " << PrefixRows.Size() << " saved rows " << Debug());
+                for (ui64 iteration = 0; ; iteration++) {
+                    for (const auto& [key, row_] : *PrefixRows.GetRowsData()) {
+                        TSerializedCellVec row(row_);
+                        Feed(key.GetCells(), row.GetCells());
+                    }
+                    if (FinishPrefixImpl()) {
+                        StartNewPrefix();
```
MBkkt, line 520: move `StartNewPrefix()` to `FinishPrefixImpl() return true`?

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -736,12 +796,6 @@ void TDataShard::HandleSafe(TEvDataShard::TEvPrefixKMeansRequest::TPtr& ev, cons
         ScanManager.Drop(id);
     }
 
-    const auto range = userTable.GetTableRange();
-    if (range.IsEmptyRange(userTable.KeyColumnTypes)) {
-        badRequest(TStringBuilder() << " requested range doesn't intersect with table range");
-        return;
-    }
```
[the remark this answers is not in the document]
kunga: Because we don't save any progress for now and these range is always empty

In secondary index we save progress and thus intersect table & request ranges https://github.com/ydb-platform/ydb/blob/main/ydb/core/tx/datashard/build_index.cpp#L573

Progress saving may be implemented later if needed https://github.com/ydb-platform/ydb/issues/16342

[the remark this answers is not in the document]
MBkkt: Ok, but in such case, it makes sense to do same for all other vector index scans?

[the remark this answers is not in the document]
kunga: Probably, for now I'm not sure how much data they read in one run

ydb/core/tx/datashard/build_index.cpp
kunga: Default log level is Notice, and I expect to at least see start/finish logs as these scan operation is quite expensive in database scope

ydb/core/tx/datashard/local_kmeans.cpp
kunga: https://github.com/ydb-platform/ydb/issues/16349

ydb/core/tx/datashard/local_kmeans.cpp
kunga: Yeah, just refactored with minimal diff

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+                    for (const auto& [key, row_] : *PrefixRows.GetRowsData()) {
+                        TSerializedCellVec row(row_);
+                        Feed(key.GetCells(), row.GetCells());
+                    }
+                    if (FinishPrefixImpl()) {
+                        StartNewPrefix();
+                        LOG_D("FinishPrefix finished in " << iteration << " iterations " << Debug());
+                        return true;
+                    } else {
+                        LOG_D("FinishPrefix not finished in " << iteration << " iterations " << Debug());
```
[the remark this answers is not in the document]
kunga: It isn't round, it feed rows iteration, theoretically they may differ

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -494,28 +470,100 @@ class TPrefixKMeansScan final: public TPrefixKMeansScanBase, private TCalculatio
...
+                return false; // retry KMEANS
+            }
+        }
+
+        if (State == UploadState) {
+            return true;
+        }
+
+        Y_ASSERT(false);
+        return true;
```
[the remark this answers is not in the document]
kunga: I'd preffer current style more as after we may have here more actions

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: Same, else looks redundant

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: Here yes, but it will be tricky to check when we need to write a row into buffer

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: We may, but as the first pass require to read all needed columns that seems not much important

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: https://github.com/ydb-platform/ydb/pull/16282#discussion_r2016877208

ydb/core/tx/datashard/prefix_kmeans.cpp
```diff
@@ -366,7 +371,7 @@ class TPrefixKMeansScanBase: public TActor<TPrefixKMeansScanBase>, public NTable
         return false;
     }
 
-    void WriteLevelRows()
+    void FormLevelRows()
```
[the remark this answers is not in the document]
kunga: TryUpload after each/all AddRow call?

It will lead to 1-row uploads, that as we know from https://github.com/ydb-platform/ydb/pull/16220 isn't effective

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: FinishPrefixImpl has 3 return true in different locations, and only responsible for state changes now, so it looks better here for me

ydb/core/tx/datashard/prefix_kmeans.cpp
MBkkt: >  but as the first pass require to read all needed columns that seems not much important

But count of other passes is more than 1 or 2, it even can be configurable see number of rounds

kunga: We can't read only embedding because when we get next prefix row we expect it to have all of them

Also reading only some tags help only is scheme has different column families

Prefer to leave it as is

MBkkt: > We can't read only embedding because when we get next prefix row we expect it to have all of them

We can for kmeans recompute steps

But ok

MBkkt (dismissed):

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: Yes, but we have multiple buffers and need multiple uploads, so Seek will be called anyway and need wait uploads logic

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: This way we check limit as soon as possible without waiting next iteration with broken limits, and that seems more sensible

But doesn't matter so let's leave as is

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: Ok, will do separately https://github.com/orgs/ydb-platform/projects/32/views/7?pane=issue&itemId=104049167

And better add limits

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: Just an optimization (no need to pass TryUpload arguments) and easier to read code 

Also somewhere it is used in another UploadInProgress / TryUpload combination

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: Now all upload-relevant method are groupped together

ydb/core/tx/datashard/prefix_kmeans.cpp
kunga: I think no, TryUpload return true when something is uploading

But here we should return true when something is uploading and some other buffer reaches its limit

MBkkt (approved):

PR #16423, opened 2025-03-28
fail gracefully on SHOW CREATE TABLE targeting a view

### LINKED ISSUE #16401: show create table: when targeting a view that reads from a table t, produces results of a 'select * from t'
## Steps to reproduce

1. Create a table `my_table`.
2. Create a view `my_view` reading from the table `my_table`.
3. Call `SHOW CREATE TABLE` on the view `my_view` (it works, but with a bug):

```sql
SHOW CREATE TABLE `my_view`;
```
4. Observe that the `SHOW CREATE` is transformed into something that produces the same results as the

```sql
SELECT * FROM `my_table`;
```

## Probable cause

See the following `KQP_YQL` [trace](https://gist.github.com/jepett0/4a0477e58a57a693139a981ec3850f8c) of the

```sql
SHOW CREATE TABLE `view`;
```

(these logs can be obtained by settings `KQP_YQL` log level to `TRACE` in your cluster's `config.yaml`)

I believe the following chain of transformations:
1) `show create table`
2) `select * from my_view` with a special marker `showCreateTable`
3)

```sql
select * from `.sys/show_create` where path = "my_view" and pathType = "Table"
```

is broken, because on the step 2 `select * from my_view` is rewritten to be `select * from my_table`, but the marker `showCreateTable` is lost in this rewrite and the step 3 is newer executed.

jepett0:
### Changelog entry 

Making `SHOW CREATE TABLE` fail on views rather than produce wrong output.

### Changelog category 

* Bugfix 

### Description for reviewers 

Issue:
- https://github.com/ydb-platform/ydb/issues/16401

### DIFF
--- a/ydb/core/kqp/provider/yql_kikimr_datasource.cpp
+++ b/ydb/core/kqp/provider/yql_kikimr_datasource.cpp
@@ -9,6 +9,7 @@
 
 #include <yql/essentials/core/yql_expr_optimize.h>
 #include <yql/essentials/core/yql_expr_type_annotation.h>
+#include <yql/essentials/core/yql_opt_utils.h>
 #include <yql/essentials/providers/common/schema/expr/yql_expr_schema.h>
 #include <ydb/library/yql/providers/dq/expr_nodes/dqs_expr_nodes.h>
 #include <ydb/library/yql/dq/expr_nodes/dq_expr_nodes.h>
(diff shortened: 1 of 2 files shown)


shnikd (approved):

CyberROFL (commented):
lgtm

gridnevvvit (approved):

PR #16759, opened 2025-04-04
Support for distconf explicit configuring of state storage and static groups

alexvru:
### Changelog entry 

Support for distconf explicit configuring of state storage and static groups

### Changelog category 

### Description for reviewers 

Stub for configuration files to support distconf explicit setting of state storage and static blobstorage configurations
through "cluster config replace" mechanism.


mregrock (approved):
LGTM

PR #16923, opened 2025-04-08
Fix recursive remove path

iddqdex:
Сломали

### DIFF
--- a/ydb/public/lib/ydb_cli/common/recursive_remove.cpp
+++ b/ydb/public/lib/ydb_cli/common/recursive_remove.cpp
@@ -222,29 +222,6 @@ TStatus RemoveDirectoryRecursive(
 
 TStatus RemovePathRecursive(
     const TDriver& driver,
-    const TString& path,
-    const TRemoveDirectoryRecursiveSettings& settings
-) {
-    TSchemeClient schemeClient(driver);
-    auto entity = schemeClient.DescribePath(path).ExtractValueSync();
-    if (!entity.IsSuccess()) {
-        if (settings.NotExistsIsOk_ && entity.GetStatus() == EStatus::SCHEME_ERROR && entity.GetIssues().ToString().find("Path not found") != TString::npos) {
-            return TStatus(EStatus::SUCCESS, {});
-        }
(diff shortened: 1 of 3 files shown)


ydb/public/lib/ydb_cli/common/recursive_remove.cpp
```diff
@@ -234,12 +234,13 @@ TStatus RemovePathRecursive(
...
-    NQuery::TQueryClient queryClient(driver);
-    NCoordination::TClient coordinationClient(driver);
-    auto remover = NInternal::CreateDefaultRemover(schemeClient, tableClient, topicClient, queryClient, coordinationClient, settings);
-    return remover(entity.GetEntry());
+    switch (entity.GetEntry().Type) {
+        case NYdb::NScheme::ESchemeEntryType::ColumnStore:
+        case NYdb::NScheme::ESchemeEntryType::Directory:
+            return RemoveDirectoryRecursive(driver, path, settings);
+        default:
+            return RemovePathRecursive(driver, entity.GetEntry(), settings);
```
jepett0, line 242: Isn't it an infinite loop? RemovePathRecursive -> RemovePathRecursive -> ...

jepett0: Ahh, ok, it calls another overload of the RemovePathRecursive function. It isn't an infinite loop

iddqdex: Нет, там вызывается другой RemovePathRecursive

jepett0 (dismissed):
Thank you for fixing it!

ydb/public/lib/ydb_cli/common/recursive_remove.cpp
jepett0, line 242: In this line we delegate the removal to the RemovePathRecursive overload that accepts scheme entry. However, there is a problem with this overload, it expects schemeEntry.Name to be equal to the full path to the object to be removed

jepett0 (approved):

pnv1 (approved):

PR #16932, opened 2025-04-08
Add option to the CMS to set the limit of FAULTY pdisks per node

SammyVimes:
### Changelog entry 

Add `MaxFaultyPDisksPerNode` to CMS config, so that Sentinel could not exceed that number. This may be useful in case if a node with a large count of disks goes down but it is known that this node will be brought back in the near future. This way, we will not be moving many disks at once (which can be rather long in case of HDDs). If `MaxFaultyPDisksPerNode` is 0, then there is no limit.

### Changelog category 

* Experimental feature

### Description for reviewers 

Self-heal detects unavailability of disks based on FAULTY status set by Sentinel and moves them out. Imagine a situation, where we have nodes with disk arrays attached, which contain 64 disks with 15TB or more of data on each. In such a configuration, a node or an array failure triggers too much data moving with weeks to reconcile without options to stop this process, while in fact the disks are operational, and the node or the shelve is likely to be restored much faster. So, to keep such a cluster in operational state self-heal should be disabled.

In this PR I introduce a way to control how many disks on one host local Self-Heal can restore. Forced FAULTY status should be respected and always allowed anyway, but is still counted. So if max faulty per node value is 10 and there are 10 faulty disks or faulty candidates (based on PDiskState) on a node (doesn't matter if they are forcefully FAULTY or this status was set by Sentinel), then another disk on this node can only become FAULTY if it was set by CMS forcefully.

### DIFF
--- a/ydb/core/cms/sentinel.cpp
+++ b/ydb/core/cms/sentinel.cpp
@@ -208,11 +208,6 @@ EPDiskStatus TPDiskStatus::GetStatus() const {
     return Current;
 }
 
-EPDiskStatus TPDiskStatus::GetNewStatus() const {
-    TString unused;
-    return Compute(Current, unused);
-}
-
 bool TPDiskStatus::IsNewStatusGood() const {
     TString unused;
     switch (Compute(Current, unused)) {
@@ -273,7 +268,7 @@ TClusterMap::TClusterMap(TSentinelState::TPtr state)
 {
 }
 
-void TClusterMap::AddPDisk(const TPDiskID& id, const EPDiskStatus status) {
+void TClusterMap::AddPDisk(const TPDiskID& id, const bool inGoodState) {
     Y_ABORT_UNLESS(State->Nodes.contains(id.NodeId));
     const auto& location = State->Nodes[id.NodeId].Location;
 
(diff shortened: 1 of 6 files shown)


ydb/core/cms/sentinel.cpp
```diff
@@ -347,6 +357,37 @@ TClusterMap::TPDiskIDSet TGuardian::GetAllowedPDisks(const TClusterMap& all, TSt
...
+                    if (++counter > allowedCount) {
+                        disallowed.emplace(pdisk, NKikimrCms::TPDiskInfo::MAX_FAULTY_PER_NODE);
+                        result.erase(pdisk);
+                    }
+                }
+
+                if (changedCount > allowedCount) {
+                    auto disallowedPdisks = disallowed | std::views::keys;
+                    issuesBuilder
+                        << "Ignore state updates due to MaxFaultyPDisksPerNode" \
```
pixcc, line 381: Слеши не нужны, это же не макрос

pixcc (changes requested):

pixcc (dismissed):

ydb/core/protos/cms.proto
```diff
@@ -452,6 +452,7 @@ message TCmsConfig {
         optional uint32 DataCenterRatio = 10 [default = 50];
         optional uint32 RoomRatio = 11 [default = 70];
         optional uint32 RackRatio = 12 [default = 90];
+        optional uint32 MaxFaultyPDisksPerNode = 17 [default = 0];
```
pixcc, line 455: Название немного misleading, можно трактовать, что при отказе всех дисков на ноде, только `N` дисков будет помечено `FAULTY`. А в нашей реализации никакой диск не будет помечен `FAULTY`. Но лучше я придумать тоже не смог.

LGTM

CyberROFL: Стоит добавить комментарий. Что это примерно то же самое, что *Ratio, только в абсолютах и на хосте (а не стойке, модуле, дц).

CyberROFL (approved):

pixcc (approved):

PR #17012, opened 2025-04-09
Clarifying the cause of the error after an authentication attempt

flown4qqqq:
### Changelog entry 

The pull request is aimed at clarifying the reason why the user could not log in.

### Changelog category 

* Improvement

### DIFF
--- a/ydb/library/login/login.cpp
+++ b/ydb/library/login/login.cpp
@@ -434,7 +434,12 @@ TLoginProvider::TCheckLockOutResponse TLoginProvider::CheckLockOutUser(const TCh
             response.Status = TCheckLockOutResponse::EStatus::RESET;
         } else {
             response.Status = TCheckLockOutResponse::EStatus::SUCCESS;
-            response.Error = TStringBuilder() << "User " << request.User << " is not permitted to log in";
+
+            if (!sid.IsEnabled) {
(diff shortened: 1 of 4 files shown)


molotkov-and (dismissed):

ydb/library/login/login.cpp
```diff
@@ -434,7 +434,12 @@ TLoginProvider::TCheckLockOutResponse TLoginProvider::CheckLockOutUser(const TCh
             response.Status = TCheckLockOutResponse::EStatus::RESET;
         } else {
             response.Status = TCheckLockOutResponse::EStatus::SUCCESS;
-            response.Error = TStringBuilder() << "User " << request.User << " is not permitted to log in";
+
+            if (!sid.IsEnabled) {
+                response.Error = TStringBuilder() << "User " << request.User << " is not permitted to log in, because it is blocked";
+            } else {
+                response.Error = TStringBuilder() << "User " << request.User << " is not permitted to log in, because the count of password attempts has been exceeded";
+            }
```
ijon, line 442: I would suggest making error messages more concise:
-> `User {request.User} login denied: account is blocked`
-> `User {request.User} login denied: too many failed password attempts`

flown4qqqq: the logic was that we wanted to mimic postgres. However, with the clarification of the reason, we are no longer doing this, so your option looks more successful.

ijon (approved):

Issue #17044, opened 2025-04-10
SVG look refactoring


Hor911:

Hor911:
![sample](https://github.com/user-attachments/assets/6f124af8-064c-458f-bc49-370fa22a58b7)

PR #17131, opened 2025-04-11
Update time in stat

Hor911:
#17044 

Visualize time with no changes in statistics

wait

Here are 25s of normal processing than hangs until 10m deadlock timeout is reached


pnv1 (approved):
CLI part OK

spuchin (approved):

PR #17133, opened 2025-04-11
pq rd: fix use-after-free (with logbroker federation)

yumkam:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Only logbroker federation was affected;
Also fixes cosmetic-only mixed-up counters and avoid type re-initialization (only used in main actor)


yumkam:
Preferred alternative pr

PR #17193, opened 2025-04-14
Fix potential use-after-free in task_ping.cpp

yumkam:
Counters here is this->Counters and this is captured by lambda. regression by 3321dfa2cccbc1dfe52c7db872a69bdce11f641e / #16437

### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


yumkam:
There are more places that access via captured `this` (`Config`, `YdbConnection`) from futures, so this should not be only problem. Closing for now

PR #17374, opened 2025-04-17
Fix shredding stalling

alexvru:
### Changelog entry 

Fix shredding stalling

### Changelog category 

### Description for reviewers 

This patch fixes randomly appearing stalling during disk shredding.


mregrock (approved):
LGTM

PR #17625, opened 2025-04-23
ydbd admin bs disk obliterate - improvement

zinal:
### Changelog entry 

The "obliterate" command has been improved:
* an error is returned if non-existing device or file is passed as the parameter, instead of creating a new small file and succeeding
* an error is returned if the file or device is too small to be the YDB storage device

### Changelog category 

* Improvement

### Description for reviewers 

...

### DIFF
--- a/ydb/core/blobstorage/pdisk/blobstorage_pdisk_tools.cpp
+++ b/ydb/core/blobstorage/pdisk/blobstorage_pdisk_tools.cpp
@@ -242,21 +242,13 @@ void ObliterateDisk(TString path) {
     ui64 diskSizeBytes = 0;
     DetectFileParameters(path, diskSizeBytes, isBlockDevice);
 
-    constexpr ui64 GB_BYTES = 4096L * 262144L;
-    if (diskSizeBytes <= GB_BYTES) {
+    constexpr size_t portionSize = NPDisk::FormatSectorSize * NPDisk::ReplicationFactor;
+    if (diskSizeBytes <= portionSize) {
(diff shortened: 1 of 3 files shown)


ydb/core/blobstorage/pdisk/blobstorage_pdisk_tools.cpp
```diff
@@ -233,10 +233,19 @@ bool ReadPDiskFormatInfo(const TString &path, const NPDisk::TMainKey &mainKey, T
...
     f.Flock(LOCK_EX | LOCK_NB);
 
-    TVector<ui8> zeros(NPDisk::FormatSectorSize * NPDisk::ReplicationFactor, 0);
+    bool isBlockDevice = false;
+    ui64 diskSizeBytes = 0;
+    DetectFileParameters(path, diskSizeBytes, isBlockDevice);
+
+    constexpr size_t portionSize = NPDisk::FormatSectorSize * NPDisk::ReplicationFactor;
+    if (diskSizeBytes <= portionSize) {
+        ythrow TFileError() << "illegal size " << diskSizeBytes << " for device " << path;
```
rosik, line 245: The PR description seems to me more clear than words "illegal size", I suggest citing it as is. Also, `yexception` seems to be more appropriate here.  

```suggestion
        ythrow yexception() << "file is too small to be the YDB storage device, path# " << path.Quote() <<
            " diskSizeBytes# " << diskSizeBytes;
```

rosik (approved):

PR #17630, opened 2025-04-23
Remove large-tests tags

verachecheneva:
Remove ya:large_tests_on_multi_slots and ya:large_tests_on_ya_make_2 tags
- These tags are no longer processed  
- Their behavior became the default


maximyurchuk (approved):

maximyurchuk:
Спасибо за ПР!

PR #17687, opened 2025-04-24
healthcheck vdisk use state2

StekPerepolnen:
### Changelog entry 

To avoid ambiguity, we switched to using the PDisk State from BSC info to determine the PDisk status instead of the PDisk status in CMS. It is already used in the healthcheck fallback logic and reflects the actual disk condition. We plan to make it the primary source for PDisk state in healthcheck.

### Changelog category 

* Improvement

### Description for reviewers 

**motivation:**
- healthcheck may report OK for a PDisk with INACTIVE CMS status, which is not always correct if the PDisk is actually dead.
- added a new field State to BSC information to improve determination of PDisk statuses. This is the same field used in fallback scenarios, allowing unified logic across all cases.

**notes:**
- in case of missing state information, there is a fallback to statusV2.

### DIFF
--- a/ydb/core/health_check/health_check.cpp
+++ b/ydb/core/health_check/health_check.cpp
@@ -2148,36 +2148,52 @@ public:
         const auto& pDisk = itPDisk->second->GetInfo();
 
         context.Location.mutable_storage()->mutable_pool()->mutable_group()->mutable_vdisk()->mutable_pdisk()->begin()->set_path(pDisk.GetPath());
-        const auto& statusString = pDisk.GetStatusV2();
-        const auto *descriptor = NKikimrBlobStorage::EDriveStatus_descriptor();
-        auto status = descriptor->FindValueByName(statusString);
-        if (!status) {
+        const auto& stateString = pDisk.GetState();
+        const auto *descriptor = NKikimrBlobStorage::TPDiskState::E_descriptor();
+        auto state = descriptor->FindValueByName(stateString);
+        if (!state) {;
             context.ReportStatus(Ydb::Monitoring::StatusFlag::RED,
-                                 TStringBuilder() << "Unknown PDisk state: " << statusString,
+                                 TStringBuilder() << "Unknown PDisk state: " << stateString,
                                  ETags::PDiskState);
             storagePDiskStatus.set_overall(context.GetOverallStatus());
             context.OverallStatus = Ydb::Monitoring::StatusFlag::ORANGE;
             return;
         }
-        switch (status->number()) {
-            case NKikimrBlobStorage::ACTIVE:
-            case NKikimrBlobStorage::INACTIVE: {
+        switch (state->number()) {
+            case NKikimrBlobStorage::TPDiskState::Normal:
+            case NKikimrBlobStorage::TPDiskState::Stopped:
                 context.ReportStatus(Ydb::Monitoring::StatusFlag::GREEN);
                 break;
-            }
-            case NKikimrBlobStorage::FAULTY: {
+            case NKikimrBlobStorage::TPDiskState::Initial:
+            case NKikimrBlobStorage::TPDiskState::InitialFormatRead:
+            case NKikimrBlobStorage::TPDiskState::InitialSysLogRead:
+            case NKikimrBlobStorage::TPDiskState::InitialCommonLogRead:
                 context.ReportStatus(Ydb::Monitoring::StatusFlag::YELLOW,
-                                     TStringBuilder() << "PDisk state is " << statusString,
-                                     ETags::PDiskState);
+                                        TStringBuilder() << "PDisk state is " << stateString,
+                                        ETags::PDiskState);
                 break;
-            }
-            case NKikimrBlobStorage::BROKEN:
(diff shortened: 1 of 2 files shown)


ydb/core/health_check/health_check.cpp
```diff
@@ -2128,6 +2128,58 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
         return TStringBuilder() << pDiskKey.GetNodeId() << "-" << pDiskKey.GetPDiskId();
     }
 
+    bool TryParsePDiskStatus(const TString& statusString, ui32& statusNumber) {
+        const auto *descriptor = NKikimrBlobStorage::EDriveStatus_descriptor();
+        const auto* valueDescriptor = descriptor->FindValueByName(statusString);
+        if (!valueDescriptor) {
+            return false;
```
adameat, line 2135: why don't return default value (UNKNOWN) in that case? it will make this function much simpler...?

ydb/core/health_check/health_check.cpp
```diff
@@ -2128,6 +2128,58 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
...
+    void ReportPDiskState(NKikimrBlobStorage::TPDiskState::E stateEnum, TSelfCheckContext& context) {
+        using E = NKikimrBlobStorage::TPDiskState;
+        const TString stateStr = E::E_Name(stateEnum);
+
+        switch (stateEnum) {
+            case E::Normal:
+            case E::Stopped:
+                context.ReportStatus(Ydb::Monitoring::StatusFlag::GREEN);
+                break;
+            case E::Initial:
```
adameat, line 2150: I advise from making reaction to Initial states - it's okay to have them during start of a node. So it would make this check to flap every time a node is starting.

ydb/core/health_check/health_check.cpp
```diff
@@ -2128,6 +2128,58 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
...
+            case E::DeviceIoError:
+            case E::Missing:
+            case E::Timeout:
+            case E::NodeDisconnected:
+            case E::Unknown:
+                context.ReportStatus(Ydb::Monitoring::StatusFlag::RED,
+                                    TStringBuilder() << "PDisk state is " << stateStr,
+                                    ETags::PDiskState);
+                break;
+            case E::Reserved15:
```
adameat, line 2175: Maybe ignore reserved statues too - they were made for compatibility between versions, and this version clearly don't know how to interpret them...?

ydb/core/health_check/health_check.cpp
```diff
@@ -2128,6 +2128,58 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
...
+            case E::InitialCommonLogReadError:
+            case E::InitialCommonLogParseError:
+            case E::CommonLoggerInitError:
+            case E::OpenFileError:
+            case E::ChunkQuotaError:
+            case E::DeviceIoError:
+            case E::Missing:
+            case E::Timeout:
+            case E::NodeDisconnected:
+            case E::Unknown:
```
adameat, line 2170: Unknown status is when status is unknown - let's not react on that? it could be during startup too...

ydb/core/health_check/health_check.cpp
```diff
@@ -2148,35 +2200,45 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
...
-                                     TStringBuilder() << "PDisk state is " << statusString,
-                                     ETags::PDiskState);
-                break;
+
+        if (pDisk.HasState()) {
+            const auto& stateString = pDisk.GetState();
+            const auto *descriptor = NKikimrBlobStorage::TPDiskState::E_descriptor();
+            auto state = descriptor->FindValueByName(stateString);
+            if (!state) {
+                context.ReportStatus(Ydb::Monitoring::StatusFlag::RED,
```
adameat, line 2209: again, it only will trigger during different versions - I don't think it's a reason to report RED alert.

adameat (changes requested):

ydb/core/health_check/health_check.cpp
```diff
@@ -2128,6 +2128,58 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
...
+        }
+        statusNumber = valueDescriptor->number();
+        return true;
+    }
+
+    void ReportPDiskState(NKikimrBlobStorage::TPDiskState::E stateEnum, TSelfCheckContext& context) {
+        using E = NKikimrBlobStorage::TPDiskState;
+        const TString stateStr = E::E_Name(stateEnum);
+
+        switch (stateEnum) {
```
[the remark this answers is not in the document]
StekPerepolnen: I'd prefer not to add a default case here intentionally — to make sure that adding a new PDisk state will trigger a compile-time warning, so it won't go unnoticed.

adameat (dismissed):

adameat (approved):

vporyadke (approved):

PR #17815, opened 2025-04-28
Fix lost broken read locks

nikvas0:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Fix lost broken read locks


nikvas0:
Don't need it. Read locks are checked during commit

PR #17896, opened 2025-04-30
Replace all mentions of Kikimr cluster with YDB cluster

blinkov:
...as it gets exposed to end users in error messages.

### Changelog category 


maximyurchuk (changes requested):
yql/essentials надо править в аркадии, здесь лишь зеркало

PR #17934, opened 2025-04-30
Fix SSId issue in domains config

alexvru:
### Changelog entry 

Fix SSId issue in domains config

### Changelog category 

### Description for reviewers 

Distconf was not working properly when Domain.SSId field was not set, although process did start correctly with such static configuration. This patch synchronizes this behaviour.


mregrock (approved):
LGTM

PR #17952, opened 2025-05-01
Add root user attribute to keep gizmo resource id

StekPerepolnen:
### Changelog entry 

Store the access resource ID for cluster information as the `container_id` user attribute in the root database. Users can check permissions on this resource to access cluster-related information.

### Changelog category 

* Improvement

### Description for reviewers 

This replaces the previous approach of storing the resource in configuration, as implemented in issue https://github.com/ydb-platform/ydb/pull/16614
For backward compatibility, if `container_id` is not present, its value is `gizmo.

### DIFF
--- a/ydb/core/grpc_services/grpc_request_check_actor.h
+++ b/ydb/core/grpc_services/grpc_request_check_actor.h
@@ -52,23 +52,27 @@ inline const TVector<TEvTicketParser::TEvAuthorizeTicket::TEntry>& GetEntriesFor
 
     auto makeEntries = [rootAttributes]() -> TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> {
         const TString& accessServiceType = AppData()->AuthConfig.GetAccessServiceType();
-        TVector<TString> permissions;
         if (accessServiceType == "Yandex_v2") {
-            permissions = {"ydb.developerApi.get", "ydb.developerApi.update"};
+            TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = {
+                    {NKikimr::TEvTicketParser::TEvAuthorizeTicket::ToPermissions({"ydb.developerApi.get", "ydb.developerApi.update"}), {{"gizmo_id", "gizmo"}}}
+            };
+            return entries;
         } else if (accessServiceType == "Nebius_v1") {
-            permissions = {"ydb.clusters.get", "ydb.clusters.monitor", "ydb.clusters.manage"};
+            TVector<TString> permissions = {"ydb.clusters.get", "ydb.clusters.monitor", "ydb.clusters.manage"};
+            auto it = std::find_if(rootAttributes.begin(), rootAttributes.end(),
+            [](const std::pair<TString, TString>& p) {
+                return p.first == "container_id";
+            });
+            if (it == rootAttributes.end()) {
+                return {};
+            }
+            TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = {
+                    {NKikimr::TEvTicketParser::TEvAuthorizeTicket::ToPermissions(permissions), {{"gizmo_id", it->second}}}
+            };
+            return entries;
         } else {
             return {};
         }
-        auto it = std::find_if(rootAttributes.begin(), rootAttributes.end(),
-        [](const std::pair<TString, TString>& p) {
-            return p.first == "container_id";
-        });
-        const TString& accessResourceId = it != rootAttributes.end() ? it->second : "gizmo";
-        TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = {
-                {NKikimr::TEvTicketParser::TEvAuthorizeTicket::ToPermissions(permissions), {{"gizmo_id", accessResourceId}}}
-        };
-        return entries;
     };
 
     static TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = makeEntries();
--- a/ydb/core/grpc_services/grpc_request_check_actor.h
+++ b/ydb/core/grpc_services/grpc_request_check_actor.h
@@ -39,44 +39,38 @@ bool TGRpcRequestProxyHandleMethods::ValidateAndReplyOnError(TCtx* ctx) {
     }
 }
 
-inline const TVector<TEvTicketParser::TEvAuthorizeTicket::TEntry>& GetEntriesForAuthAndCheckRequest(TEvRequestAuthAndCheck::TPtr& ev, const TVector<std::pair<TString, TString>>& rootAttributes) {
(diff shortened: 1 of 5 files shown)


ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -39,33 +39,40 @@ bool TGRpcRequestProxyHandleMethods::ValidateAndReplyOnError(TCtx* ctx) {
...
-            permissions = {"ydb.developerApi.get", "ydb.developerApi.update"};
+            TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = {
+                    {NKikimr::TEvTicketParser::TEvAuthorizeTicket::ToPermissions({"ydb.developerApi.get", "ydb.developerApi.update"}), {{"gizmo_id", "gizmo"}}}
+            };
+            return entries;
         } else if (accessServiceType == "Nebius_v1") {
-            permissions = {"ydb.clusters.get", "ydb.clusters.monitor", "ydb.clusters.manage"};
+            TVector<TString> permissions = {"ydb.clusters.get", "ydb.clusters.monitor", "ydb.clusters.manage"};
+            auto it = std::find_if(rootAttributes.begin(), rootAttributes.end(),
+            [](const std::pair<TString, TString>& p) {
```
UgnineSirdis, line 63: Fix indent here, please.

ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -39,33 +39,40 @@ bool TGRpcRequestProxyHandleMethods::ValidateAndReplyOnError(TCtx* ctx) {
...
             return {};
         }
-        const TString& clusterAccessResourceId = AppData()->AuthConfig.GetClusterAccessResourceId();
-        TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = {
-            {NKikimr::TEvTicketParser::TEvAuthorizeTicket::ToPermissions(permissions), {{"gizmo_id", clusterAccessResourceId}}}
-        };
-        return entries;
     };
 
     static TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = makeEntries();
```
UgnineSirdis, line 78: Now local static variable `entries` depends on input parameter `rootAttributes`. So, effectively entries will be inited only with the first call param. Is this what you initialy wanted?

Can we take something from AppData() instead of taking from rootAttributes?

It is hard to understand, where is the guarantee that on this particular node we every time receive the same `rootAttributes`

ydb/core/grpc_services/grpc_request_proxy.cpp
```diff
@@ -224,6 +225,18 @@ class TGRpcRequestProxyImpl
...
+            if (rootIt != Databases.end() && rootIt->second.IsDatabaseReady()) {
+                const TDatabaseInfo* rootDb = &it->second;
+                if (rootDb && rootDb->SchemeBoardResult) {
+                    auto& schemeData = rootDb->SchemeBoardResult->DescribeSchemeResult;
+                    rootAttributes.reserve(schemeData.GetPathDescription().UserAttributesSize());
+                    for (const auto& attr : schemeData.GetPathDescription().GetUserAttributes()) {
+                        rootAttributes.emplace_back(std::make_pair(attr.GetKey(), attr.GetValue()));
+                    }
+                }
+            }
```
UgnineSirdis, line 239: else return error `Ydb::StatusIds::UNAVAILABLE`

ydb/core/security/ticket_parser_impl.h
```diff
@@ -482,6 +482,7 @@ class TTicketParserImpl : public TActorBootstrapped<TDerived> {
 
         // Use attribute "gizmo_id" as container id that contains cluster access resource
         // IAM can link roles for cluster access resource
+        // Note: "gizmo_id" and "folder_id" are always sent in separate TEvAuthorizeTicket requests.
         if (const auto gizmoId = record.GetAttributeValue(permission, "gizmo_id"); gizmoId) {
```
UgnineSirdis, line 486: Is it OK that gizmo_id has higher priority than database folder?

ydb/core/grpc_services/grpc_request_proxy.cpp
```diff
@@ -272,6 +285,7 @@ class TGRpcRequestProxyImpl
                 event.Release(),
                 Counters,
                 skipCheckConnectRigths,
+                rootAttributes,
```
UgnineSirdis, line 288: This code is under `if (database)`. Maybe I have not understood everything, but does this mean that we process root attributes only in requests with database parameter?

ydb/core/security/ticket_parser_impl.h
StekPerepolnen: `gizmo_id` and `folder_id` are never included in the same permission check — verified on slice.
For safety, I changed the priority: now `folder_id` takes precedence over `gizmo_id`.
Worst case: UI access may be limited, but DB access will behave correctly. No issues expected.

ydb/core/grpc_services/grpc_request_check_actor.h
StekPerepolnen: I removed static intentionally so that the attribute is checked on every call — I wanted changes to user attributes (like folder_id) to be reflected in authentication without requiring node restart.
But I might have misunderstood the original idea. Are you suggesting that we should use something from AppData() instead?
As far as I see, user attributes like folder_id or gizmo_id come from SchemeBoardResult, i.e. from path-level metadata, not from AppData or any static config.
So using rootAttributes per request still seems necessary to ensure correct permission checks.

ydb/core/security/ticket_parser_impl.h
```diff
@@ -462,7 +462,7 @@ class TTicketParserImpl : public TActorBootstrapped<TDerived> {
         pathsContainer->mutable_resource_path()->add_path()->set_id(id);
     }
 
-    static void AddNebiusContainerId(nebius::iam::v1::AuthorizeCheck* pathsContainer, const TString& id) {
+    static void SetNebiusContainerId(nebius::iam::v1::AuthorizeCheck* pathsContainer, const TString& id) {
```
UgnineSirdis, line 465: 👍

ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -39,37 +39,40 @@ bool TGRpcRequestProxyHandleMethods::ValidateAndReplyOnError(TCtx* ctx) {
...
-
-    static TVector<NKikimr::TEvTicketParser::TEvAuthorizeTicket::TEntry> entries = makeEntries();
-    return entries;
+    if (!isBearerToken || !useAccessService || !needClusterAccessResourceCheck) {
+        return {};
+    }
+
+    const TString& accessServiceType = AppData()->AuthConfig.GetAccessServiceType();
+
+	if (accessServiceType == "Yandex_v2") {
```
UgnineSirdis, line 54: As I can see, the code here is unindented.

UgnineSirdis (changes requested):

ydb/core/grpc_services/grpc_request_check_actor.h
StekPerepolnen: removed tabulation

UgnineSirdis (approved):

PR #18140, opened 2025-05-07
Used default value of EnableSystemViews feature flag in SchemeShard ut tests

n00bcracker:
### Changelog entry 

Updated default behavior of SchemeShard test environment.
Used default value of EnableSystemViews feature flag instead of redefined in the constructor.

### Changelog category 


ydb/core/tx/schemeshard/ut_base/ut_base.cpp
```diff
@@ -10079,7 +10079,7 @@ Y_UNIT_TEST_SUITE(TSchemeShardTest) {
 
     Y_UNIT_TEST(RejectSystemViewPath) {
         TTestBasicRuntime runtime;
-        TTestEnv env(runtime, 4, true, &CreateFlatTxSchemeShard, true);
+        TTestEnv env(runtime);
```
ijon, line 10082: This test should explicitly set EnableSystemViews to true via `TTestEnvOptions().EnableSystemViews(true)` as the test is meaningless if `EnableSystemViews = false`.
And it does not matter that `EnableSystemViews = true` by default -- default could be changed (not in practice but in theory).

ijon (approved):

PR #18541, opened 2025-05-20
Report group statuses in SysView

alexvru:
### Changelog entry 

Report group statuses in SysView

### Changelog category 

### Description for reviewers 

BSC now reports OperatingStatus and ExpectedStatus fields of group info into the SysView.


adameat (approved):
I suggested to add comments in proto describing new statuses

PR #18560, opened 2025-05-20
Bugfix in `ydb tools restore --replace`, do not throw an error if the object is missing from the database

### LINKED ISSUE #15410: ydb tools restore: add `--replace` option
we need to be able to restore scheme objets by recreating them (i.e. drop + create)

rfc:
- https://github.com/ydb-platform/ydb-rfc/pull/149

jepett0:
### Changelog category 

### Description for reviewers 

Do not throw an error as a result of `ydb tools restore --replace` in case the object is missing from the database and the option `--verify-existence` is __not__ specified.

Issues:
- https://github.com/ydb-platform/ydb/issues/15410

### DIFF
--- a/ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
+++ b/ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
@@ -2316,6 +2316,83 @@ Y_UNIT_TEST_SUITE(BackupRestore) {
         );
     }
 
+    Y_UNIT_TEST(TestReplaceRestoreOptionOnNonExistingSchemeObjects) {
+        NKikimrConfig::TAppConfig config;
+        config.MutableQueryServiceConfig()->AddAvailableExternalDataSources("ObjectStorage");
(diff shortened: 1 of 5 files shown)


ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
```diff
@@ -2316,6 +2316,83 @@ Y_UNIT_TEST_SUITE(BackupRestore) {
         );
     }
 
+    Y_UNIT_TEST(TestReplaceRestoreOptionOnNonExistingSchemeObjects) {
```
jepett0, line 2319: Exactly the same as the already existing `TestReplaceRestoreOption`, but the objects are dropped manually in the test bodies instead of automatically by the help of the `--replace` option. It means that the restoration will happen:
1) with the `--replace` option turned on
2) on an empty database (the objects will be dropped manually)

pnv1 (approved):

CyberROFL (approved):

PR #18653, opened 2025-05-22
bugfix protobuf_printer: avoid a memory leak

ArkadyRudenko:
`google::protobuf::TextFormat::Printer::RegisterFieldValuePrinter` tooks a ownership of the `google::protobuf::TextFormat::FastFieldValuePrinter` only in the success case. If !success `TPrinter` is leaked.
[docs](https://github.com/protocolbuffers/protobuf/blob/8543a716690c03102d3dfcfd98c579001a799a77/src/google/protobuf/text_format.h#L427)


maximyurchuk:
Thank you for the PR, @UgnineSirdis will take a look at it.

UgnineSirdis (approved):

PR #18765, opened 2025-05-23
Pass TStorageConfig through pointer to prevent copying when doing subscription fan-out

alexvru:
### Changelog entry 

Pass TStorageConfig through pointer to prevent copying when doing subscription fan-out

### Changelog category 

* Performance improvement

### Description for reviewers 

This patch prevents copying TStorageConfig many times when sending changed configuration to subscribers.


mregrock (approved):
LGTM

PR #18856, opened 2025-05-26
Automatic blobstorage balancing

SammyVimes:
### Changelog entry 

Adds periodic balancing task to BlobStorageController that moves VDisks to PDisks which are less occupied to make distribution of VDisks even. 

### Changelog category 

* Experimental feature

### Description for reviewers 

...

### DIFF
--- a/ydb/core/blobstorage/ut_blobstorage/cluster_balancing.cpp
+++ b/ydb/core/blobstorage/ut_blobstorage/cluster_balancing.cpp
@@ -58,12 +58,8 @@ struct TTestEnv {
 
     bool EachPDiskHasNVDisks(ui32 n) {
         auto usageMap = BuildPDiskUsageMap();
-        for (const auto& [pdiskId, usage] : usageMap) {
-            if (usage != n) {
-                return false;
-            }
-        }
-        return true;
+
+        return std::all_of(usageMap.begin(), usageMap.end(), [&](std::pair<TPDiskId, ui32> val) { return val.second == n; });
     }
 
     template<class TCondition>
@@ -123,6 +119,17 @@ Y_UNIT_TEST_SUITE(ClusterBalancing) {
         UNIT_ASSERT(success);
 
         auto usageMap1 = env.BuildPDiskUsageMap();
+
+        env.Env.Runtime->FilterFunction = [&](ui32, std::unique_ptr<IEventHandle>& ev) {
+            if (ev->GetTypeRewrite() == TEvBlobStorage::TEvControllerConfigResponse::EventType) {
+                const auto& response = ev->Get<TEvBlobStorage::TEvControllerConfigResponse>()->Record.GetResponse();
+                    
+                const auto& status = response.GetStatus(0);                
+
+                UNIT_ASSERT_VALUES_EQUAL(0, status.ReassignedItemSize());
+            }
+            return true;
+        };
 
         env.Env.Sim(TDuration::Seconds(5));
 
--- a/ydb/core/blobstorage/ut_blobstorage/cluster_balancing.cpp
+++ b/ydb/core/blobstorage/ut_blobstorage/cluster_balancing.cpp
@@ -1,0 +1,137 @@
+#include <ydb/core/blobstorage/ut_blobstorage/lib/env.h>
+#include <ydb/core/util/random.h>
+
+#include <library/cpp/iterator/enumerate.h>
+
+struct TTestEnv {
+    TTestEnv(ui32 nodeCount, TBlobStorageGroupType erasure, ui32 pdiskPerNode, ui32 groupCount)
+    : Env({
+        .NodeCount = nodeCount,
+        .VDiskReplPausedAtStart = false,
+        .Erasure = erasure,
+        .ConfigPreprocessor = [](ui32, TNodeWardenConfig& conf) {
+            auto* bscSettings = conf.BlobStorageConfig.MutableBscSettings();
+            auto* clusterBalancingSettings = bscSettings->MutableClusterBalancingSettings();
+
+            clusterBalancingSettings->SetEnable(true);
+            clusterBalancingSettings->SetMaxReplicatingPDisks(100);
+            clusterBalancingSettings->SetMaxReplicatingVDisks(800);
+            clusterBalancingSettings->SetIterationIntervalMs(TDuration::Seconds(1).MilliSeconds());
+        },
+    })
+    {
+        Env.CreateBoxAndPool(pdiskPerNode, groupCount);
+        Env.Sim(TDuration::Minutes(1));
+        // Uncomment to see cluster rebalancing logs
+        // Env.Runtime->SetLogPriority(NKikimrServices::BS_CLUSTER_BALANCING, NActors::NLog::PRI_DEBUG);
+    }
+
+    bool IsDynamicGroup(ui32 groupId) {
+        return groupId & 0x80000000;
+    }
+
+    std::unordered_map<TPDiskId, ui32> BuildPDiskUsageMap() {
+        auto config = Env.FetchBaseConfig();
+
+        std::unordered_map<TPDiskId, ui32> pdiskUsageMap;
+
+        for (const auto& pdisk : config.pdisk()) {
+            TPDiskId pdiskId(pdisk.GetNodeId(), pdisk.GetPDiskId());
+            pdiskUsageMap[pdiskId] = pdisk.GetNumStaticSlots();
+        }
+
+        for (const auto& vslot : config.vslot()) {
+            if (!IsDynamicGroup(vslot.GetGroupId())) {
+                continue;
+            }
+
+            TPDiskId pdiskId(vslot.GetVSlotId().GetNodeId(), vslot.GetVSlotId().GetPDiskId());
+            auto it = pdiskUsageMap.find(pdiskId);
+            if (it == pdiskUsageMap.end()) {
+                continue;
+            }
+            it->second += 1;
+        }
+
+        return pdiskUsageMap;
+    }
+
+    bool EachPDiskHasNVDisks(ui32 n) {
+        auto usageMap = BuildPDiskUsageMap();
+        for (const auto& [pdiskId, usage] : usageMap) {
+            if (usage != n) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    template<class TCondition>
+    bool WaitFor(TCondition&& condition, size_t maxAttempts = 1) {
+        for (size_t attempt = 0; attempt < maxAttempts; ++attempt) {
+            if (condition()) {
+                return true;
+            }
+            Env.Sim(TDuration::Minutes(1));
+        }
+        return false;
+    }
+
+    TEnvironmentSetup* operator->() {
+        return &Env;
+    }
+
+    TEnvironmentSetup Env;
+};
+
+Y_UNIT_TEST_SUITE(ClusterBalancing) {
+
+    Y_UNIT_TEST(ClusterBalancingEvenDistribution) {
+        TTestEnv env(8, TBlobStorageGroupType::Erasure4Plus2Block, 1, 2);
+
+        UNIT_ASSERT(env.EachPDiskHasNVDisks(2));
+
+        env->AlterBox(1, 2);
+
+        bool success = env.WaitFor([&] {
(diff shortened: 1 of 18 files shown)


ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,562 @@
...
+                            return;
+                    }
+                    Yield();
+                }
+            }
+        }
+
+    public:
+        TBalancingActor(const TActorId& controllerId, const TClusterBalancingSettings& settings)
+        : TActorCoroImpl(/* stackSize */ 640_KB, /* allowUnhandledDtor */ false)
```
alexvru: 640k ought to be enough for everyone (c) ? :)
isn't it too much?

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,562 @@
...
+                NKikimrBlobStorage::EVDiskStatus status;
+                NKikimrBlobStorage::EVDiskStatus_Parse(statusStr, &status);
+
+                switch (status) {
+                    case NKikimrBlobStorage::ERROR:
+                    case NKikimrBlobStorage::READY:
+                        break;
+                    case NKikimrBlobStorage::INIT_PENDING:
+                    case NKikimrBlobStorage::REPLICATING:
+                        replicatingPDisks.insert(TPDiskId(vslot.GetVSlotId().GetNodeId(), vslot.GetVSlotId().GetPDiskId()));
```
alexvru: key.ComprisingPDiskId()

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,562 @@
...
+                    continue;
+                }
+                it->second += 1;
+            }
+
+            storageInfo.ReplicatingPDisks = replicatingPDisks.size();
+            storageInfo.ReplicatingVDisks = replicatingVDisks;
+
+            std::unordered_map<TGroupId, ui32> goodSlotsByGroup;
+            for (const auto& vslotEntry : vSlotsResponse.entries()) {
```
alexvru: I'm not sure that relying on SysViews are a good idea -- they are async and all needed information is already reported in QueryBaseConfig (vslot.IsReady is the main field)

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,562 @@
...
+
+        bool IsDynamicGroup(TGroupId groupId) {
+            return groupId & 0x80000000;
+        }
+
+        TStorageInfo BuildStorageInfo(const NKikimrBlobStorage::TBaseConfig& config, const NKikimrSysView::TEvGetVSlotsResponse& vSlotsResponse) {
+            TStorageInfo storageInfo;
+
+            // First, iterate over PDisk and initialize the map.
+            for (const auto& pdisk : config.pdisk()) {
```
alexvru: pdisk -> GetPDisk(), it is default generated field name, doesn't fit our styleguide
and the same for others

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,562 @@
...
+                    continue;
+                }
+                
+                bool isHealthy = true;
+                ui32 slotCount = 0;
+
+                for (const auto& vslotId : group.GetVSlotId()) {
+                    auto key = TVSlotId(vslotId);
+                    auto it = vslotMap.find(key);
+                    if (it != vslotMap.end()) {
```
alexvru, line 104: it would be a good point to debug abort in else branch and drop isHealthy too

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,562 @@
...
+                        const auto& vslot = it->second;
+                        const auto& statusStr = vslot->GetStatus();
+                        NKikimrBlobStorage::EVDiskStatus status;
+                        NKikimrBlobStorage::EVDiskStatus_Parse(statusStr, &status);
+
+                        slotCount++;
+
+                        if (!vslot->GetReady() || status != NKikimrBlobStorage::EVDiskStatus::READY) {
```
alexvru: GetReady() is enough

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,562 @@
...
+            }
+
+            return storageInfo;
+        }
+
+        std::vector<std::vector<const TVSlot*>> OrderVSlotsByPDiskUsage(
+            const std::vector<const TVSlot*>& vslots,
+            const TStorageInfo& storageInfo
+        ) {
+            std::map<ui32, std::vector<const TVSlot*>, std::greater<>> vslotsByPDiskSlotUsage;
```
alexvru, line 130: maybe it's better to order by number of remaining free slots, but that's just an idea

ydb/core/blobstorage/ut_blobstorage/cluster_balancing.cpp
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    bool EachPDiskHasNVDisks(ui32 n) {
+        auto usageMap = BuildPDiskUsageMap();
+        for (const auto& [pdiskId, usage] : usageMap) {
+            if (usage != n) {
+                return false;
+            }
+        }
+        return true;
```
serbel324:
```suggestion
        return std::all_of(usageMap.begin(), usageMap.end(), [&](std::pair<TPDiskId, ui32> val) { return val.second == n; });
```

ydb/core/blobstorage/ut_blobstorage/cluster_balancing.cpp
```diff
@@ -0,0 +1,137 @@
...
+        bool success = env.WaitFor(check, 10);
+
+        UNIT_ASSERT(success);
+
+        auto usageMap1 = env.BuildPDiskUsageMap();
+
+        env.Env.Sim(TDuration::Seconds(5));
+
+        // Check that the cluster balancing doesn't do anything now.
+        // Optimal distribution is already achieved, any reassignment will only move data around for no reason.
```
serbel324, line 137: I would also suggest to add a filter function here to intercept reassigns directly. Something like this:
```suggestion
        // Optimal distribution is already achieved, any reassignment will only move data around for no reason.
        evn.Env->Runtime->FilterFunction = [&](ui32, std::unique_ptr<IEventHandle>& ev) {
            if (ev->GetTypeRewrite() == TEvBlobStorage::TEvControllerConfigRequest::EventType) {
                const auto& request = ev->Get<TEvBlobStorage::TEvControllerConfigRequest>()->Record.GetRequest();
                for (const auto& command : request.GetCommand()) {
                    UNIT_ASSERT(command.GetCommandCase() != NKikimrBlobStorage::TConfigRequest::TCommand::kSanitizeGroup);
                }
            }
            return true;
        };
```

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,460 @@
...
+    // See TODO about coroutine timeout.
+    // const TDuration TIMEOUT = TDuration::Seconds(10);
+
+    struct TEvResume : TEventLocal<TEvResume, EvResume> {};
+
+    struct TPoison {};
+
+    class TClusterBalancingActor : public TActorCoroImpl {
+    private:
+        using TGroupId = ui32;
```
serbel324: There is dedicated type for group id https://github.com/ydb-platform/ydb/blob/main/ydb/core/base/blobstorage_common.h#L8

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,460 @@
...
+        using TVSlot = NKikimrBlobStorage::TBaseConfig_TVSlot;
+
+        const TActorId ControllerId;
+        const TClusterBalancingSettings Settings;
+
+        struct TStorageInfo {
+            std::unordered_set<TGroupId> HealthyGroups;
+            std::unordered_map<TPDiskId, ui32> PDiskUsageMap;
+            ui32 ReplicatingPDisks;
```
serbel324: Misleading naming: PDisk cannot be replicating, I suggest renaming to something like PDisksWithReplicatingVDisks

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,460 @@
...
+        }
+
+        template <typename TEventType>
+        THolder<typename TEventType::THandle> WaitForResponse() {
+            // TODO: Wait with deadline when CoroActor's deadline is fixed
+            // For now this doesn't return nullptr, hence this function doesn't return an empty holder.
+            return WaitForSpecificEvent<TEventType>(&ProcessUnexpectedEvent/*, NActors::TMonotonic::Now() + TIMEOUT*/);
+        }
+
+        void Yield(ui64 timeoutMs = 0) {
```
serbel324: Why do you pass ui64 if you convert it to TDuration anyways?
```suggestion
        void Yield(TDuration timeout = TDuration::Zero()) {
```
I don't insist on that change, but I think it would be better this way

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,460 @@
...
+            auto* event = new TEvResume();
+
+            if (timeoutMs > 0) {
+                Schedule(TDuration::MilliSeconds(timeoutMs), event);
```
serbel324:
```suggestion
            if (timeoutMs > TDuration::Zero()) {
                Schedule(timeout, event);
```
see above

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,460 @@
...
+        : TActorCoroImpl(/* stackSize */ 64_KB, /* allowUnhandledDtor */ false)
+        , ControllerId(controllerId)
+        , Settings(settings) {}
+
+        void Run() override {
+            try {
+                while (true) {
+                    RunBalancing();
+                    
+                    Yield(Settings.IterationIntervalMs);
```
serbel324:
```suggestion
                    Yield(TDuration::MilliSeconds(Settings.IterationIntervalMs));
```
see above

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,460 @@
...
+                }
+            }
+
+            auto groupSlotsOrdered = OrderVSlotsByPDiskUsage(candidateVSlots, storageInfo);
+
+            if (groupSlotsOrdered.empty()) {
+                // No groups to balance.
+                return;
+            }
+
```
serbel324: redundant check, if groupSlotsOrdered is empty for cycle won't do any iterations anyways
```suggestion
```

ydb/core/mind/bscontroller/cluster_balancing.cpp
```diff
@@ -0,0 +1,460 @@
...
+            TPDiskId pdiskFrom(reassigned.GetFrom().GetNodeId(), reassigned.GetFrom().GetPDiskId());
+            TPDiskId pdiskTo(reassigned.GetTo().GetNodeId(), reassigned.GetTo().GetPDiskId());
+
+            const auto itFrom = storageInfo.PDiskUsageMap.find(pdiskFrom);
+            ui32 usageFrom = (itFrom != storageInfo.PDiskUsageMap.end()) ? itFrom->second : 0;
+
+            const auto itTo = storageInfo.PDiskUsageMap.find(pdiskTo);
+            ui32 usageTo = (itTo != storageInfo.PDiskUsageMap.end()) ? itTo->second : 0;
+
+            if (usageTo + 1 > usageFrom - 1) {
```
serbel324: It's not guaranteed that reassign after this dry run will pick the same PDisk or improve balance at all. I suggest to add a new parameter to ReassignDisk request, which will discard the transaction if VDisk is assigned to PDisk with higher usage. This way we will assure that balancing strictly improves usage distribution and we won't do extra work with dry run before actual request.

serbel324 (approved):

PR #18864, opened 2025-05-26
Add cluster state info into state storage config

alexvru:
### Changelog entry 

Add cluster state info into state storage config

### Changelog category 

### Description for reviewers 

This patch adds necessary fields into StateStorage configuration to ease determination of the bridged cluster state.


mregrock (approved):
LGTM

PR #18935, opened 2025-05-28
fix stream lookup square and add simple overload checker v2

gridnevvvit:
### Changelog category

### Description for reviewers 

...


gridnevvvit:
@copilot, can you find a bug introduced in this PR?

PR #19030, opened 2025-05-29
Added processing sys view path type in GRPC handlers

### LINKED ISSUE #18896: Support system view path type in GRPC handlers
Check all GRPC handlers which get response from SchemeCache or SchemeShard for code processing ordinal tables.
Go over these places and think how to process system views there.

n00bcracker:
This PR is one from the row to support new system view path type all around YDB.
In previous PRs I added this new type. In the future we want to remove SystemViewInfo field from TableId and use the kind of Navigate entry to process only system view paths.
In this PR:
- updated conditions where we check whether the Navigate entry is system view or not
- added new Entry type in ydb_scheme proto to support describe path and table RPC requests

### DIFF
--- a/ydb/core/grpc_services/rpc_read_rows.cpp
+++ b/ydb/core/grpc_services/rpc_read_rows.cpp
@@ -385,7 +385,7 @@ public:
         OwnerId = entry.Self->Info.GetSchemeshardId();
         TableId = entry.Self->Info.GetPathId();
 
-        if (entry.TableId.IsSystemView()) {
+        if (entry.TableId.IsSystemView() || entry.Kind == NSchemeCache::TSchemeCacheNavigate::KindSysView) {
             return ReplyWithError(Ydb::StatusIds::SCHEME_ERROR,
                 Sprintf("Table '%s' is a system view. ReadRows is not supported.", GetTable().c_str()));
         }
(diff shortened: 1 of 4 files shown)


ydb/public/api/protos/ydb_scheme.proto
```diff
@@ -67,6 +67,7 @@ message Entry {
         VIEW = 20;
         RESOURCE_POOL = 21;
         TRANSFER = 23;
+        SYS_VIEW = 24;
```
azevaykin, line 70: Это используется в текущем PR?

ydb/core/grpc_services/rpc_read_rows.cpp
```diff
@@ -592,7 +592,7 @@ class TReadRowsRPC : public TActorBootstrapped<TReadRowsRPC> {
             }
             case NScheme::NTypeIds::Decimal: {
                 return NYdb::TTypeBuilder().Decimal(NYdb::TDecimalType(
-                        typeInfo.GetDecimalType().GetPrecision(), 
+                        typeInfo.GetDecimalType().GetPrecision(),
```
azevaykin, line 595:
```suggestion
                        typeInfo.GetDecimalType().GetPrecision(), 
```

Да, пробел лишний. Но предлагаю не добавлять неревантных изменений.

ydb/public/api/protos/ydb_scheme.proto
n00bcracker: Да, неявно.
Вот [тут](https://github.com/n00bcracker/ydb/blob/4ed0fabeb553fac7bb4005c832e29a5c7657bc69/ydb/core/ydb_convert/ydb_convert.cpp#L1018). Эта функция вызывается в различных rpc_describe_*

azevaykin (approved):

PR #19124, opened 2025-06-02
add support of check canonical results in `ydb workload query run` command.

iddqdex:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 
https://github.com/ydb-platform/ydb/issues/18892
...


ydb/library/workload/query/query.cpp
```diff
@@ -60,6 +59,13 @@ There is example of import directory: https://github.com/ydb-platform/ydb/tree/m
         return R"(Run load testing.
 Executing load testing using queries from files in the "run" directory or directly from the command line via the "--query" parameter.
 
+Files with "sql" and "yql" extensions will be used for generate queries. For each one can be set canonical result using file with same name and addition extension ".result". That is a CSV-with-headers formatted files with some addition syntax:
```
pnv1: Files with the "sql" and "yql" extensions will be used to generate queries. For each one, a canonical result can be set using a file with the same name and the additional ".result" extension. These are CSV files with headers and some additional syntax:

- If a query has more than one result set, the result file should contain the same number of data sets, separated by empty lines.
- The last line may be set to "...", which means the query result can have more rows, but only the first ones will be checked.
- By default, floating-point numbers are compared with a relative accuracy of 1e-3 percent, but you can specify any absolute or relative accuracy like: "1.5+-0.01", "2.4e+10+-1%".

The canonical result will not be used unless the "--check-canonical" flag is set.

pnv1 (approved):

PR #19216, opened 2025-06-03
Add basic check for group names

yurikiselev:
### Changelog entry 
Always make some checks for group names when creating or renaming them: they may consists of Latin letters in both cases, digits, underscore and hyphen.

### Changelog category 
* Improvement

### Description for reviewers
Affects on checks provided at the first launch of YDB https://ydb.tech/docs/ru/reference/configuration/security_config?version=main#security-bootstrap

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__init_root.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__init_root.cpp
@@ -69,7 +69,7 @@ struct TSchemeShard::TTxInitRoot : public TSchemeShard::TRwTxBase {
             auto response = Self->LoginProvider.CreateGroup({
                 .Group = defaultGroup.GetName(),
                 .Options = {
-                    .CheckName = false
+                    .StrongCheckName = false
                 }
             });
             if (response.Error) {
--- a/ydb/library/login/login.cpp
+++ b/ydb/library/login/login.cpp
@@ -68,8 +68,19 @@ TLoginProvider::TLoginProvider(const TPasswordComplexity& passwordComplexity,
 TLoginProvider::~TLoginProvider()
 {}
 
-bool TLoginProvider::CheckAllowedName(const TString& name) {
-    return name.find_first_not_of("abcdefghijklmnopqrstuvwxyz0123456789") == std::string::npos;
+bool TLoginProvider::StrongCheckAllowedName(const TString& name) {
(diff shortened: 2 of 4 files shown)


CyberROFL (dismissed):

ydb/core/tx/schemeshard/schemeshard__init_root.cpp
```diff
@@ -69,7 +69,7 @@ struct TSchemeShard::TTxInitRoot : public TSchemeShard::TRwTxBase {
             auto response = Self->LoginProvider.CreateGroup({
                 .Group = defaultGroup.GetName(),
                 .Options = {
-                    .CheckName = false
+                    .StrongCheckName = false
```
molotkov-and, line 72: You changed the semantics of this flag. As I understand it, you want to check names with different complexity. Then you should create new `option`

ydb/library/login/login.cpp
```diff
@@ -93,7 +104,7 @@ bool TLoginProvider::CheckPasswordOrHash(bool IsHashedPassword, const TString& u
 TLoginProvider::TBasicResponse TLoginProvider::CreateUser(const TCreateUserRequest& request) {
     TBasicResponse response;
 
-    if (!CheckAllowedName(request.User)) {
+    if (!StrongCheckAllowedName(request.User)) {
```
molotkov-and, line 107: Why do usernames can not contain upper case characters and `-`, `_` or other symbols?
I think usernames and group names must be checked in the same way

yurikiselev: > I think usernames and group names must be checked in the same way

Probably yes. I discussed with Ivan @ijon that in that issue we add basic checks for the groups. The alphabet expansion can be done separately.
So I've added checks that the default group names will pass. I can remove the underscore for now, so the checks would be more similar.

ydb/core/tx/schemeshard/schemeshard__init_root.cpp
yurikiselev: Actually `TRenameGroupRequest.TOptions` was not used at all (CheckName is true always) – only `TCreateGroupRequest.TOptions` was used. So there were no way to change name on the "forbidden" for users.
So the PR will affect only group creation and in this case it looks like fix rather then improvement with a new option suggested.

molotkov-and (approved):

CyberROFL (approved):

PR #19444, opened 2025-06-06
Introduce new fullsync protocol, which takes snapshot only once per synchronization

serbel324:
### Changelog entry 

Introduce new UnorderedData protocol of full synchronization which allows to take hull index snapshot only once per full sync session. New protocol will be used to send DoNotKeep flags from Phantom Flag Storage, which will be sent without relying on linear order. For now the new protocol is disabled by default.

### Changelog category 

* Experimental feature


ydb/core/blobstorage/vdisk/skeleton/blobstorage_syncfull.cpp
```diff
@@ -30,58 +30,87 @@ namespace NKikimr {
         TIntrusivePtr<TBarriersSnapshot::TBarriersEssence> BarriersEssence;
     };
 
+    class TFrequentlyCalledHPTimer {
```
alexvru: How about moving it to common code?

ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
```diff
@@ -1829,6 +1821,14 @@ namespace NKikimr {
         }
 
         void SkeletonIsUpAndRunning(const TActorContext &ctx, bool runRepl = false) {
+            // now, when we obtained DbBirthLsn, we can create full sync handler actor
+            // run handler in the same mailbox
+            if (DbBirthLsn) {
+                Db->SyncFullHandlerID.Set(ctx.RegisterWithSameMailbox(CreateHullSyncFullHandler(Db, HullCtx,
```
alexvru, line 1829: What happens if there is no DbBirthLsn at the time?

ydb/core/blobstorage/vdisk/skeleton/blobstorage_syncfullhandler.cpp
```diff
@@ -12,141 +12,264 @@ namespace NKikimr {
...
+                case NKikimrBlobStorage::EFullSyncProtocol::UnorderedData:
+                    if (ev->Get()->IsInitial()) {
+                        HandleInitialEventUnorderedDataProtocol(ev);
+                    } else {
+                        HandleConsequentEventUnorderedDataProtocol(ev);
+                    }
+                    break;
+                default:
+                    // unknown protocol, report incompatibility and die
+                    RespondWithErroneousStatus(ev, {}, NKikimrProto::BLOCKED);
```
alexvru: Why BLOCKED?

ydb/core/blobstorage/vdisk/skeleton/blobstorage_syncfullhandler.cpp
```diff
@@ -12,141 +12,264 @@ namespace NKikimr {
...
-            auto aid = ctx.Register(actor);
-            ActiveActors.Insert(aid, __FILE__, __LINE__, ctx, NKikimrServices::BLOBSTORAGE);
-            Become(&TThis::StateFunc);
+        std::optional<TActorId> DeleteUnorderedDataSession(const TSessionKey& sessionKey) {
+            auto it1 = UnorderedDataFullSyncSessions.find(sessionKey);
+            if (it1 != UnorderedDataFullSyncSessions.end()) {
+                // Recipient demanded fullsync restart, kill existing actor
+                TActorId actorId = it1->second;
+                Send(actorId, new TEvents::TEvPoisonPill);
+                auto it2 = UnorderedDataFullSyncSessionLookup.find(actorId);
```
alexvru: You can just call erase with a key and check return value to ensure item was actually deleted.

ydb/core/blobstorage/vdisk/skeleton/blobstorage_syncfullhandler.cpp
```diff
@@ -12,141 +12,264 @@ namespace NKikimr {
...
+                    // unknown protocol, report incompatibility and die
+                    RespondWithErroneousStatus(ev, {}, NKikimrProto::BLOCKED);
+                    
+            }
+        }
+
+
+        void Handle(const TEvVGenerationChange::TPtr& ev) {
+            SelfVDiskId = ev->Get()->NewVDiskId;
+            // We must restart all fullsync sessions when VDisk generation changes
```
alexvru, line 261: Do we really need to restart them? I'd suppose to identify clients by order number and incarnation guid and restart only when that guid changes. Make sessions more consistent.

ydb/core/blobstorage/vdisk/skeleton/blobstorage_syncfullhandler.cpp
```diff
@@ -12,141 +12,264 @@ namespace NKikimr {
...
+                // Recipient demanded fullsync restart, kill existing actor
+                TActorId actorId = it1->second;
+                Send(actorId, new TEvents::TEvPoisonPill);
+                auto it2 = UnorderedDataFullSyncSessionLookup.find(actorId);
+                Y_VERIFY(it2 != UnorderedDataFullSyncSessionLookup.end());
+                UnorderedDataFullSyncSessions.erase(it1);
+                UnorderedDataFullSyncSessionLookup.erase(it2);
+                return actorId;
+            }
+            return TActorId{};
```
alexvru: Why `optional<TActorId>` then?

alexvru (changes requested):

ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
serbel324: It only happens if VDisk is donor, but in that case we can't start syncing with this VDisk

ydb/core/blobstorage/vdisk/skeleton/blobstorage_syncfullhandler.cpp
serbel324: I just used it as previously unused erroneous status, but I guess you right, it's not semantically appropriate to use BLOCKED here

ydb/core/blobstorage/vdisk/skeleton/blobstorage_syncfullhandler.cpp
serbel324: When generation changes it restarts on client side anyways
https://github.com/ydb-platform/ydb/blob/main/ydb/core/blobstorage/vdisk/syncer/syncer_job_task.cpp#L267

alexvru (approved):

PR #19542, opened 2025-06-10
Split TLoginProvider::LoginUser to several methods

### LINKED ISSUE #15369: Extract login handler from schemeshard local transactions to separate actor
Extract login handler from schemeshard local transactions to separate actor and send to schemeshard result this operation.

Now user password is checked in schemeshard [local transaction](https://github.com/ydb-platform/ydb/blob/main/ydb/core/tx/schemeshard/schemeshard__login.cpp#L74)

In order to verify password [`argon2` algorithm](https://github.com/ydb-platform/ydb/blob/main/ydb/library/login/login.cpp#L746) is used and it uses up a lot of CPU time.

Need extract using login provider functions to separate actor and send to schemeshard result of operation to persist this state

There should be a queue of operations. OVERLOADED status should be returned if the inflight is too big.

yurikiselev:
### Changelog category

* Not for changelog

### Description for reviewers
We need to move cpu-intensive part of LoginUser from SchemeShard local transaction to a new actor. Such cpu-intensive part is password hash verifying. After that LoginProvider would be accessible from different threads.
However it was not intended to work properly in multithread mode. To avoid races we split LoginUser method to retrieve VerifyHash part into a separate method. It's made static now and will be used in a new actor later.

### DIFF
--- a/ydb/library/login/login.cpp
+++ b/ydb/library/login/login.cpp
@@ -38,13 +38,13 @@ public:
     TString GenerateHash(const TString& password) const;
     static bool VerifyHash(const TString& password, const TString& hash);
     bool VerifyHashWithCache(const TLruCache::TKey& key);
-    bool NeedVerifyHash(const TLruCache::TKey& key, TPassportCheckResult* passportCheckResult);
+    bool NeedVerifyHash(const TLruCache::TKey& key, TPasswordCheckResult* checkResult);
     void UpdateCache(const TLruCache::TKey& key, const bool isSuccessVerifying);
 
     void UpdateCacheSettings(const TLoginProvider::TCacheSettings& cacheSettings);
 
 private:
-    void ClearPasswordsCache();
+    void ClearCache();
 };
 
 const THolder<const NArgonish::IArgon2Base> TLoginProvider::TImpl::ArgonHasher = Default<NArgonish::TArgon2Factory>().Create(
@@ -470,22 +470,22 @@ TLoginProvider::TCheckLockOutResponse TLoginProvider::CheckLockOutUser(const TCh
     return response;
 }
 
-bool TLoginProvider::NeedVerifyHash(const TLoginUserRequest& request, TPassportCheckResult* passportCheckResult, TString* passwordHash) {
-    Y_ENSURE(passportCheckResult);
+bool TLoginProvider::NeedVerifyHash(const TLoginUserRequest& request, TPasswordCheckResult* checkResult, TString* passwordHash) {
+    Y_ENSURE(checkResult);
     Y_ENSURE(passwordHash);
 
-    if (FillNoKeys(passportCheckResult)) {
+    if (FillNoKeys(checkResult)) {
         return false;
     }
 
     if (!request.ExternalAuth) {
         const auto* sid = GetUserSid(request.User);
-        if (FillInvalidUser(sid, passportCheckResult)) {
+        if (FillInvalidUser(sid, checkResult)) {
             return false;
         }
 
         *passwordHash = sid->PasswordHash;
-        return Impl->NeedVerifyHash({.User = request.User, .Password = request.Password, .Hash = sid->PasswordHash}, passportCheckResult);
+        return Impl->NeedVerifyHash({.User = request.User, .Password = request.Password, .Hash = sid->PasswordHash}, checkResult);
     }
 
     return false;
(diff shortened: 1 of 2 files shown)


ydb/library/login/login.cpp
```diff
@@ -19,33 +19,41 @@
...
+    TString GenerateHash(const TString& password) const;
+    static bool VerifyHash(const TString& password, const TString& hash);
     bool VerifyHashWithCache(const TLruCache::TKey& key);
+    bool NeedVerifyHash(const TLruCache::TKey& key, TPassportCheckResult* passportCheckResult);
+    void UpdateCache(const TLruCache::TKey& key, const bool isSuccessVerifying);
 
     void UpdateCacheSettings(const TLoginProvider::TCacheSettings& cacheSettings);
+
+private:
+    void ClearPasswordsCache();
```
ijon: `UpdateCache()`, `ClearPasswordsCache()`, `UpdateCacheSettings()`, and `VerifyHashWithCache()` all refer to the same cache, but their naming is inconsistent.

I suggest removing "Passwords" from `ClearPasswordsCache()` to maintain naming consistency.

ydb/library/login/login.cpp
```diff
@@ -763,24 +833,48 @@ bool TLoginProvider::TImpl::VerifyHash(const TString& password, const TString& p
...
+    if (WrongPasswordsCache.Find(key) != WrongPasswordsCache.End()) {
+        passportCheckResult->Status = TLoginUserResponse::EStatus::INVALID_PASSWORD;
+        passportCheckResult->Error = "Invalid password";
+        return false;
+    }
+
+    return true;
+}
+
+void TLoginProvider::TImpl::UpdateCache(const TLruCache::TKey& key, const bool isSuccessVerifying) {
```
ijon, line 858: `UpdateCache()` can also be used in `VerifyHashWithCache()`.

ydb/library/login/login.cpp
```diff
@@ -532,10 +589,23 @@ TLoginProvider::TLoginUserResponse TLoginProvider::LoginUser(const TLoginUserReq
...
         sid->FailedLoginAttemptCount = 0;
     }
-
     return response;
 }
 
+TLoginProvider::TLoginUserResponse TLoginProvider::LoginUser(const TLoginUserRequest& request) {
+    TPassportCheckResult passportCheckResult;
+    TString passwordHash;
+    if (NeedVerifyHash(request, &passportCheckResult, &passwordHash)) {
```
ijon: I doubt the choice of "passportCheck" term.

The similarity between "passportCheck" and "passwordHash" creates confusion rather than clarity, with "pass-port" and "pass-word" being easily mixed up.

"passwordCheck" or simply "check" would be clearer alternatives.

ijon (approved):

ydb/library/login/login.cpp
```diff
@@ -763,33 +833,53 @@ bool TLoginProvider::TImpl::VerifyHash(const TString& password, const TString& p
...
+        return VerifyHash(key.Password, key.Hash);
+    }
+
+    if (SuccessPasswordsCache.Find(key) != SuccessPasswordsCache.End()) {
+        return true;
+    }
+
+    if (WrongPasswordsCache.Find(key) != WrongPasswordsCache.End()) {
+        return false;
+    }
```
molotkov-and, line 879: Can we replace this code by `NeedVerifyHash` function?

ydb/library/login/login.cpp
```diff
@@ -462,30 +470,78 @@ TLoginProvider::TCheckLockOutResponse TLoginProvider::CheckLockOutUser(const TCh
...
     TSidRecord* sid = nullptr;
     if (!request.ExternalAuth) {
-        auto itUser = Sids.find(request.User);
-        if (itUser == Sids.end() || itUser->second.Type != ESidType::USER) {
-            response.Status = TLoginUserResponse::EStatus::INVALID_USER;
-            response.Error = "Invalid user";
+        sid = GetUserSid(request.User);
+        if (FillInvalidUser(sid, &response)) {
             return response;
         }
```
molotkov-and, line 539: This code is repeated in `TLoginProvider::NeedVerifyHash`. Can we do these checks one time?

yurikiselev: Unfortunately, no. In the future, they may be called from different transactions. And for example, the presence of a user in the first call does not guarantee its presence in the second call.

ydb/library/login/login.cpp
yurikiselev: They look similar, but they make different verdicts under the same conditions. If you combine them, it will be quite confusing – I've tried.

molotkov-and (approved):

PR #19545, opened 2025-06-10
Fix example configs in deploy

mregrock:
### Changelog entry 

Fix example configs in deploy

### Changelog category 

### Description for reviewers 

...


blinkov (approved):
Assuming that further improvements will be in a separate PR.

PR #19688, opened 2025-06-16
Report self node pile for dynamic nodes too

alexvru:
### Changelog entry 

Report self node pile for dynamic nodes too

### Changelog category 

### Description for reviewers 

This patch allows reporting BridgePileId through SelfNodePile when checked on dynamic node too.


mregrock (approved):
LGTM

PR #19741, opened 2025-06-17
[WIP] comp topic

FloatingCrowbar:
- Enable creation of compacted topics
- Support altering of compacted topics and add tests
- Inter commit

### Changelog entry 

...

### Changelog category 

* New feature

### Description for reviewers 

...

### DIFF
--- a/ydb/services/lib/actors/pq_schema_actor.cpp
+++ b/ydb/services/lib/actors/pq_schema_actor.cpp
@@ -864,7 +864,6 @@ namespace NKikimr::NGRpcProxy::V1 {
                 haveCompConsumer = true;
             }
         }
-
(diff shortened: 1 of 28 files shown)


ydb/services/lib/actors/pq_schema_actor.cpp
```diff
@@ -1347,6 +1372,13 @@ namespace NKikimr::NGRpcProxy::V1 {
                 consumer.mutable_supported_codecs()->add_codecs(codec + 1);
             }
         }
+        if (compactedTopic && !hasCompactionConsumer) {
```
alexnick88: What if are there hasCompationcConsumer=true and comapctedTopic=false?

PR #19918, opened 2025-06-19
#9056 Complete column names in YDB CLI

vityaman:
### Changelog category 

### Description for reviewers 

Complete column names in YDB CLI

Implemented `DescribeTable` using the YDB Client. Now supported only `SELECT a| FROM table_name`. `SELECT | FROM table_name` is not working because of current limitations of the completion engine (it interprets `FROM` as a column name and then fails to parse a table name after from), so at least one letter of column name is required. Also `SELECT p.a| FROM table_name AS p` is working. 

- Related to `YQL-19747`
- Related to https://github.com/ydb-platform/ydb/issues/9056


vityaman:
@pnv1, the PR is ready for a review.

pnv1 (approved):

PR #19934, opened 2025-06-19
Save import items count after getting it from schema mapping

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

We need to save save items count setting after we read SchemaMapping, because it changes items count


CyberROFL (approved):
Можно ли воспроизвести проблему в тесте?

PR #20191, opened 2025-06-25
Correct Scan Generation in case of retries

Hor911:
#20162


gridnevvvit (changes requested):
consider implementing unit test on this

gridnevvvit (approved):

PR #20193, opened 2025-06-25
Correct Column Shard count and finished status in statistics

Hor911:
Report actual Column Shard count and their Finished status


Hor911:
```
select count(*) from lineitem
```

Before:
query-plan (12)

After:
query-plan (14)


```
select * from lineitem limit 1000
```

Before:
query-plan (13)

After:
query-plan (15)

lll-phill-lll (approved):

pnv1 (approved):

PR #20278, opened 2025-06-27
Fix single node configs

mregrock:
### Changelog entry 

Fix single node configs

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/deploy/yaml_config_examples/single-node-in-memory.yaml
+++ b/ydb/deploy/yaml_config_examples/single-node-in-memory.yaml
@@ -3,20 +3,12 @@ metadata:
   cluster: ""
   version: 0
 config:
-  erasure: none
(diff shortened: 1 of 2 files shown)


blinkov (changes requested):
This can be further simplified. Here's what they looked like in V1:
- RAM: https://github.com/ydb-platform/ydb/blob/main/ydb/library/yaml_config/ut_transform/simplified_configs/ram.yaml
- File: https://github.com/ydb-platform/ydb/blob/main/ydb/library/yaml_config/ut_transform/simplified_configs/disk.yaml
- Drive: https://github.com/ydb-platform/ydb/blob/main/ydb/library/yaml_config/ut_transform/simplified_configs/disk.yaml

blinkov (approved):

PR #19674, opened 2025-06-15
Add Extended Info field to OIDC whoami response

StekPerepolnen:
### Changelog entry 

Add Extended Info field to OIDC whoami response

### Changelog category 

* Improvement

### Description for reviewers 

This change introduces optional enrichment of `/viewer/whoami` via a new config field `whoami_extended_info_endpoint`. If set, the proxy fetches profile data from IAM concurrently with the Viewer call. Both responses are merged, adding federation, email, and other user info. The extended info is placed into the `ExtendedInfo` field of the response if available.

**Diagnostic fields from the YDB Viewer call:**
`ExtendedErrors.Ydb.ResponseStatus` - HTTP status from the Viewer, present only if a response was received.
`ExtendedErrors.Ydb.ResponseBody` - Viewer response body as JSON string; may contain diagnostics if error or no body.
`ExtendedErrors.Ydb.ClientError` - error message generated by the HTTP library

**Diagnostic fields from the IAM call:**
`ExtendedErrors.Iam.ResponseStatus` - gRPC status from failed IAM call.
`ExtendedErrors.Iam.ResponseMessage` - error message from IAM.
`ExtendedErrors.Iam.ResponseDetails` - optional error details from IAM.
`ExtendedErrors.Iam.ClientError` - timeout error message

PR #20475, opened 2025-07-01
oidc: remove actor from final extension

StekPerepolnen:
### Changelog entry 

refactoring after https://github.com/ydb-platform/ydb/pull/19674 . it's better to avoid use extra actor in final extension

### Changelog category 

* Improvement

### Description for reviewers 

In each extension, a new actor was used, which is redundant - final extension can be executed locally, and a new actor is not needed. I shifted the responsibility of executing locally or in a new actor to the individual extensions.

Each extension must have a single Execute method - when this method is called, the extension handles the context in some way: either it processes it itself, or delegates the entire work to a separate worker, like in the case of whoami.

Added Timeout override map


ydb/mvp/oidc_proxy/extension.h
```diff
@@ -1,30 +1,59 @@
...
+    TString MessageOverride;
+    TString BodyOverride;
+    THolder<NHttp::THeadersBuilder> HeadersOverride;
+};
+
+struct TExtensionsSteps: public TQueue<std::unique_ptr<TExtension>> {
+    std::unique_ptr<TExtension> Next();
+};
+
+struct TExtensionContext : public TThrRefBase {
```
UgnineSirdis, line 32: ` : public` or `: public` is both OK, but please do in one style)

ydb/mvp/oidc_proxy/extension.h
```diff
@@ -1,30 +1,59 @@
...
+    TActorId Sender;
+    TExtensionsSteps Steps;
+    THolder<TProxiedResponseParams> Params;
+};
+
+class TExtension {
```
UgnineSirdis: `IExtension` - these methods are all pure virtual (destructor is OK if default).

ydb/mvp/oidc_proxy/extension.h
```diff
@@ -1,30 +1,59 @@
...
+    TString BodyOverride;
+    THolder<NHttp::THeadersBuilder> HeadersOverride;
+};
+
+struct TExtensionsSteps: public TQueue<std::unique_ptr<TExtension>> {
+    std::unique_ptr<TExtension> Next();
+};
+
+struct TExtensionContext : public TThrRefBase {
+    TActorId Sender;
```
UgnineSirdis, line 33: Who is sender here?

ydb/mvp/oidc_proxy/extension.h
```diff
@@ -1,30 +1,59 @@
...
+class TExtensionWorker {
 protected:
     const TOpenIdConnectSettings Settings;
     TIntrusivePtr<TExtensionContext> Context;
 
+    void Reply(NHttp::THttpOutgoingResponsePtr httpResponse);
+    void Reply();
+    void Continue();
+
 public:
```
UgnineSirdis, line 43: I think that `TExtensionWorker` is whole for derivation, so we don't need public contructor.

ydb/mvp/oidc_proxy/extension.h
```diff
@@ -1,30 +1,59 @@
...
+    THolder<TProxiedResponseParams> Params;
+};
+
+class TExtension {
+public:
+    virtual ~TExtension() = default;
+    virtual void Execute(TIntrusivePtr<TExtensionContext> ctx) = 0;
+};
+
+class TExtensionWorker {
```
UgnineSirdis: TExtensionImpl?

UgnineSirdis (changes requested):
I liked this chain concept - I think it can make things simpler

ydb/mvp/oidc_proxy/extension.cpp
```diff
@@ -2,34 +2,37 @@
 
 namespace NMVP::NOIDC {
 
-void TExtension::Bootstrap() {
-    Become(&TExtension::StateWork);
+void TExtensionWorker::Reply(NHttp::THttpOutgoingResponsePtr httpResponse) {
+    NActors::TActivationContext::ActorSystem()->Send(Context->Sender, new NHttp::TEvHttpProxy::TEvHttpOutgoingResponse(std::move(httpResponse)));
```
UgnineSirdis: It can be replaced with Send from current context, like
`NActors::TActivationContext::Send(Context->Sender, new NHttp::TEvHttpProxy::TEvHttpOutgoingResponse(std::move(httpResponse)));`

UgnineSirdis (changes requested):

ydb/mvp/oidc_proxy/extension.h
StekPerepolnen: removed TExtensionWorker

ydb/mvp/oidc_proxy/extension.h
StekPerepolnen: removed TExtensionWorker

ydb/mvp/oidc_proxy/extension.cpp
StekPerepolnen: changed

ydb/mvp/oidc_proxy/extension.h
StekPerepolnen: changed

ydb/mvp/oidc_proxy/extension.h
StekPerepolnen: Sender is the original sender of the IncomingRequest

ydb/mvp/oidc_proxy/oidc_settings.h
```diff
@@ -18,27 +19,30 @@ struct TOpenIdConnectSettings {
...
     TString ClientId = DEFAULT_CLIENT_ID;
     TString SessionServiceEndpoint;
     TString SessionServiceTokenName;
     TString AuthorizationServerAddress;
     TString ClientSecret;
     std::vector<TString> AllowedProxyHosts;
     TString WhoamiExtendedInfoEndpoint;
-    TDuration RequestWithEnrichmentTimeout = DEFAULT_REQUEST_WITH_ENRICHMENT_TIMEOUT;
     TDuration RequestTimeout = DEFAULT_REQUEST_TIMEOUT;
+    THashMap<TStringBuf, TDuration> TimeoutOverrides;
```
UgnineSirdis: `THashMap<TString, TDuration>`
TStringBuf can refer to a buffer that we need to store the whole life time, it is not very convenient and practical for all cases

ydb/mvp/oidc_proxy/oidc_settings.cpp
```diff
@@ -33,6 +33,14 @@ const TVector<TStringBuf> TOpenIdConnectSettings::RESPONSE_HEADERS_WHITE_LIST =
     "traceresponse"
 };
 
+void TOpenIdConnectSettings::InitTimeoutOverrides() {
+    TimeoutOverrides.clear();
+    if (AccessServiceType == NMvp::nebius_v1) {
+        for (auto path : WHOAMI_PATHS) {
+            TimeoutOverrides[path] = TDuration::Seconds(10);
```
UgnineSirdis: Do we have settings with timeouts?
For example, like `Settings.RequestTimeout` from `THandlerSessionServiceCheck::GetRequestTimeout()` method

ydb/mvp/oidc_proxy/oidc_settings.cpp
```diff
@@ -33,6 +33,14 @@ const TVector<TStringBuf> TOpenIdConnectSettings::RESPONSE_HEADERS_WHITE_LIST =
     "traceresponse"
 };
 
+void TOpenIdConnectSettings::InitTimeoutOverrides() {
+    TimeoutOverrides.clear();
+    if (AccessServiceType == NMvp::nebius_v1) {
```
UgnineSirdis: Could you please add a comment why we set timeout like this and do it only for Nebius case?

ydb/mvp/oidc_proxy/oidc_protected_page.cpp
```diff
@@ -105,19 +105,28 @@ bool THandlerSessionServiceCheck::IsAuthorizedRequest(TStringBuf authHeader) {
     return to_lower(ToString(authHeader)).StartsWith(IAM_TOKEN_SCHEME_LOWER);
 }
 
+TDuration THandlerSessionServiceCheck::GetRequestTimeout() const {
+    auto path = TStringBuf(ProtectedPage.Url).Before('?');
+    for (const auto& [suffix, timeout] : Settings.TimeoutOverrides) {
```
UgnineSirdis: - TimeoutOverrides is already a map that allows search faster than linear. Why can't we use it here?
- Do we have case dependent paths? I bet yes, but I focus your attenttion for it, just in case :)

ydb/mvp/oidc_proxy/oidc_protected_page.cpp
```diff
@@ -105,19 +105,28 @@ bool THandlerSessionServiceCheck::IsAuthorizedRequest(TStringBuf authHeader) {
     return to_lower(ToString(authHeader)).StartsWith(IAM_TOKEN_SCHEME_LOWER);
 }
 
+TDuration THandlerSessionServiceCheck::GetRequestTimeout() const {
+    auto path = TStringBuf(ProtectedPage.Url).Before('?');
```
UgnineSirdis, line 109: There are parsers in repository (library or util) that can parse the whole url and take path robustly. Can we use them here?

ydb/mvp/oidc_proxy/oidc_protected_page.cpp
```diff
@@ -105,19 +105,28 @@ bool THandlerSessionServiceCheck::IsAuthorizedRequest(TStringBuf authHeader) {
...
+    for (const auto& [suffix, timeout] : Settings.TimeoutOverrides) {
+        if (path.EndsWith(suffix)) {
+            return timeout;
```
UgnineSirdis, line 112: I think that for future extension we wanted std::function(url) as a map value.
We wanted to be able to process different cases like:
- parsing url;
- take parameters with client timeout (example: query handler that has client timeout parameter);
- return result depending on this timeout set by client.

But it is up to you: we yet don't have such processing. Maybe it is better to add them when we will have such a case.

ydb/mvp/oidc_proxy/oidc_settings.h
```diff
@@ -18,27 +19,30 @@ struct TOpenIdConnectSettings {
...
 
     NMvp::EAccessServiceType AccessServiceType = NMvp::yandex_v2;
     TString AuthUrlPath = DEFAULT_AUTH_URL_PATH;
     TString TokenUrlPath = DEFAULT_TOKEN_URL_PATH;
     TString ExchangeUrlPath = DEFAULT_EXCHANGE_URL_PATH;
     TString ImpersonateUrlPath = DEFAULT_IMPERSONATE_URL_PATH;
 
+    void InitTimeoutOverrides();
```
UgnineSirdis: How do you think, maybe it is better to name it like simply "timeout handlers"? We don't override here anything

UgnineSirdis (changes requested):

ydb/mvp/oidc_proxy/oidc_settings.h
StekPerepolnen: changed

ydb/mvp/oidc_proxy/oidc_protected_page.cpp
StekPerepolnen: You're right, TimeoutOverrides is a map, but since the logic here matches suffixes (not full keys), we can't use find() or direct lookup - it's not a key-based match. The URL often contains dynamic or variable path segments like balancer, so suffix-based matching is intentional.

And yes, good point about case sensitivity - I will change it

ydb/mvp/oidc_proxy/oidc_settings.h
StekPerepolnen: changed to RequestTimeoutsByPath

ydb/mvp/oidc_proxy/oidc_protected_page.cpp
StekPerepolnen: I’d leave it as is for now, and we can revisit it later.

ydb/mvp/oidc_proxy/oidc_protected_page.cpp
StekPerepolnen: Honestly, this kind of straightforward parsing is quite common. Even in the actor library, it's the same approach that's used.

ydb/mvp/oidc_proxy/oidc_settings.cpp
StekPerepolnen: added

ydb/mvp/oidc_proxy/oidc_settings.cpp
StekPerepolnen: No, we don't have such settings defined elsewhere.

UgnineSirdis (approved):

PR #20514, opened 2025-07-02
Fix unexpected unique constraint violation errors from datashards

### LINKED ISSUE #19531: DataShard: EvWrite causes unique constraint violation when TLI is expected
The following test case:

* One table where a counter column is incremented
* Another table where a row based on that column is inserted

Produces unique constrain violations under load instead of TLI when EvWrite with insert is used. The problem is that old code uses reads from snapshot and doesn't observe duplicate rows, and commit fails with TLI. The new code fuses insert with commit, which performs read at commit timestamp. However, there's an inconsistency:

* Reading from the commit timestamp implies we can safely shift all reads to commit timestamp without changing observable results
* However, that cannot happen, since another transaction already committed and changed the column
* We cannot wait for locks from another shard before executing, and cannot know whether it's TLI or not
* We want to read from HEAD when possible, to minimize conflicts with other transactions
* We need to read from snapshot for insert in this case, since row didn't exist at snapshot time, and it existing now implies it's TLI, not unique constraint violation

I'm not sure, but it could be possible to have both. Try to execute local transaction body with reads at the commit timestamp. When everything is good we can start committing locally and wait for other participants. When other participants confirm, everything consistently happened at commit timestamp. Otherwise, before replying with an error, retry transaction with reads at snapshot time, chances are we will have TLI or a different error, and we should reply with that instead of the original error.

We could possibly try doing both at the same time. Use a transaction observer to note which versions are applied to rows during reads. When it's some version that is higher than the snapshot version, we know that retrying would have resulted in a TLI due to InvisibleRowSkips, so we can replace whatever it was with ABORTED. If nothing like that happened, then nothing would change as a result of our retry.

Challenges: we would need for KQP to somehow communicate this read timestamp in a special way. Also, we would need to somehow make sure that transaction observer wouldn't miss anything, i.e. `InvisibleRowSkips > 0` would always match `OnApplyCommitted(rowVersion > snapshot)`. The latter is likely, because `InvisibleRowSkips++` is always paired with `OnSkipCommitted`, which implies if it's not skipped it's applied.

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

This patch fixes unexpected unique constraint errors when duplicate key is committed after a transaction acquires its read snapshot. A couple of tests are added: one ensures transaction fails with an ABORTED error when INSERT is performed before the commit, another when INSERT is performed together with the commit (behavior changes when KQP doesn't force flush INSERT before the commit).

To work correctly datashards need MvccSnapshot specified now, so it can make sure writes are performed over the correct version. I found out that EvWrite didn't work correctly when writing over a snapshot (which is chosen by datashard even when MvccSnapshot is not specified), and could produce inconsistent results when more recent changes are present. This could affect CDC and async index correctness.

I'm marking this as "not for changelog" since EvWrite is not enabled in production yet.

Fixes #19531.

### DIFF
--- a/ydb/core/kqp/executer_actor/kqp_executer_impl.h
+++ b/ydb/core/kqp/executer_actor/kqp_executer_impl.h
@@ -1001,7 +1001,6 @@ protected:
                 ActorIdToProto(BufferActorId, settings.MutableBufferActorId());
             }
             if (!settings.GetInconsistentTx()
-                    && TasksGraph.GetMeta().LockMode == NKikimrDataEvents::OPTIMISTIC_SNAPSHOT_ISOLATION
                     && GetSnapshot().IsValid()) {
                 settings.MutableMvccSnapshot()->SetStep(GetSnapshot().Step);
                 settings.MutableMvccSnapshot()->SetTxId(GetSnapshot().TxId);
--- a/ydb/core/kqp/runtime/kqp_write_actor.cpp
+++ b/ydb/core/kqp/runtime/kqp_write_actor.cpp
@@ -996,12 +996,20 @@ public:
         } else if (isPrepare) {
             YQL_ENSURE(TxId);
             FillEvWritePrepare(evWrite.get(), shardId, *TxId, TxManager);
+            // Note: we may need to send MvccSnapshot for certain operations
+            // like INSERT to not produce phantom unique constraint errors.
+            // Older datashards would hard error when MvccSnapshot is specified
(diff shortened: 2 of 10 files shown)


nikvas0 (dismissed):
kqp ok

azevaykin (dismissed):

nikvas0 (approved):

azevaykin (approved):

PR #20539, opened 2025-07-02
Support quorum of specific bridge pile ids when changing cluster state in Bridge mode

alexvru:
### Changelog entry 

Support quorum of specific bridge pile ids when changing cluster state in Bridge mode

### Changelog category 

### Description for reviewers 

This patch allows specifying explicit set of piles to make up quorum when applying cluster state change command.


mregrock (approved):
LGTM

Gazizonoki (approved):
SDK - ok

PR #20744, opened 2025-07-07
Fix cluster permissions check under grpc

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

The new Nebius scheme for checking permissions for cluster root container worked only for http requests (because TEvRequestAuthAndCheck works only in case of http requests). This PR fixes grpc methods (like getting/setting cluster config), if we use AdministrationAllowedSIDs together with folder_id cluster attribute.

### DIFF
--- a/ydb/core/grpc_services/grpc_request_check_actor.h
+++ b/ydb/core/grpc_services/grpc_request_check_actor.h
@@ -141,8 +141,8 @@ public:
         }
 
         if constexpr (std::is_same_v<TEvent, TEvRequestAuthAndCheck>) {
-            TVector<TEvTicketParser::TEvAuthorizeTicket::TEntry> e = GetEntriesForAuthAndCheckRequest(Request_);
-            entries.insert(entries.end(), e.begin(), e.end());
+            TVector<TEvTicketParser::TEvAuthorizeTicket::TEntry> authCheckRequestEntries = GetEntriesForAuthAndCheckRequest(Request_);
+            entries.insert(entries.end(), authCheckRequestEntries.begin(), authCheckRequestEntries.end());
         }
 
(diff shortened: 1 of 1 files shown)


ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -124,7 +141,11 @@ class TGrpcRequestCheckActor
         }
 
         if constexpr (std::is_same_v<TEvent, TEvRequestAuthAndCheck>) {
-            const auto& e = GetEntriesForAuthAndCheckRequest(Request_, rootAttributes);
+            TVector<TEvTicketParser::TEvAuthorizeTicket::TEntry> e = GetEntriesForAuthAndCheckRequest(Request_);
+            entries.insert(entries.end(), e.begin(), e.end());
+        }
+
+        if (TVector<TEvTicketParser::TEvAuthorizeTicket::TEntry> e = GetEntriesForClusterAccessCheck(rootAttributes)) {
```
molotkov-and: Is there need "if" stateman?

UgnineSirdis: "If" statements are in the function. I would suggest not to show this internal if here, because it is too detailed.

ydb/core/grpc_services/grpc_request_check_actor.h
```diff
@@ -124,7 +141,11 @@ class TGrpcRequestCheckActor
...
             entries.insert(entries.end(), e.begin(), e.end());
         }
```
molotkov-and:
```suggestion
        TVector<TEvTicketParser::TEvAuthorizeTicket::TEntry> e = GetEntriesForClusterAccessCheck(rootAttributes);
        entries.insert(entries.end(), e.begin(), e.end());
```

molotkov-and (approved):

StekPerepolnen (approved):

PR #20935, opened 2025-07-10
[CLI] Reverse statistics and operation order in plan print, reduced table print info, fixed child arrows

pashandor789:
[CLI] Reverse statistics and operation order in plan print, reduced table print info, fixed child arrows

### Changelog category 


pavelvelikhov (approved):

pnv1 (approved):
CLI part LGTM

PR #20948, opened 2025-07-10
Invalidate tablet resolver cache on tablet pipe disconnect

### LINKED ISSUE #20931: Invalidate tablet resolver caches more aggressively
We have a problem with VM-based installations, where new VMs are started, tablets are migrated to new VMs and then old VMs shut down. Multiple stable nodes (e.g. storage) will have old tablet addresses still present in the tablet resolver cache, which causes an extremely long first reconnection attempt (since the old IP address no longer responds).

Ideally we would want tablet resolver cache to be coherent and updated/invalidated on all nodes when tablets move, but this is a very complicated protocol change, and we need a quick and dirty solution yesterday.

I think we just need to aggressively invalidate addresses to nodes which (potentially) no longer exist:

1. Pipes should invalidate tablet resolver cache on any node disconnect (already happening).
2. Pipes should invalidate tablet resolver cache on any tablet-initiated pipe disconnect (since that implies tablet restart anyway).
3. Tablet resolver should subscribe to node status for resolved addresses and invalidate those addresses whenever node disconnects.

This shouldn't impact normal operations, since node disconnections without corresponding tablet restarts are too rare.

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

When tablets restart gracefully, e.g. during version upgrades, we may leave old leader addresses cached for a long time, and connecting to those addresses for the first time may take 10s of seconds after a corresponding VM is destroyed. Invalidate addresses when disconnect is initiated from the tablet, which covers most system pipe cache use cases.

Fixes #20931.

### DIFF
--- a/ydb/core/tablet/tablet_resolver.cpp
+++ b/ydb/core/tablet/tablet_resolver.cpp
@@ -427,6 +427,11 @@ class TTabletResolver : public TActorBootstrapped<TTabletResolver> {
                     LOG_DEBUG(ctx, NKikimrServices::TABLET_RESOLVER,
                             "Delayed invalidation of tabletId: %" PRIu64
                             " leader: %s by NodeId", tabletId, entry.KnownLeader.ToString().c_str());
+                    if (entry.KnownFollowers.empty()) {
+                        // Avoid resolving preemptively until the next request
+                        DropEntry(tabletId, entry, ctx);
+                        return;
+                    }
                     ResolveRequest(tabletId, ctx);
                     entry.State = TEntry::StProblemResolve;
                     MoveEntryToUnresolved(tabletId, *entryHolder);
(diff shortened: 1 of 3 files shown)


vporyadke (approved):

ydb/core/tablet/tablet_resolver.cpp
```diff
@@ -427,6 +427,11 @@ class TTabletResolver : public TActorBootstrapped<TTabletResolver> {
                     LOG_DEBUG(ctx, NKikimrServices::TABLET_RESOLVER,
                             "Delayed invalidation of tabletId: %" PRIu64
                             " leader: %s by NodeId", tabletId, entry.KnownLeader.ToString().c_str());
+                    if (entry.KnownFollowers.empty()) {
+                        // Avoid resolving preemptively until the next request
+                        DropEntry(tabletId, entry, ctx);
```
snaury, line 432: This actually introduced a subtle bug: DropEntry adds tabletId to the stop list for 300ms.

ydb/core/tablet/tablet_resolver.cpp
```diff
@@ -541,6 +546,11 @@ class TTabletResolver : public TActorBootstrapped<TTabletResolver> {
             break;
         case TEntry::StNormal:
             if (!msg->Actor || entry.KnownLeader == msg->Actor || entry.KnownLeaderTablet == msg->Actor) {
+                if (entry.KnownFollowers.empty()) {
+                    // Avoid resolving preemptively until the next request
+                    DropEntry(tabletId, entry, ctx);
```
snaury, line 551: This actually introduced a subtle bug: DropEntry adds tabletId to the stop list for 300ms.

Issue #19525, opened 2025-06-09
DataShard: EvWrite in non-volatile mode should use volatile commits anyway


snaury:
We have complex logic in EvWrite, e.g. insert operations may fail with unique constrain violation errors, etc. The old transaction system doesn't expect that, it generates outgoing readsets immediately and waits for incoming readsets, but successful outgoing readsets imply the transaction will commit without errors, which is not true.

We should just always use volatile commit for EvWrite, because it's much more resilient to execution errors. On error we should generate aborting outgoing readsets, and eventually commit when all other participants send us committing readsets. One caveat is that such volatile commit must not use expectations and/or nodata readsets, since e.g. pq tablets might not be ready for them, which might complicate things a little.

PR #20958, opened 2025-07-10
25-1-2: Support volatile transactions without expectations

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

We want to convert persistent distributed transactions to optimistic volatile execution when possible, however persistent transactions don't have any concept of readset expectations (and don't need it), and non-datashard participants might break when receiving them.

This patch doesn't change existing behavior in any way, but will support persistent behavior change in the future. Intended for backporting into current stable versions.

Related to #19525.


snaury:
Decided not to merge to 25-1-2.

PR #20959, opened 2025-07-10
25-1-2: Relax unnecessary checks for TKqpLocks proto fields

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

We historically used SendingShards/ReceivingShards in TKqpLocks for exchanging lock validation results. However, it has long been repurposed for generic commit/abort decision exchange. Now that we have operations (EvWrite with INSERT) which may fail due to constraint violation unrelated to locks, checking that SendingShards must have a lock is harmful. Relax these checks and decouple locks from sending/receiving shards. Now locks listed in the proto are always validated when present on commit (even when shard is not in the sending set for some reason), and readsets are always sent from shards in SendingShards.

This relaxation is safe to perform since current stable versions crash when encountering such data, and this patch doesn't change behavior otherwise. Intended for backporting into current stable versions.

Related to #19525.


snaury:
Decided not to merge to 25-1-2.

PR #21048, opened 2025-07-14
Fix VDisk compaction race

alexvru:
### Changelog entry 

Fix VDisk compaction race

### Changelog category 

### Description for reviewers 

This patch fixes vdisk compaction race leading to possible data corruption when two compactions are started concurrently.


mregrock (approved):
LGTM

PR #21191, opened 2025-07-16
Add pdisk error handler for NoDevice(ENODEV)

va-kuznecov:
### Changelog entry 

Add new handler for error we saw on our cluster. After change we will get a readable error instead of VERIFY

### Changelog category 

* Not for changelog

### Description for reviewers 

...


SammyVimes (approved):
LGTM

PR #21262, opened 2025-07-17
audit logging for http api

StekPerepolnen:
### Changelog entry 

add audit loggint for monitoring HTTP API https://github.com/ydb-platform/ydb/issues/22123

### Changelog category 

* New feature

### Description for reviewers 

Audit logging of all modifying operations in the monitoring HTTP API is based on:
- HTTP method: all requests with POST, DELETE, or PUT are always logged
- A predefined list of requests that are known to be modifying

Audit log is enabled in the AuditConfig. All request parameters and body are logged without filtering. The request body size is limited to 2 MB for logging.

<details><summary>audit log example</summary>
<p>

```
{
    "destination": "eventdestination-e0t-audit-events-ydb",
    "event": {
        "text_data": "{\"@timestamp\":\"2025-07-25T12:18:50.678280Z\",\"@log_type\":\"audit\",\"component\":\"monitoring\",\"remote_address\":\"2a13:5947:111:10::cc7c:545b\",\"operation\":\"HTTP REQUEST\",\"method\":\"POST\",\"url\":\"\\/viewer\\/json\\/query\",\"params\":\"schema=multi&base64=false&timeout=60000\",\"body\":\"{\\\"query\\\":\\\"SELECT `id`, `value`\\\\nFROM `simple_table`\\\\nLIMIT 10;\\\",\\\"database\\\":\\\"\\/ydb-slice-1\\/arykov-test\\\",\\\"action\\\":\\\"execute-query\\\",\\\"syntax\\\":\\\"yql_v1\\\",\\\"stats\\\":\\\"full\\\",\\\"tracingLevel\\\":0,\\\"limit_rows\\\":10000,\\\"timeout\\\":60000,\\\"query_id\\\":\\\"8fd71c3d-4537-498f-b85b-d4b2a0c74928\\\",\\\"base64\\\":false}\",\"subject\":\"tenantuseraccount-e0tg829s16770fh6qkha6@as\",\"sanitized_token\":\"ne1CoQCCh5hY2Nlc3N0b2tlbi1lMHRrcmExdzE4ZDdjeDY4bXgSIXVzZXJhY2NvdW50LWUwdHI3OGJjZ2Z6bng0eTQwOXM1Nhp_ChpzZXNzaW9uLWUwdGh4YjN0ZGUzN2doMHhoORAEGl8KGnNlc3Npb24tZTB0ZG1wYThldnpwN3QyMDBxEAQaPwoZc2VydmljZWFjY291bnQtZTB0aWFtLWNwbBADGiAKHHB1YmxpY2tleS1lMHRwdGNyMjNmMzR2aGdna3gQASocb2lkY2NsaWVudC1lMHRvaWRjLXByb3h5LWRldjIMCKrsjcQGEMfizcACOgsI_4uQxAYQ3ra6dEgCWgNlMHQ.**\",\"status\":\"IN-PROCESS\",\"reason\":\"Temporary Redirect\"}\n",
        "type": "ydb-audit"
    }
}
```

</p>
</details>


ydb/core/mon/audit/url_matcher.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include "url_matcher.h"
+#include <library/cpp/cgiparam/cgiparam.h>
+
+#include <util/generic/hash.h>
+#include <util/generic/string.h>
+#include <util/generic/vector.h>
+#include <optional>
+
+namespace NActors::NAudit {
```
UgnineSirdis: NActor is a common namespace in common library. I am not sure that it is a good idea to extend this namespace here

ydb/core/mon/audit/url_matcher.h
```diff
@@ -0,0 +1,40 @@
...
+    TString Path;
+    TString ParamName;
+    TString ParamValue;
+};
+
+class TUrlMatcher {
+public:
+    void AddPattern(const TUrlPattern& rule);
+    bool Match(const TString& url, const TCgiParameters& params) const;
+    bool Match(const TString& url, const TString& params = "") const;
```
UgnineSirdis: `bool Match(const TString& url, const TString& cgiParams = {}) const;` - to accent what params here mean

ydb/core/mon/audit/ut/ya.make
```diff
@@ -0,0 +1,17 @@
...
+FORK_SUBTESTS()
+
+SIZE(MEDIUM)
+
+PEERDIR(
+    ydb/core/testlib/default
+)
+
+SRCS(
+    audit_ut.cpp
```
UgnineSirdis: It is a test for url_matcher.cpp
So it will be better to name it url_matcher_ut.cpp

ydb/core/mon/audit/audit_ut.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <library/cpp/testing/unittest/tests_data.h>
+
+using namespace NActors;
+
+Y_UNIT_TEST_SUITE(TAuditTest) {
+    Y_UNIT_TEST(MatchExactPathOnly) {
+        NActors::NAudit::TUrlMatcher matcher;
+        matcher.AddPattern({.Path = "/a/b/c"});
+
+        UNIT_ASSERT(matcher.Match("/a/b/c"));
```
UgnineSirdis: `UNIT_ASSERT(matcher.Match("a/b/c"));` ??

ydb/core/mon/audit/audit_ut.cpp
```diff
@@ -0,0 +1,68 @@
...
+        UNIT_ASSERT(matcher.Match("/a/b", "mode="));
+        UNIT_ASSERT(matcher.Match("/a/b", "mode=1"));
+        UNIT_ASSERT(matcher.Match("/a/b", "other=1&mode=1"));
+
+        UNIT_ASSERT(!matcher.Match("/a/b"));
+        UNIT_ASSERT(!matcher.Match("/a", "mode=1"));
+        UNIT_ASSERT(!matcher.Match("/a/b", "other=1"));
+    }
+
+    Y_UNIT_TEST(MatchWithParamNameAndValue) {
```
UgnineSirdis: What if we need more complex matching?
`/a/b/c?action=danger&dry_run=0`?

ydb/core/mon/audit/audit_ut.cpp
```diff
@@ -0,0 +1,68 @@
...
+using namespace NActors;
+
+Y_UNIT_TEST_SUITE(TAuditTest) {
+    Y_UNIT_TEST(MatchExactPathOnly) {
+        NActors::NAudit::TUrlMatcher matcher;
+        matcher.AddPattern({.Path = "/a/b/c"});
+
+        UNIT_ASSERT(matcher.Match("/a/b/c"));
+        UNIT_ASSERT(matcher.Match("/a/b/c", "action=start"));
+        UNIT_ASSERT(matcher.Match("/a/b/c/d"));
```
UgnineSirdis: Do we want to match also `/a/b//c`?

ydb/core/mon/audit/url_matcher.cpp
```diff
@@ -0,0 +1,82 @@
...
+        path = path.Skip(1);
+    }
+
+    while (path) {
+        TStringBuf part;
+        path.NextTok('/', part);
+
+        TString key(part);
+
+        auto& child = node->Children[key];
```
UgnineSirdis: We don't need key to be exactly TString type, TStringBuf is also OK

ydb/core/mon/audit/url_matcher.h
```diff
@@ -0,0 +1,40 @@
...
+
+private:
+    struct TParamCondition {
+        TString Name;
+        TString ExpectedValue;
+    };
+
+    struct TNode {
+        THashMap<TString, THolder<TNode>> Children;
+        bool MatchedWithoutParams = false;
```
UgnineSirdis: `MatchWithoutParams`

ydb/core/mon/audit/url_matcher.h
```diff
@@ -0,0 +1,40 @@
...
+        TVector<TParamCondition> MatchedParams;
+    };
+
+    THolder<TNode> Root = MakeHolder<TNode>();
```
UgnineSirdis: Why here is only one root? Why not "set"/"vector"?

UgnineSirdis: Also a question why is it a pointer? Do we expect nullptr here?

ydb/core/mon/audit/url_matcher.h
```diff
@@ -0,0 +1,40 @@
...
+    bool Match(const TString& url, const TString& params = "") const;
+
+private:
+    struct TParamCondition {
+        TString Name;
+        TString ExpectedValue;
+    };
+
+    struct TNode {
+        THashMap<TString, THolder<TNode>> Children;
```
UgnineSirdis: As far as I see, we can't have here nullptr, so why pointer?

ydb/core/mon/audit/url_matcher.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace NActors::NAudit {
+
+const TString ANY_PATH = "*";
+
+void TUrlMatcher::AddPattern(const TUrlPattern& pattern) {
+    TStringBuf path = pattern.Path;
+    TNode* node = Root.Get();
+    if (path.StartsWith('/')) {
+        path = path.Skip(1);
```
UgnineSirdis, line 12: How about many '/'?

ydb/core/mon/audit/url_matcher.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+        auto& child = node->Children[key];
+        if (!child) {
+            child = MakeHolder<TNode>();
+        }
+        node = child.Get();
+    }
+
+    if (pattern.ParamName.empty()) {
+        node->MatchedWithoutParams = true;
```
UgnineSirdis: Is it not enough here to have empty param name?

ydb/core/mon/audit/url_matcher.cpp
```diff
@@ -0,0 +1,82 @@
...
+        path = path.Skip(1);
+    }
+
+    const TNode* node = Root.Get();
+
+    do {
+        TStringBuf part;
+        if (!path.NextTok('/', part)) {
+            part = path;
+            path = TStringBuf();
```
UgnineSirdis, line 43: What if part is empty because of two `/`?

ydb/core/mon/audit/url_matcher.cpp
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        auto it = node->Children.find(part);
+        if (it == node->Children.end()) {
+            it = node->Children.find(ANY_PATH);
```
UgnineSirdis, line 48: If a user want to match something like `/schemeshard/modify_*`, I think it won't work...

ydb/core/mon/audit/url_matcher.cpp
```diff
@@ -0,0 +1,82 @@
...
+    const TNode* node = Root.Get();
+
+    do {
+        TStringBuf part;
+        if (!path.NextTok('/', part)) {
+            part = path;
+            path = TStringBuf();
+        }
+
+        auto it = node->Children.find(part);
```
UgnineSirdis, line 46:
```cpp
auto it = node->Children.find(part);
if (it == node->Children.end()) {
    it = node->Children.find(ANY_PATH);
}
if (it == node->Children.end()) {
    return false;
}
node = it->second.Get();
```

ydb/core/mon/audit/url_matcher.cpp
UgnineSirdis:
```cpp
if (pattern.ParamName.empty()) {
    node->MatchedParams.clear();
}
```

ydb/core/mon/audit/url_matcher.cpp
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            node = it->second.Get();
+        }
+
+        if (node->MatchedWithoutParams) {
+            return true;
+        }
+        for (const auto& p : node->MatchedParams) {
+            if (params.Has(p.Name) && (p.ExpectedValue.empty() || params.Get(p.Name) == p.ExpectedValue)) {
+                return true;
```
UgnineSirdis: I can't understand... Am I right that if we have matcher like `/a/b/c?x=y`, we will match `/a/b/c/d/e/f?x=y`?

ydb/core/mon/audit/audit_force.h
```diff
@@ -0,0 +1,62 @@
...
+#include <util/generic/vector.h>
+#include <util/generic/string.h>
+
+namespace NActors::NAudit {
+
+// Audit logging is based on HTTP methods like POST or PUT
+// but some handlers use GET for changes
+// FORCE_AUDIT lists URLs that always require audit logging
+// despite the HTTP methods associated with getting information
+static const TVector<TUrlPattern> FORCE_AUDIT = {
```
UgnineSirdis: static in header is a bad idea, because it will create a static variable in every cpp file that includes the header

ydb/core/mon/audit/audit_force.h
```diff
@@ -0,0 +1,62 @@
...
+    {.Path = "/trace"},
+    {.Path = "/fq_diag/quotas", .ParamName = "submit"},
+    {.Path = "/login"},
+    {.Path = "/logout"},
+    {.Path = "/tablets", .ParamName = "KillTabletID"},
+    {.Path = "/tablets", .ParamName = "RestartTabletID"},
+    {.Path = "/tablets/app"},
+};
+
+inline TUrlMatcher CreateForceAuditUrlMatcher() {
```
UgnineSirdis: We wanted to assign audit log class from this config https://github.com/ydb-platform/ydb/pull/21786 to every action. That was the plan.

ydb/core/mon/audit/url_matcher.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace NActors::NAudit {
+
+struct TUrlPattern {
+    TString Path;
+    TString ParamName;
+    TString ParamValue;
+};
+
+class TUrlMatcher {
```
UgnineSirdis, line 19: What if we will ned in the future two matchers - more specific and less specific?
Mayb we need some priorities for them...

ydb/core/mon/audit/audit.cpp
```diff
@@ -0,0 +1,127 @@
...
+    const TString method(request->Method);
+    const TString url(request->URL.Before('?'));
+    const auto params = request->URL.After('?');
+    const auto cgiParams = TCgiParameters(params);
+    if (!(AuditEnabled = AuditableRequest(method, url, cgiParams))) {
+        return;
+    }
+
+    NHttp::THeaders headers(request->Headers);
+    auto remote_address = ToString(headers.Get(X_FORWARDED_FOR_HEADER).Before(',')); // Get the first address in the list
```
UgnineSirdis: remoteAddress - it it the local var

ydb/core/mon/audit/audit.cpp
```diff
@@ -0,0 +1,127 @@
...
+        return true;
+    }
+
+    // OPTIONS are not audited
+    if (method == "OPTIONS") {
+        return false;
+    }
+
+    // force audit for specific URLs
+    static auto FORCE_AUDIT_MATCHER = CreateForceAuditUrlMatcher();
```
UgnineSirdis: It is incorrect naming. Why force audit? Matcher is expected to define action type

ydb/core/mon/audit/audit.cpp
```diff
@@ -0,0 +1,127 @@
...
+    AddAuditLogPart("component", MONITORING_COMPONENT_NAME);
+    AddAuditLogPart("remote_address", remote_address);
+    AddAuditLogPart("operation", DEFAULT_OPERATION);
+    AddAuditLogPart("method", method);
+    AddAuditLogPart("url", url);
+    if (!params.Empty()) {
+        AddAuditLogPart("params", ToString(params));
+    }
+    if (!request->Body.Empty()) {
+        AddAuditLogPart("body", ToString(request->Body));
```
UgnineSirdis: Even body? At least in that case we should limit its size

ydb/core/mon/audit/audit.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+void TAuditCtx::AddAuditLogParts(const TIntrusiveConstPtr<NACLib::TUserToken>& userToken) {
+    if (!AuditEnabled) {
+        return;
+    }
+    SubjectType = userToken ? userToken->GetSubjectType() : NACLibProto::SUBJECT_TYPE_ANONYMOUS;
+    AddAuditLogPart("subject", userToken->GetUserSID());
```
UgnineSirdis: Segfault detected!

```cpp
AuditEnabled &= NKikimr::AppData()->AuditConfig.EnableLogging(NKikimrConfig::TAuditConfig::TLogClassConfig::ClusterAdmin, SubjectType);
if (!AuditEnabled) {
    return;
}
```

UgnineSirdis (changes requested):

ydb/core/mon/audit/url_matcher.h
StekPerepolnen: changed to NMonitoring

ydb/core/mon/audit/url_matcher.h
StekPerepolnen: changed

ydb/core/mon/audit/ut/ya.make
StekPerepolnen: renamed

ydb/core/mon/audit/audit_ut.cpp
StekPerepolnen: added

ydb/core/mon/audit/audit_ut.cpp
StekPerepolnen: Each rule checks only one parameter. Currently, there are no handlers that require a combination of different parameters in the URL. I did it this way for simplicity. I will add comment on this

ydb/core/mon/audit/audit_ut.cpp
StekPerepolnen: no, it doesn't match to /a/b/c pattern, added to test too

ydb/core/mon/audit/url_matcher.h
StekPerepolnen: renamed

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: removed extra `TString`

ydb/core/mon/audit/url_matcher.h
StekPerepolnen: We use a single Root because all patterns are added into one tree.

StekPerepolnen: no, we don't expect nullptr - changed to `TNode Root`

ydb/core/mon/audit/url_matcher.h
StekPerepolnen: changed  `THolder<TNode>` to `TNode`

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: Multiple slashes produce empty tokens. We skip only the leading slash to normalize paths like `/a/b/c`; repeated slashes like in `//a/b` or `/a//b` are preserved and must produce empty tokens.

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: I wanted to avoid making the tree structure depend on the order in which rules are added.

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: Technically, a URL can contain multiple slashes in a row, and in that case I handle the empty segment accordingly. In the list of rules I added, multiple consecutive slashes are not expected.

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: Right now, there's no need to support overly complex rules. The ability to mark a path segment with a wildcard `*` as "any value" is sufficient.

In your example, I would just use the pattern `/schemeshard/*`.
If someone tries to access something unexpected like `/schemeshard/bad_*`, monitoring won't match it to any handler, and audit logs simply won't be recorded.

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: Yeah, I overcomplicated it. Removed the unnecessary part.

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: Yes, that's correct - it's even covered by tests.
I initially assumed there would be scenarios where we’d want to log all paths matching a prefix.
In practice, this scenario isn’t used right now.
But I don’t expect this behavior to cause problems, like adding audit logs when it’s not needed (explained in the comments above).

ydb/core/mon/audit/audit_force.h
StekPerepolnen: moved to cpp

ydb/core/mon/audit/url_matcher.h
StekPerepolnen: I don't want to add anything else to this structure for now.

I'd like each service to be responsible for determining which requests are important for the audit log and which are not. This might be necessary if I decide to add differentiation of values for the `operation` field.

If that works out, I would eventually remove this `UrlMatcher`.

ydb/core/mon/audit/audit.cpp
StekPerepolnen: renamed

ydb/core/mon/audit/audit.cpp
StekPerepolnen: renamed to AUDITABLE_ACTIONS

ydb/core/mon/audit/audit.cpp
StekPerepolnen: limit the size 2MB

ydb/core/mon/audit/audit.cpp
StekPerepolnen: I create a separate context for each request, and the header sets the default value to ANONYMOUS. 

Moreover, the AddAuditLogParts method may not be called in some cases if the user fails authentication, which is why I perform the configuration check at the very end.

UgnineSirdis: Yes, I don't mind, but I was trying to say that here according to the code we accept userToken to be nullptr

ydb/core/mon/audit/url_matcher.cpp
UgnineSirdis: Currently the behaviour is unexpected, because we simply will exit from the cycle in case of two consecutive slashes.

ydb/core/mon/audit/url_matcher.cpp
UgnineSirdis: But it is a bit contrintuitive, that we give a possibility to match subtrees, but with concrete parameters. If we make it this way now, we can't remove it in the future

UgnineSirdis (changes requested):

ydb/core/mon/audit/audit.cpp
StekPerepolnen: got it..

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: I'll add tests for this, as I consider consecutive slashes in the path - which result in empty segments - a valid case that should be handled correctly.

ydb/core/mon/audit/url_matcher.cpp
StekPerepolnen: I've changed the behavior to avoid confusion - parameters are now only checked after the full path matches exactly.

ydb/core/mon/audit/auditable_actions.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace NMonitoring::NAudit {
+
+// Audit logging is based on HTTP methods like POST or PUT
+// but some handlers use GET for changes
+// AUDITABLE_ACTIONS lists URLs that always require audit logging
+// despite the HTTP methods associated with getting information
+extern const TVector<TUrlPattern> AUDITABLE_ACTIONS;
+
+inline TUrlMatcher CreateAuditableActionsMatcher() {
```
UgnineSirdis: We have a cpp-file. Why not there?

ydb/core/mon/audit/audit.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+namespace {
+    const TString MONITORING_COMPONENT_NAME = "monitoring";
+    const TString DEFAULT_OPERATION = "HTTP REQUEST";
+    const TString EMPTY_VALUE = "{none}";
+    const TString X_FORWARDED_FOR_HEADER = "X-Forwarded-For";
+    const TStringBuf TRUNCATED_SUFFIX = "**TRUNCATED_BY_YDB**";
+
+    // audit event has limit of 4 MB, but we limit body size to 2 MB
+    const size_t MAX_AUDIT_BODY_SIZE = 2 * 1024 * 1024 - TRUNCATED_SUFFIX.size();
```
UgnineSirdis:
```cpp
#include <util/generic/size_literals.h>

constexpr TStringBuf TRUNCATED_SUFFIX = "**TRUNCATED_BY_YDB**";
constexpr size_t MAX_AUDIT_BODY_SIZE = 2_MB - TRUNCATED_SUFFIX.size();
```

ydb/core/mon/audit/audit.cpp
```diff
@@ -0,0 +1,140 @@
...
+    AddAuditLogPart("operation", DEFAULT_OPERATION);
+    AddAuditLogPart("method", method);
+    AddAuditLogPart("url", url);
+    if (!params.Empty()) {
+        AddAuditLogPart("params", ToString(params));
+    }
+    if (!request->Body.Empty()) {
+        TStringBuilder auditBody;
+        if (request->Body.size() > MAX_AUDIT_BODY_SIZE) {
+            auditBody << request->Body.substr(0, MAX_AUDIT_BODY_SIZE) << TRUNCATED_SUFFIX;
```
UgnineSirdis: Here we can optimize one copy:
```cpp
TStringBuf body = request->Body;
if (body.size() > MAX_AUDIT_BODY_SIZE) {
    auditBody << body.SubString(0, MAX_AUDIT_BODY_SIZE) << TRUNCATED_SUFFIX;
} else {
    auditBody << body;
}
```

TStringBuf::SubString will not allocate and copy data

ydb/core/mon/audit/audit.cpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+void TAuditCtx::AddAuditLogParts(const TIntrusiveConstPtr<NACLib::TUserToken>& userToken) {
+    if (!AuditEnabled) {
+        return;
+    }
+    SubjectType = userToken ? userToken->GetSubjectType() : NACLibProto::SUBJECT_TYPE_ANONYMOUS;
+    if (userToken) {
+        AddAuditLogPart("subject", userToken->GetUserSID());
```
UgnineSirdis: Do we need here explicit `subject={none}`?

ydb/core/mon/audit/auditable_actions.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <util/generic/vector.h>
+
+namespace NMonitoring::NAudit {
+
+// Audit logging is based on HTTP methods like POST or PUT
+// but some handlers use GET for changes
+// AUDITABLE_ACTIONS lists URLs that always require audit logging
+// despite the HTTP methods associated with getting information
+extern const TVector<TUrlPattern> AUDITABLE_ACTIONS;
```
UgnineSirdis: Do we need it here? I can't see the usage except from CreateAuditableActionsMatcher

UgnineSirdis (changes requested):

UgnineSirdis (approved):

PR #21283, opened 2025-07-17
[stable-25-1-2] Fix cpu counters for cases with stuck actors 

kruall:
### Changelog entry 

When an actor started taking too long to process, the ElapsedMicrosec and CpuMicrosec metrics wouldn't update. As a result, the harmonizer could behave incorrectly and, even under CPU shortage, reduce the number of threads.

### Changelog category 

### Description for reviewers 

...


maximyurchuk (changes requested):
непонятно зачем 25-1-2

PR #21786, opened 2025-07-28
Divide audit logs API writes on several classes

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/protos/config.proto
+++ b/ydb/core/protos/config.proto
@@ -1798,9 +1798,9 @@ message TAuditConfig {
             ServiceImpersonatedFromUser = 4;
         }
 
-        optional EApiClass ApiClass = 1;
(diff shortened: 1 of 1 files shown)


ydb/core/protos/config.proto
```diff
@@ -1779,9 +1779,34 @@ message TAuditConfig {
         optional string LogJsonEnvelope = 3; // Json template with text field containing %message% placeholder. For example {"my_enveloped_message": "%message%"}. %message% will be replaced with real audit log message
     }
 
+    message TApiConfig {
```
flown4qqqq: maybe we should replace the word "api"? For example, on the "log"

flown4qqqq (approved):

PR #21819, opened 2025-07-28
schemeshard: cleanup TTxState declaration, move out enums

ijon:
Clear `TTxState` declaration by moving out all macro generated enums and methods into separate files. Additional headers are included back to `schemeshard_tx_infly.h` in order to be conservative and avoid updating other core and operation files. For now. Final goal will be to narrow the scope of influence of those declarations as much as possible and to replace code generation by macros with compile or build time code generation.  

Stacked on:
- https://github.com/ydb-platform/ydb/pull/21817


ydb/core/tx/schemeshard/schemeshard_subop_state_types.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <util/generic/string.h>
+
+namespace NKikimr::NSchemeShard {
+
+// WARNING: DO NOT REORDER this constants
+// reordering breaks update
+enum ETxState {
```
Enjection, line 9: Not a blocker, but it would be great to finally have this warning above enforced with canonized tests

Enjection (dismissed):

CyberROFL (approved):

PR #21870, opened 2025-07-29
Audit config class for fast audit processing

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/audit/audit_config/audit_config.cpp
+++ b/ydb/core/audit/audit_config/audit_config.cpp
@@ -2,6 +2,23 @@
 
 namespace NKikimr::NAudit {
 
+namespace {
+
+NACLibProto::ESubjectType ToSubjectType(NKikimrConfig::TAuditConfig::TLogClassConfig::EAccountType accountType) {
+    switch (accountType) {
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::Anonymous:
+        return NACLibProto::SUBJECT_TYPE_ANONYMOUS;
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::User:
+        return NACLibProto::SUBJECT_TYPE_USER;
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::Service:
+        return NACLibProto::SUBJECT_TYPE_SERVICE;
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::ServiceImpersonatedFromUser:
+        return NACLibProto::SUBJECT_TYPE_SERVICE_IMPERSONATED_FROM_USER;
+    }
+}
+
+}
+
 TAuditConfig::TAuditConfig(const NKikimrConfig::TAuditConfig& cfg)
     : NKikimrConfig::TAuditConfig(cfg)
 {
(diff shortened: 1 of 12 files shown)


ydb/core/audit/audit_config/audit_config.cpp
```diff
@@ -0,0 +1,69 @@
...
+    switch (accountType) {
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::Anonymous:
+        return NACLibProto::SUBJECT_TYPE_ANONYMOUS;
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::User:
+        return NACLibProto::SUBJECT_TYPE_USER;
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::Service:
+        return NACLibProto::SUBJECT_TYPE_SERVICE;
+    case NKikimrConfig::TAuditConfig::TLogClassConfig::ServiceImpersonatedFromUser:
+        return NACLibProto::SUBJECT_TYPE_SERVICE_IMPERSONATED_FROM_USER;
+    }
```
StekPerepolnen: Does it make sense to add a default: case with Y_FAIL here, so that the system always crashes on incorrect configuration?

ydb/core/audit/audit_config/audit_config.cpp
```diff
@@ -0,0 +1,69 @@
...
+    for (NACLibProto::ESubjectType type : cfg->second.ExcludeSubjectTypes) {
+        if (type == subjectType) {
+            return false;
+        }
+    }
+    return true;
+}
+
+static NACLibProto::ESubjectType ToSubjectType(NKikimrConfig::TAuditConfig::TLogClassConfig::EAccountType accountType) {
```
StekPerepolnen: Can we move `ToSubjectType` into an anonymous namespace to explicitly limit its visibility?

ydb/core/audit/audit_config/audit_config.h
```diff
@@ -0,0 +1,34 @@
...
+        std::vector<NACLibProto::ESubjectType> ExcludeSubjectTypes;
+        bool EnableLogging;
+    };
+
+public:
+    TAuditConfig() = default;
+    TAuditConfig(const NKikimrConfig::TAuditConfig&);
+    TAuditConfig& operator=(const NKikimrConfig::TAuditConfig&);
+
+    bool EnableLogging(NKikimrConfig::TAuditConfig::TLogClassConfig::ELogClass logClass, NACLibProto::ESubjectType subjectType) const;
```
StekPerepolnen, line 23: Is a default value needed here? For example, in the case of a cluster-level API, it might make sense not to specify subjectType.

ydb/core/audit/audit_config/audit_config.cpp
UgnineSirdis: No, here the compiler will say us if we don't process explicitly a new case

ydb/core/audit/audit_config/audit_config.cpp
UgnineSirdis: Done. But it was already static, so, now it is the same

ydb/core/audit/audit_config/audit_config.h
UgnineSirdis: I did not understand. Can you provide an example?

StekPerepolnen (approved):

Issue #21358, opened 2025-07-18
Account GroupSizeInUnits in VDiskMetrics.AvailableSize calculus


rosik:
К ознакомлению:

- https://github.com/ydb-platform/ydb/commit/ac7fab5882732770d08bdc98e5fd8b9a0d291cd5
- https://github.com/ydb-platform/ydb/blob/56f819e7418c497b27575538f66fc921b4d0f807/ydb/core/blobstorage/pdisk/blobstorage_pdisk_chunk_tracker.h#L432-L436

PR #21872, opened 2025-07-29
Remove unused fields from Whiteboard::TBSGroupStateInfo

rosik:
### Changelog category

### Description for reviewers

Those values are set, but never read

Related to #21358


alexvru (changes requested):
Оно используется как минимум в JS.

PR #22205, opened 2025-08-01
conveyor_composite: remove stray unhandled event

yumkam:
regression by 3849a7e5cde6fbbb1a61a417b485b2c6c8dd3513
event handler was removed, but event is still emitted (once);
largely cosmetic (only nasty-looking, but harmless log message)

### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


yumkam:
solved independently in #37539

PR #22263, opened 2025-08-04
async ca: rework watermark handling and fix inputtransform and watermark interaction

yumkam:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Change watermark logic to be more similar to checkpoint.
Unlike checkpoints, watermarks can be delayed. There are important difference: checkpoints cannot be moved around data; watermark processing can be delayed, watermark don't pause channels until they are ready for processing (so, we are only waiting for data inside of channels to be processed and don't wait for input).
Somewhat special case is input transform: we must wait until it finishes input processing.
There are some possible optimizations and short-circuits, but it will be improved later.


Hor911 (changes requested):
PR does not match RFC https://nda.ya.ru/t/I8zckxs17HpogX

Hor911 (dismissed):

vitstn (approved):

uzhastik (approved):

PR #22298, opened 2025-08-04
QueryService: graceful shutdown session using attach stream

KirillKurdyukov:
### Changelog entry 

Fixed bug: Attach Stream does not send a signal on session shutdown [#20812](https://github.com/ydb-platform/ydb/issues/20812)

### Changelog category 

* Bugfix

### Description for reviewers 

Now, the session will send a SUSSES and then close the attached stream. When session is shutting down.


ydb/core/kqp/proxy_service/kqp_proxy_service.cpp
```diff
@@ -1791,6 +1793,14 @@ class TKqpProxyService : public TActorBootstrapped<TKqpProxyService> {
 
     TResourcePoolsCache ResourcePoolsCache;
     TDatabasesCache DatabasesCache;
+
+    TEvKqp::TEvCloseSessionResponse* CreateEvCloseSessionResponse(const TString& sessionId) {
```
gridnevvvit: std::unique_ptr<TEvKqp::TEvCloseSessionResponse> вернуть лучше. а release сделать по месту. в целом PR окей.

gridnevvvit (approved):

PR #22445, opened 2025-08-06
25-1-1: schemeshard: quick-fix tenant system tablets deletion

### LINKED ISSUE #19842: database drop may leave system tablets in root hive
YDBOPS-12120

Investigate the problem, and find if it can be related to enabling feature flag `enable_alter_database_create_hive_first` in 25-1.

UPD: No it is not. It seams that the issue is in race condition between deletion of tenant's system tablets (including tenant hive), which require tenant hive to be alive, and the actual disappearance of the tenant hive itself (due to tenant nodes shutdown for example).

ijon:
Merge from `main`:
- a721d56 -- https://github.com/ydb-platform/ydb/pull/21377
- 1f8359c -- https://github.com/ydb-platform/ydb/pull/22360

Currently, drop-extsubdomain operations may not completely remove dedicated databases from the system:
- system tablets of the tenant may remain in the root hive
- traces of database paths and subdomain objects may persist in the root schemeshard

Root cause: The cleanup procedure depends on the tenant hive's existence - delete-tablet requests are sent to the tenant hive for redirection to the root hive. This works for most databases but fails with dedicated databases, where the tenant hive cannot outlive drop-extsubdomain operations, and there's no guarantee that tenant nodes and tablets will be available during deletion.

Changes:
- Implement quick fix for subdomain cleanup by directly sending system tablets delete-tablet requests to the root hive (only for dedicated databases)
- Add unit and reboot tests to verify proper extsubdomain cleanup
- Fix aggressive reconnect loop in coordinator's mediator_queue
- Fix handling of TEvSchemeShard::TEvSyncTenantSchemeShard in schemeshard

Closes #19842.

### Changelog entry

Fix issue where dedicated database deletion may leave database system tablets improperly cleaned.

### Changelog category

* Bugfix

### DIFF
--- a/ydb/core/testlib/tablet_helpers.cpp
+++ b/ydb/core/testlib/tablet_helpers.cpp
@@ -1151,6 +1151,7 @@ namespace NKikimr {
                 HFunc(TEvHive::TEvAdoptTablet, Handle);
                 HFunc(TEvHive::TEvDeleteTablet, Handle);
                 HFunc(TEvHive::TEvDeleteOwnerTablets, Handle);
+                HFunc(TEvHive::TEvStopTablet, Handle);
                 HFunc(TEvHive::TEvRequestHiveInfo, Handle);
                 HFunc(TEvHive::TEvInitiateTabletExternalBoot, Handle);
                 HFunc(TEvHive::TEvUpdateTabletsObject, Handle);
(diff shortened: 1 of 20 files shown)


ijon:
Close PR unmerged as minor 25-1-1 is already out of service.

PR #22529, opened 2025-08-07
audit log on incoming request

StekPerepolnen:
### Changelog entry 

An audit log entry is required upon request reception (without waiting for the response)

### Changelog category 

* Improvement

### Description for reviewers 

Audit logging should happen twice for each request:
- **On request receipt** - to record the start of processing.
- **When the result is known** - to record the final status.

When registering a http handler, you can choose one of two audit logging strategies:
- **AutoAudit = true** - Audit logging is triggered automatically when the handler is called. The monitoring service determines when to log based on the request method and URL.
- **AutoAudit = false** - The handler explicitly controls when audit logging starts. To initiate logging, it sends a `TEvAuditExecute` message to the monitoring service request actor. The actor replies with `TEvAuditExecuteConfirmed`, ensuring the start log entry is written before request processing begins.

If audit logging was started, the monitoring service request actor records a final entry with the request outcome.

In this PR, `AuditDirectedStart` is added to YDB Viewer. It is set for all modifying handlers immediately after a redirect attempt, to avoid logging requests that return a redirect as unnecessary.

An audit log for a single user operation might look like this now:
```
2025-08-07T13:36:51.053111Z: {"status":"IN-PROCESS", "reason":"Execute", "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring", ...}
2025-08-07T13:36:51.053433Z: {"status":"SUCCESS", "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring" ...}
```
Requests with a redirect in the response will be skipped.

### DIFF
--- a/ydb/core/mon/audit/audit.cpp
+++ b/ydb/core/mon/audit/audit.cpp
@@ -17,15 +17,10 @@ namespace {
     const TString EMPTY_VALUE = "{none}";
     const TString X_FORWARDED_FOR_HEADER = "X-Forwarded-For";
     const TStringBuf TRUNCATED_SUFFIX = "**TRUNCATED_BY_YDB**";
+    const TString REASON_EXECUTE = "Execute";
 
     // audit event has limit of 4 MB, but we limit body size to 2 MB
     const size_t MAX_AUDIT_BODY_SIZE = 2_MB - TRUNCATED_SUFFIX.size();
-
-    enum ERequestStatus {
-        Success,
-        Process,
-        Error,
-    };
 
     ERequestStatus GetStatus(const NHttp::THttpOutgoingResponsePtr response) {
         auto status = response.Get()->Status;
@@ -45,6 +40,12 @@ namespace {
             case ERequestStatus::Error: return "ERROR";
         }
         return EMPTY_VALUE;
+    }
+
+    TString GetReason(const NHttp::THttpOutgoingResponsePtr& response) {
+        TStringBuilder reason;
+        reason << response.Get()->Status << " " << response.Get()->Message;
+        return reason;
     }
 
     inline TUrlMatcher CreateAuditableActionsMatcher() {
@@ -126,28 +127,37 @@ void TAuditCtx::AddAuditLogParts(const TIntrusiveConstPtr<NACLib::TUserToken>& u
     }
 }
 
-void TAuditCtx::FinishAudit(const NHttp::THttpOutgoingResponsePtr& response) {
-    AuditEnabled &= NKikimr::AppData()->AuditConfig.EnableLogging(NKikimrConfig::TAuditConfig::TLogClassConfig::ClusterAdmin, SubjectType);
+void TAuditCtx::LogAudit(ERequestStatus status, const TString& reason) {
+    AuditEnabled &= NKikimr::AppData()->AuditConfig.EnableLogging(
+        NKikimrConfig::TAuditConfig::TLogClassConfig::ClusterAdmin, SubjectType);
 
     if (!AuditEnabled) {
         return;
     }
-
-    auto status = GetStatus(response);
 
     AUDIT_LOG(
         AddAuditLogPart("subject", (Subject ? Subject : EMPTY_VALUE));
         AddAuditLogPart("sanitized_token", (SanitizedToken ? SanitizedToken : EMPTY_VALUE));
 
         for (const auto& [name, value] : Parts) {
-            AUDIT_PART(name, (!value.empty() ? value : EMPTY_VALUE))
+            AUDIT_PART(name, (!value.empty() ? value : EMPTY_VALUE));
         }
 
         AUDIT_PART("status", ToString(status));
         if (status != ERequestStatus::Success) {
-            AUDIT_PART("reason", response.Get()->Message);
+            AUDIT_PART("reason", (!reason.empty() ? reason : EMPTY_VALUE));
         }
     );
+}
+
+void TAuditCtx::LogOnExecute() {
+    LogAudit(ERequestStatus::Process, REASON_EXECUTE);
+}
+
+void TAuditCtx::LogOnResult(const NHttp::THttpOutgoingResponsePtr& response) {
+    auto status = GetStatus(response);
+    auto reason = GetReason(response);
+    LogAudit(status, reason);
 }
 
 }
--- a/ydb/core/mon/audit/audit.cpp
+++ b/ydb/core/mon/audit/audit.cpp
@@ -22,32 +22,6 @@ namespace {
     // audit event has limit of 4 MB, but we limit body size to 2 MB
     const size_t MAX_AUDIT_BODY_SIZE = 2_MB - TRUNCATED_SUFFIX.size();
 
-    ERequestStatus GetStatus(const NHttp::THttpOutgoingResponsePtr response) {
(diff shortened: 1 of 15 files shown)


ydb/core/viewer/json_pipe_req.cpp
```diff
@@ -1226,6 +1234,27 @@ bool TViewerPipeClient::NeedToRedirect() {
...
+    }
+    if (NeedAuditLog) {
+        NeedAuditLog = false;
+        auto event = std::make_unique<NHttp::TEvHttpProxy::TEvAuditExecute>();
+        if (Event) {
+            Send(Event->Sender, event.release(), IEventHandle::FlagTrackDelivery);
+        } else if (HttpEvent) {
+            Send(HttpEvent->Sender, event.release(), IEventHandle::FlagTrackDelivery);
+        }
+        Become(&TViewerPipeClient::StateAudit);
```
UgnineSirdis, line 1252: What is done here, is not consistent with the name of this function - `NeedToWriteAuditLog`. If I would read the code and meet the call of NeedToWriteAuditLog, I would think that this function does not change any state, only checks whether audit record is needed here. It may cause problems with reading and understading this code. Rename the function please.

ydb/library/actors/http/http_proxy.h
```diff
@@ -294,6 +296,8 @@ struct TEvHttpProxy {
 
     struct TEvSubscribeForCancel : NActors::TEventLocal<TEvSubscribeForCancel, EvSubscribeForCancel> {};
     struct TEvRequestCancelled : NActors::TEventLocal<TEvRequestCancelled, EvRequestCancelled> {};
+    struct TEvAuditExecute : NActors::TEventLocal<TEvAuditExecute, EvAuditExecute> {};
+    struct TEvAuditExecuteConfirmed : NActors::TEventLocal<TEvAuditExecuteConfirmed, EvAuditExecuteConfirmed> {};
```
UgnineSirdis, line 300: If this is an answer to the `TEvAuditExecute` event, we name it `TEvAuditExecuteResponse` in common.

UgnineSirdis (changes requested):

adameat (changes requested):
This PR is overly complex for its intended goal. I’m strongly against making any related changes to the actors/http library. Additionally, the amount of changes in the viewer library is excessive. I recommend reconsidering the overall architecture of the solution.

ydb/core/mon/audit/audit.cpp
```diff
@@ -17,43 +17,53 @@ namespace {
...
+        return TAuditCtx::ERequestStatus::Process;
+    }
+
+    return TAuditCtx::ERequestStatus::Error;
+}
+
+TString TAuditCtx::ToString(const TAuditCtx::ERequestStatus value) {
+    switch (value) {
+        case TAuditCtx::ERequestStatus::Success: return "SUCCESS";
+        case TAuditCtx::ERequestStatus::Process: return "IN-PROCESS";
```
UgnineSirdis, line 57: Maybe `IN-PROGRESS` would suit better. How do you think?

alexv-smirnov:
> This PR is overly complex for its intended goal.

You're right. The description lacks an important part. The problem is that the simple solution causes multiple records about taking a request for processing when viewer handlers return redirection instead of taking a request for processing. As in our installation we use redirection in most cases, we would like to avoid it, to avoid triggering multiple alerts for a single request.

So, the idea is to differentiate between:

1. Handlers which can return redirections -- they should also take on responsibility to send the audit event before actual processing, if they decided to perform it. In fact, these are the viewer handlers.
2. Other handlers which have no functionality for redirection. For them, we send the audit event before handing over the processing to the handler.

Hope it made it clearer. @adameat

adameat:
> Hope it made it clearer. @adameat

Thanks, but I think this way to solve the problem is completely wrong.
I think the approach should be very similar to, for example, http caching - when we have all source changes in one 1 source file, which injects into http proxy interface transparently and controls all requests and corresponding responses.

FYI @StekPerepolnen

alexv-smirnov:
There's another important issue here, in fact it was the motivation behind this change...
We need to log the audit record before processing, not waiting for the completion, as the completion may never occur. In fact, the processing node can crash or perform the execution for unpredictable time because of some malicious request, but we need to ensure the record exists in the audit log. @adameat

alexv-smirnov:
We can probably start with committing a simpler solution which logs every request at the entry, but would cause double records in case of redirection inside viewer handlers. So we can close the gap first, and then discuss how to make it easier to use for alerts. What do you think? @adameat

adameat:
> There's another important issue here, in fact it was the motivation behind this change... We need to log the audit record before processing, not waiting for the completion, as the completion may never occur. In fact, the processing node can crash or perform the execution for unpredictable time because of some malicious request, but we need to ensure the record exists in the audit log. @adameat

this doesn't contradict at all, it's even simpler

adameat:
> We can probably start with committing a simpler solution which logs every request at the entry, but would cause double records in case of redirection. So we can close the gap first, and then discuss how to make it easier to use for alerts. What do you think? @adameat

you can easily filter redirected responses, in my opinion.
as well as to wait for some time to finish a request...

and redirect could be easily predicted - any request with database parameter and without /node/ in url will be redirected

alexv-smirnov:
> redirect could be easily predicted - any request with database parameter and without /node/ in url will be redirected

This is much simpler than what was in the PR) Thank you! Will update the PR shortly.

StekPerepolnen:
@adameat I’m afraid predicting nodes won’t be that straightforward. Here are the rules I’ve identified for when a redirection occurs:

1) The request targets a viewer handler.
2) The `X-Forwarded-From-Node` header is missing.
3) The request includes a `database` or `tenant` parameter (for `/viewer/json/query` – only `database` parameter).
4) The request is not made to the database’s dynnode.

(the `/node` path segment is removed before the audit log stage, so it’s not considered.)

The issue with points 1) and 3) is that without an explicit list of viewer handlers and their redirection conditions, a malicious user could craft a request that bypasses audit logging.

**Our immediate goal** is to ensure the audit record is written before executing the operation, while ignoring redirection. This is critical for cases where the processing node may crash.
FYI @alexv-smirnov

adameat:
> @adameat I’m afraid predicting nodes won’t be that straightforward. Here are the rules I’ve identified for when a redirection occurs:
> 
> 1. The request targets a viewer handler.
> 2. The `X-Forwarded-From-Node` header is missing.
> 3. The request includes a `database` or `tenant` parameter (for `/viewer/json/query` – only `database` parameter).
> 4. The request is not made to the database’s dynnode.
> 
> (the `/node` path segment is removed before the audit log stage, so it’s not considered.)
> 
> The issue with points 1) and 3) is that without an explicit list of viewer handlers and their redirection conditions, a malicious user could craft a request that bypasses audit logging.
> 
> **Our immediate goal** is to ensure the audit record is written before executing the operation, while ignoring redirection. This is critical for cases where the processing node may crash. FYI @alexv-smirnov

i thought about it and realized that this is conceptually wrong. any http handler can return any status, redirect or not, it doesn’t matter. trying to play along with their logic won’t give a 100% reliable result and isn’t a viable solution either. it seems we just need to choose: either log all requests regardless of the response, or wait for the response result. there’s no other good solution here.

PR #22685, opened 2025-08-11
Add audit log on incoming request

StekPerepolnen:
### Changelog entry 

An audit log entry is required upon request reception.

### Changelog category 

* Improvement

### Description for reviewers 

In Audit Configuration you could define at which audit log entries should be created:
- **Received** - when the request is accepted, before the operation is executed. Such entries will have `status` - `IN-PROCESS` and `reason` - `Execute`.
- **Completed** - when sending the response, after the final status is determined. The `status` field will be `IN-PROCESS` for redirects, `SUCCESS` for successful executions, or `ERROR` for failures. The `reason` field will contain the response status and message.

Example: configuration with `Completed` only
```
{"status":"IN-PROCESS", "reason":"307 Temporary Redirect", "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring" ...}
{"status":"SUCCESS", "reason":"200 Ok" "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring" ...}
```

Example: configuration with `Received` and `Completed` (redirect logged only at completion phase)
```
{"status":"IN-PROCESS", "reason":"Execute", "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring", ...}
{"status":"IN-PROCESS", "reason":"307 Temporary Redirect", "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring" ...}
{"status":"IN-PROCESS", "reason":"Execute", "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring", ...}
{"status":"SUCCESS", "reason":"200 Ok" "operation":"HTTP REQUEST","url":"/viewer/query","component":"monitoring" ...}
```


ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,98 @@ namespace {
...
+        } else if (contentType == "application/x-www-form-urlencoded") {
+            TCgiParameters bodyParams(request->Body);
+            for (const auto& [key, value] : bodyParams) {
+                cgiParams.InsertUnescaped(key, value);
+            }
+        }
+    }
+}
+
+bool CheckViewerRedirect(const NHttp::THttpIncomingRequestPtr& request, const TString& currentDatabase) {
```
StekPerepolnen: I extracted the currentDatabase parameter to test the method. It was the fastest way.

ydb/core/mon/audit/redirect_ut.cpp
```diff
@@ -0,0 +1,292 @@
...
+    struct RequestParams {
+        ui32 Mask;
+        TString Path;
+        TString Url;
+        TString Body;
+        TString ContentType;
+        bool XForwardedFromNode;
+    };
+
+    class TestEvn {
```
UgnineSirdis: TestEnv

UgnineSirdis (changes requested):

ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,98 @@ namespace {
...
+    BuildParamsFromBody(cgiParams, request);
+
+    auto database = cgiParams.Get("database");
+    if (!database) {
+        database = cgiParams.Get("tenant");
+    }
+
+    auto directStr = cgiParams.Get("direct");
+
+    auto direct = FromStringWithDefault<bool>(TString(cgiParams.Get("direct")), false);
```
UgnineSirdis: auto direct = FromStringWithDefault<bool>(directStr, false);

UgnineSirdis (changes requested):

ydb/core/mon/audit/audit.h
```diff
@@ -27,4 +35,6 @@ class TAuditCtx {
     TString SanitizedToken;
 };
 
+bool CheckViewerRedirect(const NHttp::THttpIncomingRequestPtr& request, const TString& tenantName);
```
UgnineSirdis: IsViewerRedirectRequest - something like this would better tell us what this function returns

UgnineSirdis (changes requested):

ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,96 @@ namespace {
...
     }
+
+    void BuildParamsFromBody(TCgiParameters& cgiParams, const NHttp::THttpIncomingRequestPtr& request) {
+        NHttp::THeaders headers(request->Headers);
+
+        auto contentType = NHttp::Trim(headers.Get("Content-Type").Before(';'), ' ');
+        if (contentType == "application/json") {
+            NJson::TJsonValue jsonData;
+            if (NJson::ReadJsonTree(request->Body, &jsonData) && jsonData.IsMap()) {
+                for (const auto& [key, value] : jsonData.GetMap()) {
```
UgnineSirdis: Is it relevant for all cases of our http interface (seeing json map's fields as parameters)?
Or it does not matter?

ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,96 @@ namespace {
...
+
+    static const THashSet<TString> VIEWER_HANDLERS = {"viewer", "operation", "query", "scheme", "storage"};
+
+    TStringBuf part;
+    path.NextTok('/', part);
+    if (!VIEWER_HANDLERS.contains(part)) {
+        return false;
+    }
+
+    if (headers.Has("X-Forwarded-From-Node")) {
```
UgnineSirdis: This is a mark that we make when redirecting, right?

ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,96 @@ namespace {
...
+    if (!database) {
+        database = cgiParams.Get("tenant");
+    }
+
+    auto direct = FromStringWithDefault<bool>(cgiParams.Get("direct"), false);
+    if (direct) {
+        return false;
+    }
+
+    if ((database == currentDatabase) || database.empty()) {
```
UgnineSirdis: Is this because we redirect only requests for database?

ydb/core/mon/audit/audit.cpp
```diff
@@ -126,28 +200,37 @@ void TAuditCtx::AddAuditLogParts(const TIntrusiveConstPtr<NACLib::TUserToken>& u
...
         AUDIT_PART("status", ToString(status));
         if (status != ERequestStatus::Success) {
-            AUDIT_PART("reason", response.Get()->Message);
+            AUDIT_PART("reason", (!reason.empty() ? reason : EMPTY_VALUE));
         }
     );
 }
 
+void TAuditCtx::LogOnExecute() {
+    LogAudit(ERequestStatus::Process, REASON_EXECUTE);
```
UgnineSirdis: What does this reason == Execute?
I thought that the reason field is about error message, something like this. But... before finish status we don't have any error message. Maybe it's not useful field for IN-PROCESS log record?

ydb/core/mon/audit/audit.cpp
```diff
@@ -126,28 +200,37 @@ void TAuditCtx::AddAuditLogParts(const TIntrusiveConstPtr<NACLib::TUserToken>& u
...
 
         for (const auto& [name, value] : Parts) {
-            AUDIT_PART(name, (!value.empty() ? value : EMPTY_VALUE))
+            AUDIT_PART(name, (!value.empty() ? value : EMPTY_VALUE));
         }
 
         AUDIT_PART("status", ToString(status));
         if (status != ERequestStatus::Success) {
-            AUDIT_PART("reason", response.Get()->Message);
+            AUDIT_PART("reason", (!reason.empty() ? reason : EMPTY_VALUE));
```
UgnineSirdis: Do we need to log reason at all when we actually don't have any reason?

ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,96 @@ namespace {
...
+    void BuildParamsFromBody(TCgiParameters& cgiParams, const NHttp::THttpIncomingRequestPtr& request) {
+        NHttp::THeaders headers(request->Headers);
+
+        auto contentType = NHttp::Trim(headers.Get("Content-Type").Before(';'), ' ');
+        if (contentType == "application/json") {
+            NJson::TJsonValue jsonData;
+            if (NJson::ReadJsonTree(request->Body, &jsonData) && jsonData.IsMap()) {
+                for (const auto& [key, value] : jsonData.GetMap()) {
+                    if (value.IsString()) {
+                        cgiParams.InsertUnescaped(key, value.GetStringRobust());
```
UgnineSirdis: Actually we need here only two parameters: database & tenant. Can we process only them? It will be more effective, but with the same effect.

ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,96 @@ namespace {
...
+                    }
+                }
+            }
+        } else if (contentType == "application/x-www-form-urlencoded") {
+            TCgiParameters bodyParams(request->Body);
+            for (const auto& [key, value] : bodyParams) {
```
UgnineSirdis: `cgiParams.insert(bodyParams.begin(), bodyParams.end());`

UgnineSirdis (changes requested):
A couple of small issues + we already can use LogPhase from config: https://github.com/ydb-platform/ydb/pull/22695

ydb/core/mon/audit/audit.cpp
UgnineSirdis: We already can use LogPhase from config: https://github.com/ydb-platform/ydb/pull/22695

ydb/core/mon/audit/audit.cpp
```diff
@@ -47,21 +45,96 @@ namespace {
...
+    return true;
 }
 
 void TAuditCtx::AddAuditLogPart(TStringBuf name, const TString& value) {
     Parts.emplace_back(name, value);
 }
 
-bool TAuditCtx::AuditableRequest(const TString& method, const TString& url, const TCgiParameters& cgiParams) {
+bool TAuditCtx::AuditableRequest(const NHttp::THttpIncomingRequestPtr& request) {
+    // we don't log redirects
```
adameat: No, you’re not handling redirects - you’re just skipping random requests based on a chosen set of parameters. A redirect can only be handled in one way: by checking the response. And this approach is extremely dangerous because it creates the illusion of having an audit… but in reality, what will happen, for example, when the viewer starts proxying calls to the databases?

adameat (changes requested):

adameat:
Such an implementation won’t last a year. It will simply stop working silently, creating the illusion of having an audit.

ydb/core/mon/audit/audit.cpp
StekPerepolnen: We discussed the meaning of this field and that it needs to be set when message received

ydb/core/mon/audit/audit.cpp
StekPerepolnen: I looked at how it’s done in the same SS and ended up removing it. Added reason to success.

ydb/core/mon/audit/audit.cpp
StekPerepolnen: it is relevate for http viewer handlers only

ydb/core/mon/audit/audit.cpp
StekPerepolnen: yes (it is relevate for http viewer handlers only though)

ydb/core/mon/audit/audit.cpp
StekPerepolnen: we redirect only if it’s to another database and only if the database parameter is in the URL.

ydb/core/mon/audit/audit.cpp
StekPerepolnen: we user `InsertUnescaped` method here

ydb/core/mon/audit/audit.cpp
StekPerepolnen: You’re right - from a strict compatibility perspective, the audit log entry flow may change.
I’ve rebased the PR to show the current behavior with the updated configuration model.

The configuration now allows explicitly specifying one or both phases for audit logging: `Received` and `Completed`.
By default, the `Completed` phase is always logged - and at this stage, every request is recorded, regardless of whether it involved a redirect, so **no** events are lost.

The idea is that, in addition to the usual logging at `Completed`, we can optionally enable logging at `Received`. In this case, the received entry is skipped for redirect cases. Please, see examples in the description.

This avoids creating duplicate alerts for the same user request while still ensuring that all redirect events appear in the `Completed` phase.

Hope it made it clearer. @adameat

**updated:**
I’m removing all redirect-related logic for now; we agreed to go with a simpler solution. Apologies if this causes any confusion.

ydb/core/mon/audit/audit.cpp
```diff
@@ -126,28 +128,34 @@ void TAuditCtx::AddAuditLogParts(const TIntrusiveConstPtr<NACLib::TUserToken>& u
     }
 }
 
-void TAuditCtx::FinishAudit(const NHttp::THttpOutgoingResponsePtr& response) {
-    AuditEnabled &= NKikimr::AppData()->AuditConfig.EnableLogging(NKikimrConfig::TAuditConfig::TLogClassConfig::ClusterAdmin, NKikimrConfig::TAuditConfig::TLogClassConfig::Completed, SubjectType);
+void TAuditCtx::LogAudit(ERequestStatus status, const TString& reason, NKikimrConfig::TAuditConfig::TLogClassConfig::ELogPhase logPhase) {
+    auto AuditEnabled = NKikimr::AppData()->AuditConfig.EnableLogging(NKikimrConfig::TAuditConfig::TLogClassConfig::ClusterAdmin, logPhase, SubjectType);
```
UgnineSirdis, line 132: AuditEnabled is local variable and should be written in different case: auditEnabled

StekPerepolnen:
> Such an implementation won’t last a year. It will simply stop working silently, creating the illusion of having an audit.

@adameat I’ve removed all the redirect-related logic. In this PR, the solution is straightforward: an audit log entry is **always** added when a request is received.

Could you please explicitly give it a ship-it if you have no objections?

adameat (approved):

UgnineSirdis (approved):

PR #22695, opened 2025-08-11
Support IN-PROCESS record for grpc calls

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### DIFF
--- a/ydb/core/grpc_services/audit_log.cpp
+++ b/ydb/core/grpc_services/audit_log.cpp
@@ -35,7 +35,7 @@ void AuditLogConn(const IRequestProxyCtx* ctx, const TString& database, const TS
     );
 }
 
-void AuditLog(ui32 status, const TAuditLogParts& parts)
+void AuditLog(std::optional<ui32> status, const TAuditLogParts& parts)
 {
     static const TString GrpcProxyComponentName = "grpc-proxy";
 
@@ -47,11 +47,15 @@ void AuditLog(ui32 status, const TAuditLogParts& parts)
             AUDIT_PART(name, (!value.empty() ? value : EmptyValue))
         }
 
-        AUDIT_PART("status", (status == Ydb::StatusIds::SUCCESS ? TString("SUCCESS") : TString("ERROR")))
-        AUDIT_PART("detailed_status", (Ydb::StatusIds::StatusCode_IsValid(status)
-            ? TString(Ydb::StatusIds::StatusCode_Name(status))
-            : ToString(status)
-        ))
+        if (status) {
+            AUDIT_PART("status", (*status == Ydb::StatusIds::SUCCESS ? TString("SUCCESS") : TString("ERROR")))
+            AUDIT_PART("detailed_status", (Ydb::StatusIds::StatusCode_IsValid(*status)
+                ? TString(Ydb::StatusIds::StatusCode_Name(*status))
+                : ToString(*status)
+            ))
+        } else {
+            AUDIT_PART("status", TString("IN-PROGRESS"))
+        }
     );
 }
 
--- a/ydb/core/grpc_services/audit_log.cpp
+++ b/ydb/core/grpc_services/audit_log.cpp
@@ -54,7 +54,7 @@ void AuditLog(std::optional<ui32> status, const TAuditLogParts& parts)
                 : ToString(*status)
             ))
         } else {
-            AUDIT_PART("status", TString("IN-PROGRESS"))
+            AUDIT_PART("status", TString("IN-PROCESS"))
         }
     );
 }
--- a/ydb/core/protos/config.proto
+++ b/ydb/core/protos/config.proto
@@ -1823,6 +1823,7 @@ message TAuditConfig {
     optional TFileBackend FileBackend = 2;
     optional TUnifiedAgentBackend UnifiedAgentBackend = 3;
     repeated TLogClassConfig LogClassConfig = 4;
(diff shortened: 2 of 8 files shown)


ydb/core/grpc_services/audit_log.cpp
```diff
@@ -47,11 +47,15 @@ void AuditLog(ui32 status, const TAuditLogParts& parts)
...
-            : ToString(status)
-        ))
+        if (status) {
+            AUDIT_PART("status", (*status == Ydb::StatusIds::SUCCESS ? TString("SUCCESS") : TString("ERROR")))
+            AUDIT_PART("detailed_status", (Ydb::StatusIds::StatusCode_IsValid(*status)
+                ? TString(Ydb::StatusIds::StatusCode_Name(*status))
+                : ToString(*status)
+            ))
+        } else {
+            AUDIT_PART("status", TString("IN-PROCESS"))
```
flown4qqqq, line 57: Is `IN-PROCESS` exactly meant? It seemed to me that in this case you are doing an audit of an operation that has just arrived. Or am I wrong?

ydb/core/protos/config.proto
```diff
@@ -1823,6 +1823,7 @@ message TAuditConfig {
     optional TFileBackend FileBackend = 2;
     optional TUnifiedAgentBackend UnifiedAgentBackend = 3;
     repeated TLogClassConfig LogClassConfig = 4;
+    optional bool LogRequestsOnReceiving = 5;
```
flown4qqqq: To be honestly, I dont like this fields's name. I would like to discuss name of it.

ydb/core/grpc_services/audit_log.cpp
UgnineSirdis: Sorry, I did not understand your question. Such a name described in doc: https://ydb.tech/docs/en/security/audit-log.
As see it as an operation that has just arrived and is to execute. So, in terms of status it has no finish status, so I use the status variable to define that the opration is not finished yet

flown4qqqq (dismissed):

ydb/core/blobstorage/nodewarden/distconf_invoke_common.cpp
```diff
@@ -252,41 +254,43 @@ namespace NKikimr::NStorage {
...
-                    AUDIT_PART("sanitized_token", wrapEmpty(userToken.GetSanitizedToken()))
-                    AUDIT_PART("status", TString("SUCCESS"))
-                    AUDIT_PART("reason", TString(), false)
-                    AUDIT_PART("operation", TString("REPLACE CONFIG"))
-                    AUDIT_PART("old_config", oldConfig)
-                    AUDIT_PART("new_config", newConfig)
-                );
+                if (AppData()->AuditConfig.EnableLogPhase(NKikimrConfig::TAuditConfig::Completed)) {
+                    AUDIT_LOG(
+                        const auto& replaceConfig = record.GetReplaceStorageConfig();
```
flown4qqqq: let's not put the code inside a macro?

ydb/core/cms/console/console_tenants_manager.cpp
```diff
@@ -650,13 +654,15 @@ class TSubDomainManip : public TActorBootstrapped<TSubDomainManip> {
...
                     << request->ToString());
 
-        AuditLogBeginRemoveDatabase(
-            Tenant->PeerName,
-            Tenant->UserToken.GetUserSID(),
-            Tenant->UserToken.GetSanitizedToken(),
-            Tenant->Path
-        );
-            
+        if (AppData()->AuditConfig.EnableLogPhase(NKikimrConfig::TAuditConfig::Completed)) {
```
flown4qqqq: At all, there is should be `received`.

but if we really want mark it like `received`, then the logs that have already appeared will stop appearing to the user during the update.

This applies to all audit logs in the console tablet.

flown4qqqq (changes requested):

ydb/core/blobstorage/nodewarden/distconf_invoke_common.cpp
UgnineSirdis: Do you want to refactor it in this PR?

ydb/core/cms/console/console_tenants_manager.cpp
UgnineSirdis: Yes, you're right. How do you think, can we refactor it now to the proper type?

flown4qqqq (approved):

ydb/core/protos/config.proto
```diff
@@ -1814,9 +1814,15 @@ message TAuditConfig {
             ServiceImpersonatedFromUser = 4;
         }
 
+        enum ELogPhase {
+            Received = 1;  // A request is received and the initial checks and authentication are made // "status" field of audit log record is equal to "IN-PROCESS"
```
CyberROFL, line 1818: Есть какой-то смысл логировать `IN-PROCESS`, но не логировать `SUCCESS`/`ERROR`?

UgnineSirdis: Кмк, не имеет смысла. Скорее может быть требование логировать только после выполнения, без промежуточных результатов

CyberROFL: А текущая реализация позволяет логировать `IN-PROCESS`, но не логировать `SUCCESS`/`ERROR`. Даже тест такой есть.

PR #22781, opened 2025-08-12
make documentation actual

SloNN:
Verified the documentation agains current column tables status

* Documentation


blinkov:
@SloNN 
1. this needs to be merged to `main` first
2. we currently don't use `stable-25-1-3` for documentation, only `stable-25-1`

PR #22812, opened 2025-08-13
Fix get implementation and route gets through bridge proxy when in Bridge mode

alexvru:
### Changelog entry 

Fix get implementation and route gets through bridge proxy when in Bridge mode

### Changelog category 

### Description for reviewers 

This fixes gets that they ensure that all required piles contain required data upon returning result.


mregrock (approved):
LGTM

kruall (approved):

PR #22908, opened 2025-08-14
Fix some issues in distconf

alexvru:
### Changelog entry 

Fix some issues in distconf

### Changelog category 

### Description for reviewers 

This patch fixes switchover process (now promoted pile becomes primary almost instantly) and fixes primary switch-induced assertion failure.


mregrock (approved):
LGTM

PR #22927, opened 2025-08-15
Retain RACE responses in bridge proxy

alexvru:
### Changelog entry 

Retain RACE responses in bridge proxy

### Changelog category 

### Description for reviewers 

This patch correctly handles RACE responses from proxies in Bridge mode to allow automatic handling of reconfiguration without reporting faulty statuses to tablets leading to their restart.


mregrock (approved):
LGTM

PR #22968, opened 2025-08-15
Fix distconf race when updating configs

alexvru:
### Changelog entry 

Fix distconf race when updating configs

### Changelog category 

### Description for reviewers 

This fixes a race leading to assertion failure in distconf.


mregrock (approved):
LGTM

PR #22989, opened 2025-08-16
Infradocs 1539.ydb returning

Hiro-mantich:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


ydb/docs/en/core/yql/reference/syntax/delete.md
```diff
@@ -33,3 +33,39 @@ SELECT * FROM $to_delete;
 * [BATCH DELETE](batch-delete.md)
 
 {% endif %}
+
+## RETURNING
```
blinkov, line 37: These articles weren't in the best shape initially, but now the order of sections has become even more odd...

Let's reorder content like this:

1. Syntax in general
2. Each feature in a little more detail (delete where, delete from, returning)
3. Examples covering each feature and their combinations

ydb/docs/en/core/yql/reference/syntax/delete.md
```diff
@@ -33,3 +33,39 @@ SELECT * FROM $to_delete;
...
+
+`RETURNING` returns the values of the deleted rows. This allows you to get the results of the operation immediately without a separate `SELECT` query.
+
+## Examples
+
+Returning all columns of deleted rows
```
blinkov, line 43:
```suggestion
### Returning all columns of deleted rows
```
Similarly for the rest of the examples.

ydb/docs/en/core/yql/reference/syntax/delete.md
```diff
@@ -33,3 +33,39 @@ SELECT * FROM $to_delete;
...
+
+```
+DELETE FROM my_table
+WHERE Key1 = 1
+RETURNING *;
+```
+
+Result
```
blinkov, line 51:
```suggestion
Result:
```

ydb/docs/en/core/yql/reference/syntax/delete.md
```diff
@@ -33,3 +33,39 @@ SELECT * FROM $to_delete;
 * [BATCH DELETE](batch-delete.md)
 
 {% endif %}
+
+## RETURNING
+
+`RETURNING` returns the values of the deleted rows. This allows you to get the results of the operation immediately without a separate `SELECT` query.
```
blinkov, line 39: The result of the `DELETE` operation is the absence of some rows, so it's unclear what we mean by "results" here. Let's rephrase this so it allows the application to know which rows were deleted like their primary keys or any other columns.

ydb/docs/en/core/yql/reference/syntax/insert_into.md
```diff
@@ -67,4 +67,39 @@ SELECT key FROM my_table_source;
 
 {% endif %}
 
+## RETURNING
 
+`RETURNING` returns the values of the inserted rows. This allows you to get the results of the operation immediately without a separate `SELECT` query.
+
+## Example
+
+Returning specific columns
```
blinkov, line 76: In the previous article, the first example was all columns, followed by specific columns in the second example. Here, their order is reversed for no apparent reason. Let's make the order consistent.

ydb/docs/en/core/yql/reference/syntax/delete.md
```diff
@@ -33,3 +33,39 @@ SELECT * FROM $to_delete;
...
+
+## Examples
```
blinkov, line 41: An example of `DELETE ... FROM` + `RETURNING` is missing.

ydb/docs/en/core/yql/reference/syntax/insert_into.md
```diff
@@ -67,4 +67,39 @@ SELECT key FROM my_table_source;
 
 {% endif %}
 
+## RETURNING
 
+`RETURNING` returns the values of the inserted rows. This allows you to get the results of the operation immediately without a separate `SELECT` query.
```
blinkov, line 72: We need to elaborate more on what's possible to return with these operations and what the limitations are. For example,

1. Can you do calculations like `RETURNING id + 5`? If yes, what are the limitations on these calculations?
2. Can you return the columns that are defined in the table but weren't inserted? For `INSERT INTO`, they'd probably be `NULL`s, but still, for other operations, it might make more sense.
3. How it interacts with other features like `INSERT INTO ... SELECT ...` or `INSERT INTO ... WITH TRUNCATE ...`

ydb/docs/en/core/yql/reference/syntax/update.md
```diff
@@ -37,3 +37,41 @@ SELECT * FROM $to_update;
...
+|1001|2022-12-15|shipped|200|
+|1002|2022-11-20|shipped|350|
+
+Returning specific columns
+
+```
+UPDATE products
+SET price = price * 0.9
+WHERE category = 'Electronics'
+RETURNING product_id, name, price AS new_price;
```
blinkov, line 69: This is the only mention of renaming columns; the `RETURNING` clause in other statements probably supports it too.

ydb/docs/en/core/yql/reference/syntax/upsert_into.md
```diff
@@ -31,3 +31,39 @@ VALUES ( 1, 10, 'Some text', Date('2021-10-07')),
        ( 2, 10, 'Some text', Date('2021-10-08'))
 ```
 
+## RETURNING
+
+`RETURNING` returns values:
+* For new rows – all inserted values
+* For updated rows – new values after the update
```
blinkov, line 38: It's unclear whether the table columns that weren't modified are accessible. And if they do, whether getting them has a performance impact.

blinkov (changes requested):

PR #23124, opened 2025-08-19
Improve Get/Range query performance in Bridge mode

alexvru:
### Changelog entry 

Improve Get/Range query performance in Bridge mode

### Changelog category 

### Description for reviewers 

This patch allows nonrestring reads to finish as first as the first response comes (instead of waiting for all of them), thus making latency better. However, it doesn't reduce number of reads made.


mregrock (approved):
LGTM

PR #23229, opened 2025-08-20
Fix bridge mode connectivity checking

alexvru:
### Changelog entry 

Fix bridge mode connectivity checking

### Changelog category 

### Description for reviewers 

This allows fully wiped pile to connect to the rest of the cluster; also this patch fixes some issues related to config propagation checks when doing failover.


mregrock (approved):
LGTM

PR #23343, opened 2025-08-22
fixed unstable viewer test

nshestakov:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/viewer/tests/test.py
+++ b/ydb/core/viewer/tests/test.py
@@ -989,24 +989,30 @@ def test_topic_data():
 
 
 def test_topic_data_cdc():
+    call_viewer("/viewer/query", {
+        'database': dedicated_db,
+        'query': 'create table table_test_topic_data_cdc(id int64, name text, primary key(id))',
+        'schema': 'multi'
+    })
+
     alter_response = call_viewer("/viewer/query", {
         'database': dedicated_db,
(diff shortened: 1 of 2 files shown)


ydb/core/viewer/tests/test.py
```diff
@@ -1029,43 +1035,50 @@ def test_async_replication_describe():
...
 
-    call_viewer("/viewer/query", {
+    create_result = get_viewer_normalized("/viewer/query", {
         'database': dedicated_db,
         'query': 'CREATE ASYNC REPLICATION `TestAsyncReplication` FOR `TableNotExists` AS `TargetAsyncReplicationTable` WITH (CONNECTION_STRING = "{}")'.format(endpoint),
         'schema': 'multi'
     })
 
-    result = get_viewer_normalized("/viewer/describe_replication", {
+    time.sleep(3)
```
adameat: why do you need sleep here? 
it's a very bad practice, I don't want to start it here... can we replace it with some kind of wait?

adameat (changes requested):

ydb/core/viewer/tests/test.py
nshestakov: after executing create, a schema object is created. but the tablet is just starting to run asynchronously: check for a table, create a changefeed, etc. and then the desktop can still output irrelevant information.

adameat (approved):

PR #23407, opened 2025-08-22
Added local backup support for system views

### LINKED ISSUE #23094: Dump and restore system views with materialized paths
To follow FSTEK guidelines we have to save all users' permissions and attributes for each scheme entry.
All details are presented in [RFC](https://github.com/ydb-platform/ydb-rfc/blob/main/sysviews_backup_policies.md).

n00bcracker:
# Add Local Backup Support for System Views

This PR implements backup and restore functionality for system views (`.sys` directory contents) in YDB. 

**Key Changes:**
- Extended backup infrastructure to handle system directories alongside regular directories
- Added `--replace-sys-acl` CLI option to control whether system object permissions are restored from backup or preserved from target
- Added comprehensive test coverage for system view backup/restore scenarios with ACL management
- Maintains backward compatibility - existing backups continue to work without changes

The feature enables complete database backup/restore including system dirs with configurable permission handling.

### DIFF
--- a/ydb/library/backup/db_iterator.h
+++ b/ydb/library/backup/db_iterator.h
@@ -47,6 +47,7 @@ private:
             NScheme::ESchemeEntryType::Topic,
             NScheme::ESchemeEntryType::CoordinationNode,
             NScheme::ESchemeEntryType::Replication,
+            NScheme::ESchemeEntryType::SysView,
         };
 
         return values;
@@ -157,6 +158,10 @@ public:
         return GetCurrentNode()->Type == NScheme::ESchemeEntryType::Directory;
     }
 
+    bool IsSystemDir() const {
+        return NConsoleClient::IsSystemDir(*GetCurrentNode());
+    }
+
     bool IsReplication() const {
         return GetCurrentNode()->Type == NScheme::ESchemeEntryType::Replication;
     }
@@ -169,6 +174,10 @@ public:
         return GetCurrentNode()->Type == NScheme::ESchemeEntryType::ExternalTable;
     }
 
+    bool IsSystemView() const {
+        return GetCurrentNode()->Type == NScheme::ESchemeEntryType::SysView;
+    }
+
     bool IsListed() const {
         return NextNodes.front().IsListed;
     }
@@ -177,8 +186,27 @@ public:
         return bool{NextNodes};
     }
 
-    bool IsSkipped() const {
-        return NConsoleClient::IsSystemObject(*GetCurrentNode());
+    bool IsMaterializedSysDir() {
+        const auto& entry = *GetCurrentNode();
+        if (IsDir() && entry.Name == ".sys") {
+            // TODO(n00bcracker): drop this check after removing EnableRealSystemViewPaths flag
+            const TString& fullPath = GetFullPath();
+            NScheme::TListDirectoryResult listResult = Client.ListDirectory(GetFullPath()).GetValueSync();
+            Y_ENSURE(listResult.IsSuccess(), "Can't list '.sys' directory, maybe it doesn't exist, dbPath# "
+                << fullPath.Quote());
+
+            for (const auto& child : listResult.GetChildren()) {
+                if (child.Type == NScheme::ESchemeEntryType::SysView) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
+    bool IsSkipped() {
+        return IsSystemDir() && !IsMaterializedSysDir();
     }
 
     void Next() {
--- a/ydb/public/lib/ydb_cli/common/sys.cpp
+++ b/ydb/public/lib/ydb_cli/common/sys.cpp
@@ -1,17 +1,37 @@
 #include "sys.h"
 
 #include <ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/scheme/scheme.h>
+#include <util/generic/hash_set.h>
 
 namespace NYdb::NConsoleClient {
 
-bool IsSystemObject(const NScheme::TSchemeEntry& entry) {
-    if (entry.Type != NScheme::ESchemeEntryType::Directory) {
+namespace {
+
+// In future should be syncronized with 'ydb/core/tx/schemeshard/schemeshard_system_names.cpp'
+const THashSet<TStringBuf> SystemReservedNames = {
+    ".sys",
+    ".metadata",
+    ".tmp",
+    ".backups",
+};
+
+}
+
+bool IsSystemName(const TStringBuf name) {
+    return SystemReservedNames.contains(name)
+        || name.StartsWith("~");
+}
+
+bool IsSystemDir(const NScheme::TSchemeEntry& entry) {
+    if (entry.Type == NScheme::ESchemeEntryType::Directory) {
+        return IsSystemName(entry.Name);
+    } else {
         return false;
     }
+}
 
-    return entry.Name.starts_with("~")
-        || entry.Name.starts_with(".sys")
-        || entry.Name.starts_with(".metadata");
+bool IsSystemObject(const NScheme::TSchemeEntry& entry) {
+    return IsSystemDir(entry) || entry.Type == NScheme::ESchemeEntryType::SysView;
 }
 
 }
--- a/ydb/public/lib/ydb_cli/dump/restore_impl.cpp
+++ b/ydb/public/lib/ydb_cli/dump/restore_impl.cpp
@@ -4,12 +4,14 @@
 
 #include <ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/discovery/discovery.h>
 #include <ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/proto/accessor.h>
+#include <ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/value/value.h>
 #include <ydb/public/api/protos/ydb_cms.pb.h>
 #include <ydb/public/api/protos/ydb_rate_limiter.pb.h>
 #include <ydb/public/api/protos/ydb_table.pb.h>
 #include <ydb/public/lib/ydb_cli/common/recursive_list.h>
 #include <ydb/public/lib/ydb_cli/common/recursive_remove.h>
 #include <ydb/public/lib/ydb_cli/common/retry_func.h>
+#include <ydb/public/lib/ydb_cli/common/sys.h>
 #include <ydb/public/lib/ydb_cli/dump/files/files.h>
 #include <ydb/public/lib/ydb_cli/dump/util/log.h>
 #include <ydb/public/lib/ydb_cli/dump/util/query_utils.h>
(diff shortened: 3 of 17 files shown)


ydb/library/backup/db_iterator.h
```diff
@@ -177,8 +182,27 @@ class TDbIterator {
         return bool{NextNodes};
     }
 
-    bool IsSkipped() const {
-        return NConsoleClient::IsSystemObject(*GetCurrentNode());
+    bool IsRealSysDir() {
```
jepett0: minor suggestion:

```suggestion
    bool IsMaterializedSysDir() {
```

ydb/library/backup/util.h
```diff
@@ -6,7 +6,7 @@
 
 #define LOG_IMPL(log, level, message) \
     if (log->FiltrationLevel() >= level) { \
-        log->Write(level, TStringBuilder() << message); \
+        log->Write(level, TStringBuilder() << message << Endl); \
```
jepett0: Have you checked, how it looks in the terminal? I know that the newline symbol is missing from the logs, but somehow, it was __not__ missing in the terminal before. I'm afraid that there would be two newlines in the terminal now

ydb/public/lib/ydb_cli/dump/restore_impl.cpp
```diff
@@ -58,6 +58,12 @@ bool IsFileExists(const TFsPath& path) {
     return path.Exists() && path.IsFile();
 }
 
+constexpr bool IsSystemDir(const TString& dbPath) {
+    return dbPath.EndsWith("~")
```
jepett0: `EndsWith`? Are you sure? Previously, you defined a system directory as the one __starting__ with these strings

ydb/public/lib/ydb_cli/dump/restore_impl.cpp
```diff
@@ -119,10 +125,69 @@ Ydb::Cms::CreateDatabaseRequest ReadDatabaseDescription(const TFsPath& fsDirPath
...
+        if (it == actualColumnsMap.end()) {
+            return Result<TRestoreResult>(EStatus::SCHEME_ERROR, TStringBuilder()
+                << "Column " << TString(dumpedCol.name()).Quote() << " from dump is missing in actual system view");
+        }
+
+        const auto& actualCol = *it->second;
+        TString dumpedTypeStr = dumpedCol.type().SerializeAsString();
+        TString actualTypeStr = actualCol.type().SerializeAsString();
+
+        if (dumpedTypeStr != actualTypeStr) {
```
jepett0: I'm not sure this is a correct way to compare two protobufs. My AI is telling me this:

```
No, it's generally not correct to compare them as serialized wire messages for several reasons:

- Field order differences: Protobuf serialization doesn't guarantee field order
- Default values: Missing fields vs fields with default values may serialize differently
- Unknown fields: Serialization may include unknown fields that shouldn't affect equality
- Serialization format: Different serialization options (packed vs unpacked repeated fields)
- Performance: Serialization is much slower than direct field comparison
```

ydb/public/lib/ydb_cli/dump/restore_impl.cpp
```diff
@@ -1163,27 +1212,28 @@ TRestoreResult TRestoreClient::DropAndRestoreExternals(const TVector<TFsBackupEn
...
 
     for (const auto& [dbPath, i] : externalTables) {
-        if (existingEntries.contains(dbPath)) {
-            auto result = Drop(ESchemeEntryType::ExternalTable, dbPath, settings);
-            if (!result.IsSuccess()) {
+        if (ExistingEntries.contains(dbPath)) {
+            if (auto result = Drop(ESchemeEntryType::ExternalTable, dbPath, settings); !result.IsSuccess()) {
                 return result;
             }
+            ExistingEntries.erase(dbPath);
```
jepett0: `Drop` may be successful without actually dropping the object if the `--dry-run` option is enabled. Would it break something if we erase the `dbPath` from the `ExistingEntries` here?

ydb/public/lib/ydb_cli/dump/util/log.h
```diff
@@ -4,7 +4,7 @@
 
 #define LOG_IMPL(log, level, message) \
     if (log->FiltrationLevel() >= level) { \
-        log->Write(level, TStringBuilder() << message); \
+        log->Write(level, TStringBuilder() << message << Endl); \
```
jepett0: Same concern about the double newline in the terminal. Have you checked it?

ydb/services/ydb/backup_ut/ya.make
```diff
@@ -19,7 +19,7 @@ SRCS(
 PEERDIR(
     contrib/libs/fmt
     library/cpp/streams/zstd
-    ydb/core/testlib/default
+    ydb/core/testlib/pg
```
jepett0, line 22: Are you sure this is justified? Why do you need it?

ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
```diff
@@ -872,6 +876,94 @@ void TestViewDependentOnAnotherViewIsRestored(
...
+
+    result = client.DescribePath(systemDirectory).ExtractValueSync();
+    UNIT_ASSERT_C(result.IsSuccess(), result.GetIssues().ToString());
+    UNIT_ASSERT_VALUES_EQUAL(result.GetEntry().Type, ESchemeEntryType::Directory);
+    const auto replacedSystemDirPermissions = result.GetEntry().Permissions;
+    const auto replacedSystemDirOwner = result.GetEntry().Owner;
+
+    UNIT_ASSERT_VALUES_EQUAL(replacedSystemDirPermissions.size(), dumpedSystemDirPermissions.size());
+    for (size_t i = 0; i < replacedSystemDirPermissions.size(); ++i) {
+       UNIT_ASSERT_C(EqualPermissions(replacedSystemDirPermissions[i], dumpedSystemDirPermissions[i]),
```
jepett0: Is the order of the permissions guaranteed?

ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
```diff
@@ -2021,11 +2113,19 @@ Y_UNIT_TEST_SUITE(BackupRestore) {
             driverConfig.SetAuthToken("root@builtin");
         }
         auto driver = TDriver(driverConfig);
-
+        TSchemeClient schemeClient(driver);
         NQuery::TQueryClient queryClient(driver);
         auto session = queryClient.GetSession().ExtractValueSync().GetSession();
         TReplicationClient replicationClient(driver);
 
+        if (useSecret) {
```
jepett0, line 2137: Why do you need to add this?

ydb/tests/functional/ydb_cli/test_ydb_backup.py
```diff
@@ -102,20 +102,20 @@ def modify_permissions(scheme_client, path):
...
+    if not are_tables_descriptions_the_same(session, path_left, path_right):
         return False
 
-    if not is_permissions_the_same(scheme_client, path_left, path_right):
+    if not are_permissions_the_same_for_two_paths(scheme_client, path_left, path_right):
         return False
 
     if check_data:
-        return is_data_the_same(session, path_left, path_right)
+        return are_data_the_same(session, path_left, path_right)
```
jepett0: I think `data` is singular, so it should be `is_data_the_same`. My AI says the following:

```
Tech industry standard: In programming and IT, data is almost always treated as singular:

- "The data is encrypted"
- "This data structure"
- "Data validation is important"
```

ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
ijon: There are no such guarantee. And a client shouldn't rely on that anyway.

ydb/services/ydb/backup_ut/ya.make
n00bcracker: If leave it as it was I get an exception when I describe pg system views. It happens because columns of these sys views have pg types. But these pg sys views are needed during dump operation.

ydb/public/lib/ydb_cli/dump/restore_impl.cpp
ijon: I suggest moving this to `ydb/public/lib/ydb_cli/common/sys.h` and unify name check between two methods (and may be even leave only one method).

ydb/public/lib/ydb_cli/common/sys.cpp
```diff
@@ -4,14 +4,18 @@
...
-    if (entry.Type != NScheme::ESchemeEntryType::Directory) {
+bool IsSystemDir(const NScheme::TSchemeEntry& entry) {
+    if (entry.Type == NScheme::ESchemeEntryType::Directory) {
+        return entry.Name.starts_with("~")
+            || entry.Name.starts_with(".sys")
+            || entry.Name.starts_with(".metadata");
+    } else {
         return false;
     }
+}
```
ijon: FYI: This name check is outdated, there are more directories that are considered "system" and which content should not be backed up (and can't be restored).
See `ydb/core/tx/schemeshard/schemeshard_system_names.cpp`.

I don't suggest to use reserved names registry now, in this PR, but eventually it should be done somehow. Thought it should be mentioned.

ydb/services/ydb/backup_ut/ydb_backup_ut.cpp
n00bcracker: Because we won't ignore system dirs during dump/restore operations, the user we use in tests needs to have permissions to read system dirs. But he isn't owner of them, the owner of system dirs is `metadata@system`. Therefore we have to modify permissions before a test start.

ydb/public/lib/ydb_cli/dump/restore_impl.cpp
n00bcracker: Yes, you're right, because CheckExistenceAndType func use ExistingEntries to check whether a path exists

ydb/public/lib/ydb_cli/dump/util/log.h
n00bcracker: Yes, checked, all like you described. I decided to leave as it was

ydb/library/backup/util.h
n00bcracker: Yes, there is a bug and I haven't found a quick solution, so reverted these changes.

ydb/public/lib/ydb_cli/dump/restore_impl.cpp
```diff
@@ -2077,39 +2201,59 @@ TRestoreResult TRestoreClient::RestorePermissions(
...
+    }
 }
 
 } // NYdb::NDump
 
 Y_DECLARE_OUT_SPEC(, NYdb::NDump::NPrivate::TLocation, o, x) {
     return x.Out(o);
 }
+
+
```
jepett0, line 2259: minor nit-pick: extra newline

ydb/tests/functional/ydb_cli/test_ydb_backup.py
```diff
@@ -1754,6 +1758,208 @@ def test_table_replacement(self):
...
+            is_(True),
+        )
+
+        # Check the restored directory
+        assert_that(
+            self.scheme_listdir('/Root/restored'),
+            contains_inanyorder('folder')
+        )
+
+    def test_enabled_option(self):
```
jepett0, line 1878: If the tests:
- test_disabled_option
- test_enabled_option

are the same except for the option value and `.sys` existence checks, you can create a parametrized test to deduplicate the code.

ijon (approved):
lgtm

jepett0 (approved):
Overall looks good

ydb/tests/functional/ydb_cli/test_ydb_backup.py
n00bcracker: No, for enabled option I also check `dry-run` mode

Gazizonoki (approved):
SDK - ok

PR #23529, opened 2025-08-25
Fix infinite retries for TEvCollectGarbage on deleted tablets when doing Bridge sync

alexvru:
### Changelog entry 

Fix infinite retries for TEvCollectGarbage on deleted tablets when doing Bridge sync

### Changelog category 

### Description for reviewers 

This patch fixes incorrect syncer behaviour when it issues CollectGarbage command with keep flags only with maximum generation set and without setting IgnoreBlock flag, thus leading to impossibility of this command execution when tablet gets deleted and maximum generation is blocked. In the case of failure command gets retried infinitely, leading to impossibility to get in SYNCHRONIZED state.


mregrock (approved):
LGTM

PR #23567, opened 2025-08-26
Do not hop distconf requests unless session is established

alexvru:
### Changelog entry 

Do not hop distconf requests unless session is established

### Changelog category 

### Description for reviewers 

Don't try to hop requests to binding when binding is still in progress.


mregrock (approved):
LGTM

PR #23642, opened 2025-08-27
fix stream lookup join chains: keep seqno for input row

gridnevvvit:
### Changelog category 

### Description for reviewers 

...


ydb/core/kqp/opt/logical/kqp_opt_log_indexes.cpp
```diff
@@ -852,6 +852,10 @@ TExprBase KqpRewriteStreamLookupIndex(const TExprBase& node, TExprContext& ctx,
                         .Tuple("tuple")
                         .Index().Value("0").Build()
                         .Build()
+                    .Add<TCoNth>()
```
nikvas0, line 855: Also need to change comment above

nikvas0 (approved):

PR #23669, opened 2025-08-27
Protect BridgeProxy request against races with configuration update

alexvru:
### Changelog entry 

Protect BridgeProxy request against races with configuration update

### Changelog category 

### Description for reviewers 

This fixes potential data race when configuration update comes during query execution in Bridge proxy.


mregrock (approved):
LGTM

PR #23689, opened 2025-08-28
Fixing slow UPDATES for Column Tables

### LINKED ISSUE #23006: Slow UPDATEs in OLAP, Column Tables
Discovered here https://github.com/ydb-platform/ydb/issues/22255

Reproduce (ydb/tests/olap/test_cs_many_updates.py):
- Start a cluster that supports OLAP `table_service_config: enable_olap_sink: true`
- Create a column table table `WITH (STORE = COLUMN)`. Have there an id (int32) and a value (int64)
- Insert 10 random rows
- Update rows one by one with random values

You will get an average update time 2-3 seconds.

 Look at the ybd logs, you will see this for every slow update
`2025-08-11T14:40:12.187743Z :KQP_SLOW_LOG NOTICE: kqp_worker_common.cpp:132: TraceId: "01k2cs64f76q948qc1m9kb5ay6", SessionId: ydb://session/3?node_id=1&id=N2U2N2VmMjUtNzc1YWI3YTctNmQzMTdiZTYtNzU3YzY3NjA=, Slow query, duration: 2.996352s, status: STATUS_CODE_UNSPECIFIED, user: UNAUTHENTICATED, results: 0b, text: "UPDATE /Root/cs_simple_reader_many_modifications/test_many_modifications SET value = 904150 WHERE id = 0;", parameters: 0b`

and many these for every slow update
`2025-08-11T14:57:11.698595Z :TX_COLUMNSHARD WARN: log.cpp:839: tablet_id=72075186224037941;self_id=[1:7537342166418338179:2409];ev=NActors::TEvents::TEvWakeup;fline=sync.h:19;event=tx_timeout;lock=281474976710683;tx_id=281474976710684;d=2.000395s;`

kirillvasilenko:
fixes #23006

See the comment on the change with explanation.

See the `TTxWriteReceivedBrokenFlag::DoExecute() and DoComplete()` here. 
[link](https://github.com/kirillvasilenko/ydb/blob/e8ee063dddd3c989594eee63927984d7deb95440/ydb/core/tx/columnshard/transactions/operators/ev_write/primary.h#L89)

If we erase the TabletId from the set, we will get the following behavior. For large enough number of shards (> 20), the arbiter will decides that the tx is finished, and will remove the operator for the tx (see TTxController::ProgressOnExecute). And all the DoComplete after that will get "absent operator" error. So, the whole tx will hang for 2-3 seconds, until "not completed" shards woken up and and do something about it.

So, that is why we only check TabletId in the set here, and remove it in DoComplete.

### DIFF
--- a/ydb/core/tx/columnshard/transactions/operators/ev_write/primary.h
+++ b/ydb/core/tx/columnshard/transactions/operators/ev_write/primary.h
@@ -86,7 +86,13 @@ private:
             auto op = Self->GetProgressTxController().GetTxOperatorVerifiedAs<TEvWriteCommitPrimaryTransactionOperator>(TxId, true);
             if (!op) {
                 AFL_WARN(NKikimrServices::TX_COLUMNSHARD_TX)("event", "repeated shard broken_flag info")("shard_id", TabletId)("reason", "absent operation");
-            } else if (!op->WaitShardsBrokenFlags.erase(TabletId)) {
+                // looks like it is the only place where we delete shards from the set
+                // we see in the logs that we had all the shards in the set, and later there is nothing in the set
+                // what if we do delete the shards from the set here?
+                // but we get to the DoComplete method (look at the next method) for slow shards later (when the tx is finished), for some reason
+                // when and how DoExecute is called, and how and when DoComplete is called?
+            //} else if (!op->WaitShardsBrokenFlags.erase(TabletId)) {
+            } else if (!op->WaitShardsBrokenFlags.contains(TabletId)) {
                 AFL_WARN(NKikimrServices::TX_COLUMNSHARD_TX)("event", "repeated shard broken_flag info")("shard_id", TabletId);
             } else {
                 op->TxBroken = op->TxBroken.value_or(false) || BrokenFlag;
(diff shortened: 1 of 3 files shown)


ydb/core/tx/columnshard/transactions/operators/ev_write/primary.h
```diff
@@ -86,7 +86,7 @@ class TEvWriteCommitPrimaryTransactionOperator: public TEvWriteCommitSyncTransac
             auto op = Self->GetProgressTxController().GetTxOperatorVerifiedAs<TEvWriteCommitPrimaryTransactionOperator>(TxId, true);
             if (!op) {
                 AFL_WARN(NKikimrServices::TX_COLUMNSHARD_TX)("event", "repeated shard broken_flag info")("shard_id", TabletId)("reason", "absent operation");
-            } else if (!op->WaitShardsBrokenFlags.erase(TabletId)) {
+            } else if (!op->WaitShardsBrokenFlags.contains(TabletId)) {
```
kirillvasilenko, line 89: If we erase the TabletId from the set, we will get the following behavior. For large enough number of shards (> 20), the arbiter will decides that the tx is finished, and will remove the operator for the tx (see TTxController::ProgressOnExecute). And all the DoComplete after that will get "absent operator" error. So, the whole tx will hang for 2-3 seconds, until "not completed" shards woken up and and do something about it.

So, that is why we only check TabletId in the set here, and remove it in DoComplete.

dorooleg:
Отличная находка!

dorooleg (dismissed):

dorooleg (approved):

Issue #23429, opened 2025-08-22
Y_VERIFY at distconf.cpp:309


alexvru:
```
Aug 22 19:13:11 ydb-vla-dev01-004 kikimr[1473588]: ========== BEGIN PANICKING ==========                    
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]: VERIFY failed (2025-08-22T19:13:09.868719Z): GlobalQuorum# 1
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]:   ydb/core/blobstorage/nodewarden/distconf.cpp:309                                                         
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]:   ConsistencyCheck(): requirement HasConnectedNodeQuorum(*StorageConfig, false) == GlobalQuorum failed
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]: 0. /home/alexvru/work/util/system/yassert.cpp:108: NPrivate::InternalPanicImpl(int, char const*, char const
*, int, int, int, TBasicStringBuf<char, std::__y1::char_traits<char>>, char const*, unsigned long) @ 0xA6D103A                                                
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]: 1. /home/alexvru/work/util/system/yassert.cpp:55: NPrivate::Panic(NPrivate::TStaticBuf const&, int, char co
nst*, char const*, char const*, ...) @ 0xA6C9DCC                     
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]: 2. /home/alexvru/work/ydb/core/blobstorage/nodewarden/distconf.cpp:0: NKikimr::NStorage::TDistributedConfig
Keeper::ConsistencyCheck() @ 0x157249F0                                                                                                                       
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]: 3. /home/alexvru/work/ydb/core/blobstorage/nodewarden/distconf.cpp:448: NKikimr::NStorage::TDistributedConf
igKeeper::StateFunc(TAutoPtr<NActors::IEventHandle, TDelete>&) @ 0x15726408    
Aug 22 19:13:12 ydb-vla-dev01-004 kikimr[1473588]: 4. /home/alexvru/work/ydb/library/actors/core/actor.cpp:406: NActors::IActor::Receive(TAutoPtr<NActors::IEv
[...]

PR #23790, opened 2025-08-29
Fix quorum problem

alexvru:
### Changelog entry 

Fix quorum problem in distconf

### Changelog category 

* Bugfix 

### Description for reviewers 

This fixes bug #23429, allowing to peer pile's nodes to connect to root node unless it isn't root anymore.


mregrock (approved):
LGTM

PR #23855, opened 2025-09-01
EXT-1481 OIDC verify on whoami requests when actor system shutdown

StekPerepolnen:
### Changelog entry 

- fix verify on whoami requests when actor system shutdown
- fix double PassAway on 404 whoami

**refactoring:**
- changed response status check
- TProxiedResponseParams stored by value, simplifying memory management

### Changelog category 

* Bugfix 

### Description for reviewers 

...

### DIFF
--- a/ydb/mvp/core/appdata.cpp
+++ b/ydb/mvp/core/appdata.cpp
@@ -3,10 +3,14 @@
 #include <ydb/library/actors/core/actor.h>
 #include <ydb/public/sdk/cpp/src/library/grpc/client/grpc_client_low.h>
 
-TMVPAppData::TMVPAppData() {
-    GRpcClientLow = std::make_shared<NYdbGrpc::TGRpcClientLow>();
-}
+namespace NMVP{
+
+TMVPAppData::TMVPAppData()
+    : GRpcClientLow(std::make_shared<NYdbGrpc::TGRpcClientLow>())
+{}
 
 TMVPAppData* MVPAppData() {
     return NActors::TActivationContext::ActorSystem()->template AppData<TMVPAppData>();
 }
+
+} // namespace NMVP
--- a/ydb/mvp/core/appdata.cpp
+++ b/ydb/mvp/core/appdata.cpp
@@ -1,8 +1,12 @@
 #include "appdata.h"
+#include <ydb/library/actors/core/actorsystem.h>
+#include <ydb/library/actors/core/actor.h>
 #include <ydb/public/sdk/cpp/src/library/grpc/client/grpc_client_low.h>
 
-TMVPAppData::~TMVPAppData() {
-    if (GRpcClientLow) {
-        GRpcClientLow->Stop(true);
-    }
+TMVPAppData::TMVPAppData() {
+    GRpcClientLow = std::make_shared<NYdbGrpc::TGRpcClientLow>();
+}
+
+TMVPAppData* MVPAppData() {
+    return NActors::TActivationContext::ActorSystem()->template AppData<TMVPAppData>();
 }
--- a/ydb/mvp/core/appdata.cpp
+++ b/ydb/mvp/core/appdata.cpp
@@ -1,0 +1,8 @@
+#include "appdata.h"
+#include <ydb/public/sdk/cpp/src/library/grpc/client/grpc_client_low.h>
+
+TMVPAppData::~TMVPAppData() {
+    if (GRpcClientLow) {
+        GRpcClientLow->Stop(true);
+    }
+}
--- a/ydb/mvp/core/appdata.h
+++ b/ydb/mvp/core/appdata.h
@@ -1,23 +1,23 @@
 #pragma once
 #include <library/cpp/monlib/metrics/metric_registry.h>
 
-namespace NMVP {
-    // forward declaration from "mvp_tokens.h"
-    class TMvpTokenator;
-}
-
 namespace NYdbGrpc {
     inline namespace Dev {
         class TGRpcClientLow;
     }
 }
 
-struct TMVPAppData {
-    std::shared_ptr<NMonitoring::TMetricRegistry> MetricRegistry;
-    NMVP::TMvpTokenator* Tokenator = nullptr;
-    std::shared_ptr<NYdbGrpc::TGRpcClientLow> GRpcClientLow;
+namespace NMVP {
+    // forward declaration from "mvp_tokens.h"
+    class TMvpTokenator;
 
-    TMVPAppData();
-};
+    struct TMVPAppData {
+        std::shared_ptr<NMonitoring::TMetricRegistry> MetricRegistry;
+        TMvpTokenator* Tokenator = nullptr;
+        std::shared_ptr<NYdbGrpc::TGRpcClientLow> GRpcClientLow;
 
-TMVPAppData* MVPAppData();
+        TMVPAppData();
+    };
+
+    TMVPAppData* MVPAppData();
+}
--- a/ydb/mvp/core/appdata.h
+++ b/ydb/mvp/core/appdata.h
@@ -1,6 +1,4 @@
 #pragma once
-#include <ydb/library/actors/core/actorsystem.h>
-#include <ydb/library/actors/core/actor.h>
 #include <library/cpp/monlib/metrics/metric_registry.h>
 
 namespace NMVP {
@@ -18,9 +16,8 @@ struct TMVPAppData {
     std::shared_ptr<NMonitoring::TMetricRegistry> MetricRegistry;
     NMVP::TMvpTokenator* Tokenator = nullptr;
     std::shared_ptr<NYdbGrpc::TGRpcClientLow> GRpcClientLow;
-    ~TMVPAppData();
+
+    TMVPAppData();
 };
 
-inline TMVPAppData* MVPAppData() {
-    return NActors::TActivationContext::ActorSystem()->template AppData<TMVPAppData>();
-}
+TMVPAppData* MVPAppData();
--- a/ydb/mvp/core/appdata.h
+++ b/ydb/mvp/core/appdata.h
@@ -1,4 +1,6 @@
 #pragma once
+#include <ydb/library/actors/core/actorsystem.h>
+#include <ydb/library/actors/core/actor.h>
 #include <library/cpp/monlib/metrics/metric_registry.h>
 
 namespace NMVP {
@@ -6,9 +8,17 @@ namespace NMVP {
     class TMvpTokenator;
 }
 
+namespace NYdbGrpc {
+    inline namespace Dev {
+        class TGRpcClientLow;
+    }
+}
+
 struct TMVPAppData {
     std::shared_ptr<NMonitoring::TMetricRegistry> MetricRegistry;
     NMVP::TMvpTokenator* Tokenator = nullptr;
+    std::shared_ptr<NYdbGrpc::TGRpcClientLow> GRpcClientLow;
+    ~TMVPAppData();
 };
 
 inline TMVPAppData* MVPAppData() {
--- a/ydb/mvp/core/mvp_test_runtime.cpp
+++ b/ydb/mvp/core/mvp_test_runtime.cpp
@@ -45,9 +45,7 @@ void TMvpTestRuntime::InitNodeImpl(TNodeDataBase* node, size_t nodeIndex) {
     node->LogSettings->SetLevel(NActors::NLog::PRI_DEBUG, NMVP::EService::GRPC, explanation);
     node->LogSettings->SetLevel(NActors::NLog::PRI_INFO, NMVP::EService::QUERY, explanation);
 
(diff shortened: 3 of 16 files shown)


ydb/mvp/oidc_proxy/openid_connect.h
```diff
@@ -97,6 +97,11 @@ std::unique_ptr<NYdbGrpc::TServiceConnection<TSessionService>> CreateGRpcService
     return client.CreateGRpcServiceConnection<TSessionService>(config);
 }
 
+inline NYdbGrpc::IQueueClientContextPtr CreateContext() {
+    static NYdbGrpc::TGRpcClientLow client;
```
UgnineSirdis: Why can't we place it in AppData, as it is done in YDB?

The source problem was during shutdown of test or server:
- we had grpc client that was not asked to shutdown
- actor system was shutdown
- a callback provided for grpc client had a pointer to the actor system that was not existing at the time that callback was run

Global variables here are not a good way to solve such problems

ydb/mvp/oidc_proxy/extension_whoami.cpp
```diff
@@ -166,6 +166,13 @@ void TExtensionWhoamiWorker::ContinueAndPassAway() {
     PassAway();
 }
 
+void TExtensionWhoamiWorker::PassAway() {
+    if (RequestContext) {
+        RequestContext->Cancel();
```
UgnineSirdis, line 173: I'm not sure 100%, but as I remember this callback should be called nevertheless after cancel

UgnineSirdis (changes requested):

ydb/mvp/oidc_proxy/openid_connect.h
StekPerepolnen: agreed, moved to AppData

ydb/mvp/oidc_proxy/extension_whoami.cpp
StekPerepolnen: The callback is indeed invoked after Cancel(), the key point is that it happens immediately after the cancel while the actor is still alive. I also added a destructor to MvpAppData, where the client is located now, to ensure the shared client is fully stopped during final cleanup.

ydb/mvp/core/appdata.h
```diff
@@ -1,14 +1,24 @@
 #pragma once
+#include <ydb/library/actors/core/actorsystem.h>
```
UgnineSirdis: As I see we have not added anything related to these two headers.
I would suggest to add implementation of `MVPAppData()` to the newly created cpp-file

UgnineSirdis (commented):
I see two variables that are related to our problem, but not used at all: ActorSystemStoppingLock and ActorSystemStopping (in both copies of mvp.h). Can you remove them and write a proper shutdown function that will call actor system stop/cleanup in proper order as here in YDB https://github.com/ydb-platform/ydb/blob/235f5059239204ebef916dafe7b5ec52fe8b56b2/ydb/core/driver_lib/run/run.cpp#L2012 and then call grpc client stop?
I think this would be the most prefereable and already debugged way to shutdown the system.

ydb/mvp/core/appdata.cpp
```diff
@@ -0,0 +1,8 @@
+#include "appdata.h"
+#include <ydb/public/sdk/cpp/src/library/grpc/client/grpc_client_low.h>
+
+TMVPAppData::~TMVPAppData() {
+    if (GRpcClientLow) {
+        GRpcClientLow->Stop(true);
```
UgnineSirdis: Can it be stopped twice?

ydb/mvp/core/mvp_test_runtime.cpp
```diff
@@ -42,7 +45,9 @@ void TMvpTestRuntime::InitNodeImpl(TNodeDataBase* node, size_t nodeIndex) {
     node->LogSettings->SetLevel(NActors::NLog::PRI_DEBUG, NMVP::EService::GRPC, explanation);
     node->LogSettings->SetLevel(NActors::NLog::PRI_INFO, NMVP::EService::QUERY, explanation);
 
-    node->AppData0.reset(new TMVPAppData());
+    auto mvpAppData = new TMVPAppData();
+    mvpAppData->GRpcClientLow = std::make_shared<NYdbGrpc::TGRpcClientLow>();
```
UgnineSirdis: Does it make sense to create client in constructor of `TMVPAppData`?

ydb/mvp/core/mvp_test_runtime.cpp
```diff
@@ -42,7 +45,9 @@ void TMvpTestRuntime::InitNodeImpl(TNodeDataBase* node, size_t nodeIndex) {
...
+    node->AppData0.reset(mvpAppData);
```
UgnineSirdis: `node->AppData0 = std::make_shared<TMVPAppData>();`
In this case shared ptr node->AppData0 will have the proper deleter function.

ydb/mvp/oidc_proxy/extension.cpp
```diff
@@ -7,13 +7,13 @@ void TExtensionContext::Reply(NHttp::THttpOutgoingResponsePtr httpResponse) {
...
+    if (!Params.StatusOverride.empty()) {
+        return Reply(Params.Request->CreateResponse(Params.StatusOverride, Params.MessageOverride, *Params.HeadersOverride, Params.BodyOverride));
     } else {
         static constexpr size_t MAX_LOGGED_SIZE = 1024;
-        BLOG_D("Can not process request to protected resource:\n" << Params->Request->GetObfuscatedData().substr(0, MAX_LOGGED_SIZE));
-        return Reply(CreateResponseForNotExistingResponseFromProtectedResource(Params->Request, Params->ResponseError));
-    }
+        BLOG_D("Can not process request to protected resource:\n" << Params.Request->GetObfuscatedData().substr(0, MAX_LOGGED_SIZE));
+        return Reply(CreateResponseForNotExistingResponseFromProtectedResource(Params.Request, Params.ResponseError));
+}
```
UgnineSirdis: incorrect indent

ydb/mvp/oidc_proxy/openid_connect.h
```diff
@@ -92,9 +93,12 @@ std::unique_ptr<NYdbGrpc::TServiceConnection<TSessionService>> CreateGRpcService
     NYdbGrpc::TGRpcClientConfig config;
     config.Locator = host;
     config.EnableSsl = (scheme == "grpcs");
-    static NYdbGrpc::TGRpcClientLow client;
     SetGrpcKeepAlive(config);
-    return client.CreateGRpcServiceConnection<TSessionService>(config);
+    return MVPAppData()->GRpcClientLow->CreateGRpcServiceConnection<TSessionService>(config);
+}
+
+inline NYdbGrpc::IQueueClientContextPtr CreateContext() {
```
UgnineSirdis: This name is very common. At the same time the body is pretty simple: `MVPAppData()->GRpcClientLow->CreateContext();`
I would suggest to use the expression `MVPAppData()->GRpcClientLow->CreateContext();` directly. It will be more clear in code

ydb/mvp/oidc_proxy/extension_whoami.cpp
```diff
@@ -166,6 +166,13 @@ void TExtensionWhoamiWorker::ContinueAndPassAway() {
     PassAway();
 }
 
+void TExtensionWhoamiWorker::PassAway() {
+    if (RequestContext) {
```
UgnineSirdis, line 172: I think in case of reply from tris RequestContext - TEvGetProfileResponse/TEvErrorResponse - we need to write nullptr to it?
But I'm not sure, it's worth to check how to properly kill context

UgnineSirdis (changes requested):

ydb/mvp/core/appdata.h
StekPerepolnen: I removed cpp-file completely since the destructor is not needed.

ydb/mvp/oidc_proxy/extension.cpp
StekPerepolnen: changed

ydb/mvp/oidc_proxy/extension_whoami.cpp
StekPerepolnen: probably `RequestContext.reset();`
added

ydb/mvp/core/mvp_test_runtime.cpp
StekPerepolnen: moved to TMVPAppData constructor

ydb/mvp/oidc_proxy/openid_connect.h
StekPerepolnen: changed

ydb/mvp/core/mvp_test_runtime.cpp
StekPerepolnen: moved to constructor

ydb/mvp/core/appdata.cpp
StekPerepolnen: As far as I can see calling `GRpcClientLow->Stop(true)` multiple times is safe. Repeated calls just do nothing.

ydb/mvp/core/appdata.cpp
```diff
@@ -0,0 +1,12 @@
+#include "appdata.h"
+#include <ydb/library/actors/core/actorsystem.h>
+#include <ydb/library/actors/core/actor.h>
+#include <ydb/public/sdk/cpp/src/library/grpc/client/grpc_client_low.h>
+
+TMVPAppData::TMVPAppData() {
```
UgnineSirdis:
```suggestion
TMVPAppData::TMVPAppData()
    : GRpcClientLow(std::make_shared<NYdbGrpc::TGRpcClientLow>())
{
}
```

ydb/mvp/core/appdata.h
```diff
@@ -6,11 +6,18 @@ namespace NMVP {
     class TMvpTokenator;
 }
 
+namespace NYdbGrpc {
+    inline namespace Dev {
+        class TGRpcClientLow;
+    }
+}
+
 struct TMVPAppData {
```
UgnineSirdis: Shouldn't it be also in the namespace `NMVP`?
TMVPAppData class and MVPAppData() function

ydb/mvp/oidc_proxy/mvp.h
```diff
@@ -18,8 +17,6 @@ const TString& GetEServiceName(NActors::NLog::EComponent component);
 
 class TMVP {
 private:
-    TRWMutex ActorSystemStoppingLock;
```
UgnineSirdis, line 21: There is a twin code here: ydb/mvp/meta/mvp.h
It is worth to fix both

ydb/mvp/oidc_proxy/extension_whoami.cpp
```diff
@@ -166,6 +166,14 @@ void TExtensionWhoamiWorker::ContinueAndPassAway() {
     PassAway();
 }
 
+void TExtensionWhoamiWorker::PassAway() {
```
UgnineSirdis, line 171: We need also reset RequestContext when getting reply for corresponding request. In TEvExtensionRequest/TEvErrorResponse handler. In order not to call Cancel if it is not needed.

UgnineSirdis (changes requested):
We need to write similar shutdown functionality also in `~TTestActorRuntimeBase`.

ydb/mvp/core/appdata.h
StekPerepolnen: added

ydb/mvp/oidc_proxy/extension_whoami.cpp
StekPerepolnen: changed

ydb/mvp/oidc_proxy/mvp.h
StekPerepolnen: changed

ydb/mvp/core/mvp_test_runtime.cpp
```diff
@@ -1,8 +1,11 @@
...
+
+#include "appdata.h"
+#include "mvp_log.h"
+
 #include <ydb/library/actors/core/mailbox.h>
 #include <ydb/library/actors/core/executor_thread.h>
-#include "mvp_log.h"
-#include "appdata.h"
-#include "mvp_test_runtime.h"
+#include <ydb/public/sdk/cpp/src/library/grpc/client/grpc_client_low.h>
```
UgnineSirdis: We don't use here `TGRpcClientLow`

UgnineSirdis (changes requested):

ydb/mvp/core/mvp_test_runtime.cpp
StekPerepolnen: removed

UgnineSirdis (approved):

PR #23913, opened 2025-09-01
Do not allow non-bootstrap command as first config-changing for cluster

alexvru:
### Changelog entry 

Do not allow non-bootstrap command as first config-changing for cluster

### Changelog category 

### Description for reviewers 

This patch makes it impossible to execute any config-changing command as first for the cluster.


mregrock (approved):
LGTM

PR #23966, opened 2025-09-02
Rollback transaction fix (#23628)

XJIE6:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


maximyurchuk (changes requested):
Не надо в 25-1-3

PR #24058, opened 2025-09-03
Make sure that index type is valid as soon as possible

kungasc:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Make index types checks as soon as possible to prevent unsupported index from being handled (for example in `CommonCheck` or table manipulations)

Also I propose to delete 

```
const NKikimrSchemeOp::EIndexType indexType = indexDesc.HasType() 
    ? indexDesc.GetType() 
    : NKikimrSchemeOp::EIndexTypeGlobal;
```

logic, because it's hard to keep in mind that we should always consider empty index type as `EIndexTypeGlobal` (for example, also in `CommonCheck`).

Unknown values in proto2 are ignored, so we could create a global index by mistake, instead of rejecting a request with unknown enum value.

Hope that making index type mandatory won't break anything (but we can easily fix it later if needed by explicitly filling it in requests). Added a few SQL/SDK tests to make sure they still work when using a short form statements.

Move invalid index handle to `default` case for better readability.

### DIFF
--- a/ydb/core/base/table_index.cpp
+++ b/ydb/core/base/table_index.cpp
@@ -85,6 +85,10 @@ TTableColumns CalcTableImplDescription(NKikimrSchemeOp::EIndexType type, const T
     return result;
 }
 
+TString InvalidIndexType(NKikimrSchemeOp::EIndexType type) {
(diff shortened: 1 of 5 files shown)


kungasc:
doesn't work :(

PR #24359, opened 2025-09-07
EXT-1486 Heartbeat setting for audit logging

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

We need to know if something goes wrong with audit events process - for example, if someone (occasionally or on their purpose) disabled audit logging. At the same time, it can be normal that for some cluster there are no records for a long time. With this setting we can setup alerts on audit and not to get false positive on them.

### DIFF
--- a/ydb/core/audit/audit_log.h
+++ b/ydb/core/audit/audit_log.h
@@ -22,7 +22,7 @@
 #define AUDIT_PART_NO_COND(key, value) AUDIT_PART_COND(key, value, true)
 #define AUDIT_PART_COND(key, value, condition)                                                                                    \
     do {                                                                                                                          \
-        if (condition && !value.empty()) {                                                                                        \
+        if (condition && !TStringBuf(value).empty()) {                                                                            \
             auditParts.emplace_back(key, value);                                                                                  \
         }                                                                                                                         \
     } while (0);
--- a/ydb/core/protos/config.proto
+++ b/ydb/core/protos/config.proto
@@ -1851,6 +1851,7 @@ message TAuditConfig {
             Operations = 7; // Operations API. See ydb/public/api/grpc/ydb_operation_v1.proto
             ExportImport = 8;
             Acl = 9;
+            AuditHeartbeat = 10;
             Default = 1000; // For default settings for all other nonconfigured log classes
         }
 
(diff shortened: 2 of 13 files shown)


ydb/core/audit/audit_log.h
```diff
@@ -22,7 +22,7 @@
 #define AUDIT_PART_NO_COND(key, value) AUDIT_PART_COND(key, value, true)
 #define AUDIT_PART_COND(key, value, condition)                                                                                    \
     do {                                                                                                                          \
-        if (condition && !value.empty()) {                                                                                        \
+        if (condition && !TStringBuf(value).empty()) {                                                                            \
```
flown4qqqq, line 25: Why is it nessasary?

ydb/core/protos/config.proto
```diff
@@ -1876,6 +1876,7 @@ message TAuditConfig {
     optional TFileBackend FileBackend = 2;
     optional TUnifiedAgentBackend UnifiedAgentBackend = 3;
     repeated TLogClassConfig LogClassConfig = 4;
+    optional uint32 HeartbeatIntervalSeconds = 5; // Make heartbeat records to audit log every HeartbeatInterval seconds. 0 means that heartbeat is disabled
```
flown4qqqq: Wouldn't it be better create new field in ELogClass (`HeartBeat` as example) and verify condition 'HeartbeatIntervalSeconds > 0'?

flown4qqqq (changes requested):

ydb/core/audit/audit_log.h
UgnineSirdis: It allows to specify in AUDIT_LOG macro literals.
  Now we can write:
```
AUDIT_PART("status", "SUCCESS")
```

While before this "feature" we had to write:
```
AUDIT_PART("status", TString("SUCCESS"))
```
And if we call the macro in this way, the code will be less effective, because it  creates TString twice: during the check and while inserting to the `auditParts`

ydb/core/protos/config.proto
UgnineSirdis: I think that it will not be very convenient and will not work as expected. If I were a person that sets up the cluster, I would expect that specifying `HeartbeatIntervalSeconds` setting should run the heartbeat feature.

flown4qqqq: On the one hand, you are right. But on the other hand, I don't like the loss of uniformity in the config. We have all types of logs listed in the ELogClass enum, except for one special one.

Of course, HeartBeat audit is special case, but other types can have special setting in future too

flown4qqqq: Discussed. It was decided that there will be new protobuf message with options for HeatBeat logs. Moreover, we agreed that adding new field in ELogClass is better.

flown4qqqq (changes requested):
I think it's worth doing the following:

1) add a check that IntervalSeconds > 0;
2) set the default IntervalSeconds value, for example, 30

flown4qqqq (approved):

PR #24508, opened 2025-09-09
Transaction error messages

Alek5andr-Kotov:
### Changelog entry 

Currently, the PQ tablet returns an error message only for initial checks. The text of the error message is lost during the transaction.

Fixed it. If the transaction completes with an error, the PQ tablet returns a description of the error. The topic, partition, SourceID, and SeqNo are specified for the write operation. The topic, partition, and consumer are specified for the read operation.

### Changelog category 

### Description for reviewers 

...


nikvas0 (approved):
kqp ok

nshestakov (approved):

PR #24552, opened 2025-09-10
Correct FreeUpToLsn interpretation

alexvru:
### Changelog entry 

Correct FreeUpToLsn interpretation

### Changelog category 

### Description for reviewers 

This patch fixes cases when FreeUpToLsn was misinterpreted as inclusive one.


mregrock (approved):
LGTM

PR #24833, opened 2025-09-12
added memory reporter for spilling

lll-phill-lll:
### Changelog entry 

The use of these functions has already been added in this commit: https://github.com/ydb-platform/ydb/commit/a6de3ae0a30f46f7f8fa776b9b512190a9a6dde8
...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
+++ b/ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
@@ -73,6 +73,21 @@ namespace NYql::NDq {
             });
         }
 
+        void ReportAllocated(ui64 bytes, NKikimr::NMiniKQL::TScopedAlloc* alloc) {
+            RequestExtraMemory(bytes, alloc);
+        }
+
+        void ReportFreed(ui64 bytes, NKikimr::NMiniKQL::TScopedAlloc* alloc) {
+            if (!MkqlMemoryLimit) {
+                return;
+            }
+            MemoryLimits.MemoryQuotaManager->FreeQuota(bytes);
+            UpdateSpillingZones(alloc);
+            if (MkqlMemoryQuota) {
+                MkqlMemoryQuota->Sub(bytes);
+            }
+        }
+
         void TryShrinkMemory(NKikimr::NMiniKQL::TScopedAlloc* alloc) {
             if (alloc->GetAllocated() - alloc->GetUsed() > MemoryLimits.MinMemFreeSize) {
                 alloc->ReleaseFreePages();
@@ -154,16 +169,17 @@ namespace NYql::NDq {
                 //                << ", requested: " << memory << ", host: " << HostName();
             }
 
-            if (MemoryLimits.MemoryQuotaManager->IsReasonableToUseSpilling()) {
-                alloc->SetMaximumLimitValueReached(true);
-            } else {
-                alloc->SetMaximumLimitValueReached(false);
-            }
+            UpdateSpillingZones(alloc);
 
             if (Y_UNLIKELY(ProfileStats)) {
                 ProfileStats->MkqlExtraMemoryBytes += memory;
                 ProfileStats->MkqlExtraMemoryRequests++;
             }
+        }
+
+        void UpdateSpillingZones(NKikimr::NMiniKQL::TScopedAlloc* alloc) {
+            bool isMaximumLimitValueReached = MemoryLimits.MemoryQuotaManager->IsReasonableToUseSpilling();
+            alloc->SetMaximumLimitValueReached(isMaximumLimitValueReached);
         }
 
         ui64 AlignMemorySizeToMbBoundary(ui64 memory) {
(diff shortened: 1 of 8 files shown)


Hor911 (approved):

ydb/library/yql/dq/actors/spilling/spiller_memory_reporter.cpp
```diff
@@ -0,0 +1,48 @@
...
+    Y_ENSURE(reportFreeCallback != nullptr);
+}
+
+void TSpillerMemoryUsageReporter::ReportAlloc(ui64 bytes) {
+    ReportAllocCallback_(bytes);
+    BytesAllocated_ += bytes;
+}
+
+void TSpillerMemoryUsageReporter::ReportFree(ui64 bytes) {
+    // TODO(mfilitov): generate error if bytes > BytesAllocates_
```
abyss7, line 21: Maybe implement right away?

ydb/library/yql/dq/actors/spilling/spiller_memory_reporter.cpp
```diff
@@ -0,0 +1,48 @@
...
+    ui64 toFree = std::min(BytesAllocated_, bytes);
+    if (toFree) {
+        ReportFreeCallback_(toFree);
+    }
+    BytesAllocated_ -= toFree;
+}
+
+TSpillerMemoryUsageReporter::~TSpillerMemoryUsageReporter() {
+    // TODO(mfilitov): generate error if BytesAllocated_ != 0
```
abyss7, line 30: If it's somehow legal, then it's better to describe behavior in the comments - why cancelled request with spilling doesn't report freed memory?

ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
```diff
@@ -73,6 +73,21 @@ namespace NYql::NDq {
             });
         }
 
+        void ReportAllocated(ui64 bytes, NKikimr::NMiniKQL::TScopedAlloc* alloc) {
```
abyss7, line 76: The method name suggests that we already have allocated some memory, but the `RequestExtraMemory` suggests that we're just making a request. Does it mean we can potentially over-allocate memory before getting approvement from the resource manager?

ydb/library/yql/dq/actors/spilling/compute_storage.cpp
```diff
@@ -8,7 +8,10 @@ namespace NYql::NDq {
 using namespace NActors;
 
 TDqComputeStorage::TDqComputeStorage(TTxId txId, TWakeUpCallback wakeUpCallback, TErrorCallback errorCallback,
-    TIntrusivePtr<TSpillingTaskCounters> spillingTaskCounters, TActorSystem* actorSystem) : ActorSystem_(actorSystem) {
+    TIntrusivePtr<TSpillingTaskCounters> spillingTaskCounters, TSpillerMemoryUsageReporter::TPtr memoryUsageReporter, TActorSystem* actorSystem)
+        : ActorSystem_(actorSystem)
+        , MemoryUsageReporter_(memoryUsageReporter)
```
abyss7, line 13: Check not nullptr?

abyss7 (changes requested):

ydb/library/yql/dq/actors/spilling/spiller_memory_reporter.cpp
lll-phill-lll: Currently there is no way to gracefully stop the request. I can only add verify here (which is an overkill). I think I can add a debug verify here

ydb/library/yql/dq/actors/compute/dq_compute_memory_quota.h
lll-phill-lll: Yes, you're correct. At the moment, all allocations that spilling does when packing values go through tcmalloc. These allocations are not tracked in memory at all, which can lead to (and already does lead to) OOMs.

The logic now works like this:
when we pack values, we directly tell RM that we’ve taken some memory. Once we get confirmation that the blob has been written to disk, we tell RM that the memory was released.

Because of this, there can be cases where we report more memory than we actually have. In that case, an exception will be thrown (instead of today’s OOM).

Still, since we now report this memory, spilling can control its own usage. When RM signals the “yellow zone” after the limit is reached, spilling will flush buffers more often and free memory sooner. This helps lower overall RSS and prevents OOMs.

ydb/library/yql/dq/actors/spilling/spiller_memory_reporter.cpp
lll-phill-lll: As I mentioned in the other comment, spilling reports memory as Allocated when it packs values, and reports it as Freed once it gets confirmation that the blob was written to disk.

In this case, if the query is cancelled for any reason between these two events, the memory release won’t be reported. Even though the memory itself will actually be freed.

PR #24939, opened 2025-09-15
Stop gRPC server when in DISCONNECTED state

alexvru:
### Changelog entry 

Stop gRPC server when in DISCONNECTED state

### Changelog category 

* Experimental feature

### Description for reviewers 

This patch makes gRPC server shut down on every node in DISCONNECTED pile. It is started again when pile gets reconnected.


mregrock (approved):
LGTM

PR #25070, opened 2025-09-16
add delete perf test for olap

swalrus1:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/tests/olap/test_cs_many_updates.py
+++ b/ydb/tests/olap/test_cs_many_updates.py
@@ -264,6 +264,7 @@ class TestCSManyUpdates(object):
             result_sets = self.ydb_client.query(
                 f"SELECT id, value FROM `{table_path}`;"
             )
+        print(timer.report())
 
         rows = [row for result_set in result_sets for row in result_set.rows]
         assert len(rows) == rows_num
@@ -273,17 +274,17 @@ class TestCSManyUpdates(object):
         timer = Timer("delete from table")
         with timer:
             self._delete_all(table_path, timer)
+        print(timer.report())
 
         timer = Timer("select all after delete")
         with timer:
             result_sets = self.ydb_client.query(
                 f"SELECT id, value FROM `{table_path}`;"
             )
+        print(timer.report())
 
         rows = [row for result_set in result_sets for row in result_set.rows]
         assert not rows
-
-        print(timer.report())
 
     @pytest.mark.parametrize(
         "rows_num,operation_sequence",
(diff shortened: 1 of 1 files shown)


ydb/tests/olap/test_cs_many_updates.py
```diff
@@ -132,6 +132,14 @@ def _do_update(self, table_path, pks, mods, timer):
                 self.ydb_client.query(update_query)
                 timer.one_more()
 
+    def _delete_all(self, table_path, timer):
+        """Delete all rows using DELETE statement."""
+        update_query = f"""
+        DELETE FROM `{table_path}`
+        """
+        self.ydb_client.query(update_query)
+        timer.one_more()
```
kirillvasilenko: It is a single operation, so we should not do timer.one_more() here. Otherwise, it counts 2 opertaions.

ydb/tests/olap/test_cs_many_updates.py
```diff
@@ -257,13 +265,27 @@ def _test_many_updates_impl(self, rows_num, operation_sequence):
             result_sets = self.ydb_client.query(
                 f"SELECT id, value FROM `{table_path}`;"
             )
-        print(timer.report())
```
kirillvasilenko: Why don't we print here the report?

ydb/tests/olap/test_cs_many_updates.py
```diff
@@ -257,13 +265,27 @@ def _test_many_updates_impl(self, rows_num, operation_sequence):
...
 
         rows = [row for result_set in result_sets for row in result_set.rows]
         assert len(rows) == rows_num
         actual_data = {row["id"]: row["value"] for row in rows}
         assert actual_data == expected_data
 
+        timer = Timer("delete from table")
+        with timer:
+            self._delete_all(table_path, timer)
+
```
kirillvasilenko, line 278: let's print the report here.

ydb/tests/olap/test_cs_many_updates.py
```diff
@@ -257,13 +265,27 @@ def _test_many_updates_impl(self, rows_num, operation_sequence):
...
+        timer = Timer("select all after delete")
+        with timer:
+            result_sets = self.ydb_client.query(
+                f"SELECT id, value FROM `{table_path}`;"
+            )
+
+        rows = [row for result_set in result_sets for row in result_set.rows]
+        assert not rows
+
+        print(timer.report())
```
kirillvasilenko: It's better to move the print() right under the 'with timer:'. In this case we will see the report even if the result is incorrect.

ydb/tests/olap/test_cs_many_updates.py
swalrus1: fixed

ydb/tests/olap/test_cs_many_updates.py
swalrus1: as far as I understand, report must be before `assert` in order to be printed if the result is incorrect, not necessarily under `with`?

kirillvasilenko (approved):

PR #25084, opened 2025-09-16
Optimize TEvAssimilate

alexvru:
### Changelog entry 

Optimize TEvAssimilate

### Changelog category 

* Performance improvement

### Description for reviewers 

This patch makes TEvVAssimilate processor checking whether blob is going to be collected before returning it.


mregrock (approved):
LGTM

PR #25184, opened 2025-09-17
Update CI to main version

SammyVimes:


.github/CODEOWNERS
```diff
@@ -1,2 +1,92 @@
-* @mvgorbunov @dcherednik @va-kuznecov @alexv-smirnov
```
UgnineSirdis, line 1: This file I think should not be changed

UgnineSirdis (changes requested):

PR #25306, opened 2025-09-18
Automatic checksumming decision for DSProxy/VDisk interaction

alexvru:
### Changelog entry 

Automatic checksumming decision for DSProxy/VDisk interaction

### Changelog category 

### Description for reviewers 

This patch makes DS proxy calculate blob checksums only when VDisks report this feature turned on.


mregrock (approved):
LGTM

PR #25387, opened 2025-09-19
EXT-1523 Write audit records always except for denylist

StekPerepolnen:
### Changelog entry 

Change the audit logging logic from AllowedList checking to DenyList checking

### Changelog category 

* Improvement

### Description for reviewers 

`TUrlMatcher` now requires explicit use of the `Recursive` parameter instead of `*` suffix.
When `Recursive` is set, the rule applies to all subpaths.

An audit record must be created if either of the following is true:
- the HTTP method is POST, PUT, or DELETE (modifying method)
- the endpoint is not in the denylist (counters or viewer endpoints)


ydb/core/mon/audit/audit_ut.cpp
```diff
@@ -5,8 +5,59 @@
...
+        return Storage.back();
+    }
+};
+
+NHttp::THttpIncomingRequestPtr MakeRequest(TString method, TString url) {
+    NHttp::THttpIncomingRequestPtr Request = new NHttp::THttpIncomingRequest();
+    Request->Method = std::move(method);
+    Request->URL = std::move(url);
+    return Request;
+}
```
UgnineSirdis:
```suggestion
struct TRequestHolder : public THttpIncomingRequest {
    TStringBuf Store(TString value) {
        return Storage.emplace_back(std::move(value));
    }

private:
    TVector<TString> Storage;
};

NHttp::THttpIncomingRequestPtr MakeRequest(TString method, TString url) {
    auto request = MakeIntrusive<TRequestHolder>();
    request->Method = request->Store(std::move(method));
    request->URL = request->Store(std::move(url));
    return std::move(request);
}
```

ydb/core/mon/audit/url_matcher_ut.cpp
```diff
@@ -24,48 +24,18 @@ Y_UNIT_TEST_SUITE(TUrlMatcherTest) {
...
+        UNIT_ASSERT(matcher.Match("/actors/blobstorageproxies"));
+        UNIT_ASSERT(matcher.Match("/actors/blobstorageproxies/blobstorageproxy2181038080"));
         UNIT_ASSERT(matcher.Match("/actors/blobstorageproxies/blobstorageproxy2181038080?PutSamplingRate=1"));
         UNIT_ASSERT(matcher.Match("/actors/blobstorageproxies/somethingelse?PutSamplingRate=123"));
 
-        UNIT_ASSERT(!matcher.Match("/actors/blobstorageproxies"));
-        UNIT_ASSERT(!matcher.Match("/actors/blobstorageproxies/blobstorageproxy2181038080"));
-        UNIT_ASSERT(!matcher.Match("/actors/blobstorageproxies/blobstorageproxy2181038080?OtherParam=1"));
         UNIT_ASSERT(!matcher.Match("/actors/otherproxy/blobstorageproxy2181038080?PutSamplingRate=1"));
     }
```
UgnineSirdis, line 41: `UNIT_ASSERT(!matcher.Match("/actors/blobstorageproxies123"));`

UgnineSirdis (changes requested):

UgnineSirdis (approved):

PR #25459, opened 2025-09-20
Actually terminate YDB CLI on double Ctrl+C

abyss7:
### Changelog entry 

Actually terminate YDB CLI on double Ctrl+C

### Changelog category 

* New feature

### DIFF
--- a/ydb/public/lib/ydb_cli/common/interruptible.cpp
+++ b/ydb/public/lib/ydb_cli/common/interruptible.cpp
@@ -1,11 +1,13 @@
 #include "interruptible.h"
 
-namespace NYdb {
-namespace NConsoleClient {
+namespace NYdb::NConsoleClient {
 
 TAtomic TInterruptibleCommand::Interrupted = false;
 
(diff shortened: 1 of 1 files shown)


ydb/public/lib/ydb_cli/common/interruptible.cpp
```diff
@@ -1,11 +1,13 @@
...
 
-namespace NYdb {
-namespace NConsoleClient {
+namespace NYdb::NConsoleClient {
 
 TAtomic TInterruptibleCommand::Interrupted = false;
 
 void TInterruptibleCommand::OnTerminate(int) {
+    if (Interrupted) {
+        std::exit(0);
```
pnv1, line 9: We avoid using `exit()` if possible.
In what command CLI does not actually terminate on ctrl+c?

abyss7: `ydb -e grpc://localhost:1924 -d /local sql -f ~/tpch-q21.sql`

PR #25700, opened 2025-09-24
Fix gRPC server runner

alexvru:
### Changelog entry 

Fix gRPC server runner

### Changelog category 

### Description for reviewers 

This fixes the case when gRPC server doesn't need to be started.


mregrock (approved):
LGTM

PR #25759, opened 2025-09-24
Add ICB setting for ChunkBaseLimit in PDisk

serbel324:
### Changelog entry 

ChunkBaseLimit parameter is configured via PDiskConfig, which can be inconvenient to manage. This change adds ICB control to manipulate this parameter. When set, ICB option has priority over config.

### Changelog category 

* Improvement


the-ancient-1 (approved):
LGTM

PR #25867, opened 2025-09-26
Propagate StorageConfig up the distconf tree

alexvru:
### Changelog entry 

Propagate StorageConfig up the distconf tree

### Changelog category 

### Description for reviewers 

This patch allows configuration to be propagated from the bottom node to the top.


mregrock (approved):
LGTM

PR #25991, opened 2025-09-29
Deduplicate sequence copying code

### LINKED ISSUE #22913: Prefixed vector index update
As global vector index update is now implemented (#21264), now we need to implement prefixed vector index update.

The main problem is that inserted rows may contain unknown prefixes. So we should add them into the indexImplPrefixTable, probably with 1 cluster first. Then, at some point, clusters for the new prefix should probably be recalculated.

It's also slightly similar to inserting rows into an empty global vector index ~~which is also not supported yet~~. UPD: It's already supported, but in another way: empty global vector index now contains just 1 cluster from the beginning.

vitalif:
### Changelog category

### Description for reviewers

A bit of refactoring extracted from #25505

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__operation_consistent_copy_tables.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__operation_consistent_copy_tables.cpp
@@ -150,24 +150,7 @@ bool CreateConsistentCopyTables(
         TPath dstPath = TPath::Resolve(dstStr, context.SS);
         TPath dstParentPath = dstPath.Parent();
 
-        THashSet<TString> sequences;
-        for (const auto& child: srcPath.Base()->GetChildren()) {
-            auto name = child.first;
-            auto pathId = child.second;
-
-            TPath childPath = srcPath.Child(name);
-            if (!childPath.IsSequence() || childPath.IsDeleted()) {
-                continue;
-            }
-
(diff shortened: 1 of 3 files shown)


ydb/core/tx/schemeshard/schemeshard__operation_consistent_copy_tables.cpp
```diff
@@ -230,21 +209,50 @@ bool CreateConsistentCopyTables(
...
+        const auto& sequenceName = sequenceDesc.GetName();
+
+        sequences.emplace(sequenceName);
+    }
+    return sequences;
+}
+
+
+void AddCopySequences(TOperationId nextId, const TTxTransaction& tx, TOperationContext& context,
+    TVector<ISubOperation::TPtr>& result, const TPath& srcTable, const TString& dstPath) {
```
CyberROFL:
```suggestion
    TVector<ISubOperation::TPtr>& result, const TPath& srcTable, const TString& dstPath)
{
```
Иначе сливается.

CyberROFL (dismissed):

ydb/core/tx/schemeshard/schemeshard__operation_consistent_copy_tables.cpp
vitalif: Ок, поменял)

CyberROFL (approved):

PR #26088, opened 2025-09-30
better format

nfrmtk:
...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/kqp/tools/.clang-format
+++ b/ydb/core/kqp/tools/.clang-format
@@ -8,3 +8,5 @@ SeparateDefinitionBlocks: Always
 ColumnLimit: 120
 AllowShortLambdasOnASingleLine: Inline
 AllowShortFunctionsOnASingleLine: Empty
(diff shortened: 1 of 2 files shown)


ydb/core/kqp/tools/fmt.bashfn
nfrmtk, line 1: ROOT=$(ya dump root) line was moved in function because it was annoying to see error when calling source om fmt.bashfn from home directory

SCRIPT_DIR was moved to unify

ydb/core/kqp/tools/.clang-format
nfrmtk, line 1: {Pack, Break}ConstructorInitializers are to unify with current style, for example see https://github.com/ydb-platform/ydb/blob/074268bc87be8c4b05691f12389f2400c0ea1584/ydb/library/yql/dq/comp_nodes/dq_scalar_hash_join.cpp#L22-L30

pzuev (approved):

PR #26125, opened 2025-09-30
lower initial delay before sending base statistics from schemeshard to statistics aggregator

ztlpn:
### Changelog category 

### Description for reviewers 

* lower initial delay before sending base statistics from schemeshard to statistics aggregator.
* improve logging
* add test with incomplete table base statistics after schemeshard reboot


azevaykin (approved):

ijon (approved):
schemeshard ok

PR #26178, opened 2025-10-01
Revert "Compatible Scalar and Block HashV2 Shuffles"

iddqdex:
Reverts ydb-platform/ydb#25357


iddqdex:
https://github.com/ydb-platform/ydb/issues/26182

Issue #26012, opened 2025-09-29
Create scheduler tests


maVovk:
Implement unit-tests for KQP scheduler

Necessary scenarios:

1. Fair share calculation

Basic:
- [x] One database, one pool, one query. Demand is under a limit
- [x] Same tree structure but now demand is over a limit and FairShare should be set depending on flag
- [x] Check correct weighted distribution between childs
- [x] More complex structure still passes previous checks 
Advanced:
- [x] Structure and limits are changed between UpdateFairShare calls. Queries are added and deleted

Corner-cases:
- [x] Empty root, pool, database etc.
- [x] Zero demand, weight, limit

2. Nodes insertion/deletion

PR #26210, opened 2025-10-02
Scheduler tests

maVovk:
Implemented unit-tests for KQP scheduler. Related issue - #26012 

Necessary scenarios:

1. Fair share calculation

Basic:

- [x] One database, one pool, one query. Demand is under a limit
- [x] Same tree structure but now demand is over a limit and FairShare should be set depending on flag
- [x] Check correct weighted distribution between childs
- [x] More complex structure still passes previous checks

Advanced:
- [x] Structure and limits are changed between UpdateFairShare calls. Queries are added and deleted

Corner-cases:

- [x] Empty root, pool, database etc.
- [x] Zero demand, weight, limit
- [x] Nodes insertion/deletion

### DIFF
--- a/ydb/core/kqp/runtime/scheduler/kqp_compute_scheduler_service_ut.cpp
+++ b/ydb/core/kqp/runtime/scheduler/kqp_compute_scheduler_service_ut.cpp
@@ -7,54 +7,255 @@
 
 namespace NKikimr::NKqp::NScheduler {
 
-    Y_UNIT_TEST_SUITE(TKqpScheduler) {
-        Y_UNIT_TEST(SingleDatabasePoolQueryStructure) {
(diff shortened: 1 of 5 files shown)


ydb/core/kqp/runtime/scheduler/kqp_compute_scheduler_service_ut.cpp
```diff
@@ -0,0 +1,419 @@
+#include "kqp_compute_scheduler_service.h"
+
+#include "fwd.h"
+#include "tree/dynamic.h"
+
+#include <library/cpp/testing/unittest/registar.h>
+
+namespace NKikimr::NKqp::NScheduler {
+
+Y_UNIT_TEST_SUITE(TKqpScheduler) {
```
abyss7, line 10: Let's name it `TKqpCpuScheduler` for clarity - there are other schedulers in Kqp

PR #26325, opened 2025-10-03
tx lock ranges soft limit

swalrus1:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


ydb/core/tx/columnshard/transactions/locks/read_start.cpp
```diff
@@ -33,9 +35,16 @@ void TEvReadStart::DoSerializeToProto(NKikimrColumnShardTxProto::TEvent& proto)
 }
 
 void TEvReadStart::DoAddToInteraction(const ui64 txId, TInteractionsContext& context) const {
+    const ui64 softLimit = AppDataVerified().ColumnShardConfig.GetTxLockRangesSoftLimit();
     for (auto&& i : *Filter) {
+        if (context.GetIntervalCount(txId) >= softLimit) {
+            context.AddInterval(txId, PathId.InternalPathId, TIntervalPoint::From(i.GetPredicateFrom(), Schema),
```
kirillvasilenko, line 41: I am not sure that it fixes the problem. 

It adds a single interval for a read if the limit is reached.

Are we sure that in the problematic scenario number_of_reads << number_of_intervals? If number_of_reads almost = number_of_intervals, the change does not help us.

PR #26353, opened 2025-10-03
Reuse common space checks for S3 upload rows

kungasc:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Overall, it seems that having a separate S3 upload rows is a mistake because it doesn't handle overload subscribed (and before this PR hasn't handled replicated and out of space), but it requires a huge refactoring I'm not going to do right now.


snaury (approved):

CyberROFL:
lgtm

PR #26573, opened 2025-10-08
24-4-enterprise: Make it possible to backup out-of-range decimal values

snaury:
### Changelog entry 

Fix backup of out-of-range decimal values which may have been accidentally inserted into tables. Fixes #26470.

### Changelog category 

* Bugfix 

### Description for reviewers 

We now use MaxPrecision when transforming binary decimal values into strings, so it does not result in unexpected `nullptr` as long as the binary value is correct (less than ±10^35 or one of special values). Also make sure to handle `nullptr` result as an error.


zinal (approved):
Lgtm

PR #26596, opened 2025-10-09
ci: change bt for san to relwithdebinfo

PavelEfarinov:
### Changelog category 

### Description for reviewers 

Changed build type for sanitizer build to `relwithdebinfo`

Refs: #25375


.github/actions/test_ya/action.yml
```diff
@@ -194,7 +194,7 @@ runs:
             ;;
           release-asan)
```
maximyurchuk: Не release-asan уже

PR #26713, opened 2025-10-11
Incremental restore smoke test

LebedevOleg1:
### Changelog entry 
Add full cycle local backup restore with incrementals
...

### Changelog category 

### Description for reviewers 
Add full cycle local backup restore with incremen
tals
...


Enjection (approved):
Generally good, though a bit wordy and lacking extraction of common framework patterns. The main concern is the numerous sleeps without documented rationale. Are they needed to prevent duplicate backup names, or because we lack proper operation polling? This makes our setup flaky and dependent on executor speed. Let's address these issues in a separate PR.

PR #26817, opened 2025-10-14
Add polling to tests

LebedevOleg1:
### Changelog entry 
change time.sleep(1) for polling-function
...

### Changelog category 

### Description for reviewers 
add polling
...


Enjection (approved):
Great! Not exactly what I meant by polling, but this helper is great as well.

PR #26869, opened 2025-10-15
Fix BSC vs NW interaction protocol

alexvru:
### Changelog entry 

Fix BSC vs NW interaction protocol

### Changelog category 

### Description for reviewers 

This fixes cases when BSC may send service set update before issuing actual response to RegisterNode leading to an incorrect behaviour of some subsystems in NW.


mregrock (approved):
LGTM

PR #27022, opened 2025-10-16
Add new ICB params for SectorMap

mregrock:
### Changelog entry 

Add new ICB params for SectorMap

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
+++ b/ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
@@ -86,10 +86,10 @@ TPDisk::TPDisk(std::shared_ptr<TPDiskCtx> pCtx, const TIntrusivePtr<TPDiskConfig
         }
         auto failureProbs = Cfg->SectorMap->GetFailureProbabilities();
         if (failureProbs) {
-            SectorMapWriteErrorProbability = TControlWrapper(0, 0, 1000000);
-            SectorMapReadErrorProbability = TControlWrapper(0, 0, 1000000);
-            SectorMapSilentWriteFailProbability = TControlWrapper(0, 0, 1000000);
-            SectorMapReadReplayProbability = TControlWrapper(0, 0, 1000000);
+            SectorMapWriteErrorProbability = TControlWrapper(0, 0, 1000000000);
+            SectorMapReadErrorProbability = TControlWrapper(0, 0, 1000000000);
+            SectorMapSilentWriteFailProbability = TControlWrapper(0, 0, 1000000000);
+            SectorMapReadReplayProbability = TControlWrapper(0, 0, 1000000000);
         }
     }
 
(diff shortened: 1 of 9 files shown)


ydb/library/pdisk_io/sector_map.h
```diff
@@ -249,15 +272,35 @@ class TSectorMap : public TThrRefBase {
...
+                    char tmp[4 * NSectorMap::SECTOR_SIZE];
+                    int processed = LZ4_decompress_safe(it->second.data(), tmp, it->second.size(), 4 * NSectorMap::SECTOR_SIZE);
+                    Y_VERIFY_S(processed == NSectorMap::SECTOR_SIZE, "processed# " << processed);
+                    memcpy(data + pos, tmp, NSectorMap::SECTOR_SIZE);
+                }
+                if (SectorOperationThrottler.Get() != nullptr) {
+                    SectorOperationThrottler->ThrottleRead(dataSize, dataOffset, prevOperationIsInProgress, timer.Passed() * 1000);
+                }
+                return true;
+            }
```
serbel324: You didn't call ReadCallback() after completion
I think faulty operation should behave the same way the correct does, except for reading different data
I also suggest you get rid of copy-paste

ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
```diff
@@ -84,6 +84,13 @@ TPDisk::TPDisk(std::shared_ptr<TPDiskCtx> pCtx, const TIntrusivePtr<TPDiskConfig
             SectorMapLastSectorWriteRate = TControlWrapper(diskModeParams->LastSectorWriteRate, 0, 100000ull * 1024 * 1024);
             SectorMapSeekSleepMicroSeconds = TControlWrapper(diskModeParams->SeekSleepMicroSeconds, 0, 100ul * 1000 * 1000);
         }
+        auto failureProbs = Cfg->SectorMap->GetFailureProbabilities();
+        if (failureProbs) {
+            SectorMapWriteErrorProbability = TControlWrapper(0, 0, 1000000);
+            SectorMapReadErrorProbability = TControlWrapper(0, 0, 1000000);
+            SectorMapSilentWriteFailProbability = TControlWrapper(0, 0, 1000000);
+            SectorMapReadReplayProbability = TControlWrapper(0, 0, 1000000);
```
serbel324: Maybe it's better to use [0, 1'000'000'000] range? NVMe device theoretically can perform 1mil operations per second, and we may want to emulate less frequent errors than 1 every second

ydb/library/pdisk_io/sector_map.h
```diff
@@ -27,6 +32,13 @@ namespace NPDisk {
 
 namespace NSectorMap {
 
+struct TFailureProbabilities {
+    std::atomic<double> WriteErrorProbability;
+    std::atomic<double> ReadErrorProbability;
+    std::atomic<double> SilentWriteFailProbability;
+    std::atomic<double> ReadReplayProbability;
```
serbel324, line 40: I suggest you add std::atomic<ui64> counters and increment them each time SectorMap emulates an error and then report it to metrics in PDisk main cycle

serbel324 (approved):

PR #27104, opened 2025-10-18
Fix the notify about commit logic in Column Shards

### LINKED ISSUE #26463: Fix the notify about commit logic in Column Shards
The "notify about commit" logic does the following:
given:
- tx1 writes 1=1, not commit yet, the new portion is visible as uncommitted for reads
- tx2 wants to read everything, sees the uncommitted portion by tx1, remembers that there might be conflict. It will know for sure whether there is a conflict or not only when it actually reads the portion
- tx2 goes to read everything (it happens in separate actors to the column shard tablet), it reads very very long time
- tx2 during the actual read, understands that there is a conflict for sure between tx1 and tx2, it will be able to "register" this conflict only when the actual read is finished and the column shard processes the result
- meanwhile, the tx1 commits successfully and disappears completely from the shard
- tx2 finishes the read, gets back to the column shard and wants to register the conflict. But there is no tx2 anymore. So tx2 has to understand somehow whether tx1 committed successfully and tx2 must break itself, or tx1 aborted and tx2 may proceed.

The "notify about commit" logic lets tx2 understand whether tx1 has committed or not.

Currently, this logic does not work completely due to various bugs.

kirillvasilenko:
Fixes #26463

There is some renaming. It does not fixes the logic but makes it much easier to understand and harder to make a mistake.

The actual fix is in:
- TReadMetadata, I added clear separation between "maybe conflicting writes" (writes that we may be conflicting but we have not checked for sure yet) and "conflicting writes" (writes that we have checked and found out they are conflicting). Previously, there was no way to get "maybe conflicting writes" but they are crucial for the notify about commit logic.
-  TOperationsManager::AddEventForLock(), there were a tricky bug. We check if there is a lock for conflicting tx and "tell" it to break the current tx, when it commits. If the lock was present, we did not check if it is committed already. And that was a mistake, because the lock may be present and its tx may be already committed. Now we check whether the conflicting tx is committed or not always.
-  TReadMetadata::DoOnBeforeStartReading(). Here we must pass "maybe conflicting writes" to evWriter and "ask" all the maybe conflicting txs to notify the given tx when they commit. We do not now which are really conflicting so we have to "ask" all of them.

### DIFF
--- a/ydb/core/tx/columnshard/engines/reader/common_reader/constructor/read_metadata.h
+++ b/ydb/core/tx/columnshard/engines/reader/common_reader/constructor/read_metadata.h
@@ -81,7 +81,7 @@ class TReadMetadata: public TReadMetadataBase {
     using TBase = TReadMetadataBase;
 
 private:
-    std::shared_ptr<TAtomicCounter> BrokenWithCommitted = std::make_shared<TAtomicCounter>();
+    mutable TAtomicCounter BreakLockOnReadFinished = TAtomicCounter();
     std::shared_ptr<NColumnShard::TLockSharingInfo> LockSharingInfo;
 
     class TWriteIdInfo {
@@ -99,17 +99,17 @@ private:
             return LockId;
         }
 
-        void MarkAsConflictable() const {
+        void MarkAsConflicting() const {
             Conflicts->Inc();
         }
 
-        bool IsConflictable() const {
+        bool IsConflicting() const {
             return Conflicts->Val();
         }
     };
 
     THashMap<ui64, std::shared_ptr<TAtomicCounter>> LockConflictCounters;
-    THashMap<TInsertWriteId, TWriteIdInfo> ConflictedWriteIds;
+    THashMap<TInsertWriteId, TWriteIdInfo> ConflictingWrites;
 
     virtual void DoOnReadFinished(NColumnShard::TColumnShard& owner) const override;
     virtual void DoOnBeforeStartReading(NColumnShard::TColumnShard& owner) const override;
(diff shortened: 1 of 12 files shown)


ydb/core/tx/columnshard/engines/reader/common_reader/constructor/read_metadata.h
```diff
@@ -81,7 +81,7 @@ class TReadMetadata: public TReadMetadataBase {
     using TBase = TReadMetadataBase;
 
 private:
-    std::shared_ptr<TAtomicCounter> BrokenWithCommitted = std::make_shared<TAtomicCounter>();
+    mutable TAtomicCounter BreakLockOnReadFinished = TAtomicCounter();
```
dorooleg, line 84: mutable? Из-за избавления от shared_ptr?

ydb/core/tx/columnshard/engines/reader/common_reader/constructor/read_metadata.h
```diff
@@ -132,51 +132,48 @@ class TReadMetadata: public TReadMetadataBase {
...
     }
 
-    bool GetBrokenWithCommitted() const {
-        return BrokenWithCommitted->Val();
+    bool GetBreakLockOnReadFinished() const {
+        return BreakLockOnReadFinished.Val();
     }
-    THashSet<ui64> GetConflictableLockIds() const {
+
+    void SetBreakLockOnReadFinished() const {
```
dorooleg, line 139: А почему не Add или Inc?

ydb/core/tx/columnshard/engines/reader/common_reader/constructor/read_metadata.h
kirillvasilenko: Так точно. Там метод который это поле меняет const, и на этот const там полагаются вызывающие методы и т.д. В общем, я не стал все это править, просто локально упростил. shared_ptr убрал т.к. это внутренняя переменная, никуда мы ее не шарим, а сам класс не копируемый.

ydb/core/tx/columnshard/engines/reader/common_reader/constructor/read_metadata.h
kirillvasilenko: Ну оно по смыслу не Add. Мы именно просим «сломай свой лок когда закончится чтение».

kirillvasilenko: То что оно там add внутри - это детали реализации, не вижу смысла «показывать» их наружу.

dorooleg: Странно выглядит, что на самом деле это не Set

dorooleg (approved):

PR #27202, opened 2025-10-20
Full restore test with complex schema change

LebedevOleg1:
### Changelog entry 
Full cycle local backup restore with complex schema change
...

### Changelog category 

### Description for reviewers 
Full cycle local backup restore with complex schema change
...


ydb/tests/functional/backup_collection/basic_user_scenarios.py
```diff
@@ -577,6 +577,95 @@ def poll_restore_by_count(self, start_total: int, start_success: int, timeout_s:
...
+        finally:
+            # remove temporary directory if we created it here
+            if tmp_created:
+                rmdir_res = run_cli(["scheme", "rmdir", "-rf", tmp_dir_rel])
+                if rmdir_res.exit_code != 0:
+                    logger.warning("Failed to rmdir tmp dir %s: code=%s stdout=%s stderr=%s",
+                                   tmp_dir_rel,
+                                   rmdir_res.exit_code,
+                                   getattr(rmdir_res, "std_out", b"").decode("utf-8", "ignore"),
+                                   getattr(rmdir_res, "std_err", b"").decode("utf-8", "ignore"))
```
Enjection, line 668: Extract to helper

ydb/tests/functional/backup_collection/basic_user_scenarios.py
```diff
@@ -577,6 +577,95 @@ def poll_restore_by_count(self, start_total: int, start_success: int, timeout_s:
             "last_success": last_success,
         }
 
+    def _rearrange_table(self, from_name: str, to_name: str):
+        full_from = f"/Root/{from_name}"
```
Enjection, line 582: a bit unclear name, so leave some docstring what exactly happens with table. Is it just copied or what

Enjection: or the tables are swapped, then name it swap

ydb/tests/functional/backup_collection/basic_user_scenarios.py
```diff
@@ -1683,3 +1772,202 @@ def test_incremental_chain_restore_when_tables_deleted(self):
...
+            self._rearrange_table("orders", "orders_rearr")
+
+            create_table_with_data(session, "other_place_topic")
+
+        snapshot_stage2_t1 = self._capture_snapshot("orders_rearr")
+        # products may be dropped; try capture but tolerate errors
+        try:
+            snapshot_stage2_t2 = self._capture_snapshot(t2)
+        except Exception:
+            snapshot_stage2_t2 = None
```
Enjection: maybe some helper if this "safe capture" is reusable pattern

Enjection: Actually this safeguards probably not the best way to do testing. It is good when test is deterministic and we know exactly what will happen next instad of guessing.

ydb/tests/functional/backup_collection/basic_user_scenarios.py
```diff
@@ -1683,3 +1772,202 @@ def test_incremental_chain_restore_when_tables_deleted(self):
...
+        schema_stage2_t1 = self._capture_schema("/Root/orders_rearr")
```
Enjection: Repetition of `/Root/orders_rearr` and other names is not right, store them in some variables

ydb/tests/functional/backup_collection/basic_user_scenarios.py
```diff
@@ -577,6 +577,95 @@ def poll_restore_by_count(self, start_total: int, start_success: int, timeout_s:
...
+                tmp_created = False
+            except Exception:
+                res_mk = run_cli(["scheme", "mkdir", tmp_dir_rel])
+                if res_mk.exit_code != 0:
+                    logger.debug("tmp dir mkdir returned code=%s stdout=%s stderr=%s",
+                                 res_mk.exit_code,
+                                 getattr(res_mk, "std_out", b"").decode("utf-8", "ignore"),
+                                 getattr(res_mk, "std_err", b"").decode("utf-8", "ignore"))
+                else:
+                    tmp_created = True
```
Enjection, line 639: rename `tmp_created` to `need_tmp_cleanup` or something like this for readability. But I would prefer to clean it up unconditionally

ydb/tests/functional/backup_collection/basic_user_scenarios.py
```diff
@@ -577,6 +577,95 @@ def poll_restore_by_count(self, start_total: int, start_success: int, timeout_s:
...
+
+        # perform tools copy via CLI to the requested destination
+        item_arg = f"destination={dst_rel},source={src_rel}"
+        res = run_cli(["tools", "copy", "--item", item_arg])
+        if res.exit_code != 0:
+            out = (res.std_out or b"").decode("utf-8", "ignore")
+            err = (res.std_err or b"").decode("utf-8", "ignore")
+            raise AssertionError(f"tools copy failed: from={full_from} to={full_to} code={res.exit_code} STDOUT: {out} STDERR: {err}")
+
+        tmp_dir_rel = f"tmp_rearr_{int(time.time())}"
```
Enjection, line 623: time.time() is antipattern here. Use some uuid or something like this, that guarantees to never repeat in consequent calls.

Enjection (changes requested):
I think we need to spend more time on design of our checks. Now they are very superficial and could be unreliable.

ydb/tests/functional/backup_collection/basic_user_scenarios.py
LebedevOleg1: > a bit unclear name, so leave some docstring what exactly happens with table. Is it just copied or what

I used copying instead of `ALTER TABLE ... RENAME TO ...` so as not to break the logic of the test and it did not crash because the backup collection does not see where the table has moved

Enjection (approved):
Good for now, but gathering all the pieces in framework becomes absolutely essential.

PR #27972, opened 2025-10-30
added scalar layout converter

lll-phill-lll:
### Changelog entry 
Added converter for scalar inputs/outputs of dq_hash_join
...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/library/yql/dq/comp_nodes/hash_join_utils/layout_converter_common.h
+++ b/ydb/library/yql/dq/comp_nodes/hash_join_utils/layout_converter_common.h
@@ -8,8 +8,8 @@ namespace NKikimr::NMiniKQL {
 struct TPackResult {
     std::vector<ui8, TMKQLAllocator<ui8>> PackedTuples;
     std::vector<ui8, TMKQLAllocator<ui8>> Overflow;
-    int64_t NTuples{0};
-    int64_t AllocatedBytes() const;
+    i64 NTuples{0};
+    i64 AllocatedBytes() const;
 };
 
 using TPackedTuple = std::vector<ui8, TMKQLAllocator<ui8>>;
--- a/ydb/library/yql/dq/comp_nodes/hash_join_utils/scalar_layout_converter.cpp
+++ b/ydb/library/yql/dq/comp_nodes/hash_join_utils/scalar_layout_converter.cpp
@@ -583,10 +583,12 @@ class TScalarLayoutConverter : public IScalarLayoutConverter {
 public:
     TScalarLayoutConverter(
         TVector<IColumnDataExtractor::TPtr>&& packers,
-        const TVector<NPackedTuple::EColumnRole>& roles
+        const TVector<NPackedTuple::EColumnRole>& roles,
+        const THolderFactory& holderFactory
     )
         : Extractors_(std::move(packers))
         , InnerMapping_(Extractors_.size())
+        , HolderFactory_(holderFactory)
     {
         Y_ENSURE(roles.size() == Extractors_.size());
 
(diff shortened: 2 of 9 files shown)


ydb/library/yql/dq/comp_nodes/hash_join_utils/layout_converter_common.h
```diff
@@ -8,8 +8,8 @@ namespace NKikimr::NMiniKQL {
 struct TPackResult {
     std::vector<ui8, TMKQLAllocator<ui8>> PackedTuples;
     std::vector<ui8, TMKQLAllocator<ui8>> Overflow;
-    int64_t NTuples{0};
-    int64_t AllocatedBytes() const;
+    i64 NTuples{0};
+    i64 AllocatedBytes() const;
```
lll-phill-lll, line 12: Why not ui here?

nfrmtk: https://google.github.io/styleguide/cppguide.html#Integer_Types

lll-phill-lll: i64 is a platform independent alias for int64_t. 
int64_t typically breaks the build while importing the code into Arcadia.
The question here is why signed? Ntuples and allocated bytes seem to be non-negative

sorry, my previous comment was a bit misleading

ydb/library/yql/dq/comp_nodes/hash_join_utils/scalar_layout_converter.cpp
```diff
@@ -0,0 +1,650 @@
...
+namespace NKikimr::NMiniKQL {
+
+namespace {
+
+struct IColumnDataExtractor {
+    using TPtr = std::unique_ptr<IColumnDataExtractor>;
+
+    virtual ~IColumnDataExtractor() = default;
+
+    virtual void ExtractForPack(const NYql::NUdf::TUnboxedValue& value, TVector<const ui8*>& columnsData, TVector<const ui8*>& columnsNullBitmap, TVector<TVector<ui8>>& tempStorage) = 0;
```
nfrmtk, line 29: ExtractForPack is called for each element of every tuple, has virtual function call overhead and called on hot path 

so i think something should be changed

I suggest moving to batch processing here too, pack more than 1 tuple at time

ydb/library/yql/dq/comp_nodes/hash_join_utils/scalar_layout_converter.h
```diff
@@ -0,0 +1,35 @@
...
+    using TPtr = std::unique_ptr<IScalarLayoutConverter>;
+
+public:
+    virtual ~IScalarLayoutConverter() = default;
+
+    // Pack single tuple from scalar values
+    virtual void Pack(const NYql::NUdf::TUnboxedValue* values, TPackResult& packed) = 0;
+
+    // Unpack single tuple to scalar values
+    virtual void Unpack(const TPackResult& packed, ui32 tupleIndex, NYql::NUdf::TUnboxedValue* values, const THolderFactory& holderFactory) = 0;
```
nfrmtk: Is it neccesary to pass THolderFactory for every Unpack call as an argument? IMO moving it to MakeScalarLayoutConverter would make code more ergonomic

nfrmtk: Also, current hash table api returns const ui8* on match(which is a pointer in TPackResult::PackedTuples array,  not index).

It is possible to calculate index based on pointer, but it is also possible to just pass TNeumannJoinTable::Tuple here instead of const TPackResult& + ui32

nfrmtk (changes requested):

ydb/library/yql/dq/comp_nodes/hash_join_utils/scalar_layout_converter.h
lll-phill-lll: Thanks for the comment

Moved THolderFactory as you suggested. Left the tuple index as it was. I will change it while the integration to the scalar join.

ydb/library/yql/dq/comp_nodes/hash_join_utils/scalar_layout_converter.cpp
lll-phill-lll: Thanks!
Added batch pack and buckets batch pack

pzuev (approved):

nfrmtk (approved):

PR #28048, opened 2025-10-31
[incr backups] testing micro-framework

LebedevOleg1:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/tests/functional/backup_collection/basic_user_scenarios.py
+++ b/ydb/tests/functional/backup_collection/basic_user_scenarios.py
@@ -1123,6 +1123,206 @@ def backup_lifecycle(test_instance, collection_name: str, tables: List[str]):
 
 
 
+class TestFullCycleLocalBackupRestoreWIncr(BaseTestBackupInFiles):
+    def test_full_cycle_local_backup_restore_with_incrementals(self):
+        # Setup
+        t_orders = "orders"
+        t_products = "products"
+        full_orders = f"/Root/{t_orders}"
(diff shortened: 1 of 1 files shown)


ydb/tests/functional/backup_collection/basic_user_scenarios.py
Enjection, line 464: silent failure is not expected here:
1) log error
2) rename method to `_try_drop_table` or something like this, to express "best effort" nature of this method

Enjection, line 222: complex search for table can confuse two different tables with same name, but in different dirs, consider documenting this behavior

Enjection, line 869: Consider explicit type instead of dict. Make all builders consistent (e.g. Tuple[bool, Restore], Tuple[bool, SnapshotName])

Enjection: or maybe
```py
@dataclass
class BackupResult:
    success: bool
    snapshot_name: str
    
@dataclass  
class RestoreResult:
    success: bool
    data_verified: bool
    schema_verified: bool
    acl_verified: bool
```

Enjection, line 180: You already inherited from str, I guess you don't need this override

Enjection (approved):
I like it. There are couple little improvements but I believe we will address them later.

PR #28242, opened 2025-11-05
[ICRDMA] Rdma support for load actor EXT-1506 (#20815)

dcherednik:
### Changelog entry 

load actor should be able to allocate buffers using RDMA memory.

### Changelog category 

### Description for reviewers 

...


dcherednik:
load actor should be able to allocate buffers using RDMA memory.

robdrynkin (approved):

PR #28455, opened 2025-11-10
Correct distconf quorum for scenario when user adds new nodes to cluster

alexvru:
### Changelog entry 

Correct distconf quorum for scenario when user adds new nodes to cluster

### Changelog category 

### Description for reviewers 

This patch corrects adding new nodes scenario. Now distconf connects to these nodes and writes config to them before making quorum decision.


mregrock (approved):
LGTM

stanislav-shchetinin (approved):

PR #28553, opened 2025-11-10
Add option to BSC to create groups on INACTIVE disks

serbel324:
### Changelog entry 

Add new option to BSC config which allows group creation on non-ACTIVE disks. This change will allow to create groups on offline PDisks

### Changelog category 

* New feature


alexvru (changes requested):
Need to discuss.

PR #28557, opened 2025-11-10
[ICRDMA] Support for XDC transver via RDMA. EXT-1506

dcherednik:
### Changelog entry 

Allow to transfer XDC event via RDMA using READ verbs

RDMA XDC integration: robdrynkin@ydb.tech

### Changelog category 

* Experimental feature

### Description for reviewers 

...

### DIFF
--- a/ydb/library/actors/core/event_load.h
+++ b/ydb/library/actors/core/event_load.h
@@ -25,7 +25,7 @@ namespace NActors {
         size_t Tailroom = 0; // tailroom for the chunk
         size_t Alignment = 0; // required alignment
         bool IsInline = false; // if true, goes through ordinary channel
-        bool IsRdma = false; // if true, could go through RDMA
+        bool IsRdmaCapable = false; // if true, could go through RDMA
     };
 
     struct TEventSerializationInfo {
--- a/ydb/library/actors/core/event_load.h
+++ b/ydb/library/actors/core/event_load.h
@@ -25,6 +25,7 @@ namespace NActors {
         size_t Tailroom = 0; // tailroom for the chunk
         size_t Alignment = 0; // required alignment
         bool IsInline = false; // if true, goes through ordinary channel
+        bool IsRdma = false; // if true, could go through RDMA
     };
 
     struct TEventSerializationInfo {
--- a/ydb/library/actors/core/event_pb.cpp
+++ b/ydb/library/actors/core/event_pb.cpp
@@ -267,7 +267,8 @@ namespace NActors {
         return res;
     }
 
-    bool SerializeHeaderCommon(const TVector<TRope> &payload, std::function<bool(const char *p, size_t len)> append) {
+    template<typename TCb>
+    bool SerializeHeaderCommon(const TVector<TRope>& payload, TCb& append) {
         if (payload) {
             auto appendNumber = [&](size_t number) {
                 char buf[MaxNumberBytes];
@@ -339,7 +340,7 @@ namespace NActors {
 
     std::optional<TRope> SerializeToRopeImpl(std::function<TRcBuf(ui32 size)> alloc, const TVector<TRope> &payload) {
         TRope result;
-        auto sz = CalculateSerilizedHeaderSizeImpl(payload);
+        auto sz = CalculateSerializedHeaderSizeImpl(payload);
         if (!sz) {
             return result;
         }
@@ -348,7 +349,7 @@ namespace NActors {
             return {};
         }
         char* data = headerBuf.GetDataMut();
-        auto append = [&](const char *p, size_t len) {
+        auto append = [&data](const char *p, size_t len) {
             std::memcpy(data, p, len);
             data += len;
             return true;
@@ -413,7 +414,7 @@ namespace NActors {
         }
     }
 
-    ui32 CalculateSerilizedHeaderSizeImpl(const TVector<TRope> &payload) {
+    ui32 CalculateSerializedHeaderSizeImpl(const TVector<TRope> &payload) {
         ui32 result = 0;
         if (payload) {
             ++result; // marker
@@ -430,12 +431,10 @@ namespace NActors {
     ui32 CalculateSerializedSizeImpl(const TVector<TRope> &payload, ssize_t recordSize) {
         ssize_t result = recordSize;
         if (result >= 0 && payload) {
-            result += CalculateSerilizedHeaderSizeImpl(payload);
-            size_t totalPayloadSize = 0;
+            result += CalculateSerializedHeaderSizeImpl(payload);
             for (const TRope& rope : payload) {
-                totalPayloadSize += rope.GetSize();
+                result += rope.GetSize();
             }
-            result += totalPayloadSize;
         }
         return result;
     }
--- a/ydb/library/actors/core/event_pb.cpp
+++ b/ydb/library/actors/core/event_pb.cpp
@@ -1,5 +1,8 @@
 #include "event_pb.h"
 
+#include <ydb/library/actors/interconnect/rdma/mem_pool.h>
+#include <ydb/library/actors/protos/interconnect.pb.h>
+
 namespace NActors {
     TString EventPBBaseToString(const TString& header, const TString& dbgStr) {
         TString res;
@@ -264,26 +267,8 @@ namespace NActors {
         return res;
     }
 
-    bool SerializeToArcadiaStreamImpl(TChunkSerializer* chunker, const TVector<TRope> &payload) {
-        // serialize payload first
+    bool SerializeHeaderCommon(const TVector<TRope> &payload, std::function<bool(const char *p, size_t len)> append) {
         if (payload) {
-            void *data;
-            int size = 0;
-            auto append = [&](const char *p, size_t len) {
-                while (len) {
-                    if (size) {
-                        const size_t numBytesToCopy = std::min<size_t>(size, len);
-                        memcpy(data, p, numBytesToCopy);
-                        data = static_cast<char*>(data) + numBytesToCopy;
-                        size -= numBytesToCopy;
-                        p += numBytesToCopy;
-                        len -= numBytesToCopy;
-                    } else if (!chunker->Next(&data, &size)) {
-                        return false;
-                    }
-                }
-                return true;
-            };
             auto appendNumber = [&](size_t number) {
                 char buf[MaxNumberBytes];
                 return append(buf, SerializeNumber(number, buf));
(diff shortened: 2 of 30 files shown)


ydb/library/actors/core/event_pb.cpp
```diff
@@ -264,26 +267,8 @@ namespace NActors {
         return res;
     }
 
-    bool SerializeToArcadiaStreamImpl(TChunkSerializer* chunker, const TVector<TRope> &payload) {
-        // serialize payload first
+    bool SerializeHeaderCommon(const TVector<TRope> &payload, std::function<bool(const char *p, size_t len)> append) {
```
alexvru: Do we really need indirect call overhead here? For every little chunk of data?
I'd prefer function with template callback and other one calling it through std::function when necessary.

ydb/library/actors/core/event_pb.cpp
```diff
@@ -361,23 +413,43 @@ namespace NActors {
         }
     }
 
-    ui32 CalculateSerializedSizeImpl(const TVector<TRope> &payload, ssize_t recordSize) {
-        ssize_t result = recordSize;
-        if (result >= 0 && payload) {
+    ui32 CalculateSerilizedHeaderSizeImpl(const TVector<TRope> &payload) {
```
alexvru: seriAlized

ydb/library/actors/core/event_pb.cpp
```diff
@@ -361,23 +413,43 @@ namespace NActors {
...
+        return result;
+    }
+
+    ui32 CalculateSerializedSizeImpl(const TVector<TRope> &payload, ssize_t recordSize) {
+        ssize_t result = recordSize;
+        if (result >= 0 && payload) {
+            result += CalculateSerilizedHeaderSizeImpl(payload);
+            size_t totalPayloadSize = 0;
+            for (const TRope& rope : payload) {
+                totalPayloadSize += rope.GetSize();
```
alexvru: What's the use for intermediate counter? Why not add to result directly?

ydb/library/actors/core/event_load.h
```diff
@@ -25,6 +25,7 @@ namespace NActors {
         size_t Tailroom = 0; // tailroom for the chunk
         size_t Alignment = 0; // required alignment
         bool IsInline = false; // if true, goes through ordinary channel
+        bool IsRdma = false; // if true, could go through RDMA
```
alexvru: Suggest calling it IsRdmaCapable or something like that.

ydb/library/actors/core/event_pb.h
```diff
@@ -202,6 +205,21 @@ namespace NActors {
...
+            std::optional<TRope> result = SerializeToRopeImpl(alloc, Payload);
+            if (!result) {
+                return {};
+            }
+            ui32 size = Record.ByteSizeLong();
+            TRcBuf recordsSerializedBuf = alloc(size);
+            if (!recordsSerializedBuf) {
+                return {};
+            }
+            Y_ABORT_UNLESS(Record.SerializePartialToArray(recordsSerializedBuf.GetDataMut(), size));
```
alexvru: Usually side effects of an assertion are not guaranteed. So it's better to evaluate Serialize... outsize Y_ABORT_UNLESS.

ydb/library/actors/interconnect/interconnect_channel.cpp
```diff
@@ -102,11 +105,30 @@ namespace NActors {
                     } else if (Params.UseExternalDataChannel && !SerializationInfo->Sections.empty()) {
                         State = EState::SECTIONS;
                         SectionIndex = 0;
+
+                        size_t totalSize = 0;
+                        // It is possible to have event without payload. Such events has only one section.
+                        // We do not send such events via rdma.
+                        bool sendViaRdma = Params.UseRdma && RdmaMemPool && SerializationInfo->Sections.size() > 2;
```
alexvru, line 112: > 2 or >= 2?

ydb/library/actors/interconnect/interconnect_channel.cpp
```diff
@@ -147,6 +169,9 @@ namespace NActors {
                         if (section.IsInline && Params.UseXdcShuffle) {
                             type = static_cast<ui8>(EXdcCommand::DECLARE_SECTION_INLINE);
                         }
+                        if (SendViaRdma) {
```
alexvru, line 172: Maybe add a check to ensure that these two conditions are never true simultaneously?

ydb/library/actors/interconnect/interconnect_channel.cpp
```diff
@@ -294,6 +325,95 @@ namespace NActors {
         return complete;
     }
 
+    bool TEventOutputChannel::SerializeEventRdma(TEventHolder& event, NActorsInterconnect::TRdmaCreds& rdmaCreds,
+        ui32* checkSum, ssize_t rdmaDeviceIndex)
+    {
+        if (!event.Buffer && event.Event) {
+            std::optional<TRope> rope = event.Event->SerializeToRope(
```
alexvru, line 332: This inplace serialization may lead to unexpected growth of p100 latencies in the IC pool.

ydb/library/actors/interconnect/interconnect_channel.cpp
```diff
@@ -294,6 +325,95 @@ namespace NActors {
         return complete;
     }
 
+    bool TEventOutputChannel::SerializeEventRdma(TEventHolder& event, NActorsInterconnect::TRdmaCreds& rdmaCreds,
+        ui32* checkSum, ssize_t rdmaDeviceIndex)
```
alexvru: "checksum" is a single word. Like city called "Череповец", not "ЧерепОвец".

ydb/library/actors/interconnect/interconnect_channel.cpp
```diff
@@ -294,6 +325,95 @@ namespace NActors {
...
+        size_t partSize = sizeof(TChannelPart) + sizeof(ui8) + sizeof(ui16) + rdmaCreds.ByteSizeLong() + sizeof(ui32);
+        Y_ABORT_UNLESS(partSize < 4096);
+
+        if (partSize > Max<ui16>() || partSize > task.GetInternalFreeAmount()) {
+            // TODO: support split into multiple parts
+            return std::nullopt; // not enough space to serialize RDMA payload
+        }
+
+        char buffer[partSize];
+        TChannelPart *part = reinterpret_cast<TChannelPart*>(buffer);
```
alexvru, line 390: It's better to WriteUnaligned now :) ubsan will raise an issue here.

ydb/library/actors/interconnect/interconnect_channel.cpp
```diff
@@ -294,6 +325,95 @@ namespace NActors {
...
+        *part = {
+            .ChannelFlags = static_cast<ui16>(ChannelId | TChannelPart::XdcFlag),
+            .Size = static_cast<ui16>(partSize - sizeof(TChannelPart))
+        };
+        char *ptr = reinterpret_cast<char*>(part + 1);
+        *ptr++ = static_cast<ui8>(EXdcCommand::RDMA_READ);
+        ui16 credsSerializedSize = rdmaCreds.ByteSizeLong();
```
alexvru: You calculate ByteSizeLong twice in this function.

ydb/library/actors/interconnect/interconnect_channel.cpp
```diff
@@ -294,6 +325,95 @@ namespace NActors {
...
+        }
+
+        Y_ABORT_UNLESS(rdmaCreds.SerializePartialToArray(ptr, credsSerializedSize));
+        ptr += credsSerializedSize;
+        WriteUnaligned<ui32>(ptr, checkSum);
+        OutputQueueSize -= event.EventSerializedSize;
+
+        task.Write<false>(buffer, partSize);
+
+        task.AttachRdmaPayloadSize(payloadSz);
```
alexvru, line 412: Does it attach whole message at once? It may cause problems with fair bandwidth distribution between channels (in theory).

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
```diff
@@ -59,6 +63,126 @@ namespace NActors {
...
+            const auto& region = NInterconnect::NRdma::TryExtractFromRcBuf(chunk);
+            if (NeedReallocateRdma(region)) {
+                auto newBuf = TRcBuf::Copy(chunk.GetContiguousSpan(), chunk.Headroom(), chunk.Tailroom());
+                newRope.Insert(newRope.End(), TRope(std::move(newBuf)));
+            } else {
+                newRope.Insert(newRope.End(), TRope(std::move(chunk)));
+            }
+        }
+
+        rope.clear();
```
alexvru: What for?

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
```diff
@@ -59,6 +63,126 @@ namespace NActors {
...
+        }
+
+        if (!needRealloc) {
+            return;
+        }
+
+        TRope newRope;
+
+        for (TRope::TIterator it = rope.Begin(); it != rope.End(); ++it) {
+            TRcBuf chunk = it.GetChunk();
```
alexvru: GetChunk() returns mutable reference. You can modify it inplace. At least it should be possible and I suppose it is a bug if it is not.

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
```diff
@@ -143,7 +267,11 @@ namespace NActors {
     void TInputSessionTCP::Bootstrap() {
         SetPrefix(Sprintf("InputSession %s [node %" PRIu32 "]", SelfId().ToString().data(), NodeId));
         Become(&TThis::WorkingState, DeadPeerTimeout, new TEvCheckDeadPeer);
-        LOG_DEBUG_IC_SESSION("ICIS01", "InputSession created");
+        if (RdmaQp) {
+            LOG_DEBUG_IC_SESSION("ICIS01", "InputSession created, rdma qp num: %d", RdmaQp->GetQpNum());
```
alexvru: Markers must not repeat in code, despite they are obsolete :)

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
```diff
@@ -646,17 +913,25 @@ namespace NActors {
...
+            if (expectedChecksum) {
                 ui32 checksum = 0;
+                XXH3_state_t state;
+                XXH3_64bits_reset(&state);
                 for (const auto&& [data, size] : payload) {
-                    checksum = Crc32cExtendMSanCompatible(checksum, data, size);
+                    XXH3_64bits_update(&state, data, size);
                 }
-                if (checksum != descr.Checksum) {
+                checksum = XXH3_64bits_digest(&state);
```
alexvru, line 907: You can't just change checksumming algorithm, it will make this version incompatible with previous ones, won't it?

alexvru (changes requested):

ydb/library/actors/core/event_pb.cpp
dcherednik: Agree. Fixed.

ydb/library/actors/core/event_pb.cpp
dcherednik: fixed

ydb/library/actors/core/event_pb.cpp
dcherednik: hmm) fixed

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
dcherednik: Ou. It means we need to rename all ICRDMA markers we have added already. Ok)

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
dcherednik: Just to be absolutely sure we free rdma memory)
Actually this code is not needed after solving problem with "прикапыванием" rdma memory somewhere. But right now we meet some problems with it and this event copy in this function is just workaround.

ydb/library/actors/interconnect/interconnect_channel.cpp
dcherednik: In theory... But even in case of 200Gbit/s network time of transfer 1MB event (app to app via RDMA) compared or even less than time of whole one IC cycle, but we expect 400 or even 800Gbit/s network, so it is unclear does the splitting make some profit or we just fire some CPU. Moreover it is unclear how does the bandwidth will be distributed between TCP messages and RoCEv2 traffic. Probably we need to implement RDMA SEND and RECIEVE to reduce TCP overhead for control messages.
So we decided not to write some complicated code until we got some real feedback from production like load.

ydb/library/actors/interconnect/interconnect_channel.cpp
dcherednik: Fixed, thanks

ydb/library/actors/interconnect/interconnect_channel.cpp
dcherednik: Existed FeedExternalPayload function has same code right now without ubsan issue) I assume some default alignment of buffer (AI told the is a default ui16 alignment for VLA, but I am not sure))), so the ChannelFlags and Size (which is ui16) writes are ok.

Lets fix it globally to be more clear in different PR?

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
dcherednik: Oops. Thanks.

ydb/library/actors/interconnect/interconnect_channel.cpp
robdrynkin: If event has n payloads, serialized event has 2+n sections: 1 for protobuf, 1 for header, n for payloads
https://github.com/ydb-platform/ydb/blob/main/ydb/library/actors/core/event_pb.cpp#L392-L399

So if event has 1 payload, number of sections will be 3

robdrynkin: But if there are no payloads, number of sections will be equal to 1, so event can't have 2 sections. So >2 and >=2 in this case are equal conditions.

ydb/library/actors/interconnect/interconnect_channel.cpp
robdrynkin: Right now it can be true simultaneously, because CreateSerializationInfo don't have information about transport (rdma or tcp), and I think it is good thing, but shuffle is not supported for rdma yet. So for now IsRdma is the strongest condition, if we decided to send event via rdma we ignore all other conditions like IsInline, but in the near future shuffle will be supported for rdma.

dcherednik: Right now we have no shuffle support. So the section has IsInline and Params.UseXdcShuffle is true, and after we promote it to RDMA if SendViaRdma is set.
But we must implement support for UseXdcShuffle ASAP, after that we will not ignore IsInline flag here.
@robdrynkin can provide more details about this part.

ydb/library/actors/interconnect/interconnect_channel.cpp
robdrynkin: We have plans to support coroutine serializer for rdma later (after the shuffle, we see it more critical for now)

dcherednik: https://github.com/ydb-platform/ydb/issues/28724

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
dcherednik: fixed)

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
dcherednik: It works) I have rewritten this part.

alexvru (approved):

PR #28804, opened 2025-11-13
Fix config propagation

alexvru:
### Changelog entry 

Fix config propagation

### Changelog category 

### Description for reviewers 

This patch makes commits explicit, all local configuration discovery is kept local. Only leader can provide commits to storage config.


mregrock (approved):
LGTM

PR #29017, opened 2025-11-17
Fix split brain potential situation in distconf

alexvru:
### Changelog entry 

Fix split brain potential situation in distconf

### Changelog category 

### Description for reviewers 

This fixes possible case with data loss leading to split brain situation in distconf.


mregrock (approved):
LGTM

PR #29490, opened 2025-11-25
Statistics Improvement

XJIE6:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


maximyurchuk (changes requested):
в 25-3-1 не надо по инфе от xjle6

PR #29634, opened 2025-11-26
Stress test for incremental backup

LebedevOleg1:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/tests/stress/backup/workload/__init__.py
+++ b/ydb/tests/stress/backup/workload/__init__.py
@@ -2,6 +2,7 @@ import random
 import threading
 import time
 import ydb
+import uuid
 
 from ydb.tests.stress.common.common import WorkloadBase
 
@@ -48,8 +49,9 @@ class WorkloadRunnerBackup:
         self.client = client
         self.duration = duration
         self.backup_interval = backup_interval
-        self.table_path = '/'.join([self.client.database, "table", str(random.randint(100, 999))])
-        self.collection_name = "backup_col_" + str(random.randint(1000, 9999))
+        uid = uuid.uuid4().hex
+        self.table_path = f"{self.client.database}/table/{uid}"
+        self.collection_name = f"backup_col_{uid}"
         ydb.interceptor.monkey_patch_event_handler()
 
     def __enter__(self):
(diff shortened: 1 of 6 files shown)


ydb/tests/stress/backup/workload/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+        bthread = threading.Thread(target=backup_loop)
+        bthread.daemon = True
+        bthread.start()
+
+        started_at = time.time()
+        try:
+            while time.time() - started_at < self.duration:
+                print(f"Elapsed {(int)(time.time() - started_at)} seconds, stat:")
+                print(f"\t{rw.name}: {rw.get_stat()}")
+                time.sleep(10)
```
Enjection, line 117: I would make it at least configurable. Under failures we will probably face almost infinite timeouts and our logic might break.

Enjection: And of course if we are able to poll some state instead of just waiting it is preferrable

ydb/tests/stress/backup/workload/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+        return [lambda x=x: self._loop(x, self._generate_select_keys(x, count)) for x in range(count)]
+
+
+class WorkloadRunnerBackup:
+    def __init__(self, client, duration, backup_interval):
+        self.client = client
+        self.duration = duration
+        self.backup_interval = backup_interval
+        self.table_path = '/'.join([self.client.database, "table", str(random.randint(100, 999))])
+        self.collection_name = "backup_col_" + str(random.randint(1000, 9999))
```
Enjection: better use uuid for names here, less probability of clash, no cons

Enjection (changes requested):
Mostly good, not sure how will it behabe under load. One thing is -- do we have BackupCollectionService flag defined somewhere? How it works without it?

ydb/tests/stress/backup/workload/__init__.py
LebedevOleg1: I did it the same way as most tests. Should I change it in this case? Is it really a bad thing if our service crashes? Will we be able to immediately see the problem instead of endless polling?

Enjection (approved):

PR #30231, opened 2025-12-05
[C++ SDK] Added deadline propagation in retrier

Gazizonoki:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


ydb/public/sdk/cpp/src/client/table/impl/table_client.h
dcherednik, line 113: Do we really need this refactoring?? The ability to use TSession directly from low level code make it possible to capture the session somewhere and it may cause some kind of deadlock or/and stuck session in the pool.

ydb/public/sdk/cpp/src/client/query/client.cpp
dcherednik: Was it approved by KQP team? Is it a good idea to have a timeout for rollback/commit tx call as default behaviour? I this case user will have no idea was tx commiter or not in most cases.

dcherednik (commented):
Please add some

pnv1 (approved):

PR #30261, opened 2025-12-06
Separate DeleteQueue to allow reusing only current-generation locators

alexvru:
### Changelog entry 

Separate DeleteQueue to allow reusing only current-generation locators

### Changelog category 

### Description for reviewers 

This patch removes possibility for potential data race occuring with previous-generation agent trying to write blob that is being collected with fresh-generation BlobDepot (which may be reused and contain inconsistent data as a result).


mregrock (approved):
LGTM

PR #30273, opened 2025-12-07
Fix Trash table loading

alexvru:
### Changelog entry 

Fix Trash table loading

### Changelog category 

### Description for reviewers 

This patch fixes a race when Trash table is being loaded and reads an item that has just been inserted. It should not cause much trouble, but at least it can generate incorrect counters.


mregrock (approved):
LGTM

PR #30320, opened 2025-12-08
Pass column type id to statistics UDAFs

ztlpn:
### Changelog category 

### Description for reviewers 

Pass type id to Create and AddValue methods, allow Create parameters of arbitrary type.


yizenov (approved):
Looks good!

PR #30351, opened 2025-12-09
Do not reuse S3 locators in BlobDepot

alexvru:
### Changelog entry 

Do not reuse S3 locators in BlobDepot

### Changelog category 

### Description for reviewers 

Locator reuse code has been removes as it may lead to data races with long writes at agent side.


mregrock (approved):
LGTM

PR #30450, opened 2025-12-10
[NOT-FOR-MERGE-YET] dq: propagate watermark settings from sources to graph

yumkam:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

`yql/essentials` part should be integrated separately and rebased upon import.
There maybe cleanup pr to get rid of old version of FillSourceSettings.
Expected to require *both* `pragma WatermarksMode="default"` *and* `WITH (...., WATERMARK = ...)`
It makes sense to enable dq.WatermarksMode="default" to default sever flags

TODO:
- [x] Kqp (streaming/v2) is broken (does not propagate watermark settings from sources)
- [x] Not shared-reading tests are broken (requires dummy WITH WATERMARKS expression)
- [x] Passing/defaults for settings in pq dq integration (Fill*) should be rechecked and updated
- [ ] Integrate YQL part and rebase upon imported version

### DIFF
--- a/ydb/tests/fq/yds/test_watermarks.py
+++ b/ydb/tests/fq/yds/test_watermarks.py
@@ -53,23 +53,21 @@ class TestWatermarks(TestYdsBase):
         client.create_yds_connection(
             name=YDS_CONNECTION, database=os.getenv("YDB_DATABASE"), endpoint=os.getenv("YDB_ENDPOINT"), shared_reading=shared_reading
         )
-        self.init_topics(f"test_watermarks_{"shared" if shared_reading else "no_shared"}")
+        self.init_topics(f"test_watermarks_{'shared' if shared_reading else 'no_shared'}")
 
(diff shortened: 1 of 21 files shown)


ydb/tests/fq/yds/test_watermarks.py
```diff
@@ -53,39 +53,38 @@ def test_watermarks(self, kikimr: StreamingOverKikimr, client: FederatedQueryCli
...
+                    {ts} as hopTime
                 FROM {self.input_topic}
                 WITH(
                     FORMAT=json_each_row,
                     SCHEMA(
                         ts String NOT NULL,
                         pass Uint64
                     )
-                    {watermark_expr}
+                    , WATERMARK AS ({ts} - Interval("PT0.1S"))
```
APozdniakov: Предлагаю `WATERMARK = ` тут и ниже

APozdniakov (dismissed):

PR #30506, opened 2025-12-11
Do not require commit to Console to accept new configs

alexvru:
### Changelog entry 

Do not require commit to Console to accept new configs

### Changelog category 

### Description for reviewers 

This patch makes it unnecessary for previous config to be committed to Console in order to receive new config updates from the user.


mregrock (approved):
LGTM

PR #30572, opened 2025-12-11
Publish YDB Server 25.2.1.24

liruoko:
### Changelog entry 

Publish YDB Server 25.2.1.23

### Changelog category 

### DIFF
--- a/ydb/docs/en/core/changelog-server.md
+++ b/ydb/docs/en/core/changelog-server.md
@@ -4,6 +4,8 @@
 
 ### Version 25.2.1.24 {#25-2-1-24}
 
+Release date: January 28, 2026.
+
 #### Bug Fixes
 
 * [Fixed](https://github.com/ydb-platform/ydb/pull/25112) an [issue](https://github.com/ydb-platform/ydb/issues/23858) where [tablet](./concepts/glossary.md#tablet) deletion might get stuck
--- a/ydb/docs/en/core/changelog-server.md
+++ b/ydb/docs/en/core/changelog-server.md
@@ -19,8 +19,8 @@
 * Fixed a couple of cluster bootstrap related issues:
   * [Fixed](https://github.com/ydb-platform/ydb/pull/25678) an [issue](https://github.com/ydb-platform/ydb/issues/25023) where cluster bootstrap could hang when mandatory authorization was enabled.
   * [Fixed](https://github.com/ydb-platform/ydb/pull/28886) an [issue](https://github.com/ydb-platform/ydb/issues/27228) where it was impossible to create new databases for several minutes immediately after cluster deployment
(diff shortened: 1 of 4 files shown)


ydb/docs/en/core/changelog-server.md
```diff
@@ -2,6 +2,21 @@
 
 ## Version 25.2 {#25-2}
 
+### Version 25.2.1.23 {#25-2-1-23}
+
+#### Bug Fixes
+
+* Fixed an [issue](https://github.com/ydb-platform/ydb/issues/23858) where tablet deletion might get stuck
+* Fixed an [issue](https://github.com/ydb-platform/ydb/issues/20866) with follower alter
```
antonkovalenko: I miss details: when alter happens, when should I be worried and consider upgrading?
(was it an alter at all or simple follower start during version migration?)

ydb/docs/en/core/changelog-server.md
```diff
@@ -2,6 +2,21 @@
...
+* Fixed a couple of change feed related issues:
```
antonkovalenko: details would be nice - like "when importing table with changefeed and Utf8 in PK columns"
and "when importing table with changefeed that has specific naming patter"

the idea is to have information in the line that might cause recognition. I beleive in a story where user reads the log and thinks "I had issues with changefeeds and imports - let's look into this one"

ydb/docs/en/core/changelog-server.md
```diff
@@ -2,6 +2,21 @@
...
+  * [Issue](https://github.com/ydb-platform/ydb/issues/25524)
+  * [Issue](https://github.com/ydb-platform/ydb/issues/25454)
+* Fixed an [error](https://github.com/ydb-platform/ydb/issues/25869) in upsert in ColumnShard
+* Fixed heap-use-after-free [error](https://github.com/ydb-platform/ydb/issues/26225)
```
antonkovalenko: Wouldn't it be better to specify that it's related to create transfer (if it is ofc)?

ydb/docs/en/core/changelog-server.md
```diff
@@ -2,6 +2,21 @@
...
+* Fixed an [issue](https://github.com/ydb-platform/ydb/issues/23122) with duplicates in uniq secondary index
```
antonkovalenko: 1. s/uniq/unique/g ?
2. noting that bug affects upserts only seems useful (if affects only upserts)

ydb/docs/en/core/changelog-server.md
```diff
@@ -2,6 +2,21 @@
...
+* Fixed an [issue](https://github.com/ydb-platform/ydb/issues/26565) with checksum mismatch error on restoration compressed backup from s3
+* Fixed [node failure](https://github.com/ydb-platform/ydb/issues/27193) on some queries
```
antonkovalenko: seems interesting. query to column tables caused VERIFY in log.cpp? 
I would mention that it is connected with column tables
it is interesting is their a way to fix log.cpp?

fomichev3000 (approved):

Issue #30432, opened 2025-12-10
cluster went down


adameat:
```
Dec  8 10:17:17 ydb-vla-dev02-000 kikimr[208931]: VERIFY failed (2025-12-08T10:17:16.432121+0300):
Dec  8 10:17:17 ydb-vla-dev02-000 kikimr[208931]:   ydb/core/blobstorage/vdisk/synclog/blobstorage_synclogdsk.h:137
Dec  8 10:17:17 ydb-vla-dev02-000 kikimr[208931]:   GetOwnedChunks(): requirement inserted failed
```

naspirato:
встретили проблему  в stable-25-3-1
https://github.com/ydb-platform/ydb/commit/df2e9f6e74319748a6c99a149b71ea4a204720f8

запуск https://a.yandex-team.ru/projects/kikimr/ci/actions/flow?camera-state=-997%3B60%3B1&dir=kikimr%2Fci%2Farcadia_ci%2Ftasklets%2Fkikimr_nemesis&id=run_2dc_mirror_3_test_nemesis&number=319

отчет https://proxy.sandbox.yandex-team.ru/10750533454/index.html#suites/a59c1f6a75be8eb7435f91fcac1b4930/ea078a60b17016d3/

```
VERIFY failed (2026-01-05T02:24:37.552594+0300):
ydb/core/blobstorage/vdisk/synclog/blobstorage_synclogdsk.h:137
GetOwnedChunks(): requirement inserted failed
0. /home/robot-ydb-benchmark/repos/ydb-platform/ydb/util/system/yassert.cpp:83: NPrivate::InternalPanicImpl(int, char const*, char const*, int, int, int, TBasicStringBuf>, char const*, unsigned long) @ 0xB0367AE
1. /home/robot-ydb-benchmark/repos/ydb-platform/ydb/util/system/yassert.cpp:55: NPrivate::Panic(NPrivate::TStaticBuf const&, int, char const*, char const*, char const*, ...) @ 0xB031274
2. /home/robot-ydb-benchmark/repos/ydb-platform/ydb/ydb/core/blobstorage/vdisk/synclog/blobstorage_synclogdsk.h:137: NKikimr::NSyncLog::TOneChunk::GetOwnedChunks(TSet, std::__y1::allocator>&) const @ 0x15BD435A
[...]

PR #30608, opened 2025-12-12
Release snapshot in synclog when not needed

serbel324:
### Changelog entry 

Release SyncLog snapshot as soon as possible. Fixes possible root cause of #30432

### Changelog category 

* Bugfix


alexd65536 (approved):
LGTM

PR #30779, opened 2025-12-16
Fix VDisk sync log race

alexvru:
### Changelog entry 

Fix VDisk sync log race

### Changelog category 

### Description for reviewers 

This fixes race with deleting chunks leading to assertion failure and causing possible double-free corruption.


mregrock (approved):
LGTM

PR #31018, opened 2025-12-19
Fix drop backup collection with incrementals after restore

LebedevOleg1:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


CyberROFL (commented):
А тест какой-нибудь будет?

Issue #30109, opened 2025-12-04
Assert validation with string literals is always true


ubyte:
When a constant string literal is erroneously passed to one of the ASSERT, ABORT, VERIFY or ENSURE macros, it is silently converted to 'true', resulting in the check being skipped completely.

Example
```cpp

switch (version) {
  case 1: return HandleV1();
  case 2: return HandleV2();
  default: Y_ESNURE("Unsupported version"); // exception newer raised
}
```

PR #31299, opened 2025-12-25
Invert condition in the assert-like macro

ubyte:
### Changelog entry 
https://github.com/ydb-platform/ydb/issues/30109
When a constant string literal is erroneously passed to one of the ASSERT, ABORT, VERIFY or ENSURE macros, it is silently converted to 'true', resulting in the check being skipped completely.

### Changelog category 

### Description for reviewers 

...


CyberROFL (approved):
repllication & schemeshard OK

dorooleg (approved):

azevaykin (approved):

PR #31310, opened 2025-12-25
Fix encryption assertion failure in DS proxy

alexvru:
### Changelog entry 

Fix encryption assertion failure in DS proxy

### Changelog category 

### Description for reviewers 

This fixes false-positive assertion failure in DS proxy when encryption is enabled.


mregrock (approved):
LGTM

PR #31674, opened 2026-01-03
feat ydb: support load_factor in ListEndpoints handler

r142f:
### Changelog entry

Discovery service now returns actual node load in the load_factor field. Clients can use this metric for load-aware balancing when selecting database endpoints.

### Changelog category

* Improvement

### Description for reviewers

**Problem**:
The `Ydb.Discovery.V1.DiscoveryService/ListEndpoints` method always returned `load_factor = 0`, making it impossible for clients to perform load-aware balancing across database nodes.

**Solution**:
- Added new event `TEvBoardPublishUpdate` for updating data in State Storage Board 
- Added handlers for `TEvBoardPublishUpdate` in `TBoardPublishActor` and `TBoardReplicaPublishActor`
- Updated `TGRpcEndpointPublishActor`, now it:
  - Polls Node Whiteboard every 15 seconds (synchronized with Whiteboard update period)
  - Calculates `load_factor = LoadAverage[0] / NumberOfCpus`
  - Sends updates to State Storage Board when change >= 0.01 using `TEvBoardPublishUpdate`
- Added Python functional test to verify load_factor updates

### DIFF
--- a/ydb/core/base/board_publish.cpp
+++ b/ydb/core/base/board_publish.cpp
@@ -74,8 +74,15 @@ public:
         Become(&TThis::StatePublish);
     }
 
+    void Handle(TEvStateStorage::TEvBoardPublishUpdate::TPtr& ev) {
+        Payload = ev->Get()->NewPayload;
+
+        Send(Replica, new TEvStateStorage::TEvReplicaBoardPublish(Path, Payload, 0, true, PublishActor, ClusterStateGeneration, ClusterStateGuid), IEventHandle::FlagSubscribeOnSession, ++Round);
+    }
+
     STATEFN(StatePublish) {
         switch (ev->GetTypeRewrite()) {
+            hFunc(TEvStateStorage::TEvBoardPublishUpdate, Handle);
             cFunc(TEvents::TEvPoisonPill::EventType, Cleanup);
             cFunc(TEvInterconnect::TEvNodeDisconnected::EventType, NotAvailable); // no cleanup on node disconnect
             cFunc(TEvStateStorage::TEvReplicaShutdown::EventType, NotAvailableUnsubscribe);
@@ -99,7 +106,7 @@ class TBoardPublishActor : public TActorBootstrapped<TBoardPublishActor> {
     };
 
     const TString Path;
-    const TString Payload;
+    TString Payload;
     const TActorId Owner;
     const ui32 TtlMs;
     const bool Register;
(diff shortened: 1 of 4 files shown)


ydb/core/grpc_services/grpc_endpoint_publish_actor.cpp
```diff
@@ -9,33 +9,32 @@
...
 
 namespace NKikimr::NGRpcService {
 
 using namespace NActors;
 
 class TGRpcEndpointPublishActor : public TActorBootstrapped<TGRpcEndpointPublishActor> {
+    // Update period for load_factor polling from Node Whiteboard.
+    // Synchronized with Node Whiteboard's internal update period (15 seconds)
+    // to ensure we always get fresh data without unnecessary polling.
+    static constexpr int LOAD_FACTOR_UPDATE_PERIOD_SECONDS = 15;
```
Ane1y: лучше использовать TDuration

ydb/core/grpc_services/grpc_endpoint_publish_actor.cpp
```diff
@@ -83,6 +108,44 @@ class TGRpcEndpointPublishActor : public TActorBootstrapped<TGRpcEndpointPublish
...
+        if (numCpus == 0) {
+            return;
+        }
+    
+        auto newLoadFactor = loadAvg / numCpus;
+        if (std::abs(newLoadFactor - CurrentLoadFactor) >= LOAD_FACTOR_CHANGE_THRESHOLD) {
+            CurrentLoadFactor = newLoadFactor;
+
+            if (PublishActor) {
+                UpdatePublishActor();
```
Ane1y, line 139: я бы еще добавила какой нибудь лог что publishActor недоступен, а то просто теряем статистику в этом случае

ydb/core/base/board_publish.cpp
```diff
@@ -290,11 +297,24 @@ class TBoardPublishActor : public TActorBootstrapped<TBoardPublishActor> {
...
+            }
+        }
+    }
+
     STATEFN(StateCalm) {
         switch (ev->GetTypeRewrite()) {
             hFunc(TEvStateStorage::TEvResolveReplicasList, Handle);
             hFunc(TEvStateStorage::TEvPublishActorGone, CalmGone);
             hFunc(TEvPrivate::TEvRetryPublishActor, Handle);
+            hFunc(TEvStateStorage::TEvBoardPublishUpdate, Handle);
```
Ane1y, line 318: еще какой нибудь обработчик стоит добавить в stateResolve, маловероятно конечно что прилетит но все же

ydb/core/grpc_services/grpc_endpoint_publish_actor.cpp
r142f: Поправил

ydb/core/grpc_services/grpc_endpoint_publish_actor.cpp
r142f: Добавил лог

ydb/core/base/board_publish.cpp
r142f: Добавил тот же обработчик. Так как реплики еще не проинициализированы, отправки не будет, но будет сохранено актуальное значение `Payload`

Ane1y (approved):

gridnevvvit (changes requested):
Thank you for your contribution.

However, load_factor parameter is not used by our officially supported SDK and it somewhat deprecated or not supported, so it's not clear what is the specific reason to implement that. Is there some sort of feature request from SDK team? Do you have approved RFC with description how SDK will use this parameter?

The approach used to implement the load factor is not quite optimal I think.
Generally, list endpoints returns a full list of nodes and it doesn't frequently change so it means, the result of list endpoints can be cached and updated only on updates from StateStorage, and current approach probably will not be efficient because caching will not be working. So, we need some approach with rare updates here.

PR #32341, opened 2026-01-19
Made tests independent on sys views creation P.1

n00bcracker:
Some tests depend on number of sysviews created at SchemeShard initialization, so I have to compute some counters in tests dynamically


ijon (approved):

nikvas0 (approved):
kqp ok

PR #32778, opened 2026-01-26
Implement raw reads and writes in PDisk

alexvru:
### Changelog entry 

Implement raw reads and writes in PDisk

### Changelog category 

### Description for reviewers 

This patch adds TEvChunkReadRaw/TEvChunkWriteRaw events to PDisk to allow reading and writing raw device chunks without any scheduling, encyption and validation.


mregrock (approved):
LGTM

Issue #27123, opened 2025-10-20
Jepsen: incompatible order during possible volatile tx migration


snaury:
Found some extremely rare incompatible order for key 25:

<img width="866" height="449" alt="Image" src="https://github.com/user-attachments/assets/3c1548f4-ad92-47ce-83f1-296334fe5494" />

Relevant operations that appended to key 25:

```
{:index 23893, :time 33382826381, :type :invoke, :process 75, :f :txn, :value [[:append 25 399] [:r 11 nil]]}
{:index 23984, :time 34369050704, :type :invoke, :process 34, :f :txn, :value [[:r 25 nil] [:append 25 441]]}
{:index 24120, :time 34544470306, :type :info, :process 75, :f :txn, :value [[:append 25 399] [:r 11 nil]], :error [:unexpected-result "Status{code = INTERNAL_ERROR(code=400030), issues = [#1 Internal error while executing transaction. (S_FATAL)\n  #2035 (yexception) ydb/library/actors/core/log.cpp:881: verification=false;fline=kqp_write_actor.cpp:2200;unknown message=269550081; (S_ERROR), #2019 Query invalidated on scheme/internal error during Data execution (S_ERROR)]}"]}
{:index 24200, :time 34679068745, :type :ok, :process 34, :f :txn, :value [[:r 25 [18 19 51 80 106 147 192 214 253 293 313 337 368 383 374 426]] [:append 25 441]], :modified-txn [[:batch nil [[:r 25 nil] [:append 25 441]]]]}
```

The internal error suggests that `:append 25 399` migrated while shard was restarting. The tx which appended 441 first performed a read and did not observe 399, so only one of them should have successfully committed. However since other transactions observed 399 it suggests both transactions somehow committed, writing to the same index. I'm not sure how that's possible, could be some weird interaction between already migrated and committing transactions?

snaury:
Note: event `269550081` is `0x10110201` which is `TEvDataShard::TEvProposeTransactionRestart`: https://github.com/ydb-platform/ydb/blob/e0220730c8a0b82b2be53386ef4083ddacb139a5/ydb/core/tx/datashard/datashard.h#L207

PR #32779, opened 2026-01-26
DataShard: restore uncommitted persistent state on migration

snaury:
### Changelog entry 

Fix a rare serializable isolation issue during graceful shard migration. Fixes #31961.

### Changelog category 

* Bugfix 

### Description for reviewers 

Jepsen observed that serializable isolation could sometimes be violated during graceful shard migration between nodes. The issue was that when transactions try to commit breaking conflicting persistent locks, their commit may ultimately fail due to storage failure or races. Such broken locks no longer participate in acquiring new conflicts, however when migrated their persistent state won out, effectively resurrecting them without all necessary conflict edges.

We now compare loaded persistent state and migrated state, and remove/break persistent locks when their state doesn't match, attempting to replay changes which failed to commit. This makes sure restored in-memory locks and persistent locks are consistent when migration succeeds. In cases when migration fails persistent locks may restore to their pre-remove or pre-break state, however the full state is then restored from disk and remains consistent.

Related to #27123.

### DIFF
--- a/ydb/core/tx/datashard/datashard__init.cpp
+++ b/ydb/core/tx/datashard/datashard__init.cpp
@@ -44,7 +44,7 @@ public:
 private:
     THashMap<ui64, TOperation::TPtr> MigratedTxs;
     bool InMemoryStateActorStarted = false;
-    bool Restarted = false;
+    bool Rescheduled = false;
 };
 
 bool TDataShard::TTxInit::Execute(TTransactionContext& txc, const TActorContext& ctx) {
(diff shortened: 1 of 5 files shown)


ydb/core/tx/datashard/datashard_ut_read_iterator.cpp
```diff
@@ -5534,17 +5534,294 @@ Y_UNIT_TEST_SUITE(DataShardReadIteratorConsistency) {
...
+        auto readFuture = KqpSimpleBeginSend(runtime, session3, R"(
+                SELECT key, index, value FROM `/Root/table` WHERE key = 1 ORDER BY key, index;
+            )");
+        runtime.WaitFor("blocked read", [&]{ return blockedReads.size() >= 1; });
+        blockedReads.Stop();
+
+        // Make sure everything settles down
+        runtime.SimulateSleep(TDuration::Seconds(1));
+
+        // Block commits and try to break the first transaction by a blind upsert
```
ztlpn: Тут кажется комментарий неправильный, upsert-а же нет.

ydb/core/tx/locks/locks.h
```diff
@@ -966,6 +973,7 @@ class TSysLocks {
     bool Load(ILocksDb& db);
 
     void RestoreInMemoryLocks(THashMap<ui64, ILocksDb::TLockRow>&& rows);
+    bool RestorePersistentState(ILocksDb* db);
```
ztlpn, line 990: не хватает комментария про то, что значит возвращаемое значение

ydb/core/tx/datashard/datashard__init.cpp
```diff
@@ -101,6 +102,14 @@ void TDataShard::OnInMemoryStateRestored(THashMap<ui64, TOperation::TPtr> migrat
 bool TDataShard::TTxInitRestored::Execute(TTransactionContext& txc, const TActorContext& ctx) {
     LOG_DEBUG(ctx, NKikimrServices::TX_DATASHARD, "TDataShard::TTxInitRestored::Execute");
 
+    TDataShardLocksDb locksDb(*Self, txc);
+    if (Self->SysLocks.RestorePersistentState(&locksDb)) {
+        LOG_DEBUG(ctx, NKikimrServices::TX_DATASHARD, "TDataShard::TTxInitRestored::Execute: persistent lock state updated, restarting");
```
ztlpn: здесь имеется в виду рестартуем транзакцию? может тогда "restarting tx"?

ztlpn (approved):

PR #32873, opened 2026-01-28
Restore test ydb/core/tx/replication/service/ut_s3_writer from death

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

I accidentally noticed that the expression `IF (!OS_WINDOWS)` in ya.make works absolutely not as we expect from it. As a result, we didn't have `ydb/core/tx/replication/service/s3_writer.cpp` and its test at all.

### DIFF
--- a/ydb/core/tx/replication/service/s3_writer_ut.cpp
+++ b/ydb/core/tx/replication/service/s3_writer_ut.cpp
@@ -7,17 +7,15 @@
 #include <ydb/core/wrappers/ut_helpers/s3_mock.h>
 #include <ydb/core/wrappers/s3_wrapper.h>
 #include <ydb/core/wrappers/s3_storage_config.h>
+#include <ydb/core/util/aws.h>
 
 #include <library/cpp/string_utils/base64/base64.h>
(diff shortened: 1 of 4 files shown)


ydb/core/tx/replication/service/s3_writer_ut.cpp
```diff
@@ -13,12 +13,38 @@
...
 #include <util/string/printf.h>
 
+#include <contrib/libs/aws-sdk-cpp/aws-cpp-sdk-core/include/aws/core/Aws.h>
+#include <contrib/libs/curl/include/curl/curl.h>
+
+struct TAwsApiGuard {
+    Aws::SDKOptions Options;
+
+    TAwsApiGuard() {
+        InitAwsAPI();
```
CyberROFL, line 21: https://github.com/ydb-platform/ydb/blob/main/ydb/core/util/aws.h

CyberROFL (changes requested):

CyberROFL (approved):

PR #32897, opened 2026-01-28
Persistent buffer chunk allocation

Evgenik2:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


ydb/core/blobstorage/ddisk/ddisk_actor_chunks.cpp
```diff
@@ -13,6 +13,16 @@ namespace NKikimr::NDDisk {
         }
     }
 
+    void TDDiskActor::IssuePersistentBufferChunkAllocation() {
```
alexvru, line 11: Looks like copy-paste. And it gives priority to persistent buffer chunk allocation. Not that I object, but code may be reused. For example, std::queue<std::variant<TChunkForPersistentBuffer, TChunkForData>> ChunkAllocateQueue, or something like that.

alexvru (approved):

PR #32925, opened 2026-01-28
[NBS2]: introduce flush/erase operations with persistent buffers; read from ddisks; add cookie to responses

BarkovBG:
introduce flush/erase operations with persistent buffers & read from ddisks


ydb/core/nbs/cloud/blockstore/libs/storage/direct_block_group/direct_block_group.cpp
```diff
@@ -154,15 +166,29 @@ void TDirectBlockGroup::HandlePersistentBufferWriteResult(
...
+                RequestBlockFlush(
+                    ctx,
+                    request,
+                    false // isErase
+                );
+            } else {
+                RequestBlockFlush(
+                    ctx,
+                    request,
+                    true // isErase
```
vazhem, line 185: we should send erase in handler of flush response

vazhem (approved):

PR #33015, opened 2026-01-29
Fix bug in VDisk related to huge blob compaction when changing size limit and improve unit test

alexvru:
### Changelog entry 

Fix bug in VDisk related to huge blob compaction when changing size limit and improve unit test

### Changelog category 

### Description for reviewers 

This fixes very rare case when blob has been written as inplace one and still in fresh, while huge blob size suddenly changes and fresh compaction writes this blob into a huge slot.


mregrock (dismissed):
LGTM

kruall (dismissed):

kruall (approved):

PR #33022, opened 2026-01-29
[CBO] Improvements to join handling

lll-phill-lll:
### Changelog entry 

Fixes to CBO:
 - Commutative joins are handled correctly under shuffle elimination
 - Cost function for joins uses ByteSize instead of Nrows
 
### Changelog category 

### Description for reviewers 

...


alexpaniman:
Since #35116 was merged I think we can close this PR.

PR #33286, opened 2026-02-03
Fixed sasl auth compatibility issues

n00bcracker:
It turned out that during ydb slice update local users can't login because SchemeShard works on previous version.
This PR fixes that behavior.
Instead of checking absent database in CredentialProvider have to check empty hash params of a user.

This change resulted in failures of audit login tests so these tests were migrated to higher level test framework allowing to use grpc proxy.


ydb/core/security/ut_common.h
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include <ydb/core/testlib/test_client.h>
+
+#include <ydb/library/testlib/service_mocks/ldap_mock/ldap_defines.h>
+#include <ydb/library/testlib/service_mocks/ldap_mock/simple_server.h>
```
ijon, line 6: Вместо этих include'ов могли бы быть forward declarations.

ijon (approved):

ydb/core/security/ut_common.h
n00bcracker: Но это тестовый фреймворк, поэтому не особо важно, в бинарник это не попадет.

PR #33369, opened 2026-02-04
Fix key rotation when there is no pdisk encryption

eivanov89:
### Changelog entry 

...

### Changelog category 

* Bugfix 

### Description for reviewers 

...


the-ancient-1 (approved):
LGTM

PR #33521, opened 2026-02-05
[EXT-1928] Fix pdisk scheduler config is not applied

va-kuznecov:
### Changelog entry 

This PR fixes the bug that even if PDisk's schduler configs was changed, the change is not passed to the scheduler itself. The fix removes temporal variable and uses proper settings from TPDisk::Cfg object.

### Changelog category 

* Bugfix 

### Description for reviewers 

...


SammyVimes (approved):
LGTM!

PR #33710, opened 2026-02-09
Count tablet blobs in dstool

alexvru:
### Changelog entry 

Count tablet blobs in dstool

### Changelog category 

### Description for reviewers 

This patch allows counting blob counts per tablet along with sizes.


mregrock (approved):
LGTM

PR #33940, opened 2026-02-11
[ICRDMA] Bypass cpu cache during copy rdma event into regular memory.

dcherednik:
### Changelog entry 

The rdma event will be used in different thread, so no need to keep the data in cache.

### Changelog category 

### Description for reviewers 

...


dcherednik:
it allows to save up to 30% cpu in case of heavy load (compare first and last graph)
image

alexvru (approved):

PR #33951, opened 2026-02-11
Add UT for retro_tracing impl in Kikimr

serbel324:
### Changelog entry 

Add UT for DSProxy and Backpressure retro-spans

### Changelog category 


mregrock (approved):
LGTM

PR #34093, opened 2026-02-13
Propose new interface for DDisk claims through BSC

alexvru:
### Changelog entry 

Propose new interface for DDisk claims through BSC

### Changelog category 

### Description for reviewers 

This patch contains new interface for BSC-managed claims over DDisks and Persistent Buffers.


mregrock (approved):
LGTM

PR #34248, opened 2026-02-16
Workload fulltext

ghost:
### Changelog entry 

Add workload for fulltext search benchmarks

### Changelog category 

### Description for reviewers 

The `workload fulltext` command is hidden because it may be subject to changes.
It is meaningful to merge as hidden command now, because it may be useful to developers of fulltext search to validate their changes.

closes #33832


ydb/library/workload/fulltext/fulltext_workload_params.cpp
pnv1: Кажется это то, что во всех workload'ах называется --path

pnv1 (approved):

PR #34290, opened 2026-02-16
NBYDB-1939 Add MaxInFlightCount metric to vdisk SkeletonFront

usernamedt:
### Changelog entry 

This PR adds the new `MaxInFlightCount` metric to vdisk `SkeletonFront` monitoring to get more insights about the current queues state.  

### Changelog category 

* New feature

### Description for reviewers 

...

### DIFF
--- a/ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
+++ b/ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
@@ -116,7 +116,7 @@ namespace NKikimr {
                          enableThrottlingReport ? std::make_optional(OverloadHandler ? OverloadHandler->IsThrottling() : false) : std::nullopt,
                          enableThrottlingReport ? std::make_optional(OverloadHandler ? OverloadHandler->GetThrottlingRate() : 0) : std::nullopt));
             // repeat later
-            ctx.Schedule(Config->WhiteboardUpdateInterval, new TEvTimeToUpdateWhiteboard());
+            ctx.Schedule(Config->StatsUpdateInterval, new TEvTimeToUpdateStats());
         }
 
         ////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeletonfront.cpp
```diff
@@ -1139,6 +1151,8 @@ namespace NKikimr {
                                IntQueueLowGets.get(), IntQueueLogPuts.get(), IntQueueHugePutsForeground.get(),
                                IntQueueHugePutsBackground.get()}) {
                 light = Max(light, queue->GetCumulativeLight());
+
+                queue->RefreshStatistics();
```
usernamedt: This is probably not good enough  place to call `RefreshStatistics()`. Probably, we need to introduce the separate event like `EvTimeToRefreshStatistics` to update the vdisk stats in the future iterations.

ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
```diff
@@ -2924,7 +2924,7 @@ namespace NKikimr {
             HFunc(TEvTakeHullSnapshot, Handle)
             HFunc(NMon::TEvHttpInfo, Handle)
             HFunc(TEvVDiskStatRequest, Handle)
-            CFunc(TEvBlobStorage::EvTimeToUpdateWhiteboard, UpdateWhiteboard)
+            CFunc(TEvBlobStorage::EvTimeToUpdateStats, UpdateWhiteboard)
```
usernamedt, line 2927: Since we're not changing the skeleton logic, I left it as it is. But I can change here as well for consistency.

ulya-sidorina (dismissed):

ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeletonfront.cpp
```diff
@@ -1118,11 +1129,27 @@ namespace NKikimr {
             return str.Str();
         }
 
+        void UpdateStats(const TActorContext &ctx, bool schedule = true) {
```
va-kuznecov: in the PR it looks like schedule is always true. There is only once case with false (call from TEvReportScrubStatus handler) and now it's covered by explicit call to UpdateWhiteboard

```suggestion
        void UpdateStats(const TActorContext &ctx) {
```

va-kuznecov (changes requested):

va-kuznecov (approved):

PR #34395, opened 2026-02-18
Make sure truncate garbage collects removed data as soon as possible

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

When table is truncated all blobs are added to the `GcDelta.Deleted` (`GcLeft` in the commit proto) list in the same commit, so the data is eventually deleted. However, this list is lazily applied to storage, and often required at least two new log snapshots (600-1000 subsequent commits) before this data is marked for deletion.

This PR changes the logic in the following ways:

* LocalDB will make sure a new log snapshot is committed after any table is truncated (a new snapshot is also generated when restarting)
* Sys tablet will now make an additional confirmation log entry when there is a committed, but not yet persistently confirmed snapshot, and the commit queue is empty. This will make sure that eventually (either by subsequent commits, or by this confirmation log entry) the latest snapshot will become the guaranteed commit graph head that doesn't depend on earlier commits. These confirmation log entries have already been used extensively to confirm commit graph tail for followers.
* When the snapshot is confirmed in that way sys tablet notifies the executor using the new `TEvSnapshotConfirmed` event, which executor uses to garbage collect everything before the latest log snapshot.

Fixes #34367.

### DIFF
--- a/ydb/core/tablet/tablet_sys.cpp
+++ b/ydb/core/tablet/tablet_sys.cpp
@@ -225,6 +225,7 @@ void TTablet::WriteZeroEntry(TEvTablet::TDependencyGraph *graph) {
         graph->Entries.erase(graph->Entries.begin(), snapIterator); // erase head of graph
 
     Graph.Snapshot = snapshot;
+    Graph.SnapshotSource = {};
 
     const TLogoBlobID logid(TabletID(), StateStorageInfo.KnownGeneration, 0, 0, 0, 0);
     TVector<TEvTablet::TLogEntryReference> refs;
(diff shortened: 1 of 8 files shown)


ydb/core/tablet/tablet_sys.cpp
```diff
@@ -1644,7 +1647,19 @@ void TTablet::ProgressFollowerQueue() {
...
+        // We must not have another sync commit inflight
+        Graph.SyncCommit.SyncStep == 0 &&
+        (
+            // And either there are pending follower updates waiting for confirmation
+            Graph.PostponedFollowerUpdates ||
+            // Or the latest snapshot wasn't confirmed by the last commit
+            Graph.Snapshot > std::make_pair(StateStorageInfo.KnownGeneration, Graph.ConfirmedCommited)
+        ));
+
+    if (needSyncCommit) {
```
lex007in, line 1665: Похоже, теперь эта логика с SyncCommit не только фолловров касается, и имеет смысл вынести из ProgressFollowerQueue в отдельную функцию.

lex007in (dismissed):

lex007in (approved):

PR #34572, opened 2026-02-20
IgnoredTableRegexes for TLI logs

azevaykin:
### Changelog entry 

Ignoring using a regular expression over the table name in TLI logging.

```text
message TTliConfig {
    repeated string IgnoredTableRegexes = 1;
}
```
If a query references multiple tables, one can be excluded from logging while the other one will still be logged.

#### For example:
```
IgnoredTableRegexes = Table1
```
Queries:

```sql
SELECT t1.Value, t2.Value, t3.Value
FROM `/Root/Tenant1/Table1` AS t1
JOIN `/Root/Tenant1/Table2` AS t2 ON t1.Key = t2.Key
WHERE t1.Key = 1u;

UPDATE `/Root/Tenant1/Table2` SET Value = "VictimUpdate" WHERE Key = 1u;
```
or

```sql
SELECT * FROM Table1 WHERE Key = 1u;
SELECT * FROM Table2 WHERE Key = 1u;
UPDATE Table2 SET Value = "Victim2Update" WHERE Key = 1u;
```

In TLI logs, Table1 will not be tracked - only Table2 will.

The only place where Table1 will be mentioned is in the full query texts within the logs related to Table2.

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/kqp/executer_actor/kqp_tasks_graph.h
+++ b/ydb/core/kqp/executer_actor/kqp_tasks_graph.h
@@ -258,8 +258,21 @@ struct TGraphMeta {
         return QuerySpanId;  // Fall back to global QuerySpanId
     }
 
+    ui64 GetEffectiveQuerySpanId(ui64 spanId, const TString& tablePath) const {
+        if (spanId == 0) {
+            return 0;
+        }
+        for (const auto& path : IgnoredTablePaths) {
+            if (tablePath == path) {
+                return 0;
+            }
+        }
+        return spanId;
+    }
+
     ui64 QuerySpanId = 0;
     THashMap<ui32, ui64> TxQuerySpanIds;  // Per-transaction QuerySpanIds (for deferred effects)
+    TVector<TString> IgnoredTablePaths;
 };
 
 struct TTaskInputMeta {
@@ -269,6 +282,8 @@ struct TTaskInputMeta {
     NKikimrKqp::TKqpStreamLookupSettings* StreamLookupSettings = nullptr;
     NKikimrKqp::TKqpSequencerSettings* SequencerSettings = nullptr;
     NKikimrTxDataShard::TKqpVectorResolveSettings* VectorResolveSettings = nullptr;
+    // Fully-qualified table path for TLI filtering; populated for stream lookup and vector resolve inputs.
+    TString TablePath;
 };
 
 struct TTaskOutputMeta {
--- a/ydb/core/kqp/executer_actor/kqp_tasks_graph.h
+++ b/ydb/core/kqp/executer_actor/kqp_tasks_graph.h
@@ -259,20 +259,18 @@ struct TGraphMeta {
     }
 
     ui64 GetEffectiveQuerySpanId(ui64 spanId, const TString& tablePath) const {
-        if (spanId == 0) {
-            return 0;
+        if (spanId == 0 || tablePath.empty()) {
+            return spanId;
         }
-        for (const auto& path : IgnoredTablePaths) {
-            if (tablePath == path) {
-                return 0;
-            }
+        if (IgnoredTablePaths.contains(tablePath)) {
+            return 0;
         }
         return spanId;
     }
 
     ui64 QuerySpanId = 0;
     THashMap<ui32, ui64> TxQuerySpanIds;  // Per-transaction QuerySpanIds (for deferred effects)
-    TVector<TString> IgnoredTablePaths;
+    THashSet<TString> IgnoredTablePaths;
 };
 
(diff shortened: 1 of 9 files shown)


nikvas0 (dismissed):

nikvas0 (dismissed):
kqp ok

ydb/core/cms/console/tli_config_configurator.cpp
```diff
@@ -0,0 +1,46 @@
...
+            Send(MakeConfigsDispatcherID(SelfId().NodeId()),
+                    new TEvConfigsDispatcher::TEvSetConfigSubscriptionRequest(item));
+        }
+
+        void Handle(TEvConsole::TEvConfigNotificationRequest::TPtr& ev) {
+            const auto& record = ev->Get()->Record;
+            LOG_INFO_S(*TlsActivationContext, NKikimrServices::CMS_CONFIGS,
+                "TTliConfigConfigurator: new config: " << record.GetConfig().ShortDebugString());
+
+            AppDataVerified().TliConfig.CopyFrom(record.GetConfig().GetTliConfig());
```
Enjection: We can't just dynamically update AppData at runtime. The only place where this is done is FeatureFlags and icb, AFAIK, and it's done the right way (e.g., atomics and so on). We could agree that this part of the config behaves the same way as well, but:
1) I would prefer at least an explicit comment in the .proto file.
2) CopyFrom is not thread-safe, so we need some approach that makes it safe for our multithreaded environment.

---

Alternatively, I can offer another way -- instead of creating new entities, just reuse ICB with the already existing controls. However, it doesn't support arrays, at least for now.

Enjection (changes requested):

ydb/core/cms/console/tli_config_configurator.cpp
azevaykin: I reverted last commit with dynamically update. Let's do it in a following PR

azevaykin: What about moving TliConfig handling from AppData (with thread-unsafe CopyFrom) to the KQP Proxy passthrough pattern? 
The proxy subscribes to TliConfigItem, stores config locally, and passes it to sessions/executors at creation time (same pattern as TableServiceConfig).

Enjection (dismissed):

ydb/core/kqp/executer_actor/kqp_tasks_graph.h
```diff
@@ -258,8 +259,22 @@ struct TGraphMeta {
...
     }
 
+    ui64 GetEffectiveQuerySpanId(ui64 spanId, const TString& tablePath) const {
+        if (spanId == 0 || tablePath.empty()) {
+            return spanId;
+        }
+        for (const auto& pattern : IgnoredTableRegexes) {
+            std::regex regexPattern(pattern.c_str());
+            if (std::regex_match(tablePath.c_str(), regexPattern)) {
+                return 0;
```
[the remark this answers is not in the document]
azevaykin: 11d4a2b9380c6f00dac6a312b7323b70c480510b

ydb/core/kqp/executer_actor/kqp_tasks_graph.h
```diff
@@ -258,8 +259,22 @@ struct TGraphMeta {
...
+            }
+        }
```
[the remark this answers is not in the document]
azevaykin: 11d4a2b9380c6f00dac6a312b7323b70c480510b

ydb/core/kqp/executer_actor/kqp_tasks_graph.h
```diff
@@ -258,8 +259,22 @@ struct TGraphMeta {
         return QuerySpanId;  // Fall back to global QuerySpanId
     }
 
+    ui64 GetEffectiveQuerySpanId(ui64 spanId, const TString& tablePath) const {
+        if (spanId == 0 || tablePath.empty()) {
+            return spanId;
+        }
+        for (const auto& pattern : IgnoredTableRegexes) {
+            std::regex regexPattern(pattern.c_str());
+            if (std::regex_match(tablePath.c_str(), regexPattern)) {
```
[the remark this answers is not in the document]
azevaykin: 11d4a2b9380c6f00dac6a312b7323b70c480510b

nikvas0 (approved):
kqp ok

Enjection (approved):

ydb/core/cms/console/tli_config_configurator.cpp
azevaykin: https://github.com/ydb-platform/ydb/pull/34818

PR #34779, opened 2026-02-24
[ICRDMA] Fix build on arm

dcherednik:
### Changelog entry 

Fix missed compilation architecture check 

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
+++ b/ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
@@ -9,6 +9,14 @@
 
 #include <contrib/restricted/abseil-cpp-tstring/y_absl/crc/internal/non_temporal_memcpy.h>
 
+Y_FORCE_INLINE void MemcpyNoCache(void* dst, const void* src, size_t len) {
+#if defined(__x86_64__)
+    y_absl::crc_internal::non_temporal_store_memcpy_avx(dst, src, len);
+#else
+    y_absl::crc_internal::non_temporal_store_memcpy(dst, src, len);
+#endif
+}
+
 namespace NActors {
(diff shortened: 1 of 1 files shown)


ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
```diff
@@ -9,6 +9,14 @@
 
 #include <contrib/restricted/abseil-cpp-tstring/y_absl/crc/internal/non_temporal_memcpy.h>
 
+Y_FORCE_INLINE void MemcpyNoCache(void* dst, const void* src, size_t len) {
+#if defined(__x86_64__)
+    y_absl::crc_internal::non_temporal_store_memcpy_avx(dst, src, len);
```
eivanov89, line 14: I'm not sure if we can rely that we always have AVX. I understand that it's silly, but in cypher part we detect avx presence in runtime and only then we use it. I see that avx func introduced here with another rdma commit. Let's discuss it.

eivanov89 (changes requested):

va-kuznecov (approved):

ydb/library/actors/interconnect/interconnect_tcp_input_session.cpp
dcherednik: It is used only in RDMA path. I can write runtime dispatching but I don't thunk someone has RoCE card for 1k$ and cpu without AVX support.

dcherednik: I prefer to fix the compilation issue now and make ticket for runtime dispatching and even maybe move non_temporal memcpy logic into RcBuf infrastructure if it will be clear we have profit in similar pattern

eivanov89 (approved):
LGTM;

PR #35268, opened 2026-03-02
NBYDB-1930: Configure pdisk op size through configs

va-kuznecov:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
+++ b/ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
@@ -62,9 +62,8 @@ TPDisk::TPDisk(std::shared_ptr<TPDiskCtx> pCtx, const TIntrusivePtr<TPDiskConfig
             0, 64 * 1024 * 1024);
     ForsetiMaxLogBatchNs = TControlWrapper((PDiskCategory.IsSolidState() ? 50'000ll : 500'000ll), 0, 100'000'000ll);
     ForsetiMaxLogBatchNsCached = ForsetiMaxLogBatchNs;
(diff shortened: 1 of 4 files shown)


ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
```diff
@@ -2963,8 +2962,7 @@ bool TPDisk::Initialize() {
             REGISTER_LOCAL_CONTROL(ForsetiMinLogCostNsControl);
             REGISTER_LOCAL_CONTROL(ForsetiMilliBatchSize);
             REGISTER_LOCAL_CONTROL(ForsetiMaxLogBatchNs);
-            REGISTER_LOCAL_CONTROL(ForsetiOpPieceSizeSsd);
-            REGISTER_LOCAL_CONTROL(ForsetiOpPieceSizeRot);
+            REGISTER_LOCAL_CONTROL(ForsetiOpPieceSize);
```
SammyVimes, line 2965: Is it ok to have one for both types of drives?

va-kuznecov: Yes, I did it on purpose. For example, ForsetiMaxLogBatchNs and almost all Forseti* settings are different for different device kinds. And only OpPieceSize, I don't know what for, has Hdd and Ssd settings. It leads to a 'dangling' setting - if PDisk_1001 works on hdd device, then there is a PDisk_1001_ForsetiOpPieceSizeSsd setting that does nothing.

SammyVimes (approved):
LGTM

PR #35342, opened 2026-03-03
EXT-2203 BS requests sources

SammyVimes:
### Changelog entry 
Adds new pdisk metrics: for each possible blobstorage write request source count requests count and bytes count.

### Changelog category 

* Improvement

### Description for reviewers 

In `blobstorage_write_source.h` I generate a enum that is used everywhere blobstorage write requests are being sent (I consider `EvBlock` and `EvCollectGarbage` to be write requests too here). I store enum value in write request's proto (in an arbitrary `uint32` field) and pass it all the way down to the pdisk, so that pdisk mon would count requests and bytes. Note that by default these new metrics are PRIVATE (because there are a lot of them), but they can be enabled using a feature flag `extended_pdisk_sensors`

### DIFF
--- a/ydb/core/base/blobstorage.h
+++ b/ydb/core/base/blobstorage.h
@@ -3,6 +3,7 @@
 
 #include "blobstorage_pdisk_category.h"
 #include "blobstorage_relevance.h"
+#include "blobstorage_write_source.h"
 #include "boot_type.h"
 #include "events.h"
 #include "tablet_types.h"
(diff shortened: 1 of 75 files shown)


lex007in (dismissed):
datashard (tablet/table_flat) ok

dorooleg (dismissed):

dorooleg:
cs ok, остальное не смотрел

va-kuznecov (approved):

lex007in (approved):

dorooleg (approved):

Issue #25253, opened 2025-09-18
DataShard: add simplified PESSIMISTIC_EXCLUSIVE lock mode


snaury:
This would be the very first stepping stone towards implementing pessimistic locks. Some key takeaways:

* Only TEvWrite would be supported
* It would only be possible to add pessimistic locks in immediate mode, LockId, LockNodeId and LockTimestamp would be required in such a mode.
* Setting pessimistic locks in non-immediate modes doesn't make sense. Effects are applied at some unknown future timestamp and are atomic from the datashard's perspective. There would be nothing to wait for and nothing to block.
* However, when other non-locking operations want to write to exclusively locked keys they would need to wait until such keys are unlocked. Similarly exclusive lock cannot be installed while such operations are waiting and/or inflight.
* A special case when transaction sets pessimistic locks and then commits with additional side-effects. These side-effects may intersect with other locks and propose would need to wait, except it shouldn't block on its own locks. When waiting starts we might need to report new edges using the committing lock id in the future.

We would need a data structure for arranging these locks. The first implementation may be naive just to get a feel for how it's supposed to work, in the future it should be some kind of interval tree with transitive dependencies so all operations are on the order of the number of ranges in the transaction and logarithmic in the total number of ranges inflight.

PR #35378, opened 2026-03-03
Add an optional timestamp for locks in long tx service

snaury:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Since long tx service will be ultimately responsible for deadlock detection and choosing youngest transactions to abort, it makes sense to choose lock timestamp once when registering the lock, and forwarding this timestamp when subscribing to locks from other nodes.

This PR adds a lockTimestamp parameter to `TLockHandle` and forwarding it between nodes in `TEvLockStatus` messages. Related to #25253.

### DIFF
--- a/ydb/core/protos/long_tx_service.proto
+++ b/ydb/core/protos/long_tx_service.proto
@@ -91,6 +91,7 @@ message TEvLockStatus {
     optional uint64 LockId = 1;
     optional uint32 LockNode = 2;
     optional EStatus Status = 3;
+    optional uint64 LockTimestamp = 4;
 }
 
 message TEvUnsubscribeLock {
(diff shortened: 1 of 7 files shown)


ydb/core/protos/long_tx_service.proto
```diff
@@ -91,6 +91,7 @@ message TEvLockStatus {
     optional uint64 LockId = 1;
     optional uint32 LockNode = 2;
     optional EStatus Status = 3;
+    optional uint64 LockTimestamp = 4;
```
ztlpn: Это микросекунды со времени эпохи? Может в имени или в комментарии обозначить, хотя бы какая тут единица измерения?

snaury: Поправил!

ztlpn (approved):

PR #35559, opened 2026-03-06
interconnect: Use TCP_USER_TIMEOUT + SO_KEEPALIVE to check node availability without user level pings. NBYDB-1766

dcherednik:
### Changelog entry 

Enable kernel liveness as an optional per-session mode negotiated in handshake. Session enters kernel mode only when both sides requested it and both sockets successfully applied SO_KEEPALIVE/TCP_KEEP* plus mandatory TCP_USER_TIMEOUT; otherwise connection falls back to legacy user-space liveness without handshake failure.

When negotiated:
- disable user-space dead-peer watchdog in input/session paths
- preserve flow-control confirms and undelivered semantics
- keep clock-skew/ping metrics via flush-scheduled periodic pings (ClockSkewPingTimeout)

Test coverage:
- interconnect UT negotiation/fallback/idle-packet reduction/ACK/clock-skew with and without RDMA
- dedicated tcp_user_timeout blackhole tests verifying disconnect and undelivered invariants

### Changelog category 

* New feature

### Description for reviewers 

...

### DIFF
--- a/ydb/core/driver_lib/run/kikimr_services_initializers.cpp
+++ b/ydb/core/driver_lib/run/kikimr_services_initializers.cpp
@@ -498,6 +498,31 @@ static TInterconnectSettings GetInterconnectSettings(const NKikimrConfig::TInter
     }
 
     result.EnableExternalDataChannel = config.GetEnableExternalDataChannel();
+    result.EnableKernelLiveness = false;
+    if (config.GetUseKernelKeepAlive()) {
+        result.EnableKernelLiveness = true;
+
+        result.KernelUserTimeout = result.DeadPeer != TDuration::Zero()
+            ? result.DeadPeer
+            : TDuration::Seconds(10);
+        // Keep kernel liveness approximately aligned to KernelUserTimeout while keeping the setup simple.
+        // Target budget on idle links is:
+        //   KernelKeepAliveIdle + KernelKeepAliveInterval * KernelKeepAliveProbes ~= KernelUserTimeout.
+        // Interval is derived from timeout and probes are fixed for predictable behavior.
+        result.KernelKeepAliveInterval = Max(result.KernelUserTimeout / 10, TDuration::Seconds(1));
+        constexpr ui32 keepAliveProbes = 5;
+        result.KernelKeepAliveProbes = keepAliveProbes;
+
+        const ui64 userTimeoutMs = result.KernelUserTimeout.MilliSeconds();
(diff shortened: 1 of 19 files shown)


dcherednik:
This reduce number of syscalls for idle sessions
image

ydb/core/driver_lib/run/kikimr_services_initializers.cpp
```diff
@@ -507,6 +498,31 @@ static TInterconnectSettings GetInterconnectSettings(const NKikimrConfig::TInter
     }
 
     result.EnableExternalDataChannel = config.GetEnableExternalDataChannel();
+    result.EnableKernelLiveness = false;
+    if (config.GetUseKernelKeepAlive()) {
+        result.EnableKernelLiveness = true;
+
+        result.KernelUserTimeout = result.DeadPeer != TDuration::Zero()
+            ? result.DeadPeer
+            : TDuration::Seconds(10);
```
alexvru: Magic number?

ydb/library/actors/protos/interconnect.proto
```diff
@@ -107,6 +107,7 @@ message THandshakeRequest {
     optional bool RequestExternalDataChannel = 21;
     optional bool RequestXxhash = 24;
     optional bool RequestXdcShuffle = 25;
+    optional bool RequestKernelLiveness = 28;
```
alexvru: Why do we want to negotiate this over network? As far as I'm concerned, this is local-only setting. When we use kernel aliveness, we just do not generate ping requests, but peer can issue pings and we respond to them.

This way makes implementation unnecessarily complex.

alexvru (changes requested):

ydb/core/driver_lib/run/kikimr_services_initializers.cpp
dcherednik: Fixed. Used constant NActors::DEFAULT_DEADPEER_TIMEOUT

ydb/library/actors/protos/interconnect.proto
dcherednik: I have removed handshake negotiation part. But it wasn't made it mush simpler. 
Probably the most hard part which we still need is:

TSessionParams inputSessionParams = Params;
inputSessionParams.UseKernelLiveness = KernelLivenessMode;

that's code inside tcp_session

the problem that result of applying socket option may be different for initial handshake (which create session), and for resume handshake (which is part of transparent connection reestablish).

alexvru (approved):

PR #35658, opened 2026-03-10
Default fail-fast mode for IC

alexvru:
### Changelog entry 

Default fail-fast mode for IC

### Changelog category 

### Description for reviewers 

This patch prevents IC messages from stalling when they arrive during reconnection after unsuccessful attempt.


mregrock (approved):
LGTM

PR #36217, opened 2026-03-19
Fix cross-database quota interference in Kesus quoter proxy

Enjection:
When two databases share the same Kesus quoter tree (same cloud), one database hitting its RU quota causes the other's partitions to experience write throttling.

Front-policy burst-starvation. The proxy reported Available (a bucket level snapshot) as a channel rate using Front policy with ticks=2. After consumption, rate dropped to 0 for ~98ms until the next Kesus allocation, causing partition write bursts.

Fix: Dual-channel architecture. Channel 0 (Sustained, long-lived) provides steady flow at configured speed. Channel 1 (Front, short- lived) provides immediate burst of Available. The steady channel prevents starvation gaps between Kesus allocations.

### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/quoter/kesus_quoter_proxy.cpp
+++ b/ydb/core/quoter/kesus_quoter_proxy.cpp
@@ -2,8 +2,10 @@
 #include "quoter_service_impl.h"
 #include "debug_info.h"
 
+#include <ydb/core/base/appdata.h>
 #include <ydb/core/base/counters.h>
 #include <ydb/core/base/path.h>
+#include <ydb/core/control/lib/immediate_control_board_impl.h>
 #include <ydb/core/kesus/tablet/events.h>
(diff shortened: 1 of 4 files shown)


CyberROFL (commented):
Голосом договорились добавить фича-флаг.

ydb/core/quoter/kesus_quoter_proxy.cpp
CyberROFL, line 288: А это изменение-то не под флагом. На что это повлияет?

Enjection: Если ты про LastAllocAmount, то оно только в v1 используется

CyberROFL (approved):

PR #36361, opened 2026-03-20
Use sync SchemeCache request to get users hash init params

n00bcracker:
Occasionally login operations could fail since hash init params of user password haven't been updated at a dynamic node.
To avoid this behavior:
1. Remove TStaticCredentialProvider;
2. Store users hash init params in SchemeCache;
3. Retrieve users hash init params from SchemeCache in auth actors with the sync flag.


ydb/core/security/sasl/base_auth_actors.h
```diff
@@ -43,6 +43,9 @@ class TAuthActorBase : public TActorBootstrapped<TAuthActorBase> {
 protected:
     void ResolveSchemeShard(const TActorContext &ctx);
     virtual NKikimrScheme::TEvLogin CreateLoginRequest() const = 0;
+    void SendLoginRequest();
+    virtual void ProceedAuthentication(const TActorContext &ctx,
```
ijon: `ProceedWithAuthentication` or `ContinueAuthentication` or `ProcessAuthentication`.

(you "proceed with" something, not "proceed something")

ijon (approved):

PR #37053, opened 2026-04-01
Fix IsIPv4 and IsIPv6 functions

### LINKED ISSUE #37129: Refine NHttp::CrackAddress function
In https://github.com/ydb-platform/ydb/pull/37019 basic fixes were made to solve problem. Now you need to completely refine this function.

1. NHttp namespace is not suitable place for this function because it became wide used. Maybe you should move it somewhere to util or library folder.
2. Do not receive return values as reference input. it's better to return TMaybe<...> and get const TString& ... only as input.
3. There are many places where this function is used. You should check all such places and start to use your new function. You can also find another function implementation here https://github.com/ydb-platform/ydb/blob/main/ydb/public/lib/deprecated/client/msgbus_client.cpp#L7. Chech what it is and change to new implementation.
4. Fix the contract through tests.

NewBediver:
### Changelog category 

### Description for reviewers 

Use default functions to check ipv4 and ipv6 addresses

### DIFF
--- a/ydb/library/actors/http/http_proxy.cpp
+++ b/ydb/library/actors/http/http_proxy.cpp
@@ -1,7 +1,11 @@
-#include <ydb/library/actors/core/events.h>
-#include <library/cpp/monlib/metrics/metric_registry.h>
-#include <cctype>
 #include "http_proxy.h"
+
+#include <ydb/library/actors/core/events.h>
+
+#include <library/cpp/monlib/metrics/metric_registry.h>
+#include <library/cpp/ipv6_address/ipv6_address.h>
+
+#include <cctype>
 
 namespace NHttp {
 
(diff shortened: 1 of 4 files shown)


alexvru (dismissed):

dcherednik (approved):

ydb/library/actors/http/http_proxy.cpp
```diff
@@ -338,23 +338,13 @@ TActorId TUrlHandler::GetHandler(const TString& url) const {
...
 bool IsIPv6(const TString& host) {
-    if (host.find_first_not_of(":0123456789abcdef") != TString::npos) {
-        return false;
-    }
-    if (std::count(host.begin(), host.end(), ':') < 2) {
-        return false;
-    }
-    return true;
+    in6_addr addr;
+    return inet_pton(AF_INET6, host.c_str(), &addr) > 0;
```
dcherednik: Probably it will not compiled on windows without some additional headers (ws2tcpip.h). It should be rechecked

ydb/library/actors/http/http_proxy.cpp
```diff
@@ -338,23 +338,13 @@ TActorId TUrlHandler::GetHandler(const TString& url) const {
...
 bool IsIPv4(const TString& host) {
-    if (host.find_first_not_of(".0123456789") != TString::npos) {
-        return false;
-    }
-    if (std::count(host.begin(), host.end(), '.') != 3) {
-        return false;
-    }
-    return true;
+    in_addr addr;
+    return inet_pton(AF_INET, host.c_str(), &addr) > 0;
```
dcherednik: Same here

dcherednik (changes requested):

ydb/library/actors/http/http_proxy.cpp
NewBediver: I used TIpv6Address from <library/cpp/ipv6_address/ipv6_address.h> instead

ydb/library/actors/http/http_proxy.cpp
NewBediver: I used TIpv6Address from <library/cpp/ipv6_address/ipv6_address.h> instead

alexvru (approved):

alexvru (approved):

adameat (approved):

PR #37362, opened 2026-04-06
Fix Coverity issue

alexvru:
### Changelog entry 

Fix Coverity issue

### Changelog category 

### Description for reviewers 

This fixes false-positive Coverity use-after-move issue.


mregrock (approved):
LGTM

PR #37369, opened 2026-04-06
Fix false NODATA response

alexvru:
### Changelog entry 

Fix false NODATA response

### Changelog category 

* Bugfix 

### Description for reviewers 

This fixes bug #36743 leading to incorrectly generated NODATA response when pile still gets synchronized.


mregrock (approved):
LGTM

PR #37403, opened 2026-04-06
Add Ydb.sensitive annotation to appropriate protobuf fields [schemeshard]

yurikiselev:
### Changelog category


ijon (approved):

ydb/core/protos/flat_scheme_op.proto
```diff
@@ -2287,7 +2279,7 @@ message TAuth {
         TBasic Basic = 5;
         TMdbBasic MdbBasic = 6;
         TAws Aws = 7;
-        TToken Token = 8;
+        TToken Token = 8 [(Ydb.sensitive) = true];
```
yumkam, line 2282: `Token` here contains token secret name (*not* sensitive), not raw token (*would be* sensitive)

PR #37489, opened 2026-04-06
Do not mute tests, that was skipped in previues run

iddqdex:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

https://github.com/ydb-platform/ydb/issues/37488

### DIFF
--- a/.github/actions/test_ya/action.yml
+++ b/.github/actions/test_ya/action.yml
@@ -383,9 +383,6 @@ runs:
         YA_MAKE_OUTPUT_URL="$PUBLIC_DIR_URL/ya_make_output.txt"
         echo "20 [Ya make output]($YA_MAKE_OUTPUT_URL)" >> $SUMMARY_LINKS
 
-        MUTED_YAML_PATH="$PUBLIC_DIR/muted_tests.yaml"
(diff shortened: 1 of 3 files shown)


naspirato:
llm предложила подстраховаться если report будет битый
### Suggestion: make `report.json` parsing best-effort (avoid breaking retries when no skipped tests / no report)

The new logic that excludes previously `SKIPPED` tests from the muted list looks good **when `PREV_REPORT` exists and contains valid JSON**.

However, `get_previously_skipped_tests()` currently does unconditional:

- `open(report_json_path)`
- `json.load(...)`

On retries this can **fail the whole job** if `$PREV_REPORT` is missing (e.g. the previous try failed before `report.json` was generated), or if the JSON is malformed.

To keep behavior safe in the “no skipped tests” case (and generally), I suggest making this parsing **best-effort**: if the report is missing/unreadable/unparseable — just return an empty set (equivalent to “no skipped tests”), and continue generating the muted YAML normally.

Proposed change:

```python
def get_previously_skipped_tests(report_json_path: str) -> Set[Tuple[str, str]]:
    result: Set[Tuple[str, str]] = set()
    if not report_json_path:
        return result

    try:
        with open(report_json_path, "r") as f:
            report = json.load(f)
    except (OSError, json.JSONDecodeError):
        return result

    for test in report.get("results", []):
        if test.get("status", "") != "SKIPPED":
            continue
        ...
    return result

naspirato (approved):

PR #37624, opened 2026-04-08
Fix Coverity 106704: freeaddrinfo on getaddrinfo error in dnsresolver

uzhastik:
Coverity CID 106704 (CLOUD-252157): on `getaddrinfo` failure the code returned without calling `freeaddrinfo` on `res`.

- Initialize `res` to `nullptr`.
- Move `Y_DEFER { freeaddrinfo(res); }` before `getaddrinfo` so cleanup runs on every return; `freeaddrinfo(nullptr)` is allowed by POSIX.

### DIFF
--- a/ydb/library/actors/dnsresolver/dnsresolver.cpp
+++ b/ydb/library/actors/dnsresolver/dnsresolver.cpp
@@ -544,17 +544,18 @@ namespace NDnsResolver {
             name = AddTrailingDot(std::move(name), Options.AddTrailingDot);
             auto result = std::make_unique<TEvDns::TEvGetHostByNameResult>();
 
-            struct addrinfo hints, *res;
+            struct addrinfo hints, *res = nullptr;
             memset(&hints, 0, sizeof(hints));
             hints.ai_family = family;
             hints.ai_socktype = Options.ForceTcp ? SOCK_STREAM : SOCK_DGRAM;
+
+            Y_DEFER { freeaddrinfo(res); };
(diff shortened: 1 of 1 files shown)


dcherednik (changes requested):
Linux man is unclear. 

BSD man explain it better.

All of the information returned by [getaddrinfo](https://man.openbsd.org/getaddrinfo.3#getaddrinfo~6)() is dynamically allocated: the addrinfo structures themselves as well as the socket address structures and the canonical host name strings included in the addrinfo structures.

Memory allocated for the dynamically allocated structures created by a successful call to [getaddrinfo](https://man.openbsd.org/getaddrinfo.3#getaddrinfo~7)() is released by the [freeaddrinfo](https://man.openbsd.org/getaddrinfo.3#freeaddrinfo)() function. The ai pointer should be an addrinfo structure created by a call to getaddrinfo().

[RETURN VALUES](https://man.openbsd.org/getaddrinfo.3#RETURN_VALUES)

getaddrinfo() returns zero on success or one of the error codes listed in [gai_strerror(3)](https://man.openbsd.org/gai_strerror.3) if an error occurs. If an error occurs, no memory is allocated by getaddrinfo(), therefore it is not necessary to release the addrinfo structure(s).

Quick investigation of other open source code
AWS library do:
if (getaddrinfo(hostName, nullptr, &hints, &res))
            {
                return false;
            }

grpc also does not call free in case of error

PR #37645, opened 2026-04-08
Fix Coverity 106823: no std::move from const TIssues& in cancel script actor

uzhastik:
Coverity CID 106823 (CLOUD-254986): `Reply` used `Issues.AddIssues(std::move(issues))` while `issues` is `const NYql::TIssues&`, which does not move and is reported as use-after-move when the function ends.

Also fix the same pattern in `ScheduleRetry` (same actor, same defect class).

Tracker CLOUD-254986 lists multiple Coverity items; this PR targets CID 106823.


uzhastik:
Duplicate of https://github.com/ydb-platform/ydb/pull/37644 — same change (remove std::move from const TIssues& in ScheduleRetry and Reply). Closing in favor of 37644.

PR #37695, opened 2026-04-08
Fix Coverity 113581: use counters response size before std::move

uzhastik:
Coverity CID 113581 (CLOUD-259377): `TEvCountersInfoResponse` handler appended `std::move(response)` and then called `response.size()` for `CountersSize`.

Add to `CountersSize` before moving `response` into the stored pair.


Evgenik2 (approved):
LGTM

PR #37759, opened 2026-04-09
Fix missing local PDisk case

alexvru:
### Changelog entry 

Fix missing local PDisk case

### Changelog category 

### Description for reviewers 

This fixes the case when VDisk is started before actual PDisk is spinned up leading to assertion failure.


mregrock (approved):
LGTM

PR #37868, opened 2026-04-10
Proper fix for missing PDisk instead of AI-generated one

alexvru:
### Changelog entry 

Proper fix for missing PDisk instead of AI-generated one

### Changelog category 

### Description for reviewers 

Proper fix for the problem with missing PDisks in NW while running tests.


mregrock (approved):
LGTM

PR #37956, opened 2026-04-11
[NEW RBO] Enable Shuffle Elimination in new RBO

alexpaniman:
Resolves #27665. Enable Shuffle Elimination in new RBO.

What was already there in new RBO:
1. There is a stage, which finds subtrees which can be optimized with CBO, calls them "CBO Tree". These are created and expanded until they take as much operators as they can.
2. There is a rule which applies CBO by converting "CBO Tree" to CBO's internal representation. It also launches CBO and, after it runs, it converts the tree back and inlines it.

But before this PR CBO was always launched without SE.

**What's implemented in this PR to enable SE:**
1. Additional metadata added: `ShuffledByColumns`, which is initialized in `TOpRead` using information about initial table shuffles from `PartitionedByColumns` (which is historically a bit of a misnomer).
2. This metadata is propagated through all operators to reach the leaf
nodes of CBO Trees.
3. The `TOptimizeCBOTreeRule` then takes this CBO Tree, builds an orderings FSM (by adding interesting orderings and functional dependencies extracted from joins), and initializes all base nodes with appropriate shuffling from the earlier propagated `ShuffledByColumns`.
4. `TTableMapAlias` is constructed alongside the orderings FSM.
5. Then CBO runs (having everything it needs to launch SE DPHyp if it decides to).
6. The optimized join order tree is then inlined back. Information about which sides to shuffle is passed through `Shuffle(Left|Right)SideBy`. Note the distinction: `ShuffledByColumns` means "this side is already shuffled by these columns" (a statement of fact), whereas `Shuffle(Left|Right)SideBy` means "we must shuffle this side by these columns" (an imperative request).
7. Finally, `ShuffledByColumns` properties are propagated through the joins in the new RBO to other CBO Trees that are topologically further along, allowing them to be optimized next.
8. After all this finishes, additional stage runs, `TPropagateHashFuncStage`, this stage selects `ColumnShardHashV1` in all the places where initial columns shard partitioning is reused and ensures that all the joins which reuse it further use compatible partitioning on both sides.


ydb/core/kqp/opt/rbo/rules/assign_stages.cpp
```diff
@@ -91,8 +91,22 @@ bool TAssignStagesRule::MatchAndApply(TIntrusivePtr<IOperator>& input, TRBOConte
...
-            props.StageGraph.Connect(rightStage, newStageId, MakeIntrusive<TShuffleConnection>(rightShuffleKeys, rightInputStorageType, rightOutputIndex));
+            auto seHashFunc = ctx.KqpCtx.Config->GetDqDefaultHashShuffleFuncType();
+            if (join->Props.LeftShuffleEliminated || join->Props.RightShuffleEliminated) {
+                seHashFunc = NDq::EHashShuffleFuncType::ColumnShardHashV1;
+            }
+
+            if (join->Props.LeftShuffleEliminated) {
+                props.StageGraph.Connect(leftStage, newStageId, MakeIntrusive<TMapConnection>(leftInputStorageType));
+            } else {
+                props.StageGraph.Connect(leftStage, newStageId, MakeIntrusive<TShuffleConnection>(leftShuffleKeys, leftInputStorageType, 0u, seHashFunc));
```
denis0x0D: Тут нужно правильный output index указать ``leftOutputIndex```

denis0x0D (changes requested):

denis0x0D (dismissed):

pavelvelikhov (approved):

PR #38065, opened 2026-04-14
Discard per Task ResourceManager State

Hor911:
To allow more flexible and efficient memory management

Old call sequence

1. Each tasks asks RM to Allocate/Free resourcs (Memory)
2. These calls are made from AS event handler, so they are serialized and no additional sync primitives (per task) is required
3. Each tasks also allocates "tasks" in ResourceBroker (so N tasks consume N slots in Scheduler Queue)

New (these PR) call sequence

1. All tasks of the graph share single (large) quota, managed by RM in TxState
2. They may call RM simultaneously
3. But these N tasks use common "task" in RB in consume only 1 slot in Scheduler Queue
4. RB task is started with very 1st allocation and finished from ~TxState for serialization reasons

Current state is transitional. Intention is to provide single memory manager for the graph which will deal with all tasks memory requests. It enables various complex scenario, for example

- tasks may agree on the start, who is potentially needed a lot of memory for processing, and change memory allocation policy (f.e. from greedy to fair share)
- tasks may communicate in runtime to redistribute already allocated memory
- and it's good point for integration with WM

So, target call sequence (future PRs) is:

1. Task never calls RM directly. Instead it calls special memory manager for this query (using the same API)
2. This memory manager batches and clears (does netting) requests, reducing number of RM calls by the order pf magnitude

### DIFF
--- a/ydb/core/kqp/compute_actor/kqp_compute_actor_factory.cpp
+++ b/ydb/core/kqp/compute_actor/kqp_compute_actor_factory.cpp
@@ -21,7 +21,7 @@ struct TMemoryQuotaManager : public NYql::NDq::TGuaranteeQuotaManager {
     {}
 
     ~TMemoryQuotaManager() override {
-        ResourceManager->FreeResources(Tx, TaskId, NRm::TKqpResourcesRequest{
+        ResourceManager->FreeResources(*Tx, TaskId, NRm::TKqpResourcesRequest{
             .ExecutionUnits = 1,
             .Memory = Limit - Guarantee,
             .ExternalMemory = Guarantee,
@@ -29,7 +29,7 @@ struct TMemoryQuotaManager : public NYql::NDq::TGuaranteeQuotaManager {
     }
 
     bool AllocateExtraQuota(ui64 extraSize) override {
-        auto result = ResourceManager->AllocateResources(Tx, TaskId,
+        auto result = ResourceManager->AllocateResources(*Tx, TaskId,
             NRm::TKqpResourcesRequest{.Memory = extraSize});
 
         if (!result) {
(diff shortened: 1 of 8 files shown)


ydb/core/kqp/rm_service/kqp_rm_service.cpp
```diff
@@ -338,46 +341,47 @@ class TKqpResourceManager : public IKqpResourceManager {
             return result;
         }
 
-        tx->Allocated(task, resources);
-        if (!task->ResourceBrokerTaskId) {
-            task->ResourceBrokerTaskId = rbTaskId;
-        } else {
-            bool merged = ResourceBroker->MergeTasksInstant(task->ResourceBrokerTaskId, rbTaskId, SelfId);
-            Y_ABORT_UNLESS(merged);
+        with_lock(tx->Lock) {
```
gridnevvvit: а зачем тут мьютекс?

ydb/core/kqp/compute_actor/kqp_compute_actor_factory.cpp
```diff
@@ -12,27 +12,31 @@ struct TMemoryQuotaManager : public NYql::NDq::TGuaranteeQuotaManager {
 
     TMemoryQuotaManager(std::shared_ptr<NRm::IKqpResourceManager> resourceManager
         , TIntrusivePtr<NRm::TTxState> tx
-        , TIntrusivePtr<NRm::TTaskState> task
+        , ui64 taskId
```
gridnevvvit, line 15: struct TTaskState собирал статистику по таске. было удобно. зачем это?

ydb/core/kqp/rm_service/kqp_rm_service.cpp
```diff
@@ -338,46 +341,47 @@ class TKqpResourceManager : public IKqpResourceManager {
...
+            tx->Allocated(resources);
+            if (!tx->TxResourceBrokerTaskId) {
```
gridnevvvit: поскольку Id мы аллоцируем сами, пускай тогда при создании выполняется fetch_add(1) и инстанцируется TxResourceBrokerTaskId и тогда мьютекс и не нужен

gridnevvvit (approved):

PR #38149, opened 2026-04-15
Introduce detailed VDisk replication status

serbel324:
### Changelog entry 

Introduce detailed VDisk replication status. This allows to distinguish situations when replication is actively processing data, when it is waiting for token and when it can't finish due to phantoms.

### Changelog category 

* Improvement

### DIFF
--- a/ydb/core/protos/node_whiteboard.proto
+++ b/ydb/core/protos/node_whiteboard.proto
@@ -220,7 +220,7 @@ message TVDiskStateInfo {
     // The same for the non-phantom-like blobs.
     optional bool UnreplicatedNonPhantoms = 21 [default = false,  (DefaultField) = true];
     // Detailed replication status.
-    optional TVDiskDetailedReplicationStatus.E DetailedReplicationStatus = 37 [(DefaultField) = true];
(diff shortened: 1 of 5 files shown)


ydb/core/protos/node_whiteboard.proto
```diff
@@ -210,6 +219,8 @@ message TVDiskStateInfo {
     optional bool UnreplicatedPhantoms = 20 [default = false, (DefaultField) = true];
     // The same for the non-phantom-like blobs.
     optional bool UnreplicatedNonPhantoms = 21 [default = false,  (DefaultField) = true];
+    // Detailed replication status.
+    optional TVDiskDetailedReplicationStatus.E DetailedReplicationStatus = 37 [(DefaultField) = true];
```
adameat: we don't add new fields as default

```suggestion
    optional TVDiskDetailedReplicationStatus.E DetailedReplicationStatus = 37;
```

adameat (changes requested):

rosik (approved):

adameat (approved):

PR #38172, opened 2026-04-15
Added warning about non matching vector size (issue #18667)

### LINKED ISSUE #18667: Do not ignore unexpected vector column format in vector index
Currently, rows with the embedding column in invalid format (IsExpectedSize() == false) are silently ignored.
It even renders ut_vector_index_build tests incorrect - these tests try to create a vector index with vector dimension 42 while the vector column has just random string data of unexpected size.

BlazingPhoenix:
### Changelog entry 

Added warning about non matching vector size.

### Changelog category 
* Not for changelog

### Description for reviewers

### DIFF
--- a/ydb/core/tx/datashard/build_index/local_kmeans.cpp
+++ b/ydb/core/tx/datashard/build_index/local_kmeans.cpp
@@ -67,6 +67,7 @@ protected:
 
     ui64 ReadRows = 0;
     ui64 ReadBytes = 0;
+    ui64 InvalidEmbeddingRows = 0;
 
     TBatchRowsUploader Uploader;
 
@@ -201,6 +202,13 @@ public:
         }
 
         Uploader.Finish(record, status);
+
+        if (InvalidEmbeddingRows > 0) {
+            auto* issue = record.AddIssues();
+            issue->set_severity(NYql::TSeverityIds::S_WARNING);
+            issue->set_message(TStringBuilder()
+                << InvalidEmbeddingRows << " row(s) with invalid vector format were skipped during index build");
+        }
 
         if (Response->Record.GetStatus() == NKikimrIndexBuilder::DONE) {
             LOG_N("Done " << Debug() << " " << Response->Record.ShortDebugString());
@@ -485,6 +493,7 @@ protected:
 
         const auto embedding = row.at(EmbeddingPos).AsRef();
         if (!Clusters->IsExpectedFormat(embedding)) {
+            ++InvalidEmbeddingRows;
             return;
         }
 
@@ -510,6 +519,9 @@ protected:
     void FeedFinal(TArrayRef<const TCell> row, TArrayRef<const TCell> sourcePk,
         TArrayRef<const TCell> dataColumns, TArrayRef<const TCell> origKey, bool isPostingLevel)
     {
+        if (!Clusters->IsExpectedFormat(row.at(EmbeddingPos).AsRef())) {
+            return;
+        }
         Clusters->FindClusters(row.at(EmbeddingPos).AsBuf(), TmpClusters, OverlapClusters, OverlapRatio);
         if (OutForeign) {
             bool foreign = false;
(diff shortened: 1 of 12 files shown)


BlazingPhoenix:
### Code review

Found 2 issues:

1. **`InvalidEmbeddingRows` is double-counted when `EScan::Reset` re-scans rows in the UPLOAD phase** — in `local_kmeans.cpp` and `prefix_kmeans.cpp`, when `IsPrefixRowsValid == false` (the in-memory buffer overflowed), the scan issues `EScan::Reset` to re-read rows from disk for each K-means iteration. Since `FeedFinal` increments `InvalidEmbeddingRows` on every pass and the counter is never reset between iterations, a single invalid-embedding row can be counted multiple times — once per full rescan that transitions through the UPLOAD phase. The comment at the reset site makes this explicit: `// scan current prefix rows with a new state again`.

https://github.com/ydb-platform/ydb/blob/ee5e35fd69b6f548db4005751e18db96d690e554/ydb/core/tx/datashard/build_index/local_kmeans.cpp#L258-L265

https://github.com/ydb-platform/ydb/blob/ee5e35fd69b6f548db4005751e18db96d690e554/ydb/core/tx/datashard/build_index/local_kmeans.cpp#L518-L526

2. **When all rows in a prefix group have invalid embeddings, no warning is emitted** — `FeedSample` silently discards invalid-embedding rows without incrementing `InvalidEmbeddingRows`. When every row in a prefix is invalid, the sampler returns 0 rows, no clusters are formed, `FinishPrefixImpl` returns early, and `FeedFinal` is never called for those rows. The result: the most severe data-quality case (a prefix where 100% of embeddings are malformed) produces zero warning, while the warning only fires for partially-bad prefixes.

https://github.com/ydb-platform/ydb/blob/ee5e35fd69b6f548db4005751e18db96d690e554/ydb/core/tx/datashard/build_index/local_kmeans.cpp#L484-L492

https://github.com/ydb-platform/ydb/blob/ee5e35fd69b6f548db4005751e18db96d690e554/ydb/core/tx/datashard/build_index/local_kmeans.cpp#L410-L425

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

BlazingPhoenix:
### Code review

Found 1 issue:

1. `InvalidEmbeddingRowsInPrefix` is double-counted in the `FinishPrefix()` re-feed loop (`local_kmeans.cpp` and `prefix_kmeans.cpp`). When `FinishPrefixImpl()` returns `false` and `IsPrefixRowsValid` is `true`, the code re-feeds the saved `PrefixRows` in an infinite loop until K-means converges. Each iteration calls `Feed()` → `FeedSample()`, which increments `InvalidEmbeddingRowsInPrefix` for every invalid row encountered. Since the counter is never reset between loop iterations, invalid rows are counted once per iteration. Given the state machine in `FinishPrefixImpl()` (SAMPLE→KMEANS→UploadState), the loop always runs at least 2 iterations, so the reported count will be at minimum 2× the real count, and more with additional K-means rounds.

https://github.com/ydb-platform/ydb/blob/e4cb3399a62ea5e00db0a97ec1e5ee5888b45355/ydb/core/tx/datashard/build_index/local_kmeans.cpp#L395-L420

The same issue exists in `prefix_kmeans.cpp`:

https://github.com/ydb-platform/ydb/blob/e4cb3399a62ea5e00db0a97ec1e5ee5888b45355/ydb/core/tx/datashard/build_index/prefix_kmeans.cpp#L428-L461

Fix: save and restore `InvalidEmbeddingRowsInPrefix` before and after the re-feed loop, or snapshot it before the loop and only use the snapshot value for the final count.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vitalif (dismissed):

vitalif:
А эти варнинги увидит пользователь после построения индекса где-то?
Хотелось бы чтобы увидел, хотя бы в описании операции в Operations

vitalif:
Но если пока не видно то можно это отдельным пр, а этот будет за даташард

Gazizonoki (approved):
SDK lgtm

vitalif (approved):

PR #38213, opened 2026-04-16
Logic to build stats when the size exceeds the limit.

yizenov:
### Changelog entry 

Improving the logic of building columnar statistics.

### Changelog category 

* Improvement

### Description for reviewers 

Adding the following logic when the calculated size of a columnar statistic exceeds the memory limit of 8MB:
"If the computed size exceeds the limit of 8MB, build statistic with a size that can fit into 8MB."

### DIFF
--- a/ydb/core/statistics/aggregator/column_statistic_eval.cpp
+++ b/ydb/core/statistics/aggregator/column_statistic_eval.cpp
@@ -370,11 +370,6 @@ public:
         const double n = simpleStats.GetCount();
         const double ndv = simpleStats.GetCountDistinct();
 
-        // if (ndv >= 0.8 * n) {
-        //     // Too many distinct values i.e. domain is close to be PK
-        //     return TPtr{};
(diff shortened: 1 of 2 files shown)


pavelvelikhov (dismissed):

ydb/core/statistics/aggregator/column_statistic_eval.cpp
```diff
@@ -358,10 +370,10 @@ class TEWHEval : public IStage2ColumnStatisticEval {
         const double n = simpleStats.GetCount();
         const double ndv = simpleStats.GetCountDistinct();
 
-        if (ndv >= 0.8 * n) {
-            // Too many distinct values
-            return TPtr{};
-        }
+        // if (ndv >= 0.8 * n) {
```
ztlpn: Давай удалим, если не нужен этот код

ztlpn (approved):

PR #38475, opened 2026-04-21
LongTxService: deadlock detection

ztlpn:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

* Add wait graph-related metrics to LongTxService
* Implement a deadlock detection algorithm for the lock wait graph
* Add a randomized test checking that deadlock detection allows forward progress in the presence of deadlocks

Fixes https://github.com/ydb-platform/ydb/issues/35482

### DIFF
--- a/ydb/core/tx/datashard/datashard_ut_lock_rows.cpp
+++ b/ydb/core/tx/datashard/datashard_ut_lock_rows.cpp
@@ -68,6 +68,11 @@ Y_UNIT_TEST_SUITE(DataShardLockRows) {
             Reopen();
         }
 
+        TTestPipe(const TTestPipe&) = delete;
+        TTestPipe& operator=(const TTestPipe&) = delete;
+        TTestPipe(TTestPipe&&) = delete;
+        TTestPipe& operator=(TTestPipe&&) = delete;
+
         ~TTestPipe() {
             Close();
         }
--- a/ydb/core/tx/datashard/datashard_ut_lock_rows.cpp
+++ b/ydb/core/tx/datashard/datashard_ut_lock_rows.cpp
@@ -265,6 +265,26 @@ Y_UNIT_TEST_SUITE(DataShardLockRows) {
         TTestActorRuntime::TEventObserverHolder RemoveObserver;
     };
 
+    static TString GetWaitGraphStr(TTestActorRuntime& runtime, ui32 nodeIdx) {
+        ui32 nodeId = runtime.GetNodeId(nodeIdx);
+        auto sender = runtime.AllocateEdgeActor(nodeIdx);
+        runtime.Send(
+            MakeLongTxServiceID(nodeId), sender,
+            new TEvLongTxService::TEvGetLockWaitGraph, nodeIdx, true);
+        auto result = runtime.GrabEdgeEventRethrow<TEvLongTxService::TEvGetLockWaitGraphResult>(sender);
+
+        TVector<std::pair<ui64, ui64>> edges;
+        for (const auto& edge : result->Get()->WaitEdges) {
+            edges.emplace_back(edge.Awaiter.LockId, edge.Blocker.LockId);
+        }
+        std::sort(edges.begin(), edges.end());
+        TStringBuilder wgStr;
+        for (auto& [a, b] : edges) {
+            wgStr << a << " -> " << b << "\n";
+        }
+        return wgStr;
+    }
+
     Y_UNIT_TEST(Basics) {
         TPortManager pm;
 
@@ -1328,24 +1348,8 @@ Y_UNIT_TEST_SUITE(DataShardLockRows) {
         runtime.SimulateSleep(TDuration::Seconds(1));
         for (size_t nodeIdx = 0; nodeIdx < runtime.GetNodeCount(); ++nodeIdx) {
             // Check that the wait edges are instantiated in the LongTxService on all nodes.
-            ui32 nodeId = runtime.GetNodeId(nodeIdx);
-            auto sender = runtime.AllocateEdgeActor(nodeIdx);
-            runtime.Send(
-                MakeLongTxServiceID(nodeId), sender,
-                new TEvLongTxService::TEvGetLockWaitGraph, nodeIdx, true);
-            auto result = runtime.GrabEdgeEventRethrow<TEvLongTxService::TEvGetLockWaitGraphResult>(sender);
-
-            TVector<std::pair<ui64, ui64>> edges;
-            for (const auto& edge : result->Get()->WaitEdges) {
-                edges.emplace_back(edge.Awaiter.LockId, edge.Blocker.LockId);
-            }
-            std::sort(edges.begin(), edges.end());
-            TStringBuilder wgStr;
-            for (auto& [a, b] : edges) {
-                wgStr << a << " -> " << b << "\n";
-            }
             UNIT_ASSERT_VALUES_EQUAL_C(
-                wgStr,
+                GetWaitGraphStr(runtime, nodeIdx),
                 "123 -> 234\n",
                 "with nodeIdx: " << nodeIdx);
         }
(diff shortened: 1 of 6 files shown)


ydb/core/tx/datashard/datashard_ut_lock_rows.cpp
```diff
@@ -1839,6 +1842,244 @@ Y_UNIT_TEST_SUITE(DataShardLockRows) {
...
+        Cerr << "Loop iterations: " << loopIter << Endl
+            << "Successful transactions: " << successfulTxs << Endl
+            << "Failed transactions: " << failedTxs << Endl;
+        Cerr << "Response status counts:" << Endl;
+        for (const auto& [status, count] : responseStatusStats) {
+            Cerr << status << ": " << count << Endl;
+        }
+
+        UNIT_ASSERT_GT_C(
+            successfulTxs, REQUIRED_SUCCESSFUL_TXS,
```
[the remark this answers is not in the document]
ztlpn: replaced with the `actorToTxState.empty()` condition

ydb/core/tx/long_tx_service/long_tx_service_ut.cpp
```diff
@@ -508,6 +508,11 @@ Y_UNIT_TEST_SUITE(LockWaitGraph) {
             SendToLongTx(new TEvLongTxService::TEvWaitingLockRemove(reqId));
         }
 
+        template<class TEvent>
+        typename TEvent::TPtr GrabEvent(TDuration simTimeout = TDuration::Max()) {
```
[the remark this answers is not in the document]
ztlpn: It's okay

ydb/core/tx/long_tx_service/long_tx_service_impl.cpp
```diff
@@ -742,6 +749,9 @@ void TLongTxServiceActor::Handle(TEvLongTxService::TEvLockStatus::TPtr& ev) {
             TLockStateHandle{lock}, record.GetWaitEdges(),
             [&](const TWaitEdgeId& id) { return id.OwnerId.NodeId() != SelfId().NodeId(); });
 
+        if (gotTimestamp && lock.WaitNode.Island) {
+            ScheduleDeadlockDetection(*lock.WaitNode.Island, TDuration::Zero());
+        }
```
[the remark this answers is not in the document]
ztlpn: We want locks without timestamps to take part in deadlock resolution, as non-zero timestamp may never appear.

ydb/core/tx/datashard/datashard_ut_lock_rows.cpp
```diff
@@ -1839,6 +1842,244 @@ Y_UNIT_TEST_SUITE(DataShardLockRows) {
...
+        // with some idle timeout between them.
+        //
+        // Each round involves sending TEvLockRows requests to several random shards,
+        // each of them locking a few random rows. To successfully complete, a transaction
+        // must complete all rounds successfully, otherwise it is cancelled.
+        //
+        // The test checks that deadlocks are promptly detected and broken, and the system
+        // as a whole makes sufficient progress.
+
+        const ui64 RNG_SEED = 42;
```
flown4qqqq, line 1856: Like for 42 :)

flown4qqqq, line 1856: I guess, we dont need to use SCREAMING_SNAKE_CASE here

SCREAMING_SNAKE_CASE we usually use when want to create macros or smth like this

ydb/core/tx/long_tx_service/long_tx_service.h
```diff
@@ -20,6 +20,13 @@ namespace NLongTxService {
         TCounterPtr SnapshotsCollectionTimeMs;
         TCounterPtr SnapshotsPropagationTimeMs;
 
+        TCounterPtr RemoteLockSubscriptions;
+        TCounterPtr WaitGraphEdges;
+        TCounterPtr LocalWaitGraphEdges;
+        TCounterPtr WaitGraphEdgesSent;
+        TCounterPtr WaitGraphEdgesReceived;
+        TCounterPtr WaitGraphEdgesBroken;
+
```
flown4qqqq, line 29: Do we really need too many counters?

ydb/core/tx/long_tx_service/long_tx_service_impl.cpp
```diff
@@ -1426,6 +1547,184 @@ void TLongTxServiceActor::Handle(TEvLongTxService::TEvGetLockWaitGraph::TPtr& ev
...
+    }
+
+    // Kosaraju SCC algorithm, pass 1: DFS on forward graph (Awaiter -> Blocker edges),
+    // collecting nodes in finish order. Explicit stack to avoid recursion overhead.
+    // Each stack frame is (node, finishing): when finishing=true the node is being
+    // popped after all descendants have been processed.
+    TVector<TWaitNode*> finishOrder;
+    finishOrder.reserve(island.LocksCount);
+    {
+        TVector<std::pair<TWaitNode*, bool>> stack;
```
flown4qqqq, line 1612: I would suggest to separate dfs in other function.

Moreover, you can reuse this function in second step.

Something like this:

```cpp
class TSolver {
private:

std::vector<std::vector<int>> original_g;
std::vector<std::vector<int>> reversed_g;
std::vector<char> was;

void Dfs(int v, std::vector<int>& order, std::vector<std::vector<int>>& g) {
    was[v] = true;
    for (int u : g[v]) {
        if (!was[u]) {
            Dfs(u, order, g);
        }
    }

    order.push_back(v);
}

public:

void Solve() {
    int n, m;
    std::cin >> n >> m;
    original_g.resize(n);
    reversed_g.resize(n);
    was.resize(n);

    for (int i = 0; i < m; i++) {
        int u, v;
        std::cin >> u >> v;
        u--, v--;
        original_g[u].push_back(v);
        reversed_g[v].push_back(u);
    }

    std::vector<int> order;
    for (int v = 0; v < n; v++) {
        if (!was[v]) {
            Dfs(v, order, original_g);
        }
    }

    std::vector<std::vector<int>> css;
    was.assign(n, false);
    std::reverse(order.begin(), order.end());

    for (int v : order) {
        if (!was[v]) {
            std::vector<int> cs;
            Dfs(v, cs, reversed_g);
            css.push_back(cs);
        }
    }

    int cnt = 0;
    for (auto& cs : css) {
        std::cout << ++cnt << ", " << cs.size() << ": ";
        for (int v : cs) {
            std::cout << v + 1 << " ";
        }
        std::cout << std::endl;
    }
}
};
```

flown4qqqq: Of course, we shouldn't use recursive dfs in production code, but it is example only.

ydb/core/tx/long_tx_service/long_tx_service_impl.cpp
```diff
@@ -1426,6 +1547,184 @@ void TLongTxServiceActor::Handle(TEvLongTxService::TEvGetLockWaitGraph::TPtr& ev
...
+                || edge.Id.OwnerId.NodeId() != SelfId().NodeId()
+                || edge.Blocker.WaitNode().SccId != sccId) {
+                continue;
+            }
+
+            TXLOG_DEBUG("Breaking the wait edge id: " << edge.Id
+                << ", awaiter: " << edge.Awaiter.LockInfo(SelfId())
+                << ", blocker: " << edge.Blocker.LockInfo(SelfId()));
+            edge.Broken = true;
+            Send(edge.Id.OwnerId, new TEvLongTxService::TEvWaitingLockDeadlock(edge.Id.RequestId));
```
flown4qqqq, line 1718: Can we create only one message with repeated field which contains list of edges?

ydb/core/tx/long_tx_service/long_tx_service.h
ztlpn: Yes :) These counters are one-per-node, so the overhead is negligible

ydb/core/tx/long_tx_service/long_tx_service_impl.cpp
ztlpn: We can, but I don't think there is much benefit in processing these commands in batches.

ydb/core/tx/datashard/datashard_ut_lock_rows.cpp
ztlpn: The convention in this folder is to use uppercase for constants.

ydb/core/tx/long_tx_service/long_tx_service_impl.cpp
ztlpn: Two dfs runs are quite a bit different, so abstracting them can be awkward. I doubt the code would be much more readable.

ztlpn: Extracted custom logic out of the main dfs loop.

flown4qqqq (approved):

lex007in (approved):

PR #38718, opened 2026-04-23
fix a bug in interconnect

gridnevvvit:
### Changelog category 

### Description for reviewers 

...


alexvru (changes requested):
What does this fix? It doesn't violate C++ temporary object lifetime rules.

gridnevvvit:
agreed, not a true problem actually. the bug i saw on the overview page is different

PR #38742, opened 2026-04-24
Handle OUT_OF_SPACE for huge writes

alexvru:
### Changelog entry 

Handle OUT\_OF\_SPACE for huge writes

### Changelog category 

### Description for reviewers 

This allows to terminate huge blob write queries in out-of-space situation without turning VDisk into stone.


mregrock (approved):
LGTM

PR #39097, opened 2026-04-28
[IC] Suppress repeated IC handshake failure logs. NBYDB-2145

dcherednik:
### Changelog entry 

Log the first interconnect TCP handshake failure at NOTICE and demote repeated failures while the proxy is held by error to DEBUG. Promote handshake success after an error back to NOTICE so recovery is visible.

Add an interconnect unit test that stops a connected node and verifies ICP25 is emitted once at NOTICE and then at DEBUG on retries.
### Changelog category 

### Description for reviewers 

...


robdrynkin (approved):

the-ancient-1 (approved):
LGTM

PR #39100, opened 2026-04-28
Fix defrag with softchunks lock loop without progress

robdrynkin:
### Changelog entry 

Fix CPU burning loop without any progress when GarbageThresholdToRunFullCompactionPerMille > 0.

### Changelog category 

* Bugfix 

### Description for reviewers 

Previously, scheduled defrag immediately started a new planner scan after every auto quantum result, even when the quantum made no rewrite progress. With deferred full compaction enabled, this could repeatedly rescan the same Hull snapshot state and burn CPU. The scheduler now backs off when an auto quantum finishes with zero rewritten records/bytes.

Also report actual locked chunks in FreedChunks instead of planned candidates, skip useless record scans when no chunks were locked and local compaction is disabled, and add a regression test for the no-progress backoff case.

### DIFF
--- a/ydb/core/blobstorage/vdisk/defrag/defrag_quantum.cpp
+++ b/ydb/core/blobstorage/vdisk/defrag/defrag_quantum.cpp
@@ -97,7 +97,7 @@ namespace NKikimr {
                         (ActorId, SelfActorId), (LockedChunks, lockedChunks));
 
                     if (lockedChunks.empty()) {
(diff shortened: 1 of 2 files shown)


ydb/core/blobstorage/vdisk/defrag/defrag_quantum.cpp
va-kuznecov:
```suggestion
                        STLOG(PRI_NOTICE, BS_VDISK_DEFRAG, BSVDD17, DCtx->VCtx->VDiskLogPrefix << "could not lock chunks, going to run full compaction instead", (ChunksToDefrag, *ChunksToDefrag));
```

va-kuznecov (approved):

SammyVimes (approved):
LGTM

PR #39157, opened 2026-04-29
Lower logging level while creating session to unexisting rate limiter resource

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Lower logging level when the client tries to use unexisting rate limiter resource as it does not mark any problems on server side.


CyberROFL (commented):
Перепризови @Enjection после обработки комментариев от ai-ревьюверов.

Enjection (approved):

PR #39459, opened 2026-05-04
Fix static self-heal by cleaning vdisks properly and support node deletion

mregrock:
### Changelog entry 

Fix static self-heal by cleaning vdisks properly and support node deletion

### Changelog category 

### Description for reviewers 

...


serbel324 (approved):
seems ok

PR #40074, opened 2026-05-12
Split Decommit3dc test into subtests

serbel324:
### Changelog entry 

Split Decommit3dc into separate tests to fix timeout issues.

### Changelog category 


mregrock (approved):
LGTM

PR #40085, opened 2026-05-12
Add stack memory pool for coroutine actors

dcherednik:
### Changelog entry   

Introduce IStackMem/TStackMemPool and allow coroutine actors to opt into
pooled stack memory via the TUsePooledStack/UsePooledStack constructor tag.

  The new pool allocates guarded stack memory mappings. Keeps a small
thread-local cache with a global exchange for cross-thread reuse, and uses page-size based layout with explicit support for downward-growing stacks. Unix/Darwin use mmap and Windows uses VirtualAlloc.

  Switch the interconnect handshake coroutine actor to UsePooledStack<64_KB>().

  Add coroutine actor chain tests for legacy and pooled stacks,
including parallel 4/16/64-chain runs, and add a dedicated ut_mprotect target that checks guard-page protection in a child process.

Local relwithdebinfo check (executor_threads=4):

per_chain_step_us over three sequential runs, with average bars (# ~= 20us, min #):

    StackPool x1     3.187 ->   3.208 ->   3.579  avg    3.325 |#
    StackPool x4     6.017 ->   5.466 ->   5.942  avg    5.808 |#
    StackPool x16   21.890 ->  18.599 ->  21.950  avg   20.813 |#
    StackPool x64   88.554 ->  89.016 ->  80.849  avg   86.140 |####
    Legacy    x1    16.854 ->  15.749 ->  16.569  avg   16.391 |#
    Legacy    x4    69.796 ->  72.964 ->  64.190  avg   68.983 |###
    Legacy    x16  278.135 -> 285.420 -> 261.681  avg  275.079 |##############
    Legacy    x64 1003.799 ->1075.306 ->1022.903  avg 1034.003 |####################################################

### Changelog category 

* Improvement

### Description for reviewers 

...

### DIFF
--- a/ydb/library/actors/core/actor_coroutine.cpp
+++ b/ydb/library/actors/core/actor_coroutine.cpp
@@ -3,6 +3,7 @@
 
 #include <util/system/sanitizers.h>
 #include <util/system/type_name.h>
+#include <util/system/filemap.h>
 #include <util/system/info.h>
 #include <util/system/protect.h>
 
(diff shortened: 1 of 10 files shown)


dcherednik:
replaced with better version

PR #40195, opened 2026-05-13
fix(viewer): harden whiteboard↔viewer EFlag conversion

Raubzeug:
### Changelog entry 

viewer: harden whiteboard→viewer `EFlag` conversion so unknown values fall back to `Grey` instead of an arbitrary color; add regression tests against the `Yellow→Blue` numeric-aliasing bug in `/storage/groups`.

...

### Changelog category 

* Bugfix 

### Description for reviewers 

# viewer: harden whiteboard↔viewer `EFlag` conversion

## Why

`NKikimrWhiteboard::EFlag` and `NKikimrViewer::EFlag` share value names
(`Grey/Green/Yellow/Orange/Red`) but differ numerically — the viewer enum
inserts `Blue = 2`, so a numeric `static_cast` between them mis-maps
colors (`Whiteboard::Yellow (2) → Viewer::Blue (2)`). This was the root
cause of the `/storage/groups` defect where top-level `DiskSpace` showed
`"Blue"` while nested `Whiteboard.DiskSpace` showed `"Yellow"`. The
user-visible bug was already fixed by `aee6de1` (replacing the offending
`static_cast` with `GetViewerFlag()`); this PR is follow-up hardening.

## What

- **`viewer.cpp`**: drop the unsafe `default → static_cast` fallback in
  `GetViewerFlag(NKikimrWhiteboard::EFlag)`; return `Grey` instead. The
  fallback used the exact pattern that caused the original bug.
- **`viewer_ut.cpp`**: two regression tests
  - `WhiteboardFlagsConvertToViewerFlagsByNameNotByNumber` — pins down
    that the two enums are not numerically aligned (incl. the explicit
    `Whiteboard::Yellow == Viewer::Blue` alias), covers both conversion
    directions (`GetViewerFlag` *and* `GetWhiteboardFlag`, the latter
    previously untested), and checks the round-trip via `EFlag_Name`
    (same path the JSON output uses).
  - `StorageGroupVDiskDiskSpaceConvertsByNameNotByNumber` — end-to-end
    guard: a `TStorageGroups::TVDisk` built from
    `Whiteboard::EFlag::Yellow` must render as `"Yellow"`, not `"Blue"`.

## Risk

Low — pure hardening. Behavior changes only for out-of-range enum values
(previously → arbitrary color, now → `Grey`); all defined values are
already covered by the exhaustive `switch`. Test changes are additive.

...

### DIFF
--- a/ydb/core/viewer/viewer.cpp
+++ b/ydb/core/viewer/viewer.cpp
@@ -1198,6 +1198,8 @@ NKikimrWhiteboard::EFlag GetWhiteboardFlag(NKikimrViewer::EFlag flag) {
 NKikimrViewer::EFlag GetViewerFlag(NKikimrWhiteboard::EFlag flag) {
     switch (flag) {
     case NKikimrWhiteboard::EFlag::Grey:
+    case NKikimrWhiteboard::EFlag::EFlag_INT_MIN_SENTINEL_DO_NOT_USE_:
+    case NKikimrWhiteboard::EFlag::EFlag_INT_MAX_SENTINEL_DO_NOT_USE_:
         return NKikimrViewer::EFlag::Grey;
     case NKikimrWhiteboard::EFlag::Green:
         return NKikimrViewer::EFlag::Green;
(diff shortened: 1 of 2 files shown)


rosik (commented):
I suggest closing it in favor of https://github.com/ydb-platform/ydb/pull/40774

PR #40256, opened 2026-05-14
Fix handshake broker over-crediting Capacity on waiter cancel

martinevsky:
### Changelog category 
Not for changelog (changelog entry is not required)

### Description for reviewers 
THandshakeBroker::Handle(TEvHandshakeBrokerFree) unconditionally called
PermitNext(). When the freed actor was a queued waiter that never received
a Permit, no slot was consumed, but PermitNext() still fell into its
`else { Capacity += 1; }` branch and credited a phantom slot. Repeated
waiter cancellations (timeouts, poisons) drifted Capacity upward and
defeated OutgoingHandshakeInflightLimit.

Only call PermitNext() when a permitted lease was actually released.

### DIFF
--- a/ydb/library/actors/interconnect/handshake_broker.h
+++ b/ydb/library/actors/interconnect/handshake_broker.h
@@ -111,14 +111,18 @@ namespace NActors {
 
         void Handle(TEvHandshakeBrokerFree::TPtr& ev) {
             const TActorId sender = ev->Sender;
-            if (!PermittedLeases.erase(sender)) {
+            if (PermittedLeases.erase(sender)) {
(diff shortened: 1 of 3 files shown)


serbel324 (changes requested):
This change may reduce the amount of leases given by broker.
Even though it would make the behaviour of system more logical and more consistent, we cannot be sure it won't have a negative impact on system's performance in general.
I would advice against the merge until handshake broker is properly covered by various tests
https://github.com/ydb-platform/ydb/issues/42187

Verdict is "request changes" mainly to block the merge
Tests could be implemented by other people in other PRs

Issue #40341, opened 2026-05-14
TSAN: Data race in TCompositeTopicReadSession — TIntCounter::Set / signal_utils.cpp unsynchronized access


maximyurchuk:
## Summary

Multiple data races in `TCompositeTopicReadSession`: the `TIntCounter::Set` method in `signal_utils.cpp:20` and `TSignalWrapper::GetFuture` / `optional` fields are accessed both from the main thread (via `WaitEvent` -> `RefreshReadyPartitions` -> `UpdateMetrics`) and from actor threads (via `TDqPqReadBalancerActor::SendValues` -> `GetReadTime` -> `RefreshPartitionsState` -> `UpdateMetrics`).

## Build Information

- **Commit**: `acefcd6`
- **Build Preset**: `release-tsan`
- **Branch**: `prestable-26-2`

## Run Information

- **Run**: [Regression-run_Small_and_Medium #513](https://github.com/ydb-platform/ydb/actions/runs/25854299394)
- **HTML Report**: [Try 1](https://storage.yandexcloud.net/ydb-gh-logs/ydb-platform/ydb/Regression-run_Small_and_Medium/25854299394/ya-prestable-26-2-x86-64-tsan/try_1/ya-test.html)
- **Affected tests** (all from same pid=3441649, single test binary):
  - `ydb/library/yql/providers/pq/gateway/ut/TCompositeTopicReadSessionTest.AdvancePartitionTimeNoOpWhenTimeNotAdvanced`
  - `ydb/library/yql/providers/pq/gateway/ut/TCompositeTopicReadSessionTest.GetEventReturnsDataFromMock`
  - `ydb/library/yql/providers/pq/gateway/ut/TCompositeTopicReadSessionTest.GetEventsReturnsBatch`
  - `ydb/library/yql/providers/pq/gateway/ut/TCompositeTopicReadSessionTest.IdleTimeoutPartitionStillReceivesData`
  - `ydb/library/yql/providers/pq/gateway/ut/TCompositeTopicReadSessionTest.PartitionBalancingBetweenTwoSessions`
  - `ydb/library/yql/providers/pq/gateway/ut/TCompositeTopicReadSessionTest.PartitionBalancingInsideOneSession`
  - `ydb/library/yql/providers/pq/gateway/ut/TCompositeTopicReadSessionTest.ReconnectSessionRecreationWithSameAggregator`
[...]

PR #40587, opened 2026-05-18
asan fix: accesing counters only from compute actor.

martinevsky:
### Changelog entry 

...

### Changelog category 
Not for changelog (changelog entry is not required)

### Description for reviewers 
 #40341

This pull request makes a small adjustment to the `TCompositeTopicReadSession` class to improve when metrics are updated during partition state changes. The main change is moving the call to `UpdateMetrics()` from the end of the `AdvancePartitionState` method to immediately after refreshing the partition state, ensuring metrics are updated at the correct point in the process.

Most important change:

* Moved the call to `UpdateMetrics()` so that metrics are updated right after `RefreshPartitionsState` in the `AdvancePartitionState` method, and removed its previous call location from `AdvanceExternalTime` for more accurate metric reporting. [[1]](diffhunk://#diff-53feeb64c5a56fbaf5108f1a4a9d62da4835758570fb0f05fa438a96f9c5d7faR790) [[2]](diffhunk://#diff-53feeb64c5a56fbaf5108f1a4a9d62da4835758570fb0f05fa438a96f9c5d7faL898-L899)those who read this PR -->

...


martinevsky:
Акторы должны быть на одном мейлбоксе, фикс смысла не имеет, фиксать надо по другому.

PR #41103, opened 2026-05-22
Fix CREATE OR REPLACE for external data source, external table, resource pool

buhtr:
### Changelog entry 

Fixed CREATE OR REPLACE for external data sources, external tables, and resource pools

### Changelog category 

* Bugfix 

### Description for reviewers 

Fixes `SchemeShard` operations for external data source, external table, and resource pool to follow the correct patterns established by streaming query operations.

1. Migrate to` MemChanges/DbChanges` pattern

Replaced direct NIceDb persistence in Propose methods with the `DbGuard `+ `MemChanges/DbChanges` pattern, enabling proper undo on AbortPropose.

3. Use common `ReplaceIfExists` flag

Changed `CREATE OR REPLACE` to use `TModifyScheme.ReplaceIfExists` instead of per-entity `ReplaceIfExists` in: `SchemeShard`, `KQP gateway`.

5. Fix ProtectDB violation in alter_external_table

Replaced direct `NIceDb` column deletion (which violated `ProtectDB` guard) with soft-delete pattern: old columns are carried over with DeleteVersion set, matching how regular tables handle column.

fixes: YQ-4610

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__operation_create_external_data_source.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__operation_create_external_data_source.cpp
@@ -171,14 +171,6 @@ class TCreateExternalDataSource : public TSubOperation {
         return externalDataSource;
     }
 
-    void CreateTransaction(const TOperationContext &context,
-                           const TPathId &externalDataSourcePathId) const {
-        TTxState& txState = context.SS->CreateTx(OperationId,
-                                                 TTxState::TxCreateExternalDataSource,
-                                                 externalDataSourcePathId);
-        txState.Shards.clear();
-    }
-
     void RegisterParentPathDependencies(const TOperationContext& context,
                                         const TPath& parentPath) const {
         if (parentPath.Base()->HasActiveChanges()) {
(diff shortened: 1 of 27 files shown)


ydb/core/tx/schemeshard/schemeshard__operation_create_external_data_source.cpp
```diff
@@ -154,31 +154,6 @@ class TCreateExternalDataSource : public TSubOperation {
...
-
-    void CreateTransaction(const TOperationContext &context,
-                           const TPathId &externalDataSourcePathId) const {
-        TTxState& txState = context.SS->CreateTx(OperationId,
-                                                 TTxState::TxCreateExternalDataSource,
-                                                 externalDataSourcePathId);
-        txState.Shards.clear();
-    }
-
     void RegisterParentPathDependencies(const TOperationContext& context,
```
ijon: `schemeshard__operation_common_resource_pool.cpp` already has this helper function.
Same for all {create,alter} suboperation files.

ijon (approved):
schemeshard ok

abyss7 (approved):

Issue #40098, opened 2026-05-12
Mute ydb/core/kqp/ut/indexes 2 tests in main


ydbot:
Mute:
ydb/core/kqp/ut/indexes/KqpIndexes.OnlineDuplicatesStage3
ydb/core/kqp/ut/indexes/KqpIndexes.OnlineDuplicatesStage4

Branch:
main

Build type:
relwithdebinfo

Owner: TEAM:@ydb-platform/qp

**About this issue**
- Auto-created by the mute workflow.
- Bot closes when all listed tests are unmuted or no longer exist.
- Close as **Completed** to trigger fast-unmute.
- [mute_rules.md](https://github.com/ydb-platform/ydb/blob/main/.github/config/mute_rules.md) — full details.

**Summary history:** 
 KqpIndexes.OnlineDuplicatesStage3: Flaky last 1 days, at 2026-05-09:2026-05-12: success_rate 77.8%, p-7, f-2, m-0, s-0, total-9
KqpIndexes.OnlineDuplicatesStage4: Flaky last 1 days, at 2026-05-09:2026-05-12: success_rate 77.8%, p-7, f-2, m-0, s-0, total-9

**Test run history:** [link](https://datalens.yandex/34xnbsom67hcq?full_name=__in_ydb%2Fcore%2Fkqp%2Fut%2Findexes%2FKqpIndexes.OnlineDuplicatesStage3&full_name=__in_ydb%2Fcore%2Fkqp%2Fut%2Findexes%2FKqpIndexes.OnlineDuplicatesStage4&branch=main)

More info in [dashboard](https://datalens.yandex/4un3zdm0zcnyr)

ydbot:
### Links to recent failed test runs



| test | status | type | last_run | history | run | stderr | stdout | log | logsdir |
|---|---|---|---|---|---|---|---|---|---|
[...]

Issue #41083, opened 2026-05-22
Mute ydb/core/kqp/ut/indexes/KqpIndexes.OnlineDuplicatesStage1 in main


ydbot:
Mute:
ydb/core/kqp/ut/indexes/KqpIndexes.OnlineDuplicatesStage1

Branch:
main

Build type:
relwithdebinfo

Owner: TEAM:@ydb-platform/qp

**About this issue**
- Auto-created by the mute workflow.
- Bot closes when all listed tests are unmuted or no longer exist.
- Close as **Completed** to trigger fast-unmute.
- [mute_rules.md](https://github.com/ydb-platform/ydb/blob/main/.github/config/mute_rules.md) — full details.

**Summary history:** 
 KqpIndexes.OnlineDuplicatesStage1: Flaky(2026-05-19)->Passed(2026-05-20)->Flaky(2026-05-22) last 1 days, at 2026-05-19:2026-05-22: success_rate 96.6%, p-84, f-3, m-0, s-0, total-87

**Test run history:** [link](https://datalens.yandex/34xnbsom67hcq?full_name=__in_ydb%2Fcore%2Fkqp%2Fut%2Findexes%2FKqpIndexes.OnlineDuplicatesStage1&branch=main)

More info in [dashboard](https://datalens.yandex/4un3zdm0zcnyr)

ydbot:
### Links to recent failed test runs



| test | status | type | last_run | history | run | stderr | stdout | log | logsdir |
|---|---|---|---|---|---|---|---|---|---|
[...]

Issue #40957, opened 2026-05-21
MSAN: Use-of-uninitialized-value in HandleResumeRunnable — KqpIndexes.OnlineDuplicatesStage4


maximyurchuk:
## Summary

The unittest `ydb/core/kqp/ut/indexes/KqpIndexes.OnlineDuplicatesStage4` aborts under `release-msan` with `MemorySanitizer: use-of-uninitialized-value` originating from `HandleResumeRunnable` at [`ydb/library/actors/core/actor.cpp:311:34`](https://github.com/ydb-platform/ydb/blob/stable-26-2-1/ydb/library/actors/core/actor.cpp#L311). The uninitialized read is hit while the actor system dispatches a resumed runnable that was enqueued through `Async` / `TThrFuncObj` from the same `TKikimrRunner` setup path (`kqp_ut_common.cpp` → `kqp_indexes_ut.cpp:6370`).

Reproduces in `try_2` of the msan job (1/3 attempts).

## Environment

- Workflow run: https://github.com/ydb-platform/ydb/actions/runs/26185344887
- Job: `Regression-run_Small_and_Medium (release-msan) / stable-26-2-1:release-msan`
- Build preset: `release-msan`, branch `stable-26-2-1`, commit `d241a3effbedb0849c2d19d61e097cdfe92512bb`
- HTML report (try_2): https://storage.yandexcloud.net/ydb-gh-logs/ydb-platform/ydb/Regression-run_Small_and_Medium/26185344887/ya-stable-26-2-1-x86-64-msan/try_2/ya-test.html#FAIL

## Full MSan output (test `KqpIndexes.OnlineDuplicatesStage4`, try_2)

```
Test crashed (return code: 100)
==2032027==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x00001b3489a8 in HandleResumeRunnable /-S/ydb/library/actors/core/actor.cpp:311:34
    #1 0x00001b3489a8 in NActors::IActor::Receive(TAutoPtr<NActors::IEventHandle, TDelete>&) /-S/ydb/library/actors/core/actor.cpp:349:18
    #2 0x0000470daf69 in NActors::TTestActorRuntimeBase::SendInternal(TAutoPtr<NActors::IEventHandle, TDelete>, unsigned int, bool) /-S/ydb/library/actors/testlib/test_runtime.cpp:1723:33
[...]

Issue #40956, opened 2026-05-21
ASAN: stack-use-after-scope in TAdaptiveThreadPool::TImpl::Add from TKikimrRunner ctor — KqpIndexes.OnlineDuplicatesStage4


maximyurchuk:
## Summary

The unittest `ydb/core/kqp/ut/indexes/KqpIndexes.OnlineDuplicatesStage4` aborts under `release-asan` with `AddressSanitizer: stack-use-after-scope`. The READ happens inside a lambda enqueued via `NThreading::Async` and executed on the `TAdaptiveThreadPool` worker thread T52 (`DoTestOnlineDuplicates::$_1::operator()` at [`ydb/core/kqp/ut/indexes/kqp_indexes_ut.cpp:6370:17`](https://github.com/ydb-platform/ydb/blob/stable-26-2-1/ydb/core/kqp/ut/indexes/kqp_indexes_ut.cpp#L6370)). ASan localizes the access at offset 5824 inside the stack frame of `DoTestOnlineDuplicates` (`kqp_indexes_ut.cpp:6285`), which corresponds to the local variable **`runtime` declared at line 6301** (`'runtime' (line 6301) <== Memory access at offset 5824 is inside this variable`).

Root cause shape: the lambda captures a reference to `runtime` (and likely other locals from `DoTestOnlineDuplicates`), but `Async()` lets it run on a thread-pool thread after the enclosing stack frame has already returned — so the read at `kqp_indexes_ut.cpp:6370` touches freed stack memory. The thread itself is spawned from the `TKikimrRunner` constructor (`kqp_ut_common.cpp:192`) via `TAdaptiveThreadPool::Add`, which is what the "Thread T52 created by T0 here" stack shows.

## Environment

- Workflow run: https://github.com/ydb-platform/ydb/actions/runs/26185344887
- Job: `Regression-run_Small_and_Medium (release-asan) / stable-26-2-1:release-asan`
- Build preset: `release-asan`, branch `stable-26-2-1`, commit `d241a3effbedb0849c2d19d61e097cdfe92512bb`
- HTML report (try_1): https://storage.yandexcloud.net/ydb-gh-logs/ydb-platform/ydb/Regression-run_Small_and_Medium/26185344887/ya-stable-26-2-1-x86-64-asan/try_1/ya-test.html#FAIL
- Artifacts:
[...]

PR #41277, opened 2026-05-25
Do not modify capturedEvents from RunCall() thread

vitalif:
### Changelog category

### Description for reviewers

Should fix all sanitizer issues for these tests: #41256, #40957, #40956, #41083


azevaykin (approved):

vitalif:
Should also fix #40098

PR #41451, opened 2026-05-26
adding-examples-of-using-from-as-table

AlejandroMokhovani:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Adding examples of using data modification statements in combination with `FROM AS TABLE`.

### DIFF
--- a/ydb/docs/en/core/yql/reference/syntax/select/from_as_table.md
+++ b/ydb/docs/en/core/yql/reference/syntax/select/from_as_table.md
@@ -15,3 +15,24 @@ $data = AsList(
 SELECT Key, Value FROM AS_TABLE($data);
 ```
 
+When used in combination with other data modification statements, such as [UPSERT INTO](../upsert_into.md) or [INSERT INTO](../insert_into.md), it is necessary to explicitly specify the columns to be modified both in the source and the target:
+
+```yql
+$data = AsList(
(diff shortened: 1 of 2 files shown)


ydb/docs/en/core/yql/reference/syntax/select/from_as_table.md
```diff
@@ -15,3 +15,24 @@ $data = AsList(
 SELECT Key, Value FROM AS_TABLE($data);
 ```
 
+When used in combination with other data modification statements, such as [UPSERT INTO](../upsert_into.md) or [INSERT INTO](../insert_into.md), it is necessary to explicitly specify the columns to be modified both in the source and the target:
```
weres-sa: You should explicitly specify column names in both the source and the target when using expressions with modifying queries such as upsert or insert.

weres-sa: According to following sentence - “you may”, not “you should”

weres-sa: You should either explicitly specify column names in both the source and the target when using expressions with modifying queries such as upsert or insert. Or you should omit them completely.

weres-sa (dismissed):
LGTM

sintjuri (approved):

fomichev3000 (approved):

PR #41475, opened 2026-05-26
YDBDOCS2126-add deprecated note

SixOnMyface:
### Changelog entry 

add deprecated note

### Changelog category 


sintjuri (dismissed):

ydb/docs/en/core/_includes/scan_query_deprecated_warning.md
```diff
@@ -0,0 +1,6 @@
+{% note warning %}
+
+This functionality is deprecated. Its use is not recommended.
+Use the [`{{ ydb-cli }} sql`](../reference/ydb-cli/sql.md) command instead.
```
weres-sa: I think it would be better to say “you can use direct SQL queries to achieve the same result” because using the CLI may not be as popular a way to perform queries. Most users use the web interface or SDK to make requests, but right now our recommendation seems to be pushing users to use the command line interface, which is not our intention.

weres-sa (changes requested):

sintjuri (approved):

fomichev3000 (approved):

weres-sa (approved):
LGTM

PR #41616, opened 2026-05-27
Do not read ca options from profile when no TLS is needed

pnv1:
### Changelog category 


Shfdis (approved):
lgtm

PR #41655, opened 2026-05-28
Avoid hardcoded mnc patch targets in tests

kruall:
### Changelog entry 

Fix mock patching in the mnc tests

### Changelog category 

### Description for reviewers 

...


mregrock (approved):
LGTM

PR #41771, opened 2026-05-28
[NEW RBO] Fix lost IUs update

alexpaniman:
`ApplyRename` returns a value that we ignored.


alexpaniman:
Merged as part of #40533

PR #42764, opened 2026-06-07
Pace config notification delivery in configs dispatcher + shared config payload

Enjection:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Two coupled changes to `TConfigsDispatcher` so a single config update does not monopolize the user pool:

1. **Paced delivery** — per-subscriber notifications go through a `SendQueue` drained in batches per activation (`NotificationsBatchSize`, default 10, `0` = legacy single-activation fan-out), continuing via a private self-event; `NotificationsBatchDelayMs` (default 0 = just yield) optionally spaces batches. Knobs live in the previously-empty `TConfigsDispatcherConfig` (kind 33) and are read from the effective config. The first batch is sent inline, so small fan-outs behave exactly as before.
2. **Reduced protobuf copying (opt-out)** — the truncated update config is stored once as an immutable `TIntrusiveConstPtr<TEvConsole::TSharedAppConfig>` (moved, not copied, out of the merge result). By default subscribers receive it in `TEvConfigNotificationRequest::SharedConfig` with no inline config in `Record`; read through the `GetConfig()` accessor (prefers shared, falls back to `Record`). Subscribers that mutate the record inline opt out via `TEvSetConfigSubscriptionRequest::UseSharedConfig = false` and keep a private inline copy. Copies per update drop from `1 + N` to `1 move + N_opted_out`. The member is never serialized; remote (console-driven) notifications are unaffected — no dispatcher↔console protocol change.

**Commits (bisect-safe order):**
1. `tests:` — tests + inert interface surface (compiles; shared-payload tests are red).
2. `chore:` — consumer migration, behavior-neutral: read-only handlers switch to the `GetConfig()` accessor; record-mutating consumers (fq/logs, fq/checkpoint_storage, memory_controller, kqp proxy/node/compile/rm/exchanger/vector-cache) opt out at their subscribe sites; blobstorage test-env harnesses materialize the payload into re-sent copies.
3. `feat:` — dispatcher behavior (pacing + shared payload by default).

**Correctness notes:**
- `SubscribersToUpdate` is populated with the full recipient snapshot at update creation, so the ack/unsubscribe completion paths cannot fire while deliveries still wait in the queue.
- Stale queue items (superseded/completed update, unsubscribed or already-acked subscriber) are dropped at drain by cookie + membership checks.
- Superseding an update that still has unsent queued deliveries forces a fresh update to the whole subscription even if the config matches the last delivered state (re-convergence).
- Compatibility note: out-of-tree subscribers reading `Record.GetConfig()` directly should switch to the `GetConfig()` accessor or opt out.
- Mon `State` shows `PendingNotificationSends`.

**Tests** (`configs_dispatcher_ut.cpp`): `TestPacedNotificationsReachAllSubscribers`, `TestPacedNotificationsSupersededByNewerConfig` (observer-held drain self-events + supersession), `TestSharedConfigPayloadDelivery` (shared instance pointer equality, no inline config), `TestMixedSharedAndLegacySubscribers`.

Draft: opening for CI; not built locally.


mregrock (dismissed):
LGTM

CyberROFL (commented):
Обработаешь комментарии робота?

PR #42771, opened 2026-06-07
Add log message priorities to the DQ tasks runner and channels

pzuev:
### Changelog category 

### Description for reviewers 

The DQ task runner and DQ channels used to be able to only emit Debug-level messages which is terribly inconvenient in practice. E.g. TDqTaskRunner::Run should only log its invocation at the Trace level, not even at Debug, otherwise the amount of log spam is unmanageable.

The IsDebugLogEnabled gate is removed though. This will result in some additional log level checks and message formatting but we only have a reasonably bounded number of LOG(...) invocations each Run(), so this shouldn't affect performance noticeably; e.g. a short log message at the start of Run() should cost us about 100ns on our typical Xeons, and each Run() processes large batches typically. By the way, it looks like the IsDebugLogEnabled gate was slightly broken anyway - it was lowering the sample rate. The log message had to pass the sampling rate check twice with two independent random trials.

This also might enable us to log messages that weren't there before (e.g. if the component log level is configured to Warn and we start emitting Warn-level messages). But that's the point: we also want to be able to emit higher priority levels than Debug.

Note: TaskRunners created via the TLocalFactory in the ydb/library/yql/providers/dq/task_runner/tasks_runner_local.cpp use YQL logging and thus coerce all actor system log priorities to the YQL Debug priority, just as before. I can try and add actor log priorities -> YQL levels mapping but we'll need to get rid of `if (YQL_CLOG_ACTIVE(DEBUG, ProviderDq))` in this case.

Note: `static_cast<NActors::NLog::EPriority>(priority)` is still required because LOG_LOG_S only does this cast in the arguments of the `IsLogPriorityEnabled` function but not in the call to `:NActors::MemLogAdapter`. Maybe needs to be fixed, but in a separate PR.

### DIFF
--- a/ydb/library/yql/dq/runtime/dq_tasks_runner.cpp
+++ b/ydb/library/yql/dq/runtime/dq_tasks_runner.cpp
@@ -1,4 +1,3 @@
-#include "dq_channel_service.h"
 #include "dq_tasks_counters.h"
 #include "dq_tasks_runner.h"
 
@@ -37,6 +36,7 @@
 using namespace NKikimr;
 using namespace NKikimr::NMiniKQL;
 using namespace NYql::NDqProto;
+using NActors::NLog::EPrio;
 
 namespace NYql::NDq {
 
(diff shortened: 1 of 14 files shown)


evanevanevanevannnn (approved):

lll-phill-lll (approved):

ydb/library/yql/dq/runtime/dq_tasks_runner.cpp
```diff
@@ -481,7 +481,7 @@ class TDqTaskRunner : public IDqTaskRunner {
         programExplorer.Walk(programRoot.GetNode(), patternEnv);
         auto programSize = programExplorer.GetNodes().size();
 
-        LOG(TStringBuilder() << "task: " << TaskId << ", program size: " << programSize
+        LOG(EPrio::Debug, TStringBuilder() << "task: " << TaskId << ", program size: " << programSize
```
uzhastik, line 484: можем по аналогии с другими местами тут использовать LOG_D ?

PR #42920, opened 2026-06-09
NPB batch write

Evgenik2:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/blobstorage/ddisk/ddisk_actor.h
+++ b/ydb/core/blobstorage/ddisk/ddisk_actor.h
@@ -807,7 +807,7 @@ namespace NKikimr::NDDisk {
         void BarrierErasePersistentBuffer(IEventHandle& queryEv, const TQueryCredentials& creds, const std::vector<TEraseLsnId>& erases, ui64 lsn);
         void FastErasePersistentBuffer(IEventHandle& queryEv, const TQueryCredentials& creds, const std::vector<TEraseLsnId>& erases, const TFastErase& fastErase);
         void ClearPersistentBufferRecords(TPersistentBufferDiskOperationInFlight& inflight, ui64 partCookie);
-        void HandleWritePart(TPersistentBufferDiskOperationInFlight& inflight, ui64 partCookie);
+        void HandleWritePart(TPersistentBufferDiskOperationInFlight& inflight,  ui64 opCookie, ui64 partCookie);
         void HandleErasePart(TPersistentBufferDiskOperationInFlight& inflight, ui64 opCookie, ui64 partCookie, bool resultStatus);
 
         void Handle(TEvWritePersistentBuffer::TPtr ev);
--- a/ydb/core/blobstorage/ddisk/ddisk_actor.h
+++ b/ydb/core/blobstorage/ddisk/ddisk_actor.h
@@ -785,6 +785,7 @@ namespace NKikimr::NDDisk {
         void ReplyReadPersistentBuffer(ui64 operationCookie);
         void ReplyReadPersistentBuffer(TPersistentBuffer::TRecord& pr, NKikimrBlobStorage::NDDisk::TReplyStatus::E status, std::optional<TString> errorMessage);
 
+        bool PersistentBufferPreprocessWrite(TEvWritePersistentBuffer::TPtr ev);
         void ProcessPersistentBufferWrite(TEvWritePersistentBuffer::TPtr ev);
         void ProcessPersistentBufferBatchWrite(std::vector<TEvWritePersistentBuffer::TPtr> evs);
         double GetPersistentBufferFreeSpace();
--- a/ydb/core/blobstorage/ddisk/ddisk_actor.h
+++ b/ydb/core/blobstorage/ddisk/ddisk_actor.h
@@ -732,6 +732,9 @@ namespace NKikimr::NDDisk {
                 TRope JoinData(ui32 sectorSize);
             };
 
+            bool BatchWrite = false;
+            bool BatchReady = false;
+            TPersistentBufferSectorInfo BatchHeaderSectorInfo;
             std::vector<TRecord> Records;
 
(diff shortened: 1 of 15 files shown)


ydb/core/blobstorage/ddisk/ddisk_actor.h
```diff
@@ -711,42 +712,55 @@ namespace NKikimr::NDDisk {
...
-            TRope JoinData(ui32 sectorSize);
         };
 
         struct TPersistentBufferEraseInflight {
             ui64 EraseCookie;
             std::vector<ui64> OperationsCookie;
         };
 
+        ui64 PersistentBufferBatchWriteCookie = 0;
+        std::unordered_map<TPersistentBufferLocation, std::unordered_set<TPersistentBufferRecordId>> PersistentBufferHeaders;
```
eivanov89: Here you change std::vector value to std::set. And now you have std::unordered_map of sets. I would consider:
1. Elimitating set.
2. swithcing to absl hash map.

ydb/core/blobstorage/ddisk/ddisk_actor_persistent_buffer.cpp
```diff
@@ -35,7 +35,10 @@ namespace NKikimr::NDDisk {
         SectorSize = DiskFormat->SectorSize;
         Y_ABORT_UNLESS(SectorSize >= sizeof(TPersistentBufferFastErases));
         Y_ABORT_UNLESS(SectorSize >= sizeof(TPersistentBufferBarriers));
-        Y_ABORT_UNLESS(SectorSize >= sizeof(TPersistentBufferLsnRecordHeader));
+        Y_ABORT_UNLESS(SectorSize >= sizeof(TPersistentBufferHeader)
```
eivanov89: Please consider adding some constexpr to move the math to header and properly name it

ydb/core/blobstorage/ddisk/ddisk_actor.h
```diff
@@ -711,42 +712,55 @@ namespace NKikimr::NDDisk {
...
         std::unordered_map<ui64, TPersistentBufferDiskOperationInFlight> PersistentBufferDiskOperationInflight;
-        std::unordered_map<TPersistentBufferRecordId, std::vector<ui64>> PersistentBufferWriteInflightsByRecord;
+
+        // map record to operation cookie + record in inflight position
+        std::unordered_map<TPersistentBufferRecordId, std::vector<std::tuple<ui64, ui32>>> PersistentBufferWriteInflightsByRecord;
```
eivanov89: absl

ydb/core/blobstorage/ddisk/ddisk_actor_persistent_buffer.cpp
```diff
@@ -775,6 +903,186 @@ namespace NKikimr::NDDisk {
...
+            auto* partOp = static_cast<TPersistentBufferPartIoOp*>(op.get());
+            partOp->SetCookie(PersistentBufferBatchWriteCookie);
+            partOp->SetPartCookie(cookie);
+            partOp->PrepareWrite(std::move(data), diskOffset, chunkIdx, offset);
+
+            DirectUringOp(op);
+        }
+    }
+
+    void TDDiskActor::ProcessPersistentBufferBatchWrite() {
```
eivanov89, line 1020: If I understand the code correctly, you set the timer and timer event calls this handler. And here you write the batch header. However, you must write the header only when all writes in this batch succeeded. Consider the case:
1. You started n writes.
2. timeout, you write the header.
3. we crash

It might happen that you have header written while not all data in the batch has been written.

I would consider a different logic, which might simplify the overall code:
1. You should have CurrentBatch – here you accumulate the data. You close the batch by timeout (or size) and move it to InflightBatches.
2. InflightBatches – closed, being written to the disk. When all data is written, write the header. When header is written – reply OK, forget the batch.

eivanov89 (changes requested):

ydb/core/blobstorage/ddisk/ut/ddisk_actor_batch_write_ut.cpp
```diff
@@ -0,0 +1,1076 @@
...
+//     TPersistentBufferHeader { BatchSize = N, ... }
+//     N × TPersistentBufferLsnRecordHeader
+//     N × (sectorsCnt × TPersistentBufferSectorInfo)   [data sector locations]
+//   Sectors 1..K (data sectors for each record, written immediately)
+//
+// ProcessPersistentBufferBatchWrite() is called on the wakeup timer and
+// writes the unified header sector.  Only after BOTH all data parts AND the
+// header part complete does HandleWritePart() reply to the callers.
+//
+// IMPORTANT: The batch path requires an in-flight operation when the second
```
eivanov89, line 21: I would consider changing this behaviour. That might break batch logic. Currently chances you get a disk reply within 10 usec are low. But it might happen in future with faster disks. Thus, you have a default batch window 10 usec, but "opening" write might return earlier.

ydb/core/blobstorage/ddisk/ut/ddisk_actor_batch_write_ut.cpp
```diff
@@ -0,0 +1,1076 @@
...
+// write arrives.  The correct test sequence is:
+//   1. Send write A → PDisk write arrives (keep it pending — A is in-flight)
+//   2. Send write B while A is still in-flight → B goes through batch path
+//   3. Complete A's PDisk write → A replies
+//   4. Complete B's data write
+//   5. Wakeup fires → header write
+//   6. Complete header write → B replies
```
eivanov89, line 28: I would add a test with multiple tablets writing to PB. I believe I've seen some places in the code where creds is a key in batch related map. The logic is not obvious to me and the test is missing.

eivanov89 (approved):

vazhem (approved):

PR #43138, opened 2026-06-10
Fix viewer/browse_db field used

rosik:
### Changelog category 

### Description for reviewers 

`viewer/browse_db` no longer relies on `Table.SplitBoundary` to compute table partition count.
Previously, `browse_db` did not explicitly request `SetReturnBoundaries`, but still derived partition count from `SplitBoundarySize() + 1`. ReturnBoundaries is false by dafault, but is still present in responce because of a bug fixed in #43001

Now partition count is taken from `PathDescription.TableStats.PartCount` which avoids requiring boundary payloads at all.

- The problem revealed spontaneously in scope of https://github.com/ydb-platform/ydb-embedded-ui/issues/3988

### DIFF
--- a/ydb/core/viewer/browse_db.h
+++ b/ydb/core/viewer/browse_db.h
@@ -137,9 +137,6 @@ public:
                     if (pbTable.HasUniformPartitionsCount()) {
                         pbCommon.SetPartitions(pbTable.GetUniformPartitionsCount());
                     }
-                    if (pbTable.SplitBoundarySize() > 0) {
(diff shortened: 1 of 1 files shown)


alexvru (approved):

ydb/core/viewer/browse_db.h
CyberROFL, line 157: `TableStats.PartCount` — это не количество партиций, а количество партов в конкретной таблетке.

PR #43284, opened 2026-06-11
EXT-1820 Add ExpectedSlotSize to PDiskConfig

SammyVimes:
### Changelog entry 

Added support for configuring BlobStorage PDisks by expected slot size. When `ExpectedSlotSize` is set, YDB derives the expected slot count from the disk size and uses it for VDisk capacity accounting and placement.

### Changelog category 

* Experimental feature

### Description for reviewers 

This change adds `ExpectedSlotSize` as an alternative to slot-count based PDisk configuration.

`ExpectedSlotSize` is mutually exclusive with `ExpectedSlotCount` and `SlotSizeInUnits` in configuration validation paths, including YAML config, BSC host config, and distconf. When it is set, the expected slot count is inferred from the PDisk size and capped at 64 slots. If BSC has not received disk metrics yet, the inferred slot count stays zero until the disk size is known.

The new setting is propagated through NodeWarden, PDisk, BSC group mapper/storage stats, whiteboard/sys view/viewer/config APIs. Slot-size based PDisks also limit per-VDisk expected owner size by the configured expected slot size.

Added tests for validation, slot-count inference, missing-metrics behavior, slot-count cap, owner-size limiting, distconf/yaml compatibility, and sys view exposure.


rosik:
Looks like this feature duplicates `infer_pdisk_slot_count_settings` option in `blob_storage_config` section (I suppose you don't really want to duplicate the same ExpectedSlotSize for every 100 drives in config)
RFC: https://github.com/ydb-platform/ydb-rfc/blob/main/0127_config_infer_pdisk_slot_count.md
Implemented in https://github.com/ydb-platform/ydb/pull/32113
Available since 26.1

SammyVimes:
> Looks like this feature duplicates `infer_pdisk_slot_count_settings` option in `blob_storage_config` section (I suppose you don't really want to duplicate the same ExpectedSlotSize for every 100 drives in config) RFC: https://github.com/ydb-platform/ydb-rfc/blob/main/0127_config_infer_pdisk_slot_count.md Implemented in #32113 Available since 26.1

Thanks, I agree this overlaps with `InferPDiskSlotCountSettings` in the part where we derive `ExpectedSlotCount` from disk size.

However, the intended semantics here are different. `InferPDiskSlotCountSettings.UnitSize` is not a hard per-VDisk size limit: for a `6.4T` disk with `UnitSize=200G` and `MaxSlots=16` it produces `SlotSizeInUnits=2`, `ExpectedSlotCount=16`, and the effective owner quota becomes 400G.

`ExpectedSlotSize` in this PR is meant to be the actual per-VDisk hard limit. The chunk keeper gets `ExpectedOwnerSize` and `RedistributeQuotas()` uses that fixed value directly instead of `TotalSize / ExpectedOwnerCount`. So extra capacity on larger PDisks does not silently increase VDisk quota.

So yes, if all we wanted was “infer slot count from disk size”, the existing setting would be enough. But this PR is trying to add a simpler byte-based quota model: configure one expected slot size, calculate slot count from it, and enforce that size as the VDisk limit.

We don't have to set it for each pdisk, there is a `DefaultHostPDiskConfig` in `blobstorage_config.proto`

ydb/core/blobstorage/pdisk/blobstorage_pdisk_config.h
```diff
@@ -496,7 +509,7 @@ struct TInferPDiskSlotCountSettingsForDriveType {
     }
 
     explicit operator bool() const {
-        return UnitSize && MaxSlots;
+        return SlotSize || (UnitSize && MaxSlots);
```
rosik:
```suggestion
        return (UnitSize || SlotSize) && MaxSlots;
```
It's strange to ignore MaxSlots and use hardcoded value 64 for handling SlotSize.

ydb/core/base/blobstorage.h
```diff
@@ -36,6 +36,17 @@ namespace NKikimr {
...
+
+inline ui32 CalculateExpectedSlotCountFromExpectedSlotSize(ui64 diskSize, ui64 expectedSlotSize) {
+    if (!expectedSlotSize) {
+        return 0;
+    }
+    const ui64 slotCount = diskSize / expectedSlotSize;
+    return slotCount > MaxExpectedSlotCountFromExpectedSlotSize
+        ? MaxExpectedSlotCountFromExpectedSlotSize
+        : static_cast<ui32>(slotCount);
+}
```
rosik: It's inconsistent with the configurable `MaxSlots` parameter in `InferPDiskSlotCountSettings`. I suggest moving this logics to node_warden_pdisk.cpp, next to `void TNodeWarden::InferPDiskSlotCount()`

ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
```diff
@@ -1972,7 +2002,7 @@ bool TPDisk::YardInitForKnownVDisk(TYardInit &evYardInit, TOwner owner) {
     }
     ui64 writeBlockSize = ForsetiOpPieceSizeCached;
     ui64 readBlockSize = ForsetiOpPieceSizeCached;
-    ui32 ownerWeight = Cfg->GetOwnerWeight(evYardInit.GroupSizeInUnits);
+    ui32 ownerWeight = GetOwnerWeight(evYardInit.GroupSizeInUnits);
```
rosik: I didn't get the necessity of this change. What's the purpose?

ydb/core/blobstorage/base/infer_pdisk_slot_count_settings.h
```diff
@@ -0,0 +1,49 @@
...
+inline std::optional<TString> ValidateInferPDiskSlotCountSettingsForDriveType(
+        const NKikimrBlobStorage::TInferPDiskSlotCountForDriveTypeSettings& settings,
+        TStringBuf context) {
+    const bool hasSlotSize = settings.HasSlotSize() && settings.GetSlotSize();
+    const bool hasLegacySettings = settings.HasUnitSize() || settings.HasMaxSlots()
+        || settings.HasPreferInferredSettingsOverExplicit();
+    if (hasSlotSize && hasLegacySettings) {
+        return TStringBuilder() << context
+            << " SlotSize is mutually exclusive with UnitSize, MaxSlots and PreferInferredSettingsOverExplicit"
+            << " in InferPDiskSlotCountSettings";
```
rosik, line 33: I would expect that SlotSize is mutually exclusive with UnitSize, but  MaxSlots is always complementary and  mandatory

ydb/core/blobstorage/nodewarden/blobstorage_node_warden_ut.cpp
```diff
@@ -1192,6 +1239,153 @@ Y_UNIT_TEST_SUITE(TBlobStorageWardenTest) {
...
+        const ui64 updatedExpectedSlotSize = 800ull << 30;
+        UpdateInferPDiskSlotSizeSettings(runtime, realNodeWarden, updatedExpectedSlotSize);
+        CheckInferredPDiskSettings(runtime, fakeWhiteboard, fakeNodeWarden,
+            pdiskId, 3, 0u, updatedExpectedSlotSize);
+
+        pdiskConfig.SetExpectedSlotCount(17);
+        CreatePDisk(runtime, 0, pdiskPath, 0, pdiskId, 0,
+            &pdiskConfig, realNodeWarden);
+        CheckInferredPDiskSettings(runtime, fakeWhiteboard, fakeNodeWarden,
+            pdiskId, 17, 0u);
```
rosik, line 1348: I suggest adding here one more step with `PreferInferredSettingsOverExplicit=true`

ydb/core/blobstorage/nodewarden/blobstorage_node_warden_ut.cpp
```diff
@@ -1144,6 +1160,37 @@ Y_UNIT_TEST_SUITE(TBlobStorageWardenTest) {
...
+        auto* inferRotSettings = inferSettings->MutableRot();
+
+        inferRotSettings->SetSlotSize(slotSize);
+
+        TActorId sender = runtime.AllocateEdgeActor();
+        runtime.Send(new IEventHandle(realNodeWarden, sender, request.release()));
+
+        auto response = runtime.GrabEdgeEventRethrow<NConsole::TEvConsole::TEvConfigNotificationResponse>(sender);
+        Y_UNUSED(response);
+    }
```
rosik: Naming it "Infer PDisk SlotSize" is incorrect. SlotSize isn't inferred.  SlotCount is. I suggest extending the above function

ydb/core/blobstorage/nodewarden/distconf_generate.cpp
```diff
@@ -161,6 +161,9 @@ namespace NKikimr::NStorage {
         struct TPDiskInfo {
             NKikimrBlobStorage::TNodeWardenServiceSet::TPDisk Record;
             ui32 UsedSlots = 0;
+            ui32 ExpectedSlotCount = 0;
+            ui32 SlotSizeInUnits = 0;
+            ui64 ExpectedSlotSize = 0;
```
rosik: Could you explain the reasons for editing distconf? I didn't do anything similar. Supposing SlotSize and UnitSize are to be handled similarly, is it missing something important for InferPDiskSlotCountSettings support in general?

rosik (changes requested):

PR #43668, opened 2026-06-17
Do not check database for whoami

NewBediver:
### Changelog category 

### Description for reviewers 

https://nda.ya.ru/t/UBYLFQQ77fbs89

### DIFF
--- a/ydb/core/viewer/tests/canondata/result.json
+++ b/ydb/core/viewer/tests/canondata/result.json
@@ -17484,8 +17484,17 @@
     },
     "test.TestViewer.test_waiting_for_cluster_ready": {},
     "test.TestViewer.test_whoami_database": {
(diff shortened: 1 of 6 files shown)


yurikiselev (approved):

ijon (approved):

adameat:
I have strong concerns that this change will brake UI flow. Because UI primary uses whoami call to detect whenever a user has access to a database or not...

NewBediver:
> I have strong concerns that this change will brake UI flow. Because UI primary uses whoami call to detect whenever a user has access to a database or not...

We discussed further steps at the meeting 22.06.2026. This PR is no longer relevant. The necessary permissions will be issued on the problematic clusters, and the diagnostic tools will be improved.

PR #43760, opened 2026-06-17
Fix BSC data race when deleting PDisk

alexvru:
### Changelog entry 

Fix BSC data race when deleting PDisk

### Changelog category 

### Description for reviewers 

This fixes data race arising in BS controller while deleting PDisks with VSlots in deletion state.


mregrock (approved):
LGTM

PR #43848, opened 2026-06-18
fixed conflict

xyliganSereja:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


ydb/core/tx/columnshard/engines/reader/common_reader/iterator/fetching.cpp
```diff
@@ -53,6 +53,7 @@ TStepAction::TStepAction(
     }
 }
 
+NO_SANITIZE_THREAD
```
dorooleg, line 56: Это актуально?

dorooleg (changes requested):

PR #44211, opened 2026-06-22
Added authentication and database checks in config service

n00bcracker:
Added authorization check to config fetch.

Fetching a database config now requires the caller to be a cluster administrator or, when EnableDatabaseAdmin flag is set, the database administrator.


mregrock (approved):
LGTM

PR #44626, opened 2026-06-26
EXT-2313 Add timeout, precharge and continuation to the tablet_flat DbMon transaction

SammyVimes:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Opening Executor DB internals with RowsOffset=16842699 caused FLAT_EXECUTOR to consume 100% CPU for hours. 
Adding timeout should help already, since this transaction is internal UI only, it's not that important. 
Next I've added precharge (to avoid reading page by page) and continuation (in order to not start from the beginning each time we retry transaction on page fault)

### DIFF
--- a/ydb/core/tablet_flat/flat_executor_db_mon.cpp
+++ b/ydb/core/tablet_flat/flat_executor_db_mon.cpp
@@ -1,5 +1,7 @@
 #include "flat_executor.h"
 
+#include <ydb/core/base/appdata.h>
+
 #include <yql/essentials/types/binary_json/read.h>
 #include <yql/essentials/types/dynumber/dynumber.h>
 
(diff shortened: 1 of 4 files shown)


ydb/core/tablet_flat/flat_executor_db_mon.cpp
```diff
@@ -122,6 +181,30 @@ class TExecutor::TTxExecutorDbMon : public TTransactionBase<TExecutor> {
...
+                        RenderedRows.clear();
+                    }
+                    const bool canResumeByKey = lookup != NTable::ELookup::ExactMatch && !tableInfo->KeyColumns.empty();
+                    if (LastSeenKey && (LastSeenKey->GetCells().size() != tableInfo->KeyColumns.size())) {
+                        // LastSeenKey is captured from iterator GetKey() after a successful step.
+                        // It must always be a full primary key for the same table.
+                        // Better stop here in this case.
+                        Y_TABLET_ERROR("Cannot resume offset scan because key size changed: "
+                            << LastSeenKey->GetCells().size() << " vs " << tableInfo->KeyColumns.size());
+                    }
```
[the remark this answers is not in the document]
SammyVimes: No, Y_TABLET_ERROR shuts tablet down (it will be restarted) using an exception

ydb/core/tablet_flat/flat_executor_db_mon.cpp
```diff
@@ -10,16 +12,73 @@
...
+            ++itColumn;
+        }
+
+        return key;
+    }
+
+public:
     bool Execute(TTransactionContext &txc, const TActorContext &ctx) override {
+        if (IsTimedOut()) {
+            return true;
```
dcherednik, line 79: Is it save return here without any response?

SammyVimes: Yes: if we hit timeout here, monitoring proxy (forwarding actor) probably already has responded with timeout (or will do it in the nearest future if we have a clock skew between nodes)

lex007in (approved):

PR #45175, opened 2026-07-01
row_dispatcher: reuse same credentials provider for all partitions

yumkam:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Saves some cycles wasted in (re)parsing structured token and a call to CreateProvider.
Only affects first topic session creation for single read actor (assuming 4k partitions and 32 tasks-per-stage, ~100 calls)


yumkam:
With provider reuse, just very minor optimization; hence dropping

PR #45422, opened 2026-07-03
topic write_session: handle exceptions from GetAuthInfo

yumkam:
Note: write_session constructor may be called from federated_topic write session implementation in the middle of "federated session", we cannot just throw exception from there (or its handling should be also added to federated_topic code -- there are none currently)

### Changelog entry 

...

### Changelog category 

### Description for reviewers 

NOT READY-USE SOLUTION (*probably* works, but feels fragile), just point to problematic code, requires careful redesign
ALSO NOTE: `topic/*write_sessioin*` is mostly copied from persqueue_public, same problem present there too,


yumkam:
solved elsewhere

PR #45585, opened 2026-07-05
Removed filling users' old format hashes in requests to SchemeShard

n00bcracker:
Removed code for forward compatibility of new hash format since all SchemeShards were migrated to new format.


ijon (approved):

nikvas0 (approved):
kqp ok

PR #45655, opened 2026-07-06
EXT-1820 Add ExpectedSlotSize to PDiskConfig

SammyVimes:
### Changelog entry 

Added support for configuring BlobStorage PDisks by target slot size via `infer_pdisk_slot_count.<type>.slot_size` and per-drive `expected_slot_size`.

### Changelog category 

* Experimental feature

### Description for reviewers 

This change adds slot-size based PDisk configuration as an alternative to explicit slot-count configuration.

Operators can define a global policy per drive type with `infer_pdisk_slot_count.<type>.slot_size` and `max_slots`, or override individual disks with `expected_slot_size` and `max_slots`. NodeWarden derives the effective slot count as `min(drive_size / slot_size, max_slots)`.

`ExpectedSlotSize` is mutually exclusive with `ExpectedSlotCount` and `SlotSizeInUnits`, and requires `MaxSlots`. The same validation is added for YAML config, generic config validation, BSC config, and distconf.

Global inference does not override explicit per-disk settings unless `prefer_inferred_settings_over_explicit` is enabled, so migration can be done gradually.

The new setting is propagated through NodeWarden, PDisk, BSController, DistConf, whiteboard, sys views, viewer, and config APIs. PDisk uses `ExpectedSlotSize` as the per-VDisk quota, rounded down to whole chunks.

Added tests for validation, slot-count inference, migration behavior, capacity accounting, owner-size limiting, and observability.

RFC: https://github.com/ydb-platform/ydb-rfc/blob/main/174_config_expected_pdisk_slot_size.md (https://github.com/ydb-platform/ydb-rfc/pull/286)


ydb/core/blobstorage/base/pdisk_config_validation.h
```diff
@@ -0,0 +1,35 @@
...
+            << " in PDiskConfig";
+    }
+
+    if (hasExpectedSlotSize && !hasMaxSlots) {
+        return TStringBuilder() << context
+            << " ExpectedSlotSize requires MaxSlots"
+            << " in PDiskConfig";
+    }
+
+    return {};
```
rosik, line 38: It does not reject MaxSlots set without ExpectedSlotSize. Is it intentional? The analogous check is present in ValidateInferPDiskSlotCountSettingsForDriveType() (line 30):

```
    if (hasMaxSlots && !hasSlotSize && !hasUnitSize) {
        return TStringBuilder() << context
            << " MaxSlots requires SlotSize or UnitSize"
            << " in InferPDiskSlotCountSettings";
    }
```

SammyVimes: Not intentional, `MaxSlots` alone is a silent no-op everywhere, I added the symmetric check (MaxSlots requires ExpectedSlotSize) to ValidatePDiskConfig, yaml parser and cfg tool (with tests)

ydb/core/blobstorage/pdisk/blobstorage_pdisk_config.h
```diff
@@ -435,6 +439,14 @@ struct TPDiskConfig : public TThrRefBase {
             ExpectedSlotCount = cfg->GetExpectedSlotCount();
         }
 
+        if (cfg->HasExpectedSlotSize()) {
+            ExpectedSlotSize = cfg->GetExpectedSlotSize();
+        }
+
+        if (cfg->HasMaxSlots()) {
+            MaxSlots = cfg->GetMaxSlots();
+        }
```
rosik, line 450: One remaining observation

MaxSlots in config overlay here sets MaxSlots only if cfg->HasMaxSlots(). The ValidatePDiskConfig validator now rejects MaxSlots without ExpectedSlotSize. This means if an operator sets MaxSlots alone in PDiskConfigOverlay (as a system-wide connection cap), validation would reject it.

In practice, PDiskConfigOverlay is applied after ValidatePDiskConfig is called in the config validation pipeline (the validator runs on the storage config proto, not on the fully-merged TPDiskConfig runtime struct), so this should not be a problem operationally. But it is worth noting in the documentation that MaxSlots in TPDiskConfig (proto/YAML) is only meaningful with ExpectedSlotSize.

ydb/core/blobstorage/nodewarden/node_warden_mon.cpp
```diff
@@ -153,6 +153,7 @@ void TNodeWarden::RenderWholePage(IOutputStream& out) {
                     TABLEH() { out << "Category"; }
                     TABLEH() { out << "Temporary"; }
                     TABLEH() { out << "Pending"; }
+                    TABLEH() { out << "PDiskConfig warning"; }
```
rosik, line 156: I suggest a follow-up enhancement the following concern:

Current observability for `PDiskConfigWarning` is poor. **There is no Prometheus/Solomon metric.** The warning is only available through two channels today:

1. Logs (ERROR level, immediately searchable)
All four warning conditions log at `YDB_LOG_ERROR` with unique markers:

| Condition | Marker |
|---|---|
| `ExpectedSlotSize` + `ExpectedSlotCount`/`SlotSizeInUnits` conflict | `NW113` |
| `ExpectedSlotSize` without `MaxSlots` | `NW117` |
| Drive size unknown when `ExpectedSlotSize` set | `NW115` |
| Drive smaller than `ExpectedSlotSize` | `NW118` |

To find any PDisk with a warning across the cluster, search logs:
```
grep -E '"NW113"|"NW117"|"NW115"|"NW118"' /path/to/logs
```

2. NodeWarden HTML debug page (per-node)
[`node_warden_mon.cpp`](ydb/core/blobstorage/nodewarden/node_warden_mon.cpp:179) renders the warning string in the PDisk table at `http://<node>:8765/node_warden`. Not scalable for cluster-wide monitoring.

The possible implementation strategies I see:

- whiteboard → sysview pipeline that already exists for other PDisk fields
- BSController tablet counters (to ensure it's 0). My recent example: https://github.com/ydb-platform/ydb/pull/22567/changes

ydb/core/blobstorage/pdisk/blobstorage_pdisk_config.h
```diff
@@ -466,7 +478,11 @@ struct TPDiskConfig : public TThrRefBase {
     }
 
     ui32 GetOwnerWeight(ui32 groupSizeInUnits) {
-        return TPDiskConfig::GetOwnerWeight(groupSizeInUnits, SlotSizeInUnits);
+        return TPDiskConfig::GetOwnerWeight(groupSizeInUnits, SlotSizeInUnits, ExpectedSlotSize);
+    }
+
+    static ui32 GetOwnerWeight(ui32 groupSizeInUnits, ui32 slotSizeInUnits, ui64 expectedSlotSize) {
+        return expectedSlotSize ? 1 : GetOwnerWeight(groupSizeInUnits, slotSizeInUnits);
     }
```
rosik, line 492: We've already discussed it in chat, another follow-up suggestion, here is the link for history: https://gist.github.com/rosik/bce59d03afa6df385ba42c3a46a74ec6

ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
```diff
@@ -2278,12 +2302,13 @@ void TPDisk::YardResize(TYardResize &ev) {
 void TPDisk::ProcessChangeExpectedSlotCount(TChangeExpectedSlotCount& request) {
     TGuard<TMutex> guard(StateMutex);
     ExpectedSlotCount = request.ExpectedSlotCount;
-    Cfg->ExpectedSlotCount = request.ExpectedSlotCount;
+    ExpectedSlotSize = request.ExpectedSlotSize;
+    NormalizeExpectedSlotSettings();
     Cfg->SlotSizeInUnits = request.SlotSizeInUnits;
-    Keeper.SetExpectedOwnerCount(ExpectedSlotCount);
+    Keeper.SetExpectedOwnerSettings(ExpectedSlotCount, GetExpectedOwnerSizeInChunks());
```
rosik, line 2309: `NormalizeExpectedSlotSettings()` — Why and Correctness

Why it exists — `TPDisk` has **two parallel representations** of slot settings:

| Location | Purpose |
|---|---|
| [`TPDisk::ExpectedSlotCount`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.h:231), [`TPDisk::ExpectedSlotSize`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.h:232) | Authoritative runtime state, guarded by `StateMutex`, updated live |
| [`Cfg->ExpectedSlotCount`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_config.h:95), [`Cfg->ExpectedSlotSize`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_config.h:95) | Shared `TPDiskConfig` struct, readable externally |

[`NormalizeExpectedSlotSettings()`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp:127) exists to **write runtime state back into `Cfg`** so that code reading `Cfg->` sees current effective values.

It is called in exactly two places:

1. [`blobstorage_pdisk_impl_log.cpp:1735`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl_log.cpp:1735) — log restore
```cpp
NormalizeExpectedSlotSettings();                     // ← sync first
params.ExpectedOwnerCount = Cfg->ExpectedSlotCount;  // ← then read Cfg
params.ExpectedOwnerSize  = GetExpectedOwnerSizeInChunks();
```
The log restore code reads `Cfg->ExpectedSlotCount` (not the TPDisk member) to build [`TKeeperParams`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_keeper_params.h:23). The sync call before it is defensive: during normal boot (before any live updates) this is a no-op because runtime fields were initialized from `Cfg` in the constructor.

2. [`ProcessChangeExpectedSlotCount`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp:2302) — live update without restart
```cpp
ExpectedSlotCount = request.ExpectedSlotCount;   // update runtime member
ExpectedSlotSize  = request.ExpectedSlotSize;    // update runtime member
NormalizeExpectedSlotSettings();                  // ← sync Cfg
Cfg->SlotSizeInUnits = request.SlotSizeInUnits;  // update Cfg directly (no member)
Keeper.SetExpectedOwnerSettings(...);
```
This is the **primary motivation**: after a live `TEvChangeExpectedSlotCount`, external readers (tests via `GetPDiskConfig()`, HTML debug page) must see the updated values through `Cfg`.

---

Is it correct?

**Functionally yes**, but two design awkwardnesses exist:

**Awkwardness 1 — Redundant call at log init.**  
The call at [`blobstorage_pdisk_impl_log.cpp:1735`](ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl_log.cpp:1735) is superfluous in the normal boot path (values already match). A cleaner approach: read `TPDisk::ExpectedSlotCount` directly instead of `Cfg->ExpectedSlotCount` in the log restore path, eliminating the need for a sync there entirely.

**Awkwardness 2 — Inconsistent treatment of `SlotSizeInUnits`.**  
```cpp
Cfg->SlotSizeInUnits = request.SlotSizeInUnits;  // line 2307 — no TPDisk member
```
`SlotSizeInUnits` has **no corresponding `TPDisk` runtime member** — it lives only in `Cfg` and is written there directly. Meanwhile `ExpectedSlotCount` and `ExpectedSlotSize` each have both a `TPDisk` member and a `Cfg` field. So `NormalizeExpectedSlotSettings()` syncs two of the three mutable slot fields but not the third. The naming and the pattern are not uniform.

**Summary:** The function is correct and safe. Its existence is a workaround for the log-restore code reading `Cfg->ExpectedSlotCount` instead of the PDisk member directly. The dual-tracking means callers must remember to invoke it before reading from `Cfg`, and `SlotSizeInUnits` breaks the symmetry because it was never given a TPDisk runtime member counterpart.

ydb/core/protos/node_whiteboard.proto
```diff
@@ -153,6 +153,7 @@ message TPDiskStateInfo {
     optional uint32 SlotSizeInUnits = 25;
     optional double PDiskUsage = 26; // 100.0 * SharedQuota.Used / SharedQuota.HardLimit
     optional NKikimrBlobStorage.TPDiskSpaceColor.E PDiskCapacityAlert = 27; // Max(SharedQuota.Color, CommonLog.Color)
+    optional uint64 ExpectedSlotSize = 28 [(DefaultField) = true];
```
rosik, line 156: Afaik whiteboard discourages adding new fields as default

ydb/core/viewer/tests/canondata/result.json
```diff
@@ -852,6 +852,8 @@
                                 "ChangeTime": "not-zero-number-text",
                                 "Device": "Green",
                                 "EnforcedDynamicSlotSize": "not-zero-number-text",
+                                "ExpectedSlotCount": 0,
+                                "ExpectedSlotSize": "0",
```
rosik, line 856: I notice ExpectedSlotSize is a string?

rosik (dismissed):

ydb/core/viewer/tests/canondata/result.json
SammyVimes: Yup, 64 bit ints can only be represented as strings in JS, since there is only type Number in JS (which is a float actually)

CyberROFL (dismissed):

ijon (dismissed):

ydb/core/viewer/viewer_database_stats.h
```diff
@@ -424,7 +427,11 @@ class TJsonDatabaseStats : public TViewerPipeClient {
                                 }
                                 slotSize = itPDisk->second.GetTotalSize() / slotCount;
                             }
-                            storageTotal += slotSize * TPDiskConfig::GetOwnerWeight(itGroup->second.GetInfo().GetGroupSizeInUnits(), itPDisk->second.GetSlotSizeInUnits());
```
Enjection, line 427: Would be great to move this math elsewhere as a func similar to `CalcAvailableAndDiskSpace` and covered with ut to have explicit contract.

Enjection (changes requested):
No major findings on configs or viewer side, the only thing I want is small math extraction and test coverage for it.

Enjection (dismissed):

CyberROFL (dismissed):

va-kuznecov (dismissed):

ijon (dismissed):

PR #45671, opened 2026-07-06
Fix false-positive VERIFY on key deletion in testshard

serbel324:
### Changelog entry 

Fix false-positive VERIFY assertion in testshard triggered by Key deletion.
https://github.com/ydb-platform/ydb/issues/45337

### Changelog category 

* Bugfix


mregrock (approved):
LGTM

PR #45733, opened 2026-07-07
Add cpu_topology library

tindarid:
### Changelog entry

Add CPU topology library to consume data from `/sys/devices/system/cpu/cpu*/topology`

### Changelog category

* New feature

### Description for reviewers 
* Add AMD/Intel snapshots (and script for retrieval)
* Add parsing and grouping based on CPU groups (package, die, etc)
* Add PlacementGroups, which will be used for placement of related actors within 1 process (based on L3 cache grouping, which matched CCD placement for AMD)
* Add tests


ydb/library/actors/util/cpu_topology.cpp
```diff
@@ -0,0 +1,409 @@
...
+        if (!path.IsDirectory()) {
+            return {};
+        }
+
+        TVector<TString> children;
+        path.ListNames(children);
+        std::sort(children.begin(), children.end());
+        return children;
+    } catch (...) {
+        return {};
```
dcherednik: Is it ok we tread error as empty directory?

ydb/library/actors/util/cpu_topology.cpp
```diff
@@ -0,0 +1,409 @@
...
+            return std::unexpected{std::move(clusterId).error()};
+        }
+        cpu.ClusterId = *clusterId;
+        auto clusterCpus = ReadCpuMask(topologyRoot / "cluster_cpus_list");
+        if (!clusterCpus) {
+            return std::unexpected{std::move(clusterCpus).error()};
+        }
+        cpu.ClusterCpus = std::move(*clusterCpus);
+
+        auto l3Cache = ParseL3CacheGroup(root, *cpuId);
```
dcherednik, line 253: Does it work if we have no L3 cache?

dcherednik: Probably it is better add unit test with topology from VM/laptop cpu to make sure this code will work on any hardware.

ydb/library/actors/util/cpu_topology.cpp
tindarid: fixed (also for ReadFile)

ydb/library/actors/util/cpu_topology.cpp
tindarid: Added architecture with CPU without L3

Fallback: we will add it in future PRs (also tackle Windows)

dcherednik (approved):

PR #45790, opened 2026-07-07
Introduce Interconnect TCP session v2 stub

alexvru:
### Changelog entry 

Introduce Interconnect TCP session v2 stub

### Changelog category 

### Description for reviewers 

This patch introduces Interconnect TCP session v2 stub without any actual implementation.


mregrock (approved):
LGTM

PR #45828, opened 2026-07-07
EXT-2337 Fix CPU problem when subscribing to kesus quoter resource

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Here we iterate over all resources inside one quoter proxy session: https://github.com/ydb-platform/ydb/blob/87dbaabebf9b75170b08645b4c2e8431f54c489b/ydb/core/kesus/tablet/quoter_resource_tree.cpp#L1415 

So, they are all resources used by one YDB node in this Kesus. It can be a significant amount.

As for the next stage, I want to support safe deletion of these session objects. It will be made in the next PR


CyberROFL (commented):
Не собирается.

Enjection (approved):

CyberROFL (approved):

PR #46029, opened 2026-07-09
Fix decommission race bug

alexvru:
### Changelog entry 

Fix decommission race bug

### Changelog category 

### Description for reviewers 

This patch fixes decommission data race bug occuring when VDisk turns into 'deleting' state, but not yet deleted.


mregrock (approved):
LGTM

PR #46154, opened 2026-07-10
Discard .debug_aranges in ydbd binary

maximyurchuk:
### Changelog entry

Fix ydbd linker failure caused by DWARF-32 4 GB limit on `.debug_info` references from `.debug_aranges`.

### Changelog category

* Bugfix

#### Symptom

`ydb/apps/ydbd` fails to link:

```
ld.lld: error: /.../libpthread_nonshared.a(pthread_atfork.oS):(.debug_aranges+0x6):
  relocation R_X86_64_32 out of range: 4302595903 is not in [0, 4294967295]
clang++: error: linker command failed with exit code 1
```

#### Root cause

The overflowing byte at `(.debug_aranges+0x6)` is the `debug_info_offset` field in the `.debug_aranges` CU header — a 4-byte DWARF-32 offset that must fit into 32 bits (`R_X86_64_32`). Once the combined `.debug_info` section in the linked binary passes 4 GB, any CU header whose `debug_info_offset` would land beyond that mark cannot be represented, and the relocation fails.

Which specific object trips the check depends on link order: in our case the sysroot's `pthread_atfork.oS` happens to be the unlucky one, but it isn't the cause. `.debug_info` in `ydbd` has simply grown past the 4 GB DWARF-32 limit.

#### Why discarding `.debug_aranges` is the surgical fix

The R_X86_64_32 relocations that overflow live *inside* `.debug_aranges` header entries — that's the only DWARF section whose header carries a 4-byte pointer into `.debug_info`. Drop `.debug_aranges` at link time and there is nothing to relocate; the overflow disappears.

Nothing else in the binary points into `.debug_info` with a 4-byte offset:
- `.debug_line` (DWARF-4) doesn't have a `debug_info_offset` in its header.
- `.debug_pubnames` / `.debug_pubtypes` / `.debug_names` would, but ya's default clang doesn't emit them.
- `.debug_loc` / `.debug_ranges` are indexed *by* `.debug_info`, not the other way around.

So dropping `.debug_aranges` is sufficient to unblock linking and lets `.debug_info` grow past 4 GB freely (the current build already sits at ~4.006 GB after this change).

#### What `.debug_aranges` is, and who was writing to it

`.debug_aranges` is a lookup optimization: for each CU that opts in, it publishes a table of (address, length) → CU header offset, so tools can quickly answer "which CU does PC X belong to?" without scanning `.debug_info`.

In this build only two kinds of inputs actually emit it:

- **`.asm` files assembled by yasm.** `build/ymake.core.conf:5200` sets `_YASM_PREDEFINED_FLAGS_VALUE=-g dwarf2` for Linux, and yasm's DWARF-2 output always includes `.debug_aranges`. Example: `util/system/context_x86.asm` (fiber setjmp/longjmp) — this is the only member of `libyutil.a` that has an aranges section.
- **Sysroot glibc assembly routines**, built with GCC/gas, which emits aranges by default. This is where the overflowing relocation lives.

Regular C++ compiled by clang (with `-fdebug-default-version=4`) does *not* emit `.debug_aranges` at all — it stores the same address-range info in each CU's `DW_TAG_compile_unit` DIE via `DW_AT_low_pc`/`DW_AT_high_pc`/`DW_AT_ranges`. The whole `.debug_aranges` output section in the linked binary was ~2 272 bytes.

#### What we lose

Practically nothing:

- GDB, `addr2line`, `llvm-symbolizer`, `perf`, `eu-addr2line` all fall back to scanning the CU DIEs in `.debug_info` when `.debug_aranges` is absent. First lookup is microseconds slower; result is identical.
- The 2 KB of aranges we had covered a few assembly-based code paths (fiber context switch, sysroot glibc primitives). All still resolve to file:line via `.debug_line`.

#### What is preserved

Everything else — no debug info was touched for any archive:

- `.debug_info`, `.debug_loc`, `.debug_ranges`, `.debug_str`, `.debug_line`, `.debug_abbrev` — untouched, for `contrib/`, `ydb/`, `yql/`, `yt/`, `library/`, sysroot alike.
- `.symtab` / `.dynsym` — untouched.
- Full GDB experience: `ptype`, `p variable`, step into, stack traces with function names + `file.cc:line` — all intact for every module.

#### Verification

- `./ya make -r ydb/apps/ydbd` — succeeds. `.debug_info` in the resulting binary is 4 302 597 153 bytes (just over 4 GB), which would have been unlinkable before this PR.
- `readelf -S ydbd` confirms `.debug_aranges` is absent and all other `.debug_*` sections are present with expected sizes.

#### Notes for reviewers

- `INSERT AFTER .comment;` in the linker script is required. Without it, `-T script.ld` replaces the default output layout instead of augmenting it, which would produce a broken binary.
- This is a workaround. The proper long-term fix is one of: shrink `.debug_info` (via smaller PEERDIRs in ydbd or `NO_DEBUG_INFO()` on heavy libraries), land `-gdwarf64` end-to-end in the toolchain (needs sysroot rebuild), or split `ydbd` into smaller executables. This PR unblocks the build with a two-line change and no material loss of debug capability.


abyss7 (approved):
ОК - под честное слово, что ИИшка локально проверила работоспособность llvm-symbolizer и gdb.

PR #46343, opened 2026-07-13
refactor: automate SchemeShard path reference ownership

Enjection:
[YDBBUGS-452](http://st/YDBBUGS-452) · [YDBBUGS-585](http://st/YDBBUGS-585)

### Changelog entry

Internal SchemeShard path-reference ownership and cleanup refactoring.

### Description for reviewers

Automate path database-reference ownership for transactions, publications, and type-info map entries; retain restart checks and cleanup fixes. Shard reference accounting remains unchanged.

Keep cheap map-membership undo, explicit field undo, and staged table `AlterData`. Remove automatic whole-object snapshots and the associated partitioning CoW prototype: `Update()` no longer copies a table's partitioning or per-shard statistics. Existing operations that prohibit proposal rollback keep that contract; this is not an MVCC redesign.

Deferred snapshot prototype: [Enjection/ydb#45](https://github.com/Enjection/ydb/pull/45) (draft).

Pre-merge verification (`fe39a0b`): 53 related SchemeShard suites, 1,870/1,870 tests passed after reruns; three initial timeout cases each passed three serial repeats and both affected complete suites passed with lower test concurrency. Includes the 200,000-shard snapshot-free `Update()` regression and reference-count, undo-ordering, and stable-alias tests.

Merged fresh upstream `main` (`bd9080b`) without rebasing; resolved the generated-column helper conflict and adapted new metrics/topic lookups to the ownership API. Post-merge verification (`96311a4`): 15 affected SchemeShard suites, 1,033/1,033 tests passed with `hya make -T --build=relwithdebinfo -ttt --test-threads=4`; no failures, timeouts, filters, or increased timeout limits.


Enjection:
fable perf review below

```markdown
## Reducing the performance tax of the ref-handle design

Two load-bearing facts first: `TMemoryChanges::UnDo` has exactly **one** caller — `AbortOperationPropose`, i.e. rollback can only ever happen in the propose transaction (the `TMemoryChanges` instances created for the progress/plan/reply txs in `schemeshard__operation.cpp` are recorded into and then thrown away) — and the publications handoff in `DoDoneTransactions` already `std::move`s, so no hidden re-acquire there. Ranked by impact:

### 1. Biggest memory tax: the two "whole-population" side maps store zero information

`TSelfRefMap::Refs` and `ParentDbRefs` each hold one hash-map entry per object **for the lifetime of the schemeshard** (~72 bytes/entry with node + bucket overhead: 16 key + 32 `TPathRef` + pointers). On a schemeshard with millions of paths that's on the order of ~70 MB for `ParentDbRefs` plus comparable for the per-map `Refs` — and a doubled hash-op count on every insert/erase, plus slower `TTxInit`.

Key observation: in both maps, every field of the stored `TPathRef` is derivable, so the entry's *existence* is the only bit of information — and that's already encoded by the primary map:

- In `Refs`: `SS` and `Reason` are per-map constants (already members), and `PathId == key`. `DebugCheckConsistency` literally asserts Refs mirrors Map 1:1.
- In `ParentDbRefs`: `PathId == PathsById[child]->ParentPathId`, reason constant.

And there is no persistent per-entry disarmed state: disarm is always a *per-call* decision (`UndoErase`, `clear()`, dtor disarm-and-drop immediately). So:

- **`TSelfRefMap`: delete `Refs` entirely.** Call `SS->IncrementPathDbRefCount(id, Reason)` directly on new-key insert in `Set`/`SetUntracked`/`EmplaceUntracked`, decrement in `erase`, and have `UndoErase`/`clear()`/dtor simply skip the decrement (that *is* disarming). `DebugForEachRef` iterates `Map` keys instead. The structural guarantee ("insert acquires, erase releases, restore adopts") is unchanged — it just stops being materialized as a second map. Halves hash traffic, removes an entire map per container.
- **`ParentDbRefs`: replace the map with one bit on `TPathElement`** (`ParentRefHeld : 1`) or plain direct inc/dec at its three sites (`MaterializeImpl`, `PersistRemovePath`, init) — existence-of-child-row *is* the ref. 1 bit vs ~72 bytes per path, and it removes the redundant second `PathsById.find` inside `Acquire` (the old init loop incremented through the already-resolved parent pointer; the new one re-finds it).

The debug reconciliation stays a true independent oracle either way — it can recount from `Map` keysets and `ParentPathId` links, same sources of truth.

`TPathRef` itself stays for the places it genuinely carries state: `TTxState` members and publications, where the referenced path differs per instance. That's where the RAII handle earns its 32 bytes.

### 2. Biggest CPU tax: `Update()` snapshots in transactions that can never roll back

Since `UnDo` only runs from `AbortOperationPropose`, every clone recorded by `Update()` in post-plan states is dead weight — and those are the expensive ones: `Tables.Update(...)` in `TPropose::HandleReply` (copy_table), `TProposeAtTable::HandleReply` (every cdc/index version bump), `TSplitMerge` post-plan — each is a full `TTableInfo::DeepCopy`, which rebuilds the entire partition store, O(shards) with per-shard allocations. A 10k-shard table pays a 10k-entry store rebuild on plan steps, then drops it unused.

**Fix: an `Armed` flag on `TMemoryChanges`,** set true only for the propose tx. When disarmed, `NeedsUpdateSnapshot` returns false and `RecordSelfRefUndo` no-ops. Post-plan `Update()` collapses to one hash lookup, call sites don't change, and the invariant "only propose can abort" becomes explicit in code instead of folklore. ~10-line change, probably the single best CPU win.

### 3. `TTableInfo::DeepCopy` even where a snapshot *is* needed

In propose itself, copy/split still deep-copy big tables (the old shallow copy was the dangling-pointer bug the new unit test proves, so the extra cost is currently forced). Two ways to shrink it:

- **Structural (follow-up worthy):** make partitioning copy-on-write — `std::shared_ptr<const TPartitioning>` holding partitions + store together, swapped wholesale by `SetPartitioning`. The undo clone then shares the partitioning (shallow struct copy, O(1)), and the raw-pointers-into-store fragility that forced `DeepCopy` disappears with it. Also pays off everywhere else `TTableInfo` gets copied.
- **Interim, targeted:** most `Update()` sites touch only `AlterVersion`/`AlterData`. A narrow `UpdateAlter(id, changes)` that records a two-field undo closure instead of a whole-object clone covers the hottest sites without the refactor.

### 4. Smaller items (only if profiling asks)

- **`std::function` undo records:** captures (`this` + `TPathId` + snapshot ptr) exceed SSO, so each `Set`/`Update` in propose heap-allocates. A fixed-layout record with a virtual apply would fix it, but propose does only a handful per DDL — low priority, and #2 removes most of the volume anyway.
- **`NeedsUpdateSnapshot`'s two-level `THashMap<const void*, THashSet<TPathId>>`** → single `THashSet<std::pair<const void*, TPathId>>`: one hash op instead of two, one container.
- **`at()` returning the smart pointer by value:** free for the `TIntrusivePtr` maps (`TSimpleRefCount` is non-atomic), a pair of atomic ops for the `shared_ptr` maps (`OlapStores`). If a read path ever shows up in a profile, add a `const T*` accessor; not worth churn now.

### Already fine — no action

Publications handoff moves rather than copies; `DebugCheckSelfRefIntegrity` is `NDEBUG`-gated and sits on the rare abort path; log statements in Increment/Decrement are level-gated; `TPathRef` copies in `TTxState` re-acquire via non-atomic counters.

**Suggested order:** #2 (tiny diff, kills the hot-path clones) → #1 (removes the standing memory cost and simplifies `TSelfRefMap`) → #3 structural COW as the same follow-up that tackles shard refs.

```

CyberROFL (commented):
Мне нравится. Все изменения не просматривал, т.к. много просто рефакторинга, только сутевую часть.

Готов шипнуть, но если не торопишься, дождись еще @ijon.

ydb/core/tx/schemeshard/schemeshard_db_ref_map.h
```diff
@@ -0,0 +1,252 @@
...
+
+    // Debug: feed each held pathId to the DbRefCount reconciliation.
+    virtual void DebugForEachRef(const std::function<void(const TPathId&)>& fn) const = 0;
+};
+
+// THashMap<TPathId, V> holding a DbRefCount self-ref per entry: insert acquires,
+// erase releases. No operator[], so a missing-key read can't silently acquire.
+// Set/Update record undo and belong to the armed propose phase; SetUntracked/
+// UpdateUntracked/erase are for other phases (init, plan-step, progress, stats).
+template <class V>
```
flown4qqqq, line 74: Можно, пожалуйста, не использовать однобуквенные названия? Кажется, лучше использовать какой-нибудь `TValue`

ydb/core/tx/schemeshard/schemeshard_db_ref_map.h
```diff
@@ -0,0 +1,252 @@
...
+
+    // Teardown drops entries without releasing (the counters die with the shard).
+    ~TDbRefMap() override = default;
+
+    const TInner& AsMap() const {
+        return Map;
+    }
+
+    // Insert/assign (acquires on new key) and record the matching undo.
+    V& Set(TSetArgs args) {
```
flown4qqqq, line 114: Не кажется ли, что можно так в будущем случайно упустить прокидывание Value, например? Может быть, не стоит передавать сюда полноценную структуру, а задать сигнатуру функции с обязательными полями?

dorooleg (dismissed):

flown4qqqq (dismissed):

dorooleg (dismissed):

ijon:
Main concerns:
1. Do we even need this type of reference counting? Or perhaps the same goal can be achieved with something simpler?
2. Eliminating the need for explicit `Grab*` calls for operations appears to be a supplementary goal, not strictly required to achieve reference counting accuracy.
3. ~~The performance implications are unknown.~~ -- should be no performance hits from new ref counting (versioning aside)
4. The column tables case is complicated, as they still employ manual reference counting -- when would they be universally supported?

PR #46295, opened 2026-07-13
schemeshard: fix path refcount loss on reboot during CDC stream rotation

Enjection:
### Changelog entry 

...

### Changelog category 

* Not for changelog (changelog entry is not required)

### Description for reviewers 

Fixes #33764.

`TTxInit` restored the `SourcePathId` `DbRefCount` reference only for a whitelist of tx types, while `RemoveTx` decrements it unconditionally. A reboot during continuous backup CDC stream rotation (`TxRotateCdcStream`/`TxRotateCdcStreamAtTable` both source the old stream) made the post-reboot `RemoveTx` steal the references protecting the old stream's `Paths` row — including the one held by its `streamImpl` child row — so `TTxCleanDroppedPaths` erased the parent row while the child row remained, and the next `TTxInit` failed with `Parent path not found`.

Also:
- `TTxCleanDroppedPaths`: fail fast on a dropped path that still has children — the crash is self-healing (a restart recomputes the counters from the intact rows), while the row deletion would orphan the children and permanently brick the tablet
- `OnOffloadStatus`: tolerate a concurrently dropped topic (async tablet deletion makes this a legal race)
- `DropContinuousBackup`: single coordinated drop for all continuous backup streams (a drop per stream self-rejected when both the rotated-out and the current stream exist)

Regression tests reproduce the full crash chain deterministically (reboot mid-rotation → drop old stream → reboot).

Recovery mode for already-corrupted databases is split out into a follow-up PR (`SchemeShardControls.TolerateOrphanedPaths`).

PR #46361, opened 2026-07-13
schemeshard: TolerateOrphanedPaths ICB control for emergency recovery

Enjection:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Follow-up to #46295, which prevents new #33764-style corruption and makes SchemeShard fail fast on local database inconsistencies. A database where the corruption already happened (parent `Paths` row deleted, child rows remaining) still cannot boot — the parent row is physically gone.

New `SchemeShardControls.TolerateOrphanedPaths` ICB control (default off, emergency recovery only). When enabled:
- `TTxInit` synthesizes an in-memory dropped placeholder parent for orphaned path rows (nothing is persisted; the quarantined subtree is excluded from scheme board population, unreachable by name, no name conflicts — a same-name recreation supersedes the stale scheme board entry by higher path id)
- `TTxInit` skips in-flight transactions whose target path row is missing and skips missing source path references
- `TTxCleanDroppedPaths` skips a dropped path that still has children instead of stopping the tablet

Recovery runbook: enable → restart the node → boot succeeds, `LOG_ERROR`s name the orphaned path ids → force-drop the orphaned subtrees by path id → wait for the rows to be cleaned (dropped orphans and the placeholder clean up automatically once their references drain) → disable. The control must stay enabled until the orphaned rows are physically removed.

Regression test corrupts the local database via MiniKQL exactly like the issue (parent row erased, `streamImpl` child row kept) and verifies boot + operability + reboot stability with the control enabled.

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__clean_pathes.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__clean_pathes.cpp
@@ -131,12 +131,19 @@ struct TSchemeShard::TTxCleanDroppedPaths : public TTransactionBase<TSchemeShard
             TPathId pathId = *--itCandidate;
             Self->CleanDroppedPathsCandidates.erase(itCandidate);
             TPathElement::TPtr path = Self->PathsById.at(pathId);
-            if (path->DbRefCount == 0 && path->AllChildrenCount > 0) {
+            if (path->DbRefCount == 0 && path->Dropped() && path->AllChildrenCount > 0) {
                 // fail fast: a restart recomputes the counters and heals the drift
-                Y_FAIL_S("TTxCleanDroppedPaths: dropped path with children, DbRefCount is miscounted"
+                Y_VERIFY_S(Self->TolerateOrphanedPaths,
+                    "TTxCleanDroppedPaths: dropped path with children, DbRefCount is miscounted"
                     << ", PathId# " << pathId
                     << ", AllChildrenCount# " << path->AllChildrenCount
                     << ", at schemeshard: " << Self->TabletID());
+                LOG_ERROR_S(ctx, NKikimrServices::FLAT_TX_SCHEMESHARD,
+                    "TTxCleanDroppedPaths: skip dropped path with children, DbRefCount is miscounted"
+                    << ", PathId# " << pathId
+                    << ", AllChildrenCount# " << path->AllChildrenCount
+                    << ", at schemeshard: " << Self->TabletID());
+                ++SkippedCount;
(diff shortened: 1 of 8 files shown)


ydb/core/tx/schemeshard/schemeshard__clean_pathes.cpp
```diff
@@ -131,12 +131,19 @@ struct TSchemeShard::TTxCleanDroppedPaths : public TTransactionBase<TSchemeShard
             TPathId pathId = *--itCandidate;
             Self->CleanDroppedPathsCandidates.erase(itCandidate);
             TPathElement::TPtr path = Self->PathsById.at(pathId);
-            if (path->DbRefCount == 0 && path->AllChildrenCount > 0) {
+            if (path->DbRefCount == 0 && path->Dropped() && path->AllChildrenCount > 0) {
```
CyberROFL, line 134: А в https://github.com/ydb-platform/ydb/pull/46295 не должно было быть такое же условие? Как будто там не хватает `path->Dropped()`.

ydb/core/tx/schemeshard/schemeshard__init_populator.cpp
```diff
@@ -46,6 +46,25 @@ struct TSchemeShard::TTxInitPopulator : public TTransactionBase<TSchemeShard> {
...
+                }
+                auto parentIt = Self->PathsById.find(cur->ParentPathId);
+                if (parentIt == Self->PathsById.end()) {
+                    break;
+                }
+                cur = parentIt->second;
+            }
+            if (underOrphanPlaceholder) {
+                continue;
+            }
```
CyberROFL: Повторю свой вопрос из https://github.com/ydb-platform/ydb/pull/46295#discussion_r3571981161:
> Не должно ли это быть под флагом `TolerateOrphanedPaths`?

ydb/core/tx/schemeshard/schemeshard__clean_pathes.cpp
Enjection: В целом может стоит это отдельно подмёрджить тоже, но как будто почти недостижимо это условие

ydb/core/tx/schemeshard/schemeshard__init_populator.cpp
Enjection: `HasOrphanPlaceholders` должно хватить всем :)

CyberROFL (approved):

PR #46430, opened 2026-07-14
Introduce direct session ut

alexvru:
### Changelog entry 

Introduce direct session ut

### Changelog category 

### Description for reviewers 

Unit test for direct session in IC.


mregrock (approved):
LGTM

PR #46532, opened 2026-07-14
CancelAfter timeout in rate limiter must be related to quota handling only

UgnineSirdis:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

Rate Limiter API has two timeouts:
- standard OperationTimeout - timeout for grpc call, is handled the same as for every grpc method
- CancelAfter timeout - timeout for QuoterService request only: does not include all infrastructure things, such as rights checks, database resolving etc.

CancelTimeout is passed as a Deadline field in QuoterService events API, and this Deadline includes not only quota handling, but quoter and quoter resource resolving (if the request arrives for the first time). It results in problems with stability of returned status codes: we can not distinguish infrastructure errors and quota issues, especially if CancelAfter holds relatively small value.

I removed communication with Kesus from this timeout: now it handles only quota related issues.

### DIFF
--- a/ydb/core/quoter/quoter_service.cpp
+++ b/ydb/core/quoter/quoter_service.cpp
@@ -1119,7 +1119,7 @@ void TQuoterService::Handle(TEvQuota::TEvProxySession::TPtr &ev) {
         if (req.Deadline != TDuration::Max() && !ResourceInResolvingState(reqId)) {
             const TInstant now = TActivationContext::Now();
             TryTickSchedule(now);
-            ScheduleRequestDeadline(reqId, TimeToGranularity(now + req.Deadline));
+            ScheduleRequestDeadline(reqId, TimeToGranularity(now + req.Deadline)); // Can invalidate req
         }
     }
 
--- a/ydb/core/quoter/quoter_service.cpp
+++ b/ydb/core/quoter/quoter_service.cpp
@@ -460,8 +460,8 @@ void TQuoterService::ForgetRequest(TRequest &request, TRequestId reqIdx) {
             break;
         case EResourceState::ResolveResource:
             if (TQuoterState *quoter = Quoters.FindPtr(leaf.QuoterId))
-                if (TSet<TRequestId> *resWaitMap = quoter->WaitingResource.FindPtr(leaf.ResourceName))
-                    resWaitMap->erase(reqIdx);
+                if (TResourceResolve *resolve = quoter->WaitingResource.FindPtr(leaf.ResourceName))
+                    resolve->Requests.erase(reqIdx);
             break;
         }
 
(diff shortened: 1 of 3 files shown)


Enjection (changes requested):
I remember SendLeaf collapses both timeouts. So I made small claude-assisted investigation, below the results:
```markdown
`SendLeaf` sends one value, `min(operation_timeout, cancel_after)` (rpc_rate_limiter_api.cpp:518-530).
Postponing it is only coherent if it's `cancel_after`. When `cancel_after` is unset — the common case —
that value *is* `operation_timeout`, so this postpones the infrastructure timeout past the
infrastructure work: the opposite of the goal.

And once postponed, nothing else bounds the request:
- `HasCancelOperation()` is the base `false` (rpc_deferrable.h:94), not overridden here → `CancelAfterTimer`
  never armed, `OnCancelOperation` (rpc_rate_limiter_api.cpp:467) is dead.
- `Handle(TEvRpcTimeout)` only bumps a counter (quoter_service.cpp:1192-1195); the event carries two
  strings, no request identity (quoter.h:105-113).
- `TEvCancelRequest` is never constructed repo-wide (and is misdeclared at quoter.h:117 —
  `TEventLocal<TEvClearance, ...>`).

**1. Orphaned charge, now also when `cancel_after` is set.** The RPC's own `OperationTimeoutTimer`
still fires and `Reply` calls `Die`, so the client is fine. The quoter request isn't: its deadline now
lands at `resolve_time + min(op_timeout, cancel_after)`. Past `operation_timeout` the RPC answers
TIMEOUT and dies, then the session establishes, the deadline arms, and quota is spent for nobody.
Makes the comment at rpc_rate_limiter_api.cpp:460 ("always race when cancel after is not set") wrong.

**2. Stuck `ResolveQuoter` leaks.** Today: deadline fires → `ForgetRequest` erases from
`WaitingQueueResolve` → the 1-min cleanup pass (Bootstrap:387) sees `Empty()` → `CloseQuoterIfEmpty`
(:1245) → `BreakQuoter` → `QuotersIndex` erased → next request re-navigates. Indirect, but it needs the
deadline. Only `ResolveQuoter` is exposed — Kesus-down recovers via the proxy (:1151). And the navigate
is sent `FlagTrackDelivery` (:590) while `StateFunc` has no `TEvUndelivered` handler, so if the scheme
cache is unreachable the entry sticks forever, the navigate is never re-sent (:562-590), and requests
pile at :603 with no deadline — never replied, never freed. Pre-existing hole; the deadline was covering it.

**Suggestion:** pass both values, not their min — an absolute operation deadline armed at arrival and
never postponed, plus the relative `cancel_after` this PR postpones. `cancel_after` then measures only
quota time, and the hard cap still reaps orphaned/stuck requests.
```
It is funny that we are probably found another issue. It would be great to try to repro in ut.

Enjection (approved):

PR #46591, opened 2026-07-15
Leak CloseInternal session settings to fix use-after-dtor

makxenov:
`internalCloseSessionSettings` is a function-local static in `CloseInternal`, so it is constructed on the first session close -- that is, after any static `TTableClient` was constructed. Static destruction runs in reverse order, so the settings are destroyed first, and a later `~TTableClient -> Drain()` -> `CloseInternal` at exit reads them after their destructor has run. `static` was excessive, remove it.

### Changelog entry 

...

### Changelog category 
* Bugfix 

### Description for reviewers 

...

### DIFF
--- a/ydb/public/sdk/cpp/src/client/table/impl/table_client.cpp
+++ b/ydb/public/sdk/cpp/src/client/table/impl/table_client.cpp
@@ -1118,8 +1118,7 @@ TAsyncStatus TTableClient::TImpl::Close(const TKqpSessionCommon* sessionImpl, co
 }
 
 TAsyncStatus TTableClient::TImpl::CloseInternal(const TKqpSessionCommon* sessionImpl) {
-    static const auto internalCloseSessionSettings = TCloseSessionSettings()
-            .ClientTimeout(TDuration::Seconds(2));
+    const auto internalCloseSessionSettings = TCloseSessionSettings().ClientTimeout(TDuration::Seconds(2));
 
     auto driver = Connections_;
     return Close(sessionImpl, internalCloseSessionSettings)


Shfdis (approved):

ydb/public/sdk/cpp/src/client/table/impl/table_client.cpp
```diff
@@ -1118,8 +1118,10 @@ TAsyncStatus TTableClient::TImpl::Close(const TKqpSessionCommon* sessionImpl, co
 }
 
 TAsyncStatus TTableClient::TImpl::CloseInternal(const TKqpSessionCommon* sessionImpl) {
-    static const auto internalCloseSessionSettings = TCloseSessionSettings()
-            .ClientTimeout(TDuration::Seconds(2));
+    // Intentionally leaked to prevent ~TTableClient -> Drain()
+    // to read the destroyed settings at exit.
+    static const auto& internalCloseSessionSettings = *new TCloseSessionSettings(
+        TCloseSessionSettings().ClientTimeout(TDuration::Seconds(2)));
```
Shfdis: Зачем что-то утекать, следует просто убрать static, дальше эти настройки в любом случае скопируются
```suggestion
    const auto& internalCloseSessionSettings = TCloseSessionSettings().ClientTimeout(TDuration::Seconds(2)));
```

Shfdis (changes requested):

ydb/public/sdk/cpp/src/client/table/impl/table_client.cpp
makxenov: Спасибо, не увидел, что так легко static убирается

Shfdis (approved):

PR #46739, opened 2026-07-16
Fix YAML config Tribool feature flags written as booleans truncating later flags

Enjection:
### Changelog category 

### Description for reviewers 

A `Tribool` feature flag (e.g. `enable_mvcc`, proto tag 47) written as a YAML
boolean aborted the whole `TFeatureFlags` JSON→proto merge. The merge walks
fields in proto-tag order, so the abort at tag 47 silently dropped **every**
feature flag with a higher tag (`enable_arrow_format_at_datashard`, …) from the
config delivered to components. `TConfigsDispatcher::ParseYamlProtoConfig`
caught the parse exception, logged it, and returned the partially-populated
proto — so a truncated config was shipped to subscribers while the monitoring
page still showed the full resolved YAML. The effect: a dynamic-config flag
appeared set in the UI but never reached the runtime without a node restart.

Changes:
- **Accept booleans on `Tribool` fields.** `yaml_config_parser` now coerces a
  YAML/JSON boolean on any `Tribool`-style enum field to `VALUE_TRUE` /
  `VALUE_FALSE` before the merge, guided by the target proto schema
  (`CoerceBoolEnumsToNames`). This is done entirely in YDB code — the vendored
  `library/cpp/protobuf/json` is untouched.
- **Fail loudly instead of silently truncating.** `ParseYamlProtoConfig` now
  `Y_ABORT`s on a parse failure rather than delivering a partially-parsed
  config.
- **Tests:** parser unit tests (bool→`VALUE_TRUE`/`VALUE_FALSE`, enum-name form
  still works, later-tag flag survives, control) and a ConfigsDispatcher
  end-to-end test asserting the full FeatureFlags is delivered.

### DIFF
--- a/ydb/library/yaml_config/yaml_config_parser_ut.cpp
+++ b/ydb/library/yaml_config/yaml_config_parser_ut.cpp
@@ -3,6 +3,7 @@
 #include "yaml_config_helpers.h"
 
 #include <ydb/core/protos/key.pb.h>
+#include <ydb/core/protos/feature_flags.pb.h>
 
 #include <library/cpp/testing/unittest/registar.h>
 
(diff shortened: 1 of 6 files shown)


ydb/library/yaml_config/yaml_config_parser_ut.cpp
```diff
@@ -78,6 +79,71 @@ Y_UNIT_TEST_SUITE(YamlConfigParser) {
...
+        UNIT_ASSERT(cfgFalse.HasFeatureFlags());
+        UNIT_ASSERT_VALUES_EQUAL((int)cfgFalse.GetFeatureFlags().GetEnableMvcc(),
+            (int)NKikimrConfig::TFeatureFlags::VALUE_FALSE);
+    }
+
+    Y_UNIT_TEST(TriboolFeatureFlagAcceptsEnumName) {
+        // The explicit enum spelling keeps working alongside the boolean form.
+        TString config = R"(
+feature_flags:
+  enable_mvcc: VALUE_FALSE
```
kseleznyov, line 139: А почему нельзя писать true/false? зачем понадобилось VALUE_TRUE/VALUE_FALSE?

Enjection: теперь как раз можно

CyberROFL (approved):

PR #46796, opened 2026-07-16
ut_transform: redact source line numbers in golden error results

Enjection:
### Changelog category 

### Description for reviewers 

The `ut_transform` golden (canondata) results for the error cases capture the
process `stderr`, which includes the throw location as `<file>.cpp:<line>`. As a
result, **any** unrelated edit that shifts line numbers in
`yaml_config_parser.cpp` (or the deprecated parser) breaks these golden tests,
even though the observable behavior is unchanged.

This makes `cleanup_errors` redact the line number (`foo.cpp:123` →
`foo.cpp:LINE`), the same way it already redacts exception addresses, and
updates the four affected golden results accordingly. The file path and the
error message are still checked — only the volatile line number is dropped.

No production code changes; test-only.


mregrock (approved):
LGTM

PR #46998, opened 2026-07-17
NBS-7492: Deduplicate persistent buffer barrier sends per pbuffer endpoint

BarkovBG:
Deduplicate persistent buffer barrier sends per pbuffer endpoint


yudov-maksim (dismissed):

drbasic (changes requested):
давай обсудим сначала

drbasic (approved):

PR #47140, opened 2026-07-20
Fix static group self heal

mregrock:
### Changelog entry 

Fix static group self heal

### Changelog category 

### Description for reviewers 

...

### DIFF
--- a/ydb/core/blobstorage/nodewarden/distconf_invoke_static_group.cpp
+++ b/ydb/core/blobstorage/nodewarden/distconf_invoke_static_group.cpp
@@ -45,23 +45,55 @@ namespace NKikimr::NStorage {
         }
         SuccessfulVDisks.emplace(&GroupInfo->GetTopology());
 
+        TVector<ui32> nodesToConnect;
         for (ui32 i = 0, num = GroupInfo->GetTotalVDisksNum(); i < num; ++i) {
             const TVDiskID vdiskId = GroupInfo->GetVDiskId(i);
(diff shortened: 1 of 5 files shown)


ydb/core/blobstorage/nodewarden/distconf_invoke_static_group.cpp
```diff
@@ -45,24 +45,60 @@ namespace NKikimr::NStorage {
...
+                if (inserted) {
+                    nodesToConnect.push_back(nodeId);
+                }
             }
             ActorToVDisk.emplace(actorId, vdiskId);
             PendingVDiskIds.emplace(vdiskId);
         }
+
+        for (const ui32 nodeId : nodesToConnect) {
+            if (const TActorId proxyId = TActivationContext::InterconnectProxy(nodeId)) {
```
alexvru: This check is a bit excessive: there always is a proxy. Even if there isn't.

ydb/core/blobstorage/nodewarden/distconf_invoke_static_group.cpp
```diff
@@ -45,24 +45,60 @@ namespace NKikimr::NStorage {
...
+                Send(proxyId, new TEvInterconnect::TEvConnectNode, IEventHandle::FlagTrackDelivery, nodeId);
```
alexvru: Can you describe the case when it can be undelivered, plz?

alexvru (changes requested):

alexvru (approved):

PR #47335, opened 2026-07-21
Regression tests for index build without indexed column data (#26854)

lberserq:
Regression test for https://github.com/ydb-platform/ydb/issues/26854.

`IIndexMeta::BuildIndexOptional()` verifies that the produced index chunks cover exactly `recordsCount`.
`TIndexByColumns::DoBuildIndexOptional()` used to return an empty vector when the indexed column had no
data in the merged portions, so general compaction crashed the tablet with
`verification=checkRecordsCount == recordsCount;fline=meta.cpp`.

That is already fixed in main by 042f10bbf844f6ef470ceb6b3961739c2f88f943 (#31076), which returns a single
`'\0'` chunk covering all records instead. This PR only adds the missing coverage for it:

- `KqpOlapIndexes::CompactionWithoutIndexColumnData` — the ticket's table, ngramm index and lc-buckets
  planner, data written by bulk upsert, `message` and its index added afterwards, then one compaction.
  End-to-end, it exercises the whole chain: portions written without the column, the merger's
  `BuildResultFiltered()` column set, index build during compaction, and reading through the index.

The test was verified to fail with exactly the issue's VERIFY when 042f10bbf84 is reverted, and to pass on
current main.

Notes on the issue itself: the ticket's own `data.json` does not reproduce, before or after the fix — every
row carries a `message` value, so the index always has input. The indexed column can only end up without
data through a schema change: bulk upsert (what `ydb import file json` does) rejects a batch that misses any
column, and `REPLACE INTO` with a column subset stores nulls for the rest.

### DIFF
--- a/ydb/core/kqp/ut/olap/indexes/indexes_ut.cpp
+++ b/ydb/core/kqp/ut/olap/indexes/indexes_ut.cpp
@@ -1,5 +1,6 @@
 
 #include <ydb/core/base/tablet_pipecache.h>
+#include <ydb/core/formats/arrow/serializer/native.h>
 #include <ydb/core/kqp/ut/common/olap_indexes_enums.h>
 #include <ydb/core/kqp/ut/olap/combinatory/variator.h>
 #include <ydb/core/kqp/ut/olap/helpers/local.h>
@@ -10,14 +11,18 @@
 #include <ydb/core/tx/columnshard/engines/changes/with_appended.h>
 #include <ydb/core/local_indexes/bloom/const.h>
 #include <ydb/core/tx/columnshard/hooks/testing/controller.h>
+#include <ydb/core/tx/columnshard/test_helper/columnshard_ut_common.h>
 #include <ydb/core/tx/columnshard/test_helper/controllers.h>
 #include <ydb/core/tx/columnshard/test_helper/test_combinator.h>
 
 #include <ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/types/status_codes.h>
 
+#include <library/cpp/string_utils/base64/base64.h>
 #include <library/cpp/testing/unittest/registar.h>
 #include <util/datetime/base.h>
 #include <util/generic/serialized_enum.h>
+
+#include <format>
 
 namespace NKikimr::NKqp {
 static void ExecQuery(TKikimrRunner& kikimr, bool useQueryService, const TString& query) {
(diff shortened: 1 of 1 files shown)


ydb/core/kqp/ut/olap/indexes/indexes_ut.cpp
```diff
@@ -2777,6 +2782,124 @@ Y_UNIT_TEST(RenameLocalBloomIndex, EUseQueryService) {
...
+                    { "partition", NScheme::TTypeInfo(NScheme::NTypeIds::Uint32) },
+                    { "offset", NScheme::TTypeInfo(NScheme::NTypeIds::Uint64) },
+                    { "index", NScheme::TTypeInfo(NScheme::NTypeIds::Uint32) },
+                    { "level", NScheme::TTypeInfo(NScheme::NTypeIds::Int32) },
+                    { "json_payload", NScheme::TTypeInfo(NScheme::NTypeIds::Utf8) },
+                    { "ingested_at", NScheme::TTypeInfo(NScheme::NTypeIds::Timestamp) },
+                    { "saved_at", NScheme::TTypeInfo(NScheme::NTypeIds::Timestamp) },
+                    { "request_id", NScheme::TTypeInfo(NScheme::NTypeIds::Utf8) } },
+                { "timestamp", "resource_type", "resource_id", "stream_name", "partition", "offset", "index" }));
+            for (ui64 i = 0; i < 3; ++i) {
```
lberserq: magic const

ydb/core/kqp/ut/olap/indexes/indexes_ut.cpp
```diff
@@ -1,5 +1,7 @@
 
 #include <ydb/core/base/tablet_pipecache.h>
+#include <ydb/core/formats/arrow/serializer/native.h>
+#include <ydb/core/tx/columnshard/test_helper/columnshard_ut_common.h>
```
lberserq: include order

ydb/core/kqp/ut/olap/indexes/indexes_ut.cpp
lberserq: Fixed in a0ef767: the row count is `MAX_ROW_CNT` (it also drives the second batch offset and the expected total), and the repeated timestamp literal became `BASE_TIMESTAMP_US`.

ydb/core/kqp/ut/olap/indexes/indexes_ut.cpp
lberserq: Fixed in a0ef767: the whole first include block is sorted alphabetically now.

ydb/core/tx/columnshard/engines/ut/ut_index_absent_column_data.cpp
```diff
@@ -0,0 +1,110 @@
+#include <ydb/core/formats/arrow/accessor/plain/accessor.h>
```
dorooleg: А в этом тесте есть ценность? indexes_ut.cpp же достаточно? Может только его оставить?

lberserq: ок -- принято кажется с indexes_ut действительно большое пересечение

dorooleg (approved):

PR #47433, opened 2026-07-22
Tidy up in ICv2

alexvru:
### Changelog entry 

Tidy up in ICv2

### Changelog category 

### Description for reviewers 

This patch makes proper startup for uring engine in ICv2, adds some settings and ICv2 code a bit cleaner.


mregrock (approved):
LGTM

PR #47611, opened 2026-07-23
Enable kafka serverless transactions

nerhneiro:
### Changelog entry 

...

### Changelog category 

### Description for reviewers 

...


nshestakov:
## Code review: Enable kafka serverless transactions

Overall: useful direction (lazy create of Kafka metadata tables in the user DB when `EnableKafkaServerlessTransactions` is on, plus reconnect-on-flag-flip). There are a few correctness gaps around migration/races and a clear config typo in the stress test that likely means the flag is never enabled there.

### Blocking / high

1. **Wrong feature-flag name in stress test**  
   `ydb/tests/stress/kafka_serverless/tests/test_kafka_streams.py` enables `"enable_serverless_transactions"`, but the proto field is `EnableKafkaServerlessTransactions` → YAML key must be **`enable_kafka_serverless_transactions`**. As written, the stress scenario almost certainly does not enable the new path.

2. **`ALREADY_EXISTS` skips ACL and migration** (`TKafkaMetadataService::Handle(TEvTableCreated)`)  
   On `ALREADY_EXISTS` the actor increments `ProcessedRequests` and exits without:
   - applying read-only ACL  
   - running migration from `SourceDatabasePath`  
   That breaks the intended migration story under races / retries / partial previous runs (table created, then actor died before migrate/ACL). Consider continuing with “ensure ACL + migrate if needed” even when the table already exists (or use an idempotent “ensure ready” flow).

3. **Every `SCHEME_ERROR` spawns a new `TKafkaMetadataService`, with no dedup**  
   `TryRequestMetadataTablesCreation` treats any `SCHEME_ERROR` as “tables missing”. Unrelated scheme errors will also spawn creators, and concurrent clients can register many identical actors (thundering herd of CreateTable/Alter/ACL). Prefer a narrower error check and a single in-flight ensure per `(database, tables)` (e.g. published actor / pending set).

### Medium

4. **No UT that actually verifies serverless migration**  
   - `KafkaBalancingAfterServerlessMigration` runs with `serverless=false` and the flag on from start — it only checks lazy create + retry (`COORDINATOR_NOT_AVAILABLE` → success), not row migration from resource DB → user DB.  
   - Please add a case with distinct `DatabasePath` / `ResourceDatabasePath`, pre-seeded rows in the source `.metadata` tables, then assert they appear in the target after first successful request.

5. **Circular include**  
   `kafka_metadata_service.cpp` includes `kafka_balancer_actor.h` (for `MAX_SESSION_TIMEOUT_MS`), while balancer includes `kafka_metadata_service.h`. Move the timeout constants into `kafka_constants.h` (or a tiny shared header) to break the cycle.

6. **`InitialServerlessTransactionsFlagValue.value()`** in `kafka_connection.cpp`  
   Unchecked `.value()` will throw if the optional is empty (e.g. request path before the flag is snapshotted). Prefer `value_or(...)` / explicit check, or set the snapshot earlier in connection bootstrap for all modes.

7. **Duplicated `GetMetadataDatabasePath()`** across ~8 actors  
   Same ternary on the feature flag. Worth a single helper on `TContext` (or a free function taking `TContext`) to avoid drift.

### Low / nits

8. Inconsistent naming in logs/messages: `EnableServerlessTransactions` vs `EnableKafkaServerlessTransactions`.  
9. Unused include `#include <ydb/core/kafka_proxy/actors/txn_actor_response_builder.h>` in `kafka_connection.cpp`.  
10. `TKafkaListGroupsActor::Bootstrap` constructs `Kqp` before the flag-change early-return (other actors check first).  
11. Event ctor params use `bool … = 0` and PascalCase argument names — prefer `false` and style consistent with the file.  
12. PR description / changelog body is still placeholders (`...`).  
13. `TRANSACTIONAL_ID_EXPIRATION_MS` is still passed to `set_expire_after_seconds` (copied from existing initializer). Pre-existing, but worth fixing while duplicating the schema (ms vs seconds).

### What looks good

- Feature flag gated path: metadata DB = user DB when enabled, resource DB otherwise.  
- Rebind on flag flip: snapshot on connection + fail with reconnect / close connection.  
- Lazy create + client retry pattern for missing tables.  
- UT coverage for flag flip at runtime for balancer heartbeats and transactions.  
- Schema of the new tables matches existing InitManager definitions.

### Verdict

**Request changes** — at least fix the stress-test flag name and the `ALREADY_EXISTS` / migration+ACL path (and ideally add a real migration UT + creation dedup) before merge.

nshestakov:
## Failure analysis: `KafkaProtocol.TransactionsFailAfterEnablingServerlessTransactionsFlagAtRuntime`

CI (relwithdebinfo, try_3): [err log](https://storage.yandexcloud.net/ydb-gh-logs/ydb-platform/ydb/PR-check/30017195726/ya-x86-64/try_3/artifacts/logs/ydb/core/kafka_proxy/ut/test-results/unittest/testing_out_stuff/KafkaProtocol.TransactionsFailAfterEnablingServerlessTransactionsFlagAtRuntime.err)

### What happened

After enabling `EnableKafkaServerlessTransactions` at runtime, the test calls `InitProducerId` on the **old** connection and correctly gets `INVALID_TXN_STATE`. Then it immediately does `Fetch` on the **same** `kafkaClient`.

Server side (by design in this PR): on `INVALID_TXN_STATE` / `COORDINATOR_NOT_AVAILABLE` when the feature flag changed, `TKafkaConnection` sets `CloseConnection = true` and closes after sending the reply.

Log sequence:

```
EnableServerlessTransactions feature flag changed; closing connection so the client reconnects...
Building reply ... code=INVALID_TXN_STATE method=INIT_PRODUCER_ID
Sent reply ...
>>>>> TFetchRequestData          // test continues on the same client
Connection closed
unexpected end of stream         // client reads Fetch response from a closed socket
```

Final exception:

```
error on read field correlationId: unexpected end of stream
```

### Root cause

Test bug / ordering issue, not a transaction commit bug. The first txn already completed successfully; the crash is only that `Fetch` runs on a connection that the server intentionally closed after the flag-change error.

### Suggested fix

In `TransactionsFailAfterEnablingServerlessTransactionsFlagAtRuntime`:

- do `Fetch` **before** the post-flip `InitProducerId`, **or**
- use a **new** `TKafkaTestClient` for `Fetch` (same as for the second transaction).

nshestakov:
## CODESTYLE.md review

Checked against [`ydb/agents/CODESTYLE.md`](https://github.com/ydb-platform/ydb/blob/main/ydb/agents/CODESTYLE.md) (and kafka_proxy note: follow local file style when editing; for **new** files apply CODESTYLE).

### Clear violations

1. **Function / ctor arguments must be lowercase** ([Names → Variables](https://github.com/ydb-platform/ydb/blob/main/ydb/agents/CODESTYLE.md))  
   PascalCase params in several places:
   - `kafka_events.h`: `bool EnableKafkaServerlessTransactionsFlag = 0` (×4 event ctors)
   - `ut/test_server.h` / `.cpp`: `bool EnableKafkaServerlessTransactions`  
   Prefer e.g. `enableKafkaServerlessTransactions` / `enableKafkaServerlessTransactionsFlag`.

2. **Use `false`, not `0`, for `bool`** ([C++ features](https://github.com/ydb-platform/ydb/blob/main/ydb/agents/CODESTYLE.md): prefer `nullptr`/`false` over `0`)  
   Same event ctors: `= 0` → `= false`.

3. **Binary operator spacing**  
   Extra space after `!=` in `kafka_transaction_actor.cpp` (×4):
   ```cpp
   ...FlagValue !=  NKikimr::AppData()...
   ```
   Also double space before type in one ctor: `resourceDatabasePath,  bool Enable...`.

4. **Remove unused `#include`** ([Includes](https://github.com/ydb-platform/ydb/blob/main/ydb/agents/CODESTYLE.md))  
   `kafka_connection.cpp`: `#include <ydb/core/kafka_proxy/actors/txn_actor_response_builder.h>` looks unused.

5. **Include grouping / order** (new `kafka_metadata_service.*`)  
   - Header: blank line missing between local `"actors.h"` and `<ydb/...>`.  
   - Alphabetize within group: `ydb_scheme.pb.h` before `ydb_table.pb.h`.  
   - `.cpp`: same-dir include should preferably be quotes (`"kafka_balancer_actor.h"`) in the local-module group; `kafka_constants` as a separate/`ydb` path is fine with angle brackets.  
   - `kafka_offset_commit_actor.cpp`: local `"kafka_metadata_service.h"` should be with other `"..."` includes, before `<ydb/...>`.

6. **Namespace closing comment** ([Namespaces](https://github.com/ydb-platform/ydb/blob/main/ydb/agents/CODESTYLE.md))  
   `kafka_metadata_service.h`: `} // NKafka` → `} // namespace NKafka` (`.cpp` already correct).

7. **Indentation of continued log args**  
   `kafka_init_producer_id_actor.cpp`: `{LogPrefix()}` under `YDB_LOG_DEBUG` is not indented with the call (unlike neighboring sites).

### New-file class layout (soft / vs local kafka_proxy)

For **new** `TKafkaMetadataService`, CODESTYLE asks:

- private data members with `_` suffix (`DatabasePath_`, …); sibling kafka actors often omit `_` — pick one and be consistent (either CODESTYLE for the new file, or explicitly mirror nearby actors).
- Do not mix nested types / methods / data in one private section without repeating `private:` to separate methods vs members.

Not blockers if you intentionally match existing kafka_proxy actors, but worth aligning on the new file.

### OK / matching local style

- File names lowercase `.cpp`/`.h`, `#pragma once`.  
- `enum class ETables` member naming.  
- 1TBS braces in the new service.  
- Argument names like `databasePath` / `sourceDatabasePath` in the service API.  
- Existing `Y_UNIT_TEST` style in `ut_protocol.cpp` (file already uses unittest, not gtest).

nshestakov (approved):

PR #47648, opened 2026-07-23
Don't report VDisk occupancy metrics for Replication tests

serbel324:
### Changelog entry 

Fix ReplicationSpace unit tests that started to fail after enabling exporting of VDisk metrics

### Changelog category 


mregrock (approved):
LGTM

PR #47713, opened 2026-07-24
Support stored generated columns

### LINKED ISSUE #47196: Support creating table with GENERATED columns

### LINKED ISSUE #47197: Reads and writes with generated STORED columns

dahbka-lis:
### Changelog entry 

Added support for stored generated columns. A column may now be defined by an expression over other columns of the row; its value is computed automatically on write instead of being supplied by the user. Guarded behind the `EnableGeneratedStored` feature flag (disabled by default).

### Changelog category 

* New feature

### Description for reviewers 

**What was implemented:**

- **SchemeShard**: schema-level support for generated columns — create/alter/copy table, path describer, TTL validation, feature-flag gating. New `DefaultFromExpression` column attribute (renamed from `DefaultFromGenerated` to support `DEFAULT expr` later).
- **KQP provider / type annotation**: type annotation and compilation of generated expressions; validation of what an expression may reference and return (nullability checks, type coercion of the lambda result).
- **KQP optimizer**: new `kqp_opt_generated_columns` pass that injects computed values into the write path. Works for `UPSERT`/`INSERT`/`REPLACE` as well as `UPDATE ... WHERE`, `DELETE ... WHERE`, and `UPDATE|DELETE ... ON`, including `RETURNING`.
- **Write path**: generated values are computed for regular writes and secondary index writes (stream write / stream lookup).
- Extensive tests: KQP UT suite, SchemeShard UT suite, backup/export/consistent-copy coverage.

**Virtual generated columns:** only the `EnableGeneratedVirtual` feature flag and the accompanying restrictions are added in this PR — full computation support for virtual columns is not implemented yet.

**Trade-offs / restrictions (intentional):**

- **Backup is forbidden** for tables with generated columns — the public `Ydb.Table` API has no representation for generated columns, so a backup could not round-trip faithfully. Rejected with `StatusPreconditionFailed` ("Cannot backup table with generated column ...").
- Expression allow-list: **unbounded/aggregate-style functions are disabled**; non-deterministic functions are allowed.
- Referencing `TableRow()` from a generated expression is forbidden.
- Both flags default to `false`; the feature is opt-in.


Ane1y (dismissed):

Ane1y (dismissed):

ydb/core/tx/schemeshard/schemeshard__operation_backup_restore_common.h
```diff
@@ -770,6 +770,17 @@ class TBackupRestoreOperationBase: public TSubOperation {
             }
         }
 
+        if (TxType == TTxState::TxBackup && context.SS->Tables.contains(path.Base()->PathId)) {
+            TTableInfo::TPtr table = context.SS->Tables.at(path.Base()->PathId);
+            for (const auto& [_, column] : table->Columns) {
+                if (column.DefaultKind == ETableColumnDefaultKind::FromExpression && !column.IsDropped()) {
+                    result->SetError(NKikimrScheme::StatusPreconditionFailed,
+                        TStringBuilder() << "Cannot backup table with generated column '" << column.Name << "'");
```
CyberROFL, line 778: Это нужно согласовать. По результатам успешного согласования прилинковать тикет (на вас) про поддержку backup/restore.

CyberROFL (commented):
По коду ОК. Нужно получить согласование.

Ane1y (approved):

CyberROFL (approved):

PR #47864, opened 2026-07-26
Fix migration out-of-order message delivery

alexvru:
### Changelog entry 

Fix migration out-of-order message delivery

### Changelog category 

### Description for reviewers 

This fixes potential out-of-order message delivery in ICv2 when migrating sessions between worker threads.


mregrock (approved):
LGTM

PR #48072, opened 2026-07-28
Dynamic write buffer size in ICv2

alexvru:
### Changelog entry 

Dynamic write buffer size in ICv2

### Changelog category 

### Description for reviewers 

Subj.


mregrock (approved):
LGTM

PR #48075, opened 2026-07-28
Fix use-after-dtor in TWorkerThread ut

makxenov:
### Description for reviewers 
In `TFastTlsTest::ManyConcurrentKeys`, a few threads are spawned but not joined. It may happen that a thread object may be destructed BEFORE the spawned OS thread reached its `func`, so we have usage after destructor. Join the thread in `~TWorkerThread()`, as the base `~TThread()` does. The operation is idempotent, so no problem in calling `Join()` twice from both base and derived destructors.

### DIFF
--- a/ydb/core/util/ut_common.h
+++ b/ydb/core/util/ut_common.h
@@ -21,6 +21,12 @@ namespace NKikimr {
             : Func(std::move(func))
         { }
 
+        ~TWorkerThread() override {
+            // It is essential to join here, because the spawned thread may access Func
+            // after the destructor of this object if not joined.
+            Join();
+        }
+
         double GetTime() const {
             return Time;
         }


flown4qqqq (dismissed):

ydb/core/util/ut_common.h
```diff
@@ -21,6 +21,12 @@ namespace NKikimr {
             : Func(std::move(func))
         { }
 
+        ~TWorkerThread() override {
```
BlazingPhoenix, line 24: TWorkerThread inherited from ISimpleThread. ISimpleThread inherited from TThread. TThread already calls Join() in destructor. I.e. we just call Join() twise in the same place because base class destructor already called in default destructor. Why calling it twise can fix some bugs?

makxenov: Actually they are called in different places. By `~TThread()`, `TWorkerThread::Func` was already destroyed as a member of a derived class. And it may be accessed afterwards, before `Join()` of the base class. So it is essential to join the spawned thread while the whole derived class is still alive

BlazingPhoenix: Complete destructor call order
For an object of the derived class Derived, inheriting from Base, the complete sequence of calls is as follows:

1. The destructor body ~Derived(): The code you wrote in the derived class destructor is executed.
2. Destructors of Derived's member objects: All non-static data members of the derived class itself are destroyed in the reverse order of their declaration in the class.
3. Base class destructor ~Base(): The body of the base class destructor is executed.
4. Destructors of Base's member objects: After the ~Base() destructor body, the base class members are destroyed.
5. Destructors of virtual base classes: If any exist, they are called at the very end for the most derived object.

BlazingPhoenix (approved):

PR #48082, opened 2026-07-28
fix(ci): dedupe ydbdoc-review.yml env for doc_translate

sintjuri:
## Summary
- #48080 duplicated `env:` keys in `.github/workflows/ydbdoc-review.yml`, which broke parsing (workflow showed as path-only name; `doc_translate` label on #47091 did not start a job).
- Restore a clean env block including `YDB_SA_KEY` / `YDBDOC_TRANSCRIPT_BACKEND` for continue transcripts.

## Test plan
- [ ] After merge, remove+add `doc_translate` on a docs PR and confirm workflow run starts


sintjuri:
Blocks re-translate of #47091: after #48080 `doc_translate` label no longer starts a job (workflow name fell back to path). Please approve/merge this workflow-only dedupe.

PR #48100, opened 2026-07-28
Fix and enable retro tracing UT

serbel324:
### Changelog entry 

Fix and enable retro tracing UT

### Changelog category 


mregrock (approved):
LGTM

Issue #47942, opened 2026-07-27
OLAP: SIGSEGV in TAccessorsCollection::RemainOnly / TProjectionProcessor (collection.cpp) on UploadTpch1000


naspirato:
#### Фактура

| | |
|--|--|
| Suite / DB | `UploadTpch1000` / `sas_big_column` |
| Branch · Version | `main` · [`d6dd620`](https://github.com/ydb-platform/ydb/commit/d6dd6206c5217385bccddfeaa1b59f717c674526) |
| Run | `2026-07-27_d6dd620` · `2026-07-27T08:24:30` UTC |
| Allure | https://proxy.sandbox.yandex-team.ru/12931774388/index.html |
| Failed | UploadTpch1000.Query06 (SIGSEGV → node down) |

#### Что сломалось

SIGSEGV (signal 11) в `TAccessorsCollection::RemainOnly` при `TProjectionProcessor::DoExecute` на conveyor worker во время columnshard fetch/projection.

#### К чему приводит
- Crash dyn-ноды `sas9-1593` (после рестарта `Registered as 50044`)
- Соседние query в том же Allure видят **node down / connection lost**

#### Детали ошибки

Host: `sas9-1593.host.testing.ydb.yandex.net`  
Coredump: https://coredumps.yandex-team.ru/v3/cores?filter=program_type%3Dkikimr%3B+%40cluster_name%3Dolap-testing-sas-perf&since_ts=2026-07-27T12%3A01%3A02.084332%2B03%3A00&till_ts=2026-07-27T12%3A01%3A41.734595%2B03%3A00

```
Received signal 11
Backtrace:
#0 std::__y1::__function::__value_func<void (int, siginfo_t*, void*)>::operator()[abi:ne210000](int&&, siginfo_t*&&, void*&&) const at /-S/contrib/libs/cxxsupp/libcxx/include/__functional/function.h:299:0
#1 std::__y1::function<void (int, siginfo_t*, void*)>::operator()(int, siginfo_t*, void*) const at /-S/contrib/libs/cxxsupp/libcxx/include/__functional/function.h:825:0
#2 (anonymous namespace)::SetSignalHandler(int, std::__y1::function<void (int, siginfo_t*, void*)>&&, sigaction*)::$_0::operator()(int, siginfo_t*, void*) const at /-S/ydb/library/signal_backtrace/signal_backtrace.cpp:32:0
[...]

PR #48174, opened 2026-07-29
Fix double release of arena key values in DqHashCombine spilling teardown

lberserq:
Fixes the use-after-poison from #40326 and the heap-corruption coredump family behind it (#47942, #47284).

**Problem**: in `InitiateSpillingAsync`, tuples with dehydrated states (simple SUM/COUNT aggregates) release their arena-owned key references through the `HydratedBuffer` alias and only zero the buffer — the arena tuple keeps the stale pointer bits. If the graph is destroyed mid-spill (query abort / memory-limit exception with a spill write pending), `ReleaseAggregationsFromArena()` re-sweeps the arena and UnRefs the already-freed keys — a refcount write into reused heap memory, corrupting arbitrary process state. A second window of the same kind: the read-back path allocates arena tuples from recycled pages whose stale bits stay visible to the teardown sweep across pending async reads.

**Why only main**: `DqHashCombine` is enabled by default on main (`EnableDqHashCombineByDefault=true`) and explicitly disabled in stable — matching the observation in #40326 that the crash reproduces on main but not on stable-26-2.

Reproduced deterministically by the new `TDqHashCombineTest::TestTeardownDuringStateSpilling` (never-completing spiller parks the spill coroutine, graph destroyed at the first Yield): under ASAN all 4 variants crash before the fix with the exact #40326 stack, and pass after it (full suite 84/84 ASAN-clean). Also includes teardown-safety tests for WideCombiner/WideLastCombiner/CommonJoinCore written while narrowing the search (all clean, kept as regression pins).

### DIFF
--- a/ydb/library/yql/dq/comp_nodes/dq_hash_combine.cpp
+++ b/ydb/library/yql/dq/comp_nodes/dq_hash_combine.cpp
@@ -714,6 +714,12 @@ protected:
                         uv.UnRef();
                         uv = TUnboxedValuePod{};
                     }
+                    if (isDehydrated) {
+                        // The keys just released were owned by the arena tuple (HydratedBuffer
+                        // aliases them without an extra Ref); clear the stale arena copies so a
+                        // mid-spill teardown (ReleaseAggregationsFromArena) doesn't release them again
+                        std::fill_n(static_cast<TUnboxedValuePod*>(static_cast<void*>(item)), keysWidth, TUnboxedValuePod{});
+                    }
                     if (!pageFuture.has_value()) {
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


yql/essentials/minikql/comp_nodes/ut/mkql_join_ut.cpp
```diff
@@ -80,6 +80,39 @@ Y_UNIT_TEST_LLVM(InnerOrderRightFirst) {
                                                           {TMaybe<i32>{}, i32(4)},
                                                       });
 }
+// Teardown-safety check for https://github.com/ydb-platform/ydb/issues/40326: a
```
dorooleg: Только эти тесты нужно будет заносить через arcadia, через yql/essentials

risenberg:
А как ты нашел этот баг? Я по-всякому гонял llm, но ничего не нашлось.

yql/essentials/minikql/comp_nodes/ut/mkql_join_ut.cpp
dorooleg: Сейчас давай уберем эти тесты в этом pr

ydb/library/yql/dq/comp_nodes/dq_hash_combine.cpp
```diff
@@ -849,6 +855,11 @@ class TBaseAggregationState: public TComputationValue<TBaseAggregationState>
 
             while (!stateSpiller.Empty()) {
                 TUnboxedValuePod* keyAndStateBuf = static_cast<TUnboxedValuePod*>(Store->Alloc(0));
+                if (isDehydratedState) {
+                    // The arena page may be reused and hold stale value bits; the tuple stays
```
dorooleg, line 859: почему два разных вида заполнения 0, давай один выберем)

dorooleg (changes requested):

ydb/library/yql/dq/comp_nodes/dq_hash_combine.cpp
lberserq: fixed in 0f6c61737bfca71af724f378bf9997810080410a

dorooleg (approved):

pzuev (approved):

PR #48228, opened 2026-07-29
Add proto for proposed dstool workload yaml configuration

serbel324:
### Changelog entry 

Add protobuf file with a schema of dstool workload proposed yaml configuration

### Changelog category 


mregrock (approved):
LGTM

PR #48500, opened 2026-07-31
[detailed-metrics] propagate METRICS_LEVEL to DataShard via subdomain subscription

jsjant:
### Changelog category 

### Description for reviewers 

To be used later as a gate for a specific row table to report its detailed metrics at a certain level

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__init.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__init.cpp
@@ -1790,11 +1790,13 @@ struct TSchemeShard::TTxInit : public TTransactionBase<TSchemeShard> {
                         );
                     }
 
-                    alter->SetTablesMetricsLevel(
-                        rowset.GetValueOrDefault<Schema::SubDomainsAlterData::TablesMetricsLevel>(
-                            subdomainInfo->GetTablesMetricsLevel()
-                        )
-                    );
+                    if (!path->IsRoot() || !Self->IsDomainSchemeShard) {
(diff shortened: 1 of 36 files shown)


ydb/core/tx/schemeshard/schemeshard__init.cpp
```diff
@@ -1710,6 +1711,12 @@ struct TSchemeShard::TTxInit : public TTransactionBase<TSchemeShard> {
                             NKikimrSubDomains::EServerlessComputeResourcesModeShared
                         );
                     }
+
+                    domainInfo->SetTablesMetricsLevel(
+                        rowset.GetValueOrDefault<Schema::SubDomains::TablesMetricsLevel>(
+                            NKikimrSchemeOp::TTableDetailedMetricsSettings::MetricsLevelUnspecified
+                        )
+                    );
```
Enjection, line 1719: Sure, we rarely do anything on the root domain as a database. But it worth asking explicitly: was it a deliberate or an accidental choice?

Enjection (dismissed):

ydb/core/tx/schemeshard/schemeshard__init.cpp
jsjant: Yes, detailed metrics only supported for tenanted databases

It's now explicitly stated on the other schemeshard operations that were touched by this PR

Enjection (approved):

BlazingPhoenix (approved):

PR #48515, opened 2026-07-31
IC v2: configure uring engine from Common, add dead-peer detection

alexvru:
### Changelog entry 

IC v2: configure uring engine from Common, add dead-peer detection

### Changelog category 

### Description for reviewers 

The engine used to be built from a snapshot of individual V2 settings. It now
holds an intrusive pointer to the TInterconnectProxyCommon shared by every
session and derives its configuration from there; the resulting reference cycle
(Common owns the engine) is broken in Stop() once the workers are joined.

Sessions were also never actually pinged, because the shard timer only
considered sessions being migrated away, so a silently dead link stayed up
forever. Pings now go out on the shard tick, and a session that has received
nothing from its peer for DeadPeer is torn down, as in IC v1. The tick period is
derived from the ping and dead-peer settings and capped at one second so neither
reacts with a noticeable delay, while rebalancing keeps its own coarser period.


mregrock (approved):
LGTM

PR #48529, opened 2026-07-31
Fix v2 event serializer bug

alexvru:
### Changelog entry 

Fix v2 event serializer bug

### Changelog category 

### Description for reviewers 

Fix buffer reference calculation.


mregrock (approved):
LGTM

PR #48729, opened 2026-08-03
WriteSeqNum chain and gap detection for uncommitted writes

### LINKED ISSUE #2835: DataShard: support faster uncommitted writes
When flushing in-memory changes to disk (before any read from the same table) kqp writes changes to datashards to "uncommitted" changes. Currently we acknowledge those writes are they are fully persistent, but this is actually undesirable, since it doesn't add much to reliability, but it may severely increase latency when those changes are flushed multiple times. It's better to pipeline those flushes (especially with the final commit), so we should acknowledge those writes to kqp before they actually hit the disk.

However, we would need to make this safe across datashard restarts, since when we acknowledge writes they may fail to write, in which case we may successfully commit without some of those changes. To support this we would need kqp to pass information between shards and kqp about which writes are supposed to be committed. Since we don't need to support multiple writes in parallel a single "number of writes" number would be sufficient, so each uncommitted write would need to specify the last previous lock, with the last write it expects before the new write, and each write would return an incremented write number which kqp would remember in the locks list. On commit the lock would specify this last write number, and we would check that no writes have been lost between the last acknowledged write and commit. This feature would need to be opt-in per-request for backwards compatibility, i.e. datashard would need to know whether kqp supports the new write number. And kqp would need to somehow figure out whether datashard supported this new mode as well (but we can just make it a feature flag toggle, so the user is responsible for enabling it correctly).

An example of how this would work:

* First EvWrite specifies LockTxId=123, WriteIndex=1, and no previous lock
  * Shard sees there's no previous lock, adds it to the table, new lock has HasWrites=true, WriteIndex=1
  * Shard starts persisting lock and changes, but sends EvWriteResult without waiting for completion
* Second EvWrite specifies LockTxId=123, WriteIndex=2, and the last lock info (with WriteIndex=1)
  * Shard checks current lock exists and has WriteIndex==1 (otherwise it's TLI)
  * Shard changes lock to have WriteIndex=2
  * Shard starts persisting new lock info and changes, but sends EvWriteResult without waiting for completion
* The final EvWrite with commit specifies the last lock info (with WriteIndex=2)
  * Shard checks current lock exists and has WriteIndex==2 (otherwise it's TLI)
  * Shard starts (possibly volatile) commit of all the changes

In the unlikely event shard restarts without persisting any intermediate changes the above chain breaks and either the next write or the final commit will fail (some changes have been lost). As a bonus uncommitted EvWrites will become idempotent and perhaps more reliable: when EvWrite fails due to shard unavailability, that EvWrite (with the same WriteIndex) can be repeated, and shard can apply it or reply with some ALREADY status when WriteIndex already matches the current lock. Commits are not idempotent though (we could try making them idempotent too by not removing the lock at commit time, and marking it as committed instead, but it's very tricky).

azevaykin:
Number uncommitted TEvWrite operations with a per-lock WriteSeqNum (1-based, per writer, per shard). DataShard persists the seq num with the lock, rejects gaps in and between batches, treats a resend of the last batch as IsDuplicate, and compares the echoed seq num at commit in ValidateLocks.

KQP allocates seq nums, reuses them on resend, and copies them through lock echo into the prepare/commit TKqpLocks.
This does not acknowledge uncommitted writes before the tablet log commit. Flushes wait for persistence as they do today.


ydb/core/protos/feature_flags.proto
```diff
@@ -360,4 +360,5 @@ message TFeatureFlags {
     optional bool EnableKafkaServerlessTransactions = 307 [default = false];
     optional bool EnableGeneratedStored = 308 [default = false];
     optional bool EnableGeneratedVirtual = 309 [default = false];
+    optional bool EnableDataShardPipelinedUncommittedWrites = 310 [default = true];
```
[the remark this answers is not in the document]
azevaykin: Intentionally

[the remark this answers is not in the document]
azevaykin: Intentionally

ydb/core/kqp/runtime/kqp_write_actor.cpp
```diff
@@ -1667,6 +1690,11 @@ class TKqpTableWriteActor : public TActorBootstrapped<TKqpTableWriteActor> {
     const ui64 LockNodeId;
     const bool InconsistentTx;
     const bool IsOlap;
+    const bool PipelinedWrites;
```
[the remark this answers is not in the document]
azevaykin: KQP serializes query execution and one write actor exists per table

ydb/core/tx/columnshard/columnshard_schema.h
```diff
@@ -612,8 +612,12 @@ struct Schema: NIceDb::Schema {
 
         struct Flags: Column<6, NScheme::NTypeIds::Uint64> {};
 
+        struct WriteIndex: Column<7, NScheme::NTypeIds::Uint64> {};
+
+        struct WriterIndex: Column<8, NScheme::NTypeIds::Uint64> {};
+
         using TKey = TableKey<LockId>;
```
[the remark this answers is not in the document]
azevaykin: Following the suggestion would break the build. TColumnShardLocksDb derives from the shared TShardLocksDb<TColumnShard, NColumnShard::Schema>, and that template references Schema::Locks::WriterIndex/WriteIndex unconditionally in LoadLocks locks_db.h:43 and PersistAddLock locks_db.h:127. The columns are a requirement of sharing the template, not optional surface. ColumnShard rejecting WriteIndex at the request level is a separate, deliberate check.

ydb/core/tx/data_events/events.h
```diff
@@ -158,7 +166,7 @@ struct TDataEvents {
             return result;
         }
 
-        void AddTxLock(ui64 lockId, ui64 shard, ui32 generation, ui64 counter, ui64 ssId, ui64 pathId, bool hasWrites) {
+        void AddTxLock(ui64 lockId, ui64 shard, ui32 generation, ui64 counter, ui64 ssId, ui64 pathId, bool hasWrites, ui64 writerIndex = 0, ui64 writeIndex = 0) {
```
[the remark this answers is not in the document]
azevaykin: The latency claim is wrong. WriteActorWritesLatencyHistogram is fed by OnMessageReceived from client-side SendTime, not from TxStats; duplicates route through ProcessWriteCompletedShard and are recorded normally. The zero stats are also semantically correct — the duplicate branch execute_write_unit.cpp:147 reads no rows, writes no rows, and touches no lock state, so there is nothing to report. BuildCompleted sets no TxStats either; stats are filled by the caller when work actually occurred.

ydb/core/kqp/common/kqp_tx_manager.cpp
```diff
@@ -109,6 +109,16 @@ class TKqpTransactionManager : public IKqpTransactionManager {
...
             }
+            // Merge per writer so an echo from a later write can't drop another writer's entry.
+            for (const auto& writeIndex : lock.Proto.GetWriteIndexes()) {
+                auto* existing = FindIfPtr(*lockPtr->Lock.Proto.MutableWriteIndexes(),
+                    [&](const auto& entry) { return entry.GetWriterIndex() == writeIndex.GetWriterIndex(); });
+                if (existing) {
+                    existing->SetWriteIndex(writeIndex.GetWriteIndex());
+                } else {
+                    *lockPtr->Lock.Proto.AddWriteIndexes() = writeIndex;
+                }
```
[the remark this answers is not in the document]
azevaykin: Fixed

ydb/core/tx/datashard/datashard_pipeline.cpp
```diff
@@ -1725,6 +1725,25 @@ TOperation::TPtr TPipeline::BuildOperation(NEvents::TDataEvents::TEvWrite::TPtr&
             return writeOp;
     }
 
+    if (rec.HasWriteIndex()) {
+        const ui64 writerIndex = rec.GetWriteIndex().GetWriterIndex();
+        const ui64 writeIndex = rec.GetWriteIndex().GetWriteIndex();
+        if (writeIndex == 0
+            || !rec.GetLockTxId()
+            || rec.txmode() != NKikimrDataEvents::TEvWrite::MODE_IMMEDIATE
```
nikvas0, line 1797: I think PREPARE can also be safely retried using WriteIndex.

ydb/core/kqp/runtime/kqp_write_actor.cpp
```diff
@@ -918,6 +923,11 @@ class TKqpTableWriteActor : public TActorBootstrapped<TKqpTableWriteActor> {
             ProcessWriteCompletedShard(ev);
             return;
         }
+        case NKikimrDataEvents::TEvWriteResult::STATUS_ALREADY_APPLIED: {
+            // A duplicate uncommitted write; the shard reports the unchanged lock.
+            ProcessWriteCompletedShard(ev);
```
nikvas0: Need deduplication here. Both original Completed and STATUS_ALREADY_APPLIED can be received here.

ydb/core/kqp/runtime/kqp_write_actor.cpp
```diff
@@ -1667,6 +1690,11 @@ class TKqpTableWriteActor : public TActorBootstrapped<TKqpTableWriteActor> {
     const ui64 LockNodeId;
     const bool InconsistentTx;
     const bool IsOlap;
+    const bool PipelinedWrites;
+    // This writer's id in the uncommitted write chain; one write actor per table today.
+    static constexpr ui64 WriterIndex = 0;
```
nikvas0, line 1739: Maybe just use ActorId (it's just two ui64, so must be ok)

ydb/core/kqp/runtime/kqp_write_actor.cpp
```diff
@@ -1314,6 +1327,16 @@ class TKqpTableWriteActor : public TActorBootstrapped<TKqpTableWriteActor> {
             FillEvWritePrepare(evWrite.get(), shardId, *TxId, TxManager);
         } else if (!InconsistentTx) {
             evWrite->SetLockId(LockTxId, LockNodeId);
+            if (PipelinedWrites) {
+                // A resend (overload retry) must repeat the same index, not allocate a new one.
```
nikvas0: Why "overload retry"? It's for all retries.

ydb/core/kqp/common/kqp_tx_manager.cpp
```diff
@@ -109,6 +111,16 @@ class TKqpTransactionManager : public IKqpTransactionManager {
             if (lock.Proto.GetHasWrites()) {
                 lockPtr->Lock.Proto.SetHasWrites(true);
             }
+            // Merge per writer so an echo from a later write can't drop another writer's entry.
+            for (const auto& writeIndex : lock.Proto.GetWriteIndexes()) {
+                auto* existing = FindIfPtr(*lockPtr->Lock.Proto.MutableWriteIndexes(),
+                    [&](const auto& entry) { return entry.GetWriterIndex() == writeIndex.GetWriterIndex(); });
+                if (existing) {
+                    existing->SetWriteIndex(writeIndex.GetWriteIndex());
```
nikvas0: Add AFL_ENSURE(existing.WriteIndex < new.WriteIndex)

nikvas0 (changes requested):

ydb/core/tx/datashard/datashard_pipeline.cpp
nikvas0: At least for volitile tx, since it uses uncommitted changes to store data and doesn't persist itself.

ydb/core/protos/data_events.proto
```diff
@@ -6,6 +6,12 @@ import "ydb/public/api/protos/ydb_issue_message.proto";
 package NKikimrDataEvents;
 option java_package = "ru.yandex.kikimr.proto";
 
+// One writer's position in a lock's uncommitted write chain
+message TWriteIndex {
+    optional uint64 WriterIndex = 1;
+    optional uint64 WriteIndex = 2;
```
ztlpn: Вот тут надо подумать, как надо названия разнести, а то очень похожие

azevaykin: 69e88c26586b6e2b104efa2394ee851cb326987e

ydb/core/kqp/runtime/kqp_write_actor.cpp
azevaykin: 69e88c26586b6e2b104efa2394ee851cb326987e

Aggred to use WriteSeqNum

ydb/core/kqp/runtime/kqp_write_actor.cpp
azevaykin: b0e3200d108ae87d6c5a1a3406652ed38a154c01

ydb/core/kqp/common/kqp_tx_manager.cpp
azevaykin: b0e3200d108ae87d6c5a1a3406652ed38a154c01

ydb/core/kqp/runtime/kqp_write_actor.cpp
azevaykin: b0e3200d108ae87d6c5a1a3406652ed38a154c01

ydb/core/protos/data_events.proto
```diff
@@ -122,6 +129,7 @@ message TEvWrite {
     optional ETxMode TxMode = 3;
     optional uint64 LockTxId = 4;
     optional uint32 LockNodeId = 5;
+    optional TWriteSeqNum WriteSeqNum = 13;
```
nikvas0: Move WriteSeqNum to TOperation. We'll need this in future for handling splits/merges.

nikvas0: Each TOperation has it's own WriteSeqNum.
UPD: WriterIndex is the same for all operations, but let's TWriteSeqNum be struct, since we'll add another field there during split/merge/move handling implementation.

nikvas0 (changes requested):

ydb/core/protos/data_events.proto
```diff
@@ -153,6 +161,7 @@ message TEvWriteResult {
         STATUS_WRONG_SHARD_STATE = 11;
         STATUS_CONSTRAINT_VIOLATION = 12;
         STATUS_DISK_GROUP_OUT_OF_SPACE = 13;
+        STATUS_ALREADY_APPLIED = 14;
```
nikvas0: We need to store and resend (on duplicate request) some data. This data includes error text and status. For example, to handle retried INSERT correctly.
Thus I think instead of adding STATUS_ALREADY_APPLIED we need to add to EvWriteResult a new flag IsDuplicate. Status must be left the same as for original message.

nikvas0 (changes requested):

ydb/core/tx/datashard/datashard_pipeline.cpp
azevaykin: Agreed to do in the following PR

ydb/core/protos/data_events.proto
azevaykin: Is it necessary to survive datashard restart? Should this data go through "memory state migration"?

ydb/core/protos/data_events.proto
azevaykin: 79b81b728255ec033c1bddb840a4cf56ea337040

ydb/core/protos/data_events.proto
azevaykin: [Replace STATUS_ALREADY_APPLIED with IsDuplicate flag on TEvWriteResult](https://github.com/ydb-platform/ydb/pull/48729/changes/80a093af9bdb21448c2d6a9cfe9ea530e4433baa)

nikvas0: > Is it necessary to survive datashard restart? Should this data go through "memory state migration"?

No, it's fine if data about retry is lost on restart. Then the whole tx will simply fail.

However, for surviving split/merge/move this data must be moved to new datashard with inmemory state.

azevaykin: 9291421734f0951a5ed763adfc92f9a11fecd672

ydb/core/tx/datashard/execute_write_unit.cpp
```diff
@@ -117,6 +117,118 @@ class TExecuteWriteUnit : public TExecutionUnit {
...
+            const ui64 requested = op.GetWriteSeqNum().WriteSeqNum;
+            if (!requested) {
+                continue;
+            }
+
+            if (requested < current) {
+                writeOp->SetError(NKikimrDataEvents::TEvWriteResult::STATUS_BAD_REQUEST, TStringBuilder()
+                    << "Uncommitted write " << writerIndex << ":" << requested
+                    << " is already applied, writer is at " << current);
+                return EExecutionStatus::Executed;
```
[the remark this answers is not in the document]
azevaykin: 3bcd1b101c0461ef63ad5bf0e482a0e4166e5e07

ydb/core/tx/locks/locks.h
```diff
@@ -1248,8 +1279,8 @@ class TSysLocks {
     TLocksCache* Cache = nullptr;
     ILocksDb* Db = nullptr;
 
-    TLock MakeLock(ui64 lockTxId, ui32 generation, ui64 counter, const TPathId& pathId, bool hasWrites) const;
-    TLock MakeAndLogLock(ui64 lockTxId, ui32 generation, ui64 counter, const TPathId& pathId, bool hasWrites) const;
+    TLock MakeLock(ui64 lockTxId, ui32 generation, ui64 counter, const TPathId& pathId, bool hasWrites, ui64 writerIndex = 0, ui64 writeIndex = 0) const;
```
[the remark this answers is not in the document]
azevaykin: 3bcd1b101c0461ef63ad5bf0e482a0e4166e5e07

ydb/core/tx/datashard/execute_write_unit.cpp
```diff
@@ -117,6 +117,126 @@ class TExecuteWriteUnit : public TExecutionUnit {
...
+                << "Uncommitted write " << writerIndex << ":" << maxRequested
+                << " is already applied, writer is at " << current);
+            return EExecutionStatus::Executed;
+        }
+
+        // maxRequested > current: the batch advances the chain.
+        // But partial overlap with already-applied writes is a protocol error.
+        for (const auto& op : operations) {
+            const ui64 requested = op.GetWriteSeqNum().WriteSeqNum;
+            if (requested && requested <= current) {
```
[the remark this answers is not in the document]
azevaykin: a45f1444f294659728c041aa53e02c6f6380c68e

ydb/core/tx/datashard/datashard_impl.h
```diff
@@ -953,9 +953,11 @@ class TDataShard
             struct Counter : Column<4, NScheme::NTypeIds::Uint64> {};
             struct CreateTimestamp : Column<5, NScheme::NTypeIds::Uint64> {};
             struct Flags : Column<6, NScheme::NTypeIds::Uint64> {};
+            struct WriteSeqNum : Column<7, NScheme::NTypeIds::Uint64> {};
+            struct WriterIndex : Column<8, NScheme::NTypeIds::Uint64> {};
```
ztlpn: Если у каждого WriterIndex свой SeqNum, и их может быть много у одного лока, получается нужно отдельную табличку заводить?

ydb/core/kqp/common/kqp_tx_manager.cpp
```diff
@@ -685,6 +708,9 @@ class TKqpTransactionManager : public IKqpTransactionManager {
         // All QuerySpanIds of queries that wrote to this shard in insertion order.
         TVector<ui64> BreakerQuerySpanIds;
         THashSet<ui64> BreakerQuerySpanIdsSet;
+
+        // Last uncommitted write seq num sent to this shard, per writer (absent = none)
+        std::map<ui64, ui64> WriteSeqNums;
```
ztlpn: Кажется, сортированность не нужна, и THashMap вполне подходит

nikvas0, line 713: THashMap

ydb/core/kqp/runtime/kqp_write_actor.cpp
```diff
@@ -1674,6 +1713,13 @@ class TKqpTableWriteActor : public TActorBootstrapped<TKqpTableWriteActor> {
     const ui64 LockNodeId;
     const bool InconsistentTx;
     const bool IsOlap;
+    const bool PipelinedWrites;
+    // This writer's id in the uncommitted write chain; one write actor per table today.
+    static constexpr ui64 WriterIndex = 0;
```
nikvas0, line 1739: Remove static

ydb/core/kqp/runtime/kqp_write_actor.cpp
```diff
@@ -475,6 +475,11 @@ class TKqpTableWriteActor : public TActorBootstrapped<TKqpTableWriteActor> {
         , LockNodeId(lockNodeId)
         , InconsistentTx(inconsistentTx)
         , IsOlap(isOlap)
+        , PipelinedWrites(
+              AppData()->FeatureFlags.GetEnableDataShardPipelinedUncommittedWrites()
+              && !inconsistentTx
+              && !isOlap
+              && lockTxId != 0)
```
nikvas0, line 488: Can't have lockTxId=0 here.

ydb/core/kqp/runtime/kqp_write_actor.cpp
```diff
@@ -1336,6 +1353,28 @@ class TKqpTableWriteActor : public TActorBootstrapped<TKqpTableWriteActor> {
         const auto serializationResult = ShardedWriteController->SerializeMessageToPayload(shardId, *evWrite);
         YQL_ENSURE(isPrepare || isImmediateCommit || serializationResult.TotalDataSize > 0);
 
+        // Set per-operation WriteSeqNum for pipelined uncommitted writes
+        if (PipelinedWrites && !isPrepare && !isImmediateCommit && !InconsistentTx) {
+            const size_t opCount = evWrite->Record.OperationsSize();
+            auto [it, allocated] = InFlightWriteSeqNum.try_emplace(shardId);
```
nikvas0, line 1379: Better move WriteSeqNums to batches in ShardedWriteController.

ydb/core/kqp/ut/tx/kqp_sink_tx_ut.cpp
```diff
@@ -545,6 +545,165 @@ Y_UNIT_TEST_SUITE(KqpSinkTx) {
...
+    // Keys 10 and 4000000010 land on different shards, so the commit goes through ValidateLocks.
+    Y_UNIT_TEST_TWIN(PipelinedUncommittedWrites, Enabled) {
+        TPipelinedUncommittedWrites tester;
+        tester.SetEnabled(Enabled);
+        tester.SetIsOlap(false);
+        tester.Execute();
+    }
+
+    // KQP must skip ColumnShard, which does not implement WriteSeqNum
+    Y_UNIT_TEST(PipelinedUncommittedWritesOlap) {
```
nikvas0: Don't need it

ydb/core/kqp/ut/tx/kqp_sink_tx_ut.cpp
```diff
@@ -545,6 +545,165 @@ Y_UNIT_TEST_SUITE(KqpSinkTx) {
...
+        TPipelinedUncommittedWrites tester;
+        tester.SetEnabled(true);
+        tester.SetIsOlap(true);
+        tester.Execute();
+    }
+
+    // A resent uncommitted write is answered twice: with the original result and, once the
+    // shard has seen it, with IsDuplicate set. KQP must take only the first one.
```
nikvas0, line 633: There is no IsDuplicate handling in kqp in this PR)

nikvas0 (dismissed):
Approve. Kqp part isn't ready and requires some cleaning, but it's ok for merging. I'll fix it after merge.

nikvas0 (dismissed):

nikvas0 (dismissed):

ydb/core/tx/locks/locks.h
```diff
@@ -1248,8 +1305,10 @@ class TSysLocks {
     TLocksCache* Cache = nullptr;
     ILocksDb* Db = nullptr;
 
-    TLock MakeLock(ui64 lockTxId, ui32 generation, ui64 counter, const TPathId& pathId, bool hasWrites) const;
-    TLock MakeAndLogLock(ui64 lockTxId, ui32 generation, ui64 counter, const TPathId& pathId, bool hasWrites) const;
+    TLock MakeLock(ui64 lockTxId, ui32 generation, ui64 counter, const TPathId& pathId, bool hasWrites,
+        const TVector<std::pair<ui64, ui64>>& writeSeqNums = {}) const;
+    TLock MakeAndLogLock(ui64 lockTxId, ui32 generation, ui64 counter, const TPathId& pathId, bool hasWrites,
+        const TVector<std::pair<ui64, ui64>>& writeSeqNums = {}) const;
```
ztlpn: лучше будет TLockWriteSeqNum вместо пары

ztlpn: А лучше просто пару параметров, этот TLock же не поддерживает много WriterIndex

Либо сделать, чтобы поддерживал, а то шизофрения немножко.

ydb/core/protos/tx_datashard.proto
```diff
@@ -2752,6 +2759,7 @@ message TInMemoryLockInfo {
     optional uint64 VictimQuerySpanId = 10;
     optional uint64 BreakerQuerySpanId = 11;
     optional uint32 BreakerNodeId = 12;
+    repeated TWriteSeqNumResult WriteSeqNumResults = 13;
```
ztlpn: Я бы назвал TWriteSeqNumState, как в locks.h

ydb/core/tx/datashard/execute_write_unit.cpp
```diff
@@ -136,12 +278,61 @@ class TExecuteWriteUnit : public TExecutionUnit {
...
+        }
+        auto lock = DataShard.SysLocksTable().GetRawLock(guardLocks.LockTxId);
+        if (lock && lock->GetWriteSeqNum(guardLocks.SetWriteSeqNum->WriterIndex) == guardLocks.SetWriteSeqNum->WriteSeqNum) {
+            lock->SetWriteSeqNumResult(guardLocks.SetWriteSeqNum->WriterIndex,
+                SerializeWriteSeqNumResult(writeOp->GetWriteResult()->Record), db);
+        }
+    }
+
+    // Consume seq num and persist the error; skip this update's rolled-back ranges.
+    void PersistFailedWriteSeqNum(TWriteOperation* writeOp, TSetupSysLocks& guardLocks,
```
ztlpn: Вот тут мне не очень нравится поведение. Получается, при некоторых ошибках мы результат откатываем, а seqnum инкрементим (и можем даже новый лок создать), а при некоторых нет (тот же BAD_REQUEST). Предлагаю простое правило, инкрементить seqnum только в случае успешного результата. Ошибки же можно ретраить сколько угодно, шард гарантирует, что никаких записей сделано не было.

nikvas0 (dismissed):

ydb/core/tx/datashard/execute_write_unit.cpp
azevaykin: d408c304b5bb6a54165d199ce9e782362d0b740b

ydb/core/protos/tx_datashard.proto
azevaykin: d408c304b5bb6a54165d199ce9e782362d0b740b

ydb/core/tx/locks/locks.h
azevaykin: d408c304b5bb6a54165d199ce9e782362d0b740b

ztlpn (dismissed):

nikvas0 (dismissed):

ztlpn (dismissed):

dorooleg:
@kirillvasilenko посмотри пожалуйста

ydb/core/tx/columnshard/columnshard_schema.h
```diff
@@ -651,6 +652,20 @@ struct Schema: NIceDb::Schema {
         using TColumns = TableColumns<LockId, TxId>;
     };
 
+    // Per-writer uncommitted write seq num.
+    struct LockWriteSeqNums: Table<LockWriteSeqNumsTableId> {
```
kirillvasilenko: I do not understand why we add this to the CS schema now? Why not add it when (and if) we support the whole thin in CS?

azevaykin: 5c474ac51978764b813dfa25fa1e43b732613b9d

kirillvasilenko (dismissed):

nikvas0 (dismissed):

nikvas0 (approved):

ztlpn (approved):

kirillvasilenko (approved):

PR #48748, opened 2026-08-03
Support queue overflow tracking in v2

alexvru:
### Changelog entry 

Support queue overflow tracking in v2

### Changelog category 

### Description for reviewers 

This patch returns queue overflow tracking functionality into v2 as it works in v1.


mregrock (approved):
LGTM

PR #48862, opened 2026-08-04
[detailed-metrics] extract monitoring_project_id to user_attributes.h

jsjant:
### Changelog category 

### DIFF
--- a/ydb/core/tx/schemeshard/user_attributes.h
+++ b/ydb/core/tx/schemeshard/user_attributes.h
@@ -21,6 +21,7 @@ constexpr TStringBuf ATTR_DOCUMENT_API_VERSION = "__document_api_version";
 constexpr TStringBuf ATTR_ASYNC_REPLICATION = "__async_replication";
 constexpr TStringBuf ATTR_ASYNC_REPLICA = "__async_replica";
 constexpr TStringBuf ATTR_INCREMENTAL_BACKUP = "__incremental_backup";
+constexpr TStringBuf ATTR_MONITORING_PROJECT_ID = "monitoring_project_id";
 
 enum class EAttribute {
     USER,
--- a/ydb/core/protos/counters_datashard.proto
+++ b/ydb/core/protos/counters_datashard.proto
@@ -538,4 +538,5 @@ enum ETxTypes {
     TXTYPE_COMPLETE_VACUUM = 86                           [(TxTypeOpts) = {Name: "TxCompleteVacuum"}];
     TXTYPE_LOCK_ROWS = 87                                 [(TxTypeOpts) = {Name: "TxLockRows"}];
     TXTYPE_S3_DIRECT_WRITE_FINISH = 88                    [(TxTypeOpts) = {Name: "TxS3DirectWriteFinish"}];
+    TXTYPE_PERSIST_SUBDOMAIN_TABLES_METRICS_LEVEL = 89    [(TxTypeOpts) = {Name: "TxPersistSubDomainTablesMetricsLevel"}];
 }
--- a/ydb/core/protos/flat_scheme_op.proto
+++ b/ydb/core/protos/flat_scheme_op.proto
@@ -13,6 +13,7 @@ import "ydb/core/protos/s3_settings.proto";
 import "ydb/core/protos/schemeshard/operations.proto";
 import "ydb/core/protos/subdomains.proto";
 import "ydb/core/protos/sys_view_types.proto";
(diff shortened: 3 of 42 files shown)


ydb/core/tx/schemeshard/user_attributes.h
```diff
@@ -21,6 +21,7 @@ constexpr TStringBuf ATTR_DOCUMENT_API_VERSION = "__document_api_version";
 constexpr TStringBuf ATTR_ASYNC_REPLICATION = "__async_replication";
 constexpr TStringBuf ATTR_ASYNC_REPLICA = "__async_replica";
 constexpr TStringBuf ATTR_INCREMENTAL_BACKUP = "__incremental_backup";
+constexpr TStringBuf ATTR_MONITORING_PROJECT_ID = "__monitoring_project_id";
```
Enjection: I'm afraid we can't do it as simple as this. You either need to use existing one or do another meeting to decide what to do with it: https://github.com/ydb-platform/ydb/blob/fd335e5732b3f6d77d0e61f79db5074a0ab5874c/ydb/core/tx/schemeshard/schemeshard__operation_create_replication.cpp#L164-L176

Enjection: https://github.com/ydb-platform/ydb/blob/fd335e5732b3f6d77d0e61f79db5074a0ab5874c/ydb/core/tx/schemeshard/schemeshard__operation_create_pq.cpp#L172-L184

Enjection: https://github.com/ydb-platform/ydb/blob/fd335e5732b3f6d77d0e61f79db5074a0ab5874c/ydb/core/tx/schemeshard/schemeshard__operation_common_pq.cpp#L435-L451

ydb/core/tx/schemeshard/user_attributes.cpp
```diff
@@ -154,6 +157,46 @@ inline bool IsValidPathName_WeakCheck(const TString& name) {
...
+                    errStr = Sprintf("UserAttributes: attribute '%s' value is longer than %" PRIu32 " bytes",
+                        name.c_str(), TUserAttributesLimits::MaxMonitoringProjectIdLen);
+                    return false;
+                }
+                // An empty value is accepted and means "no project": the same
+                // thing removing the attribute means
+                return CheckValueStringWeak(name, value, errStr);
+        }
+
+        Y_UNREACHABLE();
```
Enjection: Never ever use Y_UNREACHABLE. Prefer Y_ABORT or Y_ENSURE.
The reason is simple: you deliberately introduce undefined behavior in a code and if someone ever forget case in switch it will lead to quite undebuggable failure. We learned it the hard way - having our production (schemeshard) crashing in a loop.
And yes I see there is below and above use of the same macro, we need to get rid of them, probably in a separate PR.

ydb/core/tx/schemeshard/user_attributes.cpp
```diff
@@ -154,6 +157,46 @@ inline bool IsValidPathName_WeakCheck(const TString& name) {
...
+                if (op == EUserAttributesOp::InitRoot) {
+                    // Tenant SchemeShard bootstrap replays what the domain
+                    // SchemeShard has already accepted, it must never fail
+                    return true;
+                }
+                if (!AppData()->FeatureFlags.GetEnableDataShardDetailedMetrics()) {
+                    errStr = Sprintf("UserAttributes: attribute '%s' cannot be set, detailed metrics are disabled",
+                        name.c_str());
+                    return false;
+                }
```
Enjection: I'm not sure I understood it correctly, but looks like flipping attribute validity with FeatureFlag is incorrect. This check can fire for attribute that is already set. Afaik console resends whole attribute set so it will become uneditable.

Enjection (changes requested):

ydb/core/tx/schemeshard/user_attributes.h
jsjant: Moved to a header with another constants to avoid this error in the future

ydb/core/tx/schemeshard/user_attributes.cpp
jsjant: Since we have no validation beforehand, I removed it

ydb/core/tx/schemeshard/user_attributes.cpp
jsjant: This is also gone now

Enjection (approved):

PR #48925, opened 2026-08-04
Fix serializer in ICv2

alexvru:
### Changelog entry 

Fix serializer in ICv2

### Changelog category 

### Description for reviewers 

This patch fixes serializer in ICv2 and also adds support for abseil cords in zero-copy output streams.


mregrock (approved):
LGTM

PR #48981, opened 2026-08-05
Add some more tests into ICv2

alexvru:
### Changelog entry 

Add some more tests into ICv2

### Description for reviewers 

Add some more tests for ICv2.


mregrock (approved):
LGTM

PR #49015, opened 2026-08-05
Move topic settings to table.h and make CDC topic settings readable

georgthegreat:
### Changelog entry 

Introduce `TChangefeedDescription::GetTopicPartitioningSettings()` for accessing CDC partitioning.

### Description for reviewers 

I am creating a new CDC topic and need to dump the result of `ALTER TABLE ADD CHANGEFEED` back from ydb.
Move TTopicSettings from `topic/control_plane.h` to `table/table.h` (where Changefeed settings are described).

### DIFF
--- a/ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/table/table_enum.h
+++ b/ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/table/table_enum.h
@@ -96,5 +96,16 @@ enum class EChangefeedState {
     Unknown = std::numeric_limits<int>::max()
 };
 
+//! Auto partitioning strategy of a changefeed's underlying topic
+enum class ETopicAutoPartitioningStrategy {
+    Unspecified /* "UNSPECIFIED" */,
+    Disabled /* "DISABLED" */,
+    ScaleUp /* "SCALE_UP" */,
+    ScaleUpAndDown /* "SCALE_UP_AND_DOWN" */,
+    Paused /* "PAUSED" */,
+
+    Unknown = std::numeric_limits<int>::max()
+};
+
 } // namespace NTable
 } // namespace NYdb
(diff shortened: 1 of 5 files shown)


Shfdis (commented):
Now topics depend on a huge table header, which can be avoided.
It is far better to add TChangefeedTopicPartitioningSettings, which will be managed by table, and leave the NTopic api untouched. Table client code can convert its own settings directly to and from Ydb::Topic::PartitioningSettings in table.cpp.
Moving existing public types between namespaces and re-exporting them with aliases changes their identity and ABI. It also changes the underlying type of EAutoPartitioningStrategy. None of that is necessary for this feature.

georgthegreat:
This would lead to code duplication.

changefeed IS a topic in the first place, isn't it? What is the problem with "hugeness" of the table.h?
It is measurable?

ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/table/table_enum.h
```diff
@@ -96,5 +96,16 @@ enum class EChangefeedState {
...
 
+//! Auto partitioning strategy of a changefeed's underlying topic
+enum class ETopicAutoPartitioningStrategy {
+    Unspecified /* "UNSPECIFIED" */,
+    Disabled /* "DISABLED" */,
+    ScaleUp /* "SCALE_UP" */,
+    ScaleUpAndDown /* "SCALE_UP_AND_DOWN" */,
+    Paused /* "PAUSED" */,
+
+    Unknown = std::numeric_limits<int>::max()
```
georgthegreat, line 107: I do not quite undestand the necessity of Unknown values here.
control_plane.h was working fine without them.

georgthegreat:
Here are the compilation time measured:

| Branch | `topic.cpp` compile time |
|---|---|
| Previous branch | ~19.44 s |
| **trunk** | **~18.66 s** |

PR #49115, opened 2026-08-06
Wait for login security state publication

KirillKurdyukov:
## Summary

- return `StatusAccepted` for successful AlterLogin operations after scheduling the security state publication
- update SchemeShard login test expectations for the asynchronous completion contract
- wait for transaction completion before checking the published security state in `UserLogin1`

## Root cause

AlterLogin updated the SchemeShard login provider and returned `StatusSuccess` before `PublishToSchemeBoard` completed. KQP treated that response as `ExecComplete`, so an immediate authentication request could resolve the database through SchemeCache while it still contained the previous security state and fail with `Unauthorized`.

Returning `StatusAccepted` makes the scheme operation handler subscribe to transaction completion. The client now observes successful completion only after the updated security state has been published.

This race was observed in the YDB .NET SDK integration test `DisableDiscovery_WhenUserIsCreatedAndPropertyIsTrue_SimpleWorking`:
https://github.com/ydb-platform/ydb-dotnet-sdk/actions/runs/31022179361/job/92361424017

## Impact

This removes the transient authorization failure when a newly created user authenticates immediately after `CREATE USER`. No client-side retry is required.

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard__operation_alter_login.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard__operation_alter_login.cpp
@@ -220,6 +220,9 @@ public:
             domainPtr->IncSecurityStateVersion();
             context.SS->PersistSubDomainSecurityStateVersion(db, subDomainPathId, *domainPtr);
             context.OnComplete.PublishToSchemeBoard(OperationId, subDomainPathId);
+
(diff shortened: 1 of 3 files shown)


KirillKurdyukov:
Superseded by #49120, which uses the branch from the contributor fork.

PR #49198, opened 2026-08-06
Update spilling root directory docs

lll-phill-lll:
### Changelog entry 
Document that `root` must be readable/writable by both static and dynamic node OS users when they differ.
...

### Description for reviewers 

...


sintjuri (commented):
## Review: Update spilling root directory docs

Overall the change is useful: the shared-`root` + different OS users case is a real operational footgun, and fixing the `#root` anchor (EN already linked to it from troubleshooting) is good. A few issues before merge.

### Blocking / should fix

1. **`1777` as the worked example is too strong for production docs.** World-writable sticky dirs (`chmod 1777`) solve “both users can create”, but they are a weak default for a spilling root that may hold query intermediate data. Prefer leading with a safer pattern and keep `1777` as a last-resort example, or drop it:
   - shared group for both node users + `chgrp` / `chmod 2770` (or ACL via `setfacl`), or
   - separate `root` per OS user / role when possible.

2. **Parent `1777` alone may still not be enough.** Spilling creates per-process dirs with owner-only mode (`S_IRWXU` / `0700` in `spilling_file.cpp`). Startup cleanup (`RemoveOldTmp`) deletes previous `node_<id>_<session>` dirs under the same `root`. If one OS user left dirs and another user later runs on the same `node_id` / shares the same `root`, that cleanup can fail even when the parent is `1777`. Worth either:
   - stating that both users must be able to remove leftover per-process dirs under `root`, or
   - recommending one OS user for all processes that use a given `root`, or a group-writable layout that matches how dirs are actually created.

3. **Tension with the note right below.** Config page says spilling runs only on [database nodes](https://ydb.tech/docs/en/concepts/glossary#database-node). The new text requires RW for both [static](https://ydb.tech/docs/en/concepts/glossary#static-node) and [dynamic](https://ydb.tech/docs/en/concepts/glossary#dynamic) node OS users. Please clarify *why* static users need access (colocated shared filesystem? config applied to all processes? hybrid nodes?). Otherwise readers will think storage-only nodes also spill. If the real case is “any OS user that runs a `ydbd` which uses this `root`”, say that instead of static/dynamic.

### Nits

4. **RU/EN parity on the config bullet is fine; troubleshooting recommendations are slightly asymmetric.** EN step 2 ends with `` `chmod 1777` ``; RU the same. Diagnostics bullets in RU keep trailing periods; EN still mostly without — pre-existing, ignore unless you touch those lines anyway.

5. **Changelog entry** is clear. Optional: mention the new `#root` anchor / troubleshooting cross-link fix in the reviewer description (the EN troubleshooting page already pointed at `#root` before the anchor existed).

6. **Minor wording (EN):** “may cause violation of …” style aside, “both users must have read and write access to this directory (for example, `rwx` for `other`, mode `1777`)” equates `other=rwx` with full `1777`; sticky bit is the important part of `1777` and is not explained. If you keep `1777`, say “sticky world-writable directory, like `/tmp` (`1777`)”.

### Verdict

**Request changes** on (1)–(3): soften/`1777` guidance, acknowledge cleanup/`0700` interaction, and resolve static-vs-database-node wording. After that this looks mergeable.

Preview: https://ydb-platform--ydb.viewer.diplodoc.com?revision=pr-49198-d7a67a731c4be8e178e0c2cc594f05cac37dc443

sintjuri (approved):

fomichev3000 (approved):

PR #49226, opened 2026-08-07
Improve ICv2 metrics and perf

alexvru:
### Changelog entry 

Improve ICv2 metrics and perf

### Description for reviewers 

This patch adds more discrete timing metrics for ICv2 and also increases performance a little bit by changing deque to vector.


mregrock (approved):
LGTM

Issue #49294, opened 2026-08-07
Optimize NBS data path


komarevtsev-d:
This issue is an umbrella for small perfomance fixes needed for NBS team. Found by these flamegraphs:

<img width="1200" height="726" alt="Image" src="https://github.com/user-attachments/assets/a39f0f94-15f5-47fb-9269-d9ad9025235f" />

<img width="1200" height="550" alt="Image" src="https://github.com/user-attachments/assets/5e2f48c6-4323-4d3a-9390-5a7f1a0754b4" />

PR #49296, opened 2026-08-07
Erasure: avoid hash lookup for erasure parameters

komarevtsev-d:
### Description for reviewers 

Replace the unordered_map in GetErasureParameters with a constexpr array and switch-based lookup, eliminating hash-table initialization and lookup overhead.

Related to #49294

### DIFF
--- a/ydb/core/erasure/erasure.cpp
+++ b/ydb/core/erasure/erasure.cpp
@@ -101,7 +101,7 @@ const TErasureParameters& GetErasureParameters(TErasureType::EErasureSpecies spe
             return ErasureParameters[5];
     }
 
-    ythrow TWithBackTrace<yexception>() << "Unknown erasure species = " << static_cast<int>(species);
+    Y_ABORT("Unknown erasure species = %d", static_cast<int>(species));
 }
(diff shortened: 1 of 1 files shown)


ydb/core/erasure/erasure.cpp
```diff
@@ -77,15 +77,31 @@ struct TErasureParameters {
...
+            return ErasureParameters[2];
+        case TErasureType::EErasureSpecies::Erasure4Plus3Block:
+            return ErasureParameters[3];
+        case TErasureType::EErasureSpecies::Erasure3Plus3Block:
+            return ErasureParameters[4];
+        case TErasureType::EErasureSpecies::ErasureMirror3of4:
+            return ErasureParameters[5];
+    }
+
+    ythrow TWithBackTrace<yexception>() << "Unknown erasure species = " << static_cast<int>(species);
```
serbel324: These exceptions are not handled most of the time
I think its better to replace it with Y_ABORT, it's easier to diagnose than unhandled exception

serbel324 (approved):

PR #49299, opened 2026-08-07
add udf store wasm implementation

akulaad:
### Changelog entry 

add udf store wasm implementation


martinevsky (changes requested):
See the two inline P1 comments.

ydb/services/udf_store/service.cpp
```diff
@@ -1,43 +1,281 @@
...
+                [&](const TString& name) { return name == libraryName; });
+            if (!dependsOnLibrary) {
+                continue;
+            }
+        } catch (...) {
+            continue;
+        }
+        LoadedUdfs.erase(md5);
+        UnloadWasmUdf(md5);
+        EnqueueWasmLoadIfNeeded(udf);
```
martinevsky, line 241: **P1:** A library change reaches this call while `CurrentSnapshot` still contains the previous library, so it queues a dependent UDF load before the replacement library artifact has been compiled. The loader does not validate artifact `source_md5` or `version`; this load can succeed with the old library and mark the UDF loaded permanently. The later successful library compile retries pending compiles, not this completed load. Please defer dependent loads until the matching library artifact is published, and validate the expected hash/version while loading.

ydb/services/udf_store/wasm/module_catalog.cpp
```diff
@@ -0,0 +1,84 @@
...
+    Y_ENSURE(artifact);
+    Y_ENSURE(!artifact->Md5.empty());
+    Y_ENSURE(!artifact->ModuleName.empty());
+
+    with_lock (Mutex_) {
+        if (auto it = ByMd5_.find(artifact->Md5); it != ByMd5_.end()) {
+            ByModuleName_.erase(it->second->ModuleName);
+        }
+        ByMd5_[artifact->Md5] = artifact;
+        ByModuleName_[artifact->ModuleName] = artifact;
```
martinevsky, line 18: **P1:** `module_name` is not made unique across queued loads: the service detects conflicts only in `LoadedWasmModuleNames`, so two same-name modules arriving in one snapshot can both be queued. The second registration overwrites this name entry but leaves the first MD5 entry. Removing that first module later unconditionally erases the name entry and removes the active replacement from the dynamic registry. Please reject duplicate module names before queueing, or make catalog replacement/removal conditional on the artifact currently mapped by name.

martinevsky (approved):

PR #49506, opened 2026-08-10
EXT-2349 report ExpectedSlotSize as SlotSizeBytes to monitoring

Maximus5:
### Changelog entry

Fix `SlotSizeBytes` PDisk monitoring counter for PDisks with configured `ExpectedSlotSize`.

### Description for reviewers

ExpectedSlotSize support was merged into stream-nb-25-1 in #47509. This PR ports the change to main and adds unit test.

When `ExpectedSlotSize` is set, `SlotSizeBytes` now reports that configured value directly. Without `ExpectedSlotSize`, the counter keeps the previous behavior and is calculated from the PDisk user chunk pool and `ExpectedSlotCount`.

Added a regression test that checks both paths:
- fallback formula: `userChunkPool / ExpectedSlotCount * ChunkSize`
- configured slot size: `SlotSizeBytes == ExpectedSlotSize`


SammyVimes (dismissed):
LGTM

SammyVimes (approved):

PR #49563, opened 2026-08-11
Raise the low watermark after truncate in datashard

flown4qqqq:
### Changelog entry 

...

### Description for reviewers 

`TRUNCATE` drops all row versions but did not move the datashard's MVCC low watermark, so a
transaction with an older snapshot silently read an empty table instead of getting an error.

Now the watermark is advanced to the truncate's own version, so such a read fails with a
retryable error — same as every other place that destroys history (GC, split, snapshot transfer).

### DIFF
--- a/ydb/core/tx/datashard/truncate_unit.cpp
+++ b/ydb/core/tx/datashard/truncate_unit.cpp
@@ -72,6 +72,16 @@ EExecutionStatus TTruncateUnit::Execute(
 
     txc.DB.Truncate(localTid);
 
+    // Truncate drops every row version, so nothing below this operation can be read any more.
+    // Advancing the low watermark keeps a stale snapshot read failing loudly instead of silently
+    // returning no rows.
+    const auto truncateVersion = DataShard.GetMvccVersion(op.Get());
+    auto& snapshotManager = DataShard.GetSnapshotManager();
+    if (snapshotManager.GetLowWatermark() < truncateVersion) {
+        NIceDb::TNiceDb db(txc.DB);
+        snapshotManager.SetLowWatermark(db, truncateVersion);
+    }
+
     auto userTable = DataShard.AlterTableSchemaVersion(actorCtx, txc, pathId, version);
 
     // We must set these flags here for the following reasons:
--- a/ydb/core/tx/datashard/truncate_unit.cpp
+++ b/ydb/core/tx/datashard/truncate_unit.cpp
@@ -75,12 +75,7 @@ EExecutionStatus TTruncateUnit::Execute(
     // Truncate drops every row version, so nothing below this operation can be read any more.
     // Advancing the low watermark keeps a stale snapshot read failing loudly instead of silently
     // returning no rows.
-    const auto truncateVersion = DataShard.GetMvccVersion(op.Get());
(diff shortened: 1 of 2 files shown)


ydb/core/tx/datashard/truncate_unit.cpp
```diff
@@ -72,6 +72,16 @@ EExecutionStatus TTruncateUnit::Execute(
...
     txc.DB.Truncate(localTid);
 
+    // Truncate drops every row version, so nothing below this operation can be read any more.
+    // Advancing the low watermark keeps a stale snapshot read failing loudly instead of silently
+    // returning no rows.
+    const auto truncateVersion = DataShard.GetMvccVersion(op.Get());
+    auto& snapshotManager = DataShard.GetSnapshotManager();
+    if (snapshotManager.GetLowWatermark() < truncateVersion) {
+        NIceDb::TNiceDb db(txc.DB);
+        snapshotManager.SetLowWatermark(db, truncateVersion);
```
azevaykin: SetLowWatermark instead of AdvanceWatermark. 
AdvanceWatermark also calls RemoveRowVersions maintaining the "everything below watermark is compacted" invariant. Post-truncate it's harmless for the user table (data already gone), but AdvanceWatermark is the idiomatic, safer call.

azevaykin (changes requested):
Persistent/volatile snapshot bypass. 
The watermark check is short-circuited by FindAvailable() at ydb/core/tx/datashard/datashard__read_iterator.cpp:1987 and ydb/core/tx/datashard/datashard__read_iterator.cpp:3059. If a persistent or volatile snapshot (ReadTable, backup, scan query snapshot) was registered before the TRUNCATE, reads through it skip the watermark check and silently see an empty table — the exact bug, unpatched. 
I found no check in check_scheme_tx_unit.cpp rejecting TRUNCATE while snapshots exist, and TTruncateUnit neither validates nor invalidates SnapshotManager entries. Scan queries (datashard__kqp_scan.cpp:656) rely on exactly such snapshots. 
This deserves a test and likely a fix (reject truncate with live snapshots, or drop/invalidate them).

flown4qqqq:
> Persistent/volatile snapshot bypass. The watermark check is short-circuited by FindAvailable() at ydb/core/tx/datashard/datashard__read_iterator.cpp:1987 and ydb/core/tx/datashard/datashard__read_iterator.cpp:3059. If a persistent or volatile snapshot (ReadTable, backup, scan query snapshot) was registered before the TRUNCATE, reads through it skip the watermark check and silently see an empty table — the exact bug, unpatched. I found no check in check_scheme_tx_unit.cpp rejecting TRUNCATE while snapshots exist, and TTruncateUnit neither validates nor invalidates SnapshotManager entries. Scan queries (datashard__kqp_scan.cpp:656) rely on exactly such snapshots. This deserves a test and likely a fix (reject truncate with live snapshots, or drop/invalidate them).

Это правда, ошибка такая есть. Но она немного отдельная, и я хотел развести этот ПР, потому что он фиксит другую проблему

sanpwc (approved):

azevaykin (approved):

PR #49686, opened 2026-08-11
Introduce chunk reserve for static group in PDisk

alexvru:
### Changelog entry 

Introduce chunk reserve for static group in PDisk

### Description for reviewers 

This patch introduces static group chunk reserve as it is already done for log chunks. The main idea is to prevent cluster from failing when PDisk gets used too much.


mregrock (approved):
LGTM

Issue #40952, opened 2026-05-21
TSAN: Data race in free in CPython runtime — stress workload binaries


maximyurchuk:
## Summary

Several `release-tsan` stress workloads invoke a Python-runtime CLI binary (`transfer`, `kafka_streams_test`, `streaming`, `workload_topic_kafka`) via `yatest.common.execute`. The binary exits with code 100 after `ThreadSanitizer` reports a `data race` whose `SUMMARY` line is `data race /-S/contrib/libs/clang20-rt/lib/tsan/rtl/tsan_interceptors_posix.cpp:741:3 in free`. The race is between the main thread tearing down the CPython interpreter (`Py_FinalizeEx` → `_Py_Finalize` → `_PyThreadState_DeleteList` → `free_threadstate` → `PyMem_RawFree` → `_PyMem_RawFree` → `free`) and a still-running Python worker thread that is still polling the GIL / event loop (`select_epoll_poll_impl` → `PyEval_SaveThread` → `_PyThreadState_Detach` → `drop_gil`).

Because the test driver treats any non-zero exit from the workload binary as a hard failure (and the embedded `ydbd` is later torn down with `Bad exit_code`), the wrapping pytest fails. 8 SANITIZER rows / 4 unique workloads across try_1/2/3.

## Environment

- Workflow run: https://github.com/ydb-platform/ydb/actions/runs/26185407422
- Job: `Regression-run_stress (release-tsan) / stable-26-2-1:release-tsan`
- Build preset: `release-tsan`, branch `stable-26-2-1`, commit `d241a3effbedb0849c2d19d61e097cdfe92512bb`
- HTML reports: [try_1](https://storage.yandexcloud.net/ydb-gh-logs/ydb-platform/ydb/Regression-run_stress/26185407422/ya-stable-26-2-1-x86-64-tsan/try_1/ya-test.html#FAIL) · [try_2](https://storage.yandexcloud.net/ydb-gh-logs/ydb-platform/ydb/Regression-run_stress/26185407422/ya-stable-26-2-1-x86-64-tsan/try_2/ya-test.html#FAIL) · [try_3](https://storage.yandexcloud.net/ydb-gh-logs/ydb-platform/ydb/Regression-run_stress/26185407422/ya-stable-26-2-1-x86-64-tsan/try_3/ya-test.html#FAIL)
[...]

PR #49726, opened 2026-08-12
Stop shared topic asyncio loop in stress workloads before exit

nshestakov:
## Summary
- Fixes TSan race from https://github.com/ydb-platform/ydb/issues/40952 in stress binaries that use sync topic clients (`kafka`, `streaming`, `transfer`): after `driver.stop()`, stop/join the leftover `"Common ydb topic event loop"` thread before process exit.
- Adds `shutdown_shared_topic_event_loop()` helper that prefers `ydb._topic_common.common._shutdown_shared_event_loop` (after python-sdk sync) and falls back to stop+join by thread name on current contrib.
- No changes under `contrib/python/ydb` (synced from elsewhere). SDK-side fix: https://github.com/ydb-platform/ydb-python-sdk/pull/874

## Test plan
- [ ] Confirm SDK PR #874 merged and/or fallback path stops the shared loop thread
- [ ] Re-run release-tsan stress for `streaming` / `transfer` / `kafka` and check that `Py_FinalizeEx` / `free_threadstate` vs `drop_gil` race is gone

### DIFF
--- a/ydb/tests/stress/common/common.py
+++ b/ydb/tests/stress/common/common.py
@@ -11,6 +11,41 @@ ydb.interceptor.monkey_patch_event_handler()
 logger = logging.getLogger(__name__)
 
 
+def shutdown_shared_topic_event_loop(timeout: Optional[float] = 30) -> None:
+    """
+    Stop the process-wide YDB topic asyncio loop thread before interpreter exit.
+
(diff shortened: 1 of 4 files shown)


nshestakov:
Closing: fix is SDK-only via atexit shutdown of the shared topic event loop (ydb-platform/ydb-python-sdk#874). Stress-side calls are not needed.

PR #49934, opened 2026-08-13
Prototype of unsafe truncate

flown4qqqq:
### Changelog entry 

...

### Status
This is a prototype. It is not intended to be merged as is and should be split into at least several separate pull requests before final integration. ***It should be draft PR forever.***

### Description for reviewers 

This adds a new operation,
```sql
TRUNCATE TABLE WITH (unsafe = true)
```
 — an unsafe truncate of a table.

Unlike the regular TRUNCATE TABLE, which is implemented as a separate schema operation, unsafe truncate is designed in a fundamentally different way: it is atomic but non-transactional.

The key idea is that the operation can run within the context of a transaction (i.e., it is aware of which transaction it belongs to), but its changes may become visible to other transactions even before the current transaction commits. This is a deliberate trade-off in favor of the performance/atomicity of this particular operation, and it may lead to a number of visibility anomalies — which should be taken into account when using this operation.


ydb/core/kqp/executer_actor/kqp_unsafe_truncate_executer.cpp
```diff
@@ -0,0 +1,642 @@
...
+
+            if (!entry.CdcStreams.empty()) {
+                ReplyError(Ydb::StatusIds::PRECONDITION_FAILED, NYql::TIssuesIds::KIKIMR_BAD_OPERATION,
+                    TStringBuilder() << "Unsafe TRUNCATE TABLE is not supported for " << table.Path
+                                     << ": it has a changefeed");
+                return;
+            }
+
+            for (const auto& index : entry.Indexes) {
+                if (index.GetType() != NKikimrSchemeOp::EIndexTypeGlobal) {
```
lex007in:
```suggestion
                if (index.GetType() != NKikimrSchemeOp::EIndexTypeGlobal
                    && index.GetType() != NKikimrSchemeOp::EIndexTypeGlobalUnique) {
```

PR #49947, opened 2026-08-13
report export failure when the multipart upload is gone

Enjection:
## Problem

`CompleteMultipartUpload` is what assembles the object in the storage. When it came back with `NoSuchUpload` the handler returned early without recording anything:

```cpp
const auto& error = result.GetError();
if (error.GetErrorType() == Aws::S3::S3Errors::NO_SUCH_UPLOAD) {
    return PassAway();
}
```

`PassAway()` reports `TEvFinish(Error.Empty(), ...)` to the scan, so an unset `Error` means success. An export whose data object was never assembled finished with `SUCCESS` and no issues while the table was missing from the bucket — a backup silently recorded as good.

## What NoSuchUpload actually means

It does not say on its own whether anything was written. The upload can be gone for two opposite reasons:

- it expired, was aborted by a lifecycle policy, or was lost across a restart — nothing was assembled, and the export must fail
- a completion of ours already succeeded and its response never reached us — the object is there and the export did finish

Reporting success loses data in the first case. Reporting failure fails a completed export in the second, which is not hypothetical: the retry path re-issues `CompleteMultipartUpload` after a lost or retryable response, and the storage answers `NoSuchUpload` because our own earlier completion consumed the upload. `TExportToS3Tests::ShouldRetryAtFinalStage` already covers exactly that sequence.

## Fix

Ask the storage rather than guess. On `NoSuchUpload` the uploader heads the data key and decides on the answer:

```cpp
if (error.GetErrorType() == Aws::S3::S3Errors::NO_SUCH_UPLOAD) {
    auto request = Aws::S3::Model::HeadObjectRequest()
        .WithKey(Settings.GetDataKey(DataFormat, CompressionCodec));
    this->Send(Client, new TEvExternalStorage::TEvHeadObjectRequest(request));
    return this->Become(&TThis::StateCheckUploadedData);
}
```

The object being present means an earlier completion succeeded and the export is done. Its absence means the upload is gone and nothing was written, which is reported as an error and cancels the export:

```
TExport::TTxProgress: issues during backing up, cancelling
Issue: 'shard: ...:2, error: s3 error: 404 ...'
```

A head that fails for a retryable reason takes the usual retry path rather than being read as absence.

Restarting the upload is not an option here, unlike `InvalidPart` in #49946: `ForceNewUpload` reuses the persisted upload id, which is exactly the one the storage no longer knows about. Recovering rather than failing would additionally require dropping the persisted `TS3Upload` row so a fresh id is minted, which is a larger change than this correctness fix.

## Tests

`TExportToS3Tests::ShouldNotSucceedWhenMultipartUploadIsLost` drops the `CompleteMultipartUpload` **request** rather than injecting an error into its response. `TS3Mock` is an http server, so by the time the response event exists the object has already been assembled; dropping the request means nothing is written and the missing data is asserted rather than assumed:

```cpp
UNIT_ASSERT_C(data.find("/data_00.csv") == data.end(), ...);   // nothing was assembled
TestGetExport(Runtime(), exportId, "/MyRoot", Ydb::StatusIds::CANCELLED);
```

Without the fix the first assertion passes and the second fails with `Unexpected status: SUCCESS, issues:` — the export claiming success over missing data.

The pre-existing `ShouldRetryAtFinalStage` covers the opposite direction. It injects a retryable error into the **response**, so the mock has already assembled the object and erased the upload; the retry then meets `NoSuchUpload` over data that exists, and the export must still succeed. The two together pin both readings of the error.

## Relation to #49946

Independent, and both branch from the same base. #49946 handles `InvalidPart`, where the upload still exists but the recorded part etags are stale, and the export fails loudly. This one handles `NoSuchUpload`, where the upload is gone. The `NO_SUCH_UPLOAD` branch short-circuits before `RequiresNewUpload()` is consulted, so #49946 does not cover it — verified by running this test against that branch.

### DIFF
--- a/ydb/core/tx/datashard/export_s3_uploader.cpp
+++ b/ydb/core/tx/datashard/export_s3_uploader.cpp
@@ -638,7 +638,10 @@ class TS3Uploader: public TActorBootstrapped<TS3Uploader<TSettings>> {
 
         const auto& error = result.GetError();
         if (error.GetErrorType() == Aws::S3::S3Errors::NO_SUCH_UPLOAD) {
-            return PassAway();
+            auto request = Aws::S3::Model::HeadObjectRequest()
+                .WithKey(Settings.GetDataKey(DataFormat, CompressionCodec));
+            this->Send(Client, new TEvExternalStorage::TEvHeadObjectRequest(request));
+            return this->Become(&TThis::StateCheckUploadedData);
         }
 
         if (CanRetry(error)) {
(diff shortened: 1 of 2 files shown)


pixcc (dismissed):

stanislav-shchetinin (dismissed):

azevaykin (dismissed):

maximyurchuk:

PR #49955, opened 2026-08-13
keep persisted import items in sync with the item list

Enjection:
## Problem

A schemeshard could not start, in a loop, after a restart:

```
VERIFY failed: Import not found: importId# 1137357911471465
  contrib/ydb/core/tx/schemeshard/schemeshard__init.cpp:5232
  ReadEverything(): requirement Self->Imports.contains(importId) failed
```

`TTxInit::ReadEverything` requires every `ImportItems` row to belong to a known import, and to have an index below the item count recorded for that import. A row that satisfies neither makes the tablet abort during boot — and it cannot be cleaned up, because cleaning it up needs a running schemeshard.

## Cause

`TImportInfo::FillItemsFromSchemaMapping` rebuilds the item list from the schema mapping and swaps it into `Items`. When the rebuilt list is shorter than the one persisted by `PersistCreateImport`, the recorded count stops covering the rows on disk: `PersistSchemaMappingImportFields` rewrites the surviving rows but never removes the rest.

From there, two ways to an unbootable tablet:

- restart directly → the surplus row's index is past the recorded count → `Invalid item's index`
- forget the import first → `PersistRemoveImport` deletes rows for the in-memory count but drops the `Imports` row unconditionally, orphaning the rest → `Import not found`

The list shrinks on two different paths.

**On failure**, the swap ran even when the mapping reported errors, so an unresolvable item shortened the list and the subsequent cancellation persisted the shorter count.

**On success**, no error is involved at all. `exclude_regexps` is matched against the destination path when items are created, and against the source object path when the mapping is applied. An item whose prefix is found in the mapping but whose every object is excluded contributes nothing, `notAllItemsFound` is never set, and the mapping succeeds with a shorter list.

The restart does not create the inconsistency; it is just when the state is read back. A cluster can carry it silently for a long time and only fail on the next start, e.g. during an upgrade.

## Fix

Swap only on success, so a failed expansion leaves the persisted state alone:

```cpp
if (result.Success) {
    Items.swap(items);
}
```

That covers the failing path only. On the succeeding path the list shrinks legitimately, so the persisted rows are kept in sync with it instead.

`PersistSchemaMappingImportFields` takes the size the list had before the mapping and deletes the rows past the new one:

```cpp
for (ui32 itemIdx = importInfo.Items.size(); itemIdx < itemsSizeBefore; ++itemIdx) {
    db.Table<Schema::ImportItems>().Key(importInfo.Id, itemIdx).Delete();
}
```

Both values are already in memory, so nothing is read back and the rows and the count that describes them are written in the same transaction. With the surplus rows gone the recorded count covers exactly the rows on disk again, so `PersistRemoveImport` deleting `[0, Items.size())` reaches all of them and leaves no orphan behind its unconditional drop of the parent row — which is what produced `Import not found`.

The checks in `TTxInit` are deliberately left alone. They guard an invariant, and skipping the rows would let genuine corruption load unnoticed — failing loudly is preferable to starting with silently incomplete state.

No stored format changes, so a database written by this code is still readable by the previous one.

## Tests

Both boot failures are reproduced on both paths, then the schemeshard is restarted:

- `TImportTests::ShouldNotCorruptItemsWhenSchemaMappingFails` — unresolvable item, restart directly
- `TImportTests::ShouldNotOrphanItemsWhenForgettingFailedImport` — unresolvable item, forget, then restart
- `TImportTests::ShouldNotCorruptItemsWhenSchemaMappingExcludesItem` — excluded item, mapping succeeds, restart directly
- `TImportTests::ShouldNotOrphanItemsWhenForgettingExcludedItemImport` — excluded item, forget, then restart

Each aborts inside `TTxInit::ReadEverything` without the fix, with `Invalid item's index` and `Import not found` respectively — the latter being the reported failure verbatim. The two excluded-item cases still abort with the swap guarded on success, which is what the counter addresses.

Reaching the schema mapping path needs a common `source_prefix` plus `EnableExportFiltering`; per-item prefixes never enter `EState::DownloadExportMetadata`.

`TestForgetImport` did not exist in `ut_helpers` and is added here, mirroring `TestForgetExport`.

## Note

This change is preventative only: clusters already holding such rows stay unbootable, and there is currently no in-product way to repair them.

### DIFF
--- a/ydb/core/tx/schemeshard/schemeshard_import.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_import.cpp
@@ -274,7 +274,11 @@ void TSchemeShard::PersistNewImportItem(NIceDb::TNiceDb& db, const TImportInfo& 
     );
 }
 
-void TSchemeShard::PersistSchemaMappingImportFields(NIceDb::TNiceDb& db, const TImportInfo& importInfo) {
+void TSchemeShard::PersistSchemaMappingImportFields(NIceDb::TNiceDb& db, const TImportInfo& importInfo, ui32 itemsSizeBefore) {
+    for (ui32 itemIdx = importInfo.Items.size(); itemIdx < itemsSizeBefore; ++itemIdx) {
+        db.Table<Schema::ImportItems>().Key(importInfo.Id, itemIdx).Delete();
+    }
+
     // There can be new items, so do at least the same as for creation
     for (ui32 itemIdx : xrange(importInfo.Items.size())) {
         const auto& item = importInfo.Items.at(itemIdx);
--- a/ydb/core/tx/schemeshard/schemeshard_impl.h
+++ b/ydb/core/tx/schemeshard/schemeshard_impl.h
@@ -1682,7 +1682,7 @@ public:
 
     static void PersistCreateImport(NIceDb::TNiceDb& db, const TImportInfo& importInfo);
     static void PersistNewImportItem(NIceDb::TNiceDb& db, const TImportInfo& importInfo, ui32 itemIdx);
-    static void PersistSchemaMappingImportFields(NIceDb::TNiceDb& db, const TImportInfo& importInfo);
+    static void PersistSchemaMappingImportFields(NIceDb::TNiceDb& db, const TImportInfo& importInfo, ui32 itemsSizeBefore);
     void PersistRemoveImport(NIceDb::TNiceDb& db, const TImportInfo& importInfo);
     static void PersistImportState(NIceDb::TNiceDb& db, const TImportInfo& importInfo);
     static void PersistImportSettings(NIceDb::TNiceDb& db, const TImportInfo& importInfo);
--- a/ydb/core/tx/schemeshard/schemeshard_import__create.cpp
+++ b/ydb/core/tx/schemeshard/schemeshard_import__create.cpp
@@ -1451,6 +1451,7 @@ private:
             }
         }
 
+        const ui32 itemsSizeBefore = importInfo->Items.size();
         const TImportInfo::TFillItemsFromSchemaMappingResult fillResult = importInfo->FillItemsFromSchemaMapping(Self);
         if (!fillResult.Success) {
             return CancelAndPersist(db, importInfo, -1, fillResult.ErrorMessage, "invalid items in schema mapping");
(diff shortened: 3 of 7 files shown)


ydb/core/tx/schemeshard/schemeshard_info_types.h
```diff
@@ -3614,6 +3614,7 @@ struct TExportInfo: public TSimpleRefCount<TExportInfo> {
     TString SanitizedToken;  // required for making audit log records
 
     TVector<TItem> Items;
+    ui32 PersistedItemCount = 0;
```
pixcc: Не понял зачем это на стороне экспорта делать

ydb/core/tx/schemeshard/schemeshard_import.cpp
```diff
@@ -272,9 +281,16 @@ void TSchemeShard::PersistNewImportItem(NIceDb::TNiceDb& db, const TImportInfo&
         NIceDb::TUpdate<Schema::ImportItems::SrcPath>(item.SrcPath),
         NIceDb::TUpdate<Schema::ImportItems::ParentIndex>(item.ParentIdx)
     );
+
+    PersistImportItemCount(db, importInfo, Max(importInfo.PersistedItemCount, itemIdx + 1));
 }
 
-void TSchemeShard::PersistSchemaMappingImportFields(NIceDb::TNiceDb& db, const TImportInfo& importInfo) {
+void TSchemeShard::PersistSchemaMappingImportFields(NIceDb::TNiceDb& db, TImportInfo& importInfo) {
+    for (ui32 itemIdx = importInfo.Items.size(); itemIdx < importInfo.PersistedItemCount; ++itemIdx) {
```
pixcc: Насколько я вижу этот PersistedItemCount используется только тут как-то осмысленно. Предлагаю не делать его полем, которое протаскивается в 99 местах. А просто сделать его локальным параметром этой функции – itemsSizeBefore

ydb/core/tx/schemeshard/schemeshard_info_types.h
Enjection: Для некоторой симметричности, но вообще согласен, давай держать фикс минимальным. Убрал.

ydb/core/tx/schemeshard/schemeshard_import.cpp
Enjection: Я делал что бы мы могли расхождение поймать, т.е. мы на старте гурзим и то и другое с диска и если математика разошлась, то мы верифайнемся тут `schemeshard_import.cpp:327`

Enjection: Вообще говоря выглядит, что действительно можно сделать так и мы ничего не потеряем. И в локальную базу лишний раз не ходим и проблем новых не наживаем

pixcc (approved):

maximyurchuk:

PR #50351, opened 2026-08-18
[detailed-metrics] wire up collection of detailed metrics and publishing it to ydb_detailed_raw group

jsjant:
### Description for reviewers 

- extend TabletCountersForgetTablet to free counter tree from leftovers when Hive moves tablet away
- limit published detailed **aggregated** metrics to SUM and MAX only
- publish all the stuff to ydb_detailed_raw group that's local per node
- **process-wide** mutex guards counters tree shape of TNodeDatabaseMetricsAggregatorImpl with lock in AddCounters and ForgetTablet to avoid interleaving Hive moves (eg when Forget thinks it's the last tablet and removes top entry from counters with table name, but at the same time another tablet of this table arrives and tries to add its counters)

### DIFF
--- a/ydb/core/tablet/tablet_counters_aggregator.cpp
+++ b/ydb/core/tablet/tablet_counters_aggregator.cpp
@@ -1,7 +1,7 @@
 #include "tablet_counters_aggregator.h"
 #include "tablet_counters_app.h"
 #include "labeled_counters_merger.h"
-#include "labeled_db_counters.h"
+#include "detailed_metrics/node_database_metrics_aggregator.h"
 #include "detailed_metrics/ydb_metrics_mapper.h"
 #include "private/aggregated_counters.h"
 #include "private/aggregated_tablet_counters.h"
@@ -17,6 +17,7 @@
 #include <ydb/core/base/feature_flags.h>
 #include <ydb/core/base/appdata.h>
 #include <ydb/core/base/counters.h>
+#include <ydb/core/base/path.h>
 #include <ydb/core/sys_view/common/events.h>
 #include <ydb/core/sys_view/service/db_counters.h>
 #include <ydb/core/sys_view/service/sysview_service.h>
@@ -44,6 +45,8 @@
 namespace NKikimr {
 
 const ui32 WAKEUP_TIMEOUT_SECONDS = 4;
+
+constexpr TDuration DATABASE_PATH_RESOLVE_TIMEOUT = TDuration::Seconds(10);
 
 TStringBuf GetHistogramAggregateSimpleName(TStringBuf name) {
     TStringBuf buffer(name);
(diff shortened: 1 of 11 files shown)


ydb/core/tablet/tablet_counters_aggregator.cpp
```diff
@@ -82,19 +86,191 @@ TActorId MakeTabletCountersAggregatorID(ui32 node, bool follower) {
...
+        }
+        if (!table) {
+            YDB_LOG_TRACE_CTX(ctx, "Skipping the counters of a tablet with no known table",
+                {"tabletId", tabletId},
+                {"followerId", followerId});
+            return;
+        }
+
+        const auto level = GetDetailedMetricsLevel(static_cast<ui32>(table->MetricsLevel));
+        if (!level) {
```
BlazingPhoenix: Returning here only stops future updates; it does not remove an existing contribution from db.Aggregator. If a table changes from Table/Partition to Disabled or Unspecified, its previous counters remain published indefinitely.
Table - Partition transitions similarly retain the original tree shape. Please reconcile the previous state at minimum call ForgetTablet when disabling, and rebuild when the effective level changes.

Required coverage: Table -> Disabled, Partition -> Disabled, Table <-> Partition, including followers.

ydb/core/tablet/tablet_counters_aggregator.cpp
```diff
@@ -984,7 +1280,8 @@ TTabletCountersAggregatorActor::Bootstrap(const TActorContext &ctx) {
         DbWatcherActorId = ctx.Register(NSysView::CreateDbWatcherActor(callback));
     }
 
-    TabletMon = new TTabletMon(appData->Counters, Follower, DbWatcherActorId);
+    TabletMon = new TTabletMon(appData->Counters, Follower, DbWatcherActorId,
```
BlazingPhoenix, line 1313: Database removal only reaches the leader aggregator. Both leader and follower actors now own DetailedMetricsByPathId, but the database watcher sends TEvRemoveDatabase only to MakeTabletCountersAggregatorID(..., false). Therefore follower state and follower-only partition leaves are not authoritatively cleaned on database deletion. Cleanup then depends on every follower’s forget event arriving successfully. Additionally, the watcher exists only when the unrelated EnableDbCounters flag is enabled.
leader-only notification https://github.com/ydb-platform/ydb/blob/2d420bedacf244ca0af68395052b1c197481ed72/ydb/core/tablet/tablet_counters_aggregator.cpp#L1024-L1027  
watcher gating https://github.com/ydb-platform/ydb/blob/2d420bedacf244ca0af68395052b1c197481ed72/ydb/core/tablet/tablet_counters_aggregator.cpp#L1278-L1284
local-only removal handler https://github.com/ydb-platform/ydb/blob/2d420bedacf244ca0af68395052b1c197481ed72/ydb/core/tablet/tablet_counters_aggregator.cpp#L1485-L1487

ydb/core/tablet/tablet_counters_aggregator.cpp
jsjant: Table/Partition → Disabled/Unspecified fixed here

Table ↔ Partition will be fixed later. Reconciliation has to happen inside the aggregator, on the table entry, not per tablet at the TCA: a table-level bucket is shared by every tablet of the table on the node, so forget-and-re-add driven per tablet thrashes — A forgets, B and C keep the entry alive at the old level, A gets re-added to the old tree.

ydb/core/tablet/tablet_counters_aggregator.cpp
jsjant: Fixed by accounting for Follower in TTabletsDbWatcherCallback

BlazingPhoenix (approved):

PR #50358, opened 2026-08-18
Harden Kafka parser against OOM and OOB on untrusted lengths

nshestakov:
### Changelog entry 

Kafka API: cap untrusted array/string/bytes/tagged-field lengths to remaining bytes, fix overflow-unsafe stream checks and v2 record-batch over-reads, skip short consumer-protocol blobs, and match Kafka ApiVersions handshake (KIP-511 fallback, Produce min=0).

### Description for reviewers 

Bug-bounty root cause: the Kafka proxy trusted a client-supplied array length and did `vector::resize(N)` before reading items. `MetadataRequest` with `Topics.length = INT_MAX` and almost no payload allocated tens of GB (ASan). Follow-up hardening and handshake compatibility:

1. **Array/string/bytes allocation cap** — require `length <= remaining`; for arrays also `count * sizeof(T) <= remaining` (fixed-width) and `<= MaxArrayBytes`. The proxy sets the cap to `max.message.bytes` (default 16MiB). Nested readers inherit it.
2. **`checkEof` without addition** — compare against `left()` so `Position + length` cannot wrap `size_t` and skip past the request buffer.
3. **v2 record-batch slice** — parse only inside the records `length` and `BatchLength` (legacy already sliced by `length`). Reject negative / oversized `BatchLength`. `WriteKafkaRecordBatch` rewrites `BatchLength` from the actual payload, like CRC.
4. **Consumer-protocol blob size** — OffsetFetch / JoinGroup used to read a version at offset 2 without checking size. Blobs shorter than 4 bytes are skipped / `nullopt`.
5. **Tagged fields** — `ReadTaggedFieldsCount` rejects `count > remaining/2`; `SkipTaggedField` / `ReadTag(size)` refuse a size larger than `left()` and skip leftover inside the declared tag (no temp `TBuffer`, `TKafkaBytes` would dangle). Regenerated `kafka_messages` and the consumer-protocol assignment parser use these helpers.
6. **ApiVersions KIP-511** — clients send their latest ApiVersions (Java v4/v5, librdkafka v3) before they know the broker range. Kafka keeps the connection and replies **v0** with `UNSUPPORTED_VERSION` plus `ApiKeys` listing only ApiVersions min/max. YDB used to `Read()` the body at the requested version and `PassAway()` on "Can't read version N". Versions outside `PresentVersions` (0–4) now skip the body, return that fallback, and serialize as v0. Supported v3/v4 still get a full `NONE` list.
7. **Produce min=0 in ApiVersions** — Kafka 4.0 still advertises Produce min=0 (KAFKA-18659) because some librdkafka versions treat min>0 as "Produce is missing". YDB advertised min=3; the parser already accepts 0–9. Report 0–9.

**Test plan**
- [x] `ydb/public/sdk/cpp/src/library/kafka` unit tests (`kafka_messages_int_ut`, `kafka_records_ut`)
- [x] `HugeArrayLengthOverSocketDoesNotCrashServer` in `ydb/core/kafka_proxy/ut`
- [x] `ConsumerProtocolBlob` tests in `ydb/core/kafka_proxy/ut`
- [x] Tagged-field count/skip tests in `kafka_messages_int_ut`
- [x] `GetApiVersionsUnsupportedVersionUsesKip511Fallback` / `ApiVersionsUnsupportedVersionKeepsConnection`
- [x] `GetApiVersionsAdvertisesProduceMinZero`


nshestakov (commented):
## Review: call chains and missing coverage

The four commits fix real parser bugs. A few of them do **not** sit on the path the bounty used, and several sibling paths are untested.

### 1. Array/string/bytes cap — this is the bounty path

```
socket
  → DoRead SIZE_PREPARE (ExpectedSize ≤ max.message.bytes)
  → MESSAGE_PROCESS
  → TKafkaReadable + SetMaxArrayBytes(max.message.bytes)
  → Header.Read / Message.Read
  → TypeStrategy::DoRead
       array  → EnsureArrayAllocationFits → resize
       string → EnsureLengthFitsRemaining → ReserveAndResize
       bytes  → EnsureLengthFitsRemaining → Bytes() / ReserveAndResize
  → yexception → PassAway()
```

Produce/Fetch `records` on the wire are **`TKafkaBytes` / `TKafkaBytesHolder`**, not `TKafkaRecords`. The array cap applies to Topics/Partitions/etc.; the records blob itself is only a length-checked view.

Covered: Metadata `Topics.length = INT_MAX` over the socket (v5 and compact v12); unit tests for array/string/bytes/holder, fixed-width leftover, `MaxArrayBytes`.

Not covered:

- Other request arrays over the socket (JoinGroup `protocols`, OffsetCommit `topics`, Fetch `topics`, nested `partitions`).
- Variable-length struct arrays that pass `length ≤ remaining` but blow `count * sizeof(T)` (only a tiny `MaxArrayBytes` unit test).
- `SetMaxArrayBytes(0)` falling back to 32MiB if `max.message.bytes` is 0.
- Secondary parsers that **do not** call `SetMaxArrayBytes`:
  - `ReadKafkaRecordBatch` / `ReadRecordBatch` (produce fallback, `TKafkaBatchCutter`, Topic `TKafkaBatchCodec`)
  - `TryReadConsumerProtocolBlob` (JoinGroup / OffsetFetch)

### 2. `checkEof` without addition

```
read / Bytes / skip → checkEof(length > left())
take(shift)         → shift >= left()
```

Unit-tested with `Max<size_t>` and an empty buffer. No socket test (hard to hit wrap with `max.message.bytes` ≤ 16MiB on 64-bit). Still the right fix.

### 3. v2 batch slice — Produce does not use `TKafkaRecords::DoRead`

```
Produce.records (TKafkaBytes, already length-checked)
  → ParseProduceRecords
  → ReadKafkaBatchHeader → TKafkaBatchHeader::Read → TakeRecordBatchBody(BatchLength)
  → if StoreRawRecords (batching + v2 or uncompressed): store blob, skip full parse
  → else: ReadRecordBatch → TKafkaRecordBatch::Read → TakeRecordBatchBody
       compressed: Bytes(body.left()) → decompress → EnsureArrayAllocationFits → resize
```

`TypeStrategy<TKafkaRecords>::DoRead` (slice by records `length`) is **not** on the proxy Produce/Fetch path. It is only exercised by `RecordsFieldReadDoesNotConsumePastLength`. Production over-read is closed by `TakeRecordBatchBody`, not by that TypeStrategy change.

`ReadKafkaRecordBatch` still uses default `MaxArrayBytes` (32MiB) and `SetAllowCompressed(true)`.

Covered (unit): extra bytes after `BatchLength`, oversized/negative `BatchLength`, truncated header, stale `BatchLength` on write, records-field `length` slice (test-only type).

Not covered:

- Produce over the socket with `BatchLength = INT_MAX` or extra garbage after a valid batch inside `records`.
- Produce compressed v2 with `recordsCount` huge vs small decompressed payload (`EnsureArrayAllocationFits` after gzip/zstd).
- Gzip/zstd bomb: `DecompressRecordBatchPayload` → `ReadAll()` has **no output cap**. With batching on, Produce stores the blob raw; `TKafkaBatchCutter::Cut` / Topic codec later call `ReadKafkaRecordBatch` and can OOM. This is outside the four commits but it is the next call on the same data.

### 4. Consumer-protocol blob — size gate only

```
JoinGroup.protocols[].metadata (TKafkaBytes)
  → GetSubscriptions → TryReadConsumerProtocolBlob
  → TKafkaReadSessionActor::TryFillTopicsToRead  (no try/catch)

SyncGroup.assignments[].assignment (TKafkaBytes)
  → balancer persists String to YDB
  → OffsetFetch KQP row
  → ParseGroupsAssignments → TryReadConsumerProtocolBlob
  → Handle()  (no try/catch)
```

`< 4` bytes → `nullopt` / skip. `≥ 4` bytes, malformed body → `result.Read()` throws → `TKafkaExceptionHandler` poisons the **TCP connection**.

Version is `memcpy` of two bytes at offset 2, **no `NormalizeNumber`**. Layout is `[ignored i16][body…]`; tests pass because the first two body bytes of an assignment/subscription are usually `0x00 0x00` (array length) → version 0.

Covered: helper unit tests for 0–3 bytes, null, valid SyncGroup-shaped assignment, GetSubscriptions short/empty/null/wrong protocol/valid.

Not covered:

- JoinGroup over the socket with 3-byte metadata (should not OOB).
- JoinGroup / OffsetFetch with `≥ 4` byte garbage (throws, drops connection).
- OffsetFetch with a short assignment **stored in YDB** (actor + `TString` → `TKafkaBytes`).
- `TKafkaTestClient::GetAssignments` / `FillTopicsFromJoinGroupMetadata` still do the old pointer arithmetic with no size check (test-only, will OOB on a short blob).

### Suggested follow-up tests (highest value)

1. Produce socket: v2 batch with `BatchLength` larger than the records blob → `INVALID_RECORD` or connection close, no crash.
2. JoinGroup socket: metadata length 3 → no OOB; metadata length ≥ 4 but invalid body → no crash (today: connection close).
3. `ReadKafkaRecordBatch` of compressed batch with huge `recordsCount` and tiny payload → exception, no huge `resize`.
4. `TryReadConsumerProtocolBlob` on a 4+ byte truncated assignment → `nullopt` (if we catch) or documented connection close.

The bounty OOM path looks correctly closed. The gaps are: Produce/Fetch records are bytes not `TKafkaRecords`, secondary readers ignore `MaxArrayBytes`, and consumer-protocol parse still throws on long malformed blobs.

Shfdis (approved):

kuzin57 (approved):

PR #51073, opened 2026-08-25
Support DEFAULT (non-literal) expressions for table columns

dahbka-lis:
### Changelog entry 

Added feature-flagged support for expression-based column `DEFAULT` values.

### Description for reviewers 

- Parses and stores `DEFAULT` expressions with their SQL context in table metadata.
- Type-checks and compiles expressions into omitted write columns; explicit values take precedence.
- Uses `EnableDefaultFromExpression`; literal defaults keep the existing path when it is disabled.
- Rejects expression defaults in `ADD/ALTER COLUMN` and column tables.
- Adds end-to-end coverage and separates generated-column-only tests from default-expression tests.


ydb/core/protos/flat_scheme_op.proto
```diff
@@ -102,10 +102,16 @@ message TColumnCompression {
...
+        GENERATED_VIRTUAL = 1;
+        DEFAULT = 2;
+    }
+
     optional string ExprText = 1;
     optional string Context = 2;
-    repeated string DependencyColumnNames = 3;
-    optional bool Stored = 4 [default = false];
+    repeated string DependencyColumnNames = 3; // Always empty for Kind == DEFAULT
+    optional EKind Kind = 4;
```
CyberROFL, line 114: Это ведь ломает обратную совместимость. Или `Stored` ни в какую ветку не попал?

dahbka-lis: Пока никуда не попал, всё ок

PR #51351, opened 2026-08-27
Fix united shared pool mode reporting

kruall:
### Changelog entry 

Fixed United shared pools being treated as legacy shared pools by the CPU harmonizer.

### Description for reviewers 

Initialize `TSharedExecutorPool` United flag from its config and cover both default and enabled modes.


serbel324 (approved):

kruall:
Superseded by #51356 after migrating the pair to native GitHub stack #51358.

PR #51352, opened 2026-08-27
Allow processing EvVBlock in VDisk running on RED

alexvru:
### Changelog entry 

Allow processing EvVBlock in VDisk running on RED

### Description for reviewers 

This patch allows system tablets to boot even on VDisks in RED space state.


mregrock (approved):
LGTM

PR #51357, opened 2026-08-27
Start shared pool for all-shared executors

kruall:
### Changelog entry 

Fixed actor systems with all executor threads shared failing to process events unless legacy shared threads were also enabled.

### Description for reviewers 

Stacked on #51356. Treat `AllThreadsAreShared` as sufficient to create shared infrastructure; thread-count arithmetic is unchanged.


mregrock (approved):
LGTM

PR #51374, opened 2026-08-27
Support SpaceDataKind for restore-get queries

alexvru:
### Changelog entry 

Support SpaceDataKind for restore-get queries

### Description for reviewers 

This patch allows to specify DataKind for any kind of queries (nest egg for the future), and currently implements support of DataKind for get queries involving writes (get with restore, range, discover, etc).


mregrock (approved):
LGTM

PR #51446, opened 2026-08-28
Fix united mode in tiny actor system auto config

kruall:
### Changelog entry 

Fixed United pool mode being ignored by actor-system auto configuration for one to three CPUs with shared threads enabled.

### Description for reviewers 

Stacked on #51357. Propagate `UseUnitedPool` through the tiny auto-config path without changing thread topology, and cover CPU 1–4 × Shared × United.


mregrock (approved):
LGTM

PR #51455, opened 2026-08-28
Fix adjacent pool wakeup after owner becomes idle

kruall:
### Changelog entry 

Fixed delayed actor events getting stuck when an adjacent executor-pool owner becomes idle.

### Description for reviewers 

With the production CPU=2 Shared+United configuration, Batch owns no worker and has no foreign slots, so its work can only run on the shared thread owned by User. Activation scheduling previously tried local workers and then the foreign fallback, leaving a parked adjacent owner unnotified. The change records the unique owner of each adjacent pool and wakes it through the shared pool before foreign workers. The regression test uses the production auto-configuration and verifies that the delayed Batch event runs on its User owner after idle.

### DIFF
--- a/ydb/core/driver_lib/run/auto_config_initializer_ut.cpp
+++ b/ydb/core/driver_lib/run/auto_config_initializer_ut.cpp
@@ -1,13 +1,153 @@
 #include "auto_config_initializer.h"
+#include "config_helpers.h"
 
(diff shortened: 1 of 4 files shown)


mregrock (dismissed):
LGTM

mregrock (approved):
LGTM

PR #51491, opened 2026-08-28
ydb-bench: stabilize throughput plateau selection

kruall:
### Changelog entry 

Improved local YDB throughput searches to select the lowest stable saturated load and show ternary-search progress clearly.

### Description for reviewers 

Plateau comparisons now use an absolute tolerance, final selection uses the lowest saturated point within that tolerance, and the run UI displays the evolving plateau candidate and search interval.


kruall:
Replaced by #51965 in GitHub Stack #51982.

PR #51537, opened 2026-08-29
ydb-bench: compare local YDB results against baseline

kruall:
### Changelog entry 

Adds baseline-oriented comparison of local YDB benchmark results, configuration, environment, and binary identity.

### Description for reviewers 

This is stacked on #51491. The comparison page now loads a bounded compact projection for local YDB profiles, lets the user choose a baseline, reports directional deltas for selected throughput, latency, and errors, and suppresses numeric deltas for semantically incompatible workloads. Configuration, affinity, host topology, and build differences remain explicit; generic charts are preserved below.

PR #51540, opened 2026-08-29
ydb-bench: overlay local YDB search curves

kruall:
### Changelog entry 

Adds synchronized comparison curves for local YDB load-search results.

### Description for reviewers 

This is stacked on #51537. Compatible local YDB profiles are grouped by dynamic-node count and overlaid by their actual searched load for throughput, latency, per-role CPU, and errors. All panels share one cursor and one color per profile geometry. Each line connects only its own measured loads, while a dedicated local-YDB chart-data query prevents unrelated malformed summaries from disabling these curves.


kruall:
Replaced by #51967 in GitHub Stack #51982.

PR #51541, opened 2026-08-29
YDB Bench: verify selected local YDB load

kruall:
## Changelog entry
Feature: Added bounded independent holdout measurements for the load selected by local YDB search.

## Description for reviewers
This is stacked on #51540. The selected load and geometry stay unchanged; completed holdout samples are stored in separate CSV artifacts and become the reported metric source. Latency uses the same aggregate SLO evaluation as search, while throughput holdout remains diagnostic. The web UI shows holdout progress and prevents deltas between Search and Holdout metric sources.

PR #51542, opened 2026-08-29
ydb-bench: show recent local YDB activity

kruall:
### Changelog entry 

Added a bounded recent-activity view for local YDB benchmark profiles.

### Description for reviewers 

Stacked on #51541. The profile page incrementally replays projected lifecycle and command events while preserving activity across reloads; event-log reads and responses remain bounded.


kruall:
Replaced by #51969 in GitHub Stack #51982.

PR #51543, opened 2026-08-29
ydb-bench: centralize local YDB workload catalog

kruall:
### Changelog entry 

Prepared local YDB benchmarks for additional workload types without changing existing KV and stock behavior.

### Description for reviewers 

Stacked on #51542. A typed workload catalog now owns config defaults, validation, schema, CLI commands, load parameters, and web-builder metadata for KV and stock.

PR #51544, opened 2026-08-29
YDB Bench: add local YDB workload lifecycle

kruall:
### Changelog entry 

Added lifecycle-aware orchestration for local YDB workloads.

### Description for reviewers 

Stacked on #51543.

Introduces declarative sample/profile dataset scopes, separate/inline warmup modes, reliable cleanup, and measurement-window-aware Linux CPU telemetry while preserving existing KV and stock behavior.


kruall:
Replaced by #51971 in GitHub Stack #51982.

PR #51545, opened 2026-08-29
YDB Bench: add local log workload

kruall:
### Changelog entry 

Added the YDB CLI log workload to local YDB benchmarks.

### Description for reviewers 

Stacked on #51544.

Adds typed log workload configuration for insert, upsert, and bulk-upsert operations, safe thread-search defaults in the web builder, workload-aware throughput units, and a runnable YAML example.


kruall:
Replaced by #51972 in GitHub Stack #51982.

PR #51546, opened 2026-08-29
YDB Bench: Reject empty local YDB samples

kruall:
### Changelog entry 

Reject local YDB measurements where any repetition completes without successful operations.

### Description for reviewers 

Stacked on #51545. Treat zero-success repetitions as infeasible independently of the error policy, preserve their validity metadata for diagnostics and verification, and avoid publishing misleading zero-latency points in summaries and UI charts.


kruall:
Replaced by #51973 in GitHub Stack #51982.

PR #51547, opened 2026-08-29
YDB Bench: Scope workload datasets by geometry

kruall:
### Changelog entry 

Recreate geometry-scoped local YDB workload datasets when the dynamic-node count changes.

### Description for reviewers 

Stacked on #51546. Add an explicit geometry dataset lifecycle: prepare once per dynamic-node stage, reuse it across search attempts and final verification, clean it before scaling, and preserve cleanup and cancellation semantics. Shared lifecycle commands are recorded under each geometry directory.


kruall:
Replaced by #51974 in GitHub Stack #51982.

PR #51548, opened 2026-08-29
YDB Bench: Add workload result adapters

kruall:
### Changelog entry 

Add extensible result adapters for local YDB CLI workloads.

### Description for reviewers 

Stacked on #51547. Move generic Total parsing behind a typed result-adapter contract, publish self-describing workload result schemas, preserve optional metric sets without fabricated values, and support adapter-provided CPU windows, details artifacts, progress durations, and latency metric mappings. KV, stock, and log retain their existing result shape.


kruall:
Replaced by #51975 in GitHub Stack #51982.

PR #51550, opened 2026-08-29
ydb-bench: consume workload result schemas in the web UI

kruall:
### Changelog entry 

Made local YDB result pages, charts, comparisons, and the profile builder use each workload’s declared result schema.

### Description for reviewers 

Consumes the workload result schema persisted by #51548 so custom workload metrics, units, SLO mappings, and error capabilities are rendered safely while historical runs retain their original generic UI. Stacked on #51548.


kruall:
Replaced by #51976 in GitHub Stack #51982.

PR #51551, opened 2026-08-29
ydb-bench: add a local TPCC workload

kruall:
### Changelog entry 

Added the YDB CLI TPCC workload to local YDB benchmark profiles with session-load search, inline warmup, and typed latency results.

### Description for reviewers 

Adds geometry-scoped TPCC init/import/run/cleanup orchestration, strict JSON metric parsing, max-session validation, Builder support, and documentation. Stacked on #51550.


kruall:
Replaced by #51977 in GitHub Stack #51982.

PR #51554, opened 2026-08-29
ydb-bench: add a local Topic workload

kruall:
### Changelog entry 

Added the YDB CLI Topic full workload to local YDB benchmark profiles with message-rate search and end-to-end p99 results.

### Description for reviewers 

Adds sample-scoped Topic init/run/cleanup orchestration, strict timestamped statistics parsing, consumer-normalized throughput, fixed p99 latency SLOs, Builder support, and documentation. Stacked on #51551.

PR #51555, opened 2026-08-29
ydb-bench: fix local dedicated node readiness

kruall:
### Changelog entry 

Fixed local YDB startup with dedicated dynamic nodes whose database status does not list computational units.

### Description for reviewers 

Treats the database RUNNING state as the control-plane readiness signal and leaves the existing per-endpoint SchemeDescribe probes to verify every dedicated dynamic node. Stacked on #51554.


kruall:
Replaced by #51979 in GitHub Stack #51982.

PR #51559, opened 2026-08-29
ydb-bench: use admitted TPCC latency for search

kruall:
### Changelog entry 

Fixed TPCC latency search to exclude the queue imposed by its own max-sessions limit.

### Description for reviewers 

Uses the CLI admitted-latency percentiles for canonical pXX metrics and SLO search, keeps full and pure latency in raw results, and bumps the result schema to prevent incompatible comparisons. Stacked on #51555.


kruall:
Replaced by #51980 in GitHub Stack #51982.

PR #51560, opened 2026-08-29
ydb-bench: verify effective TPCC threads

kruall:
### Changelog entry 

TPCC benchmarks now reject results when the CLI reduces the configured execution thread count.

### Description for reviewers 

Checks the CLI-reported effective worker count against client.threads so CPU-based clamping cannot silently label a run with the wrong configuration. Stacked on #51559.


kruall:
Replaced by #51981 in GitHub Stack #51982.

PR #51588, opened 2026-08-30
Executor GC: a pending DoNotKeep mark must pin its history entry

lberserq:
`TExecutorGCLogic::ApplyDelta` records blobs from `delta.Deleted` into `CommittedDelta` but never tells `HistoryCutter` about them — `SeenBlob` is fed only from `delta.Created`:

```cpp
for (const TLogoBlobID &blobId : delta.Deleted) {
    auto &channel = ChannelInfo[blobId.Channel()];
    channel.CommittedDelta[time].Deleted.push_back(blobId);
    // no HistoryCutter.SeenBlob(blobId)
}
```

A blob still awaiting its DoNotKeep flag therefore does not pin the history entry that resolves its generation to a group. The cutter sees the range as empty and offers it for cutting. Once the entry is gone, `GroupForGeneration` returns the `Max<ui32>()` sentinel for that generation and the flag can never be delivered.

Observed on a test cluster as a boot loop: a GC request addressed to a group that no longer resolves → BS error → `TEvPoison` → the tablet reboots into the same state, roughly 87k generations burned before it was diagnosed.

This is reachable today without any feature flag: KeyValue (`keyvalue_state.cpp:853`) and BlobDepot (`data_trash.cpp:379`) already send `TEvCutTabletHistory`, and the executor cuts its own history in `flat_executor_gclogic.cpp:150`.

### Fix

Feed `HistoryCutter::SeenBlob` from the `Deleted` loop as well, so a pending DoNotKeep mark holds its entry alive until GC has delivered the flag.

### Testing

**`THistoryCutter::DeletedBlobInApplyDeltaPinsHistoryEntry`** — history entries at generations 10 and 100; a blob at generation 50 goes into `delta.Deleted`; after the public `ApplyLogEntry` → `ApplyDelta`, `GetHistoryToCut(0)` must be empty.

Ablation: remove the added `SeenBlob` call and the cutter never observes generation 50, offers `[10, 100)` for cutting, and the assert fails.

`ya make -tt ydb/core/tablet_flat/ut -F 'THistoryCutter::*'` → all pass.

### Merge order

#50180 (ColumnShard CutHistory) depends on this and on #51684, and must not merge before both. No ColumnShard content here; the per-channel cutter guard stays in #50180.


azevaykin (changes requested):
These are two different defects.
Let's split in two PRs.

vporyadke (approved):

lex007in (approved):

azevaykin (approved):

PR #51681, opened 2026-08-31
Wake adjacent pool owner for queued activations

kruall:
### Changelog entry 

Fixed delayed actor events getting stuck when an adjacent executor-pool owner is idle.

### Description for reviewers 

Each adjacent pool is mapped to its single shared-thread owner. Activation scheduling now notifies and wakes that owner after local workers and before the foreign-slot fallback, while startup validates that the owner is unique and owns a shared thread.

This PR is stacked on #51455, which contains the production-config regression test.


kruall:
Superseded by #51455, where the regression test and runtime fix now live in one atomic commit.

PR #51683, opened 2026-08-31
Fix scheduler cookie lifetime race

kruall:
### Changelog entry 

Fixed a race that could crash a node while cancelling scheduled actor events.

### Description for reviewers 

**Failure scenario:** the scheduler fires a cookie while another participant concurrently cancels or releases the same cookie. Both may enter virtual detach calls, and the call that decrements the shared counter to zero can run `delete this` while the other virtual call has not returned yet, invalidating the object and its vtable in the scheduler path.

**Root cause:** the single atomic `Value` combined two different responsibilities: deciding the 2-way/3-way fire-versus-cancel result and managing object lifetime. Reaching zero proved that all participants had started detaching, but did not prove that their virtual calls had completed.

**Fix:** `ISchedulerCookie` now keeps separate lifetime slots. Public `Detach()` and `DetachEvent()` first complete the virtual logical-state operation and only then release the participant lifetime slot. The cookie is reclaimed after every participant has left the virtual detach implementation, while the existing fire/cancel semantics remain unchanged.

The original Query01 crash was not reproduced locally; this root cause is inferred from the reported `cookie->Detach()` crash site and the lifetime interleaving in the implementation.

### DIFF
--- a/ydb/core/util/actorsys_test/testactorsys.h
+++ b/ydb/core/util/actorsys_test/testactorsys.h
@@ -944,8 +944,12 @@ private:
 
 class TFakeSchedulerCookie : public ISchedulerCookie {
 public:
-    bool Detach() noexcept override { delete this; return false; }
-    bool DetachEvent() noexcept override { Y_ABORT(); }
+    TFakeSchedulerCookie()
+        : ISchedulerCookie(1)
+    {}
+
+    bool DetachImpl() noexcept override { return false; }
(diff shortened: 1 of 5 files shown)


kruall:
Closing this patch because the proposed lifetime race has not been reproduced, and the change does not address a scheduler queue entry that already contains a stale cookie pointer. We will first reproduce the original failure on b79d590 and establish the violated ownership invariant.

PR #51684, opened 2026-08-31
Executor GC: never build a collect request into the sentinel group

lberserq:
`TExecutorGCLogic::TChannelInfo::SendCollectGarbage` resolves each blob's generation to a group with `GroupForGeneration` and then dereferences `&affectedGroups[activeGroup]` without checking the result:

```cpp
activeGroup = channelInfo->GroupForGeneration(blobId.Generation());
vec = &affectedGroups[activeGroup].first;
...
vec->push_back(blobId);
```

For a generation below the first surviving history entry, `GroupForGeneration` returns the `Max<ui32>()` sentinel. The sentinel then becomes a key in `affectedGroups`, and the loop at the end of the bloated branch issues a `TEvCollectGarbage` against a group that does not exist.

Observed on a test cluster as a boot loop: collect to the sentinel group → BS error → `TEvPoison` → the tablet reboots into the same state, roughly 87k generations burned before it was diagnosed.

This is reachable today without any feature flag: KeyValue (`keyvalue_state.cpp:853`) and BlobDepot (`data_trash.cpp:379`) already send `TEvCutTabletHistory`, and the executor cuts its own history in `flat_executor_gclogic.cpp:150`.

### Fix

Skip blobs whose generation resolves to the sentinel, in both the Keep and DoNotKeep loops. Dropping them is correct rather than merely safe: a generation below the first surviving entry was collected together with the entry that was cut away, so there is nothing left to flag.

### Testing

**`THistoryCutter::SendCollectGarbageSkipsSentinelGroup`** reproduces the state that triggers it — one surviving history entry starting at generation 10, with GC still holding marks for generations 5 and 6 — registers edge actors as the BS proxies for the surviving and the sentinel group, and asserts both directions:

- the surviving group still receives its collect request;
- the sentinel group receives none.

The second assertion alone would be satisfied by a guard that dropped everything, which is why the first is there too.

Ablation: restore `vec = &affectedGroups[activeGroup].first/.second` and drop the `if (vec)` checks — the test fails with **2** requests to the sentinel group, one per loop.

`ya make -tt ydb/core/tablet_flat/ut -F 'THistoryCutter::*'` → 12 tests, all pass.

### Merge order

#50180 (ColumnShard CutHistory) depends on this and on #51588, and must not merge before both.

### DIFF
--- a/ydb/core/tablet_flat/flat_executor.cpp
+++ b/ydb/core/tablet_flat/flat_executor.cpp
@@ -4029,6 +4029,12 @@ void TExecutor::UpdateUsedTabletMemory() {
 }
 
 void TExecutor::UpdateCounters(const TActorContext &ctx) {
+    if (GcLogic && Counters) {
+        if (const ui64 dropped = GcLogic->TakeSentinelDroppedMarks()) {
+            Counters->Cumulative()[TExecutorCounters::GC_SENTINEL_DROPPED_MARKS].Increment(dropped);
+        }
+    }
+
     TAutoPtr<TTabletCountersBase> executorCounters;
     TAutoPtr<TTabletCountersBase> externalTabletCounters;
 
--- a/ydb/core/tablet_flat/flat_executor_counters.h
+++ b/ydb/core/tablet_flat/flat_executor_counters.h
@@ -121,7 +121,6 @@ namespace NTabletFlatExecutor {
     XX(GC_FORGOTTEN, "GcBlobsForgotten") \
     XX(GC_KEEPSET, "GcKeepFlagsSet") \
     XX(GC_NOTKEEPSET, "GcNotKeepFlagsSet") \
-    XX(GC_SENTINEL_DROPPED_MARKS, "GcSentinelDroppedMarks") \
     XX(CONSUMED_CPU, "ConsumedCPU") \
     XX(COMPACTION_READ_POSTPONED, "CompactionReadPostponed") \
     XX(COMPACTION_READ_CACHE_HITS, "CompactionReadCacheHits") \
@@ -134,6 +133,7 @@ namespace NTabletFlatExecutor {
     XX(BACKUP_CHANGELOG_ERRORS, "BackupChangelogErrors") \
     XX(BACKUP_SNAPSHOT_BYTES_WRITTEN, "BackupSnapshotBytesWritten") \
     XX(BACKUP_CHANGELOG_BYTES_WRITTEN, "BackupChangelogBytesWritten") \
+    XX(GC_SENTINEL_DROPPED_MARKS, "GcSentinelDroppedMarks") \
 
 // don't change order!
 #define FLAT_EXECUTOR_PERCENTILE_COUNTERS_MAP(XX) \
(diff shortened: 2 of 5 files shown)


ydb/core/tablet_flat/flat_executor_counters.h
```diff
@@ -121,6 +121,7 @@ namespace NTabletFlatExecutor {
     XX(GC_FORGOTTEN, "GcBlobsForgotten") \
     XX(GC_KEEPSET, "GcKeepFlagsSet") \
     XX(GC_NOTKEEPSET, "GcNotKeepFlagsSet") \
+    XX(GC_SENTINEL_DROPPED_MARKS, "GcSentinelDroppedMarks") \
```
lex007in: According to the comment above (line 80), this entry should be placed at the end

ydb/core/tablet_flat/flat_executor_gclogic.cpp
```diff
@@ -501,6 +504,12 @@ void TExecutorGCLogic::TChannelInfo::SendCollectGarbage(TGCTime uncommittedTime,
                     affectedGroups[xit->GroupID];
             }
 
+            // Below the first surviving entry GroupForGeneration returns Max<ui32>(): the entry
+            // was cut only after the barrier into its group had been advanced, so a flag sent
+            // there now would be addressed to a nonexistent group and fail forever (BS error ->
+            // TEvPoison -> boot loop). Dropping the mark is safe: live blobs in the old group
+            // are already covered by the Keep marks recorded there before the cut. A Delete for a
+            // blob deleted after the cut is lost, so the blob lingers until the group is released.
```
lex007in: Looks too verbose :) It's obvious that the sentinel should not be placed in affectedGroups. Maybe only worth mentioning that dropping is safe because those blobs have already been deleted.

ydb/core/tablet_flat/flat_executor.cpp
```diff
@@ -4029,6 +4029,12 @@ void TExecutor::UpdateUsedTabletMemory() {
 }
 
 void TExecutor::UpdateCounters(const TActorContext &ctx) {
+    if (GcLogic && Counters) {
+        if (const ui64 dropped = GcLogic->TakeSentinelDroppedMarks()) {
+            Counters->Cumulative()[TExecutorCounters::GC_SENTINEL_DROPPED_MARKS].Increment(dropped);
```
lex007in, line 4034: How will this counter be used? Do we really need it?

ydb/core/tablet_flat/flat_executor_counters.h
lberserq: 5d75b4bf59e

ydb/core/tablet_flat/flat_executor_gclogic.cpp
lberserq: 5d75b4bf59e

ydb/core/tablet_flat/flat_executor.cpp
lberserq: It is the only signal that the guard dropped marks: the blobs then linger in the old group until the whole group is released, which an operator watching a decommission needs to see.

lberserq: To distinguish "decommission is slow because data is big" from "the old group holds unreclaimable garbage from post-cut deletions".  Used by the operator who performs decommission.

About the importance. No, feature can run without it, but operator will have to read logs from the nodes if something unconventional happens.

lex007in (approved):

PR #51850, opened 2026-09-01
[ICB] Track overrides independently of default values

nisqatsi:
# [ICB] Track overrides independently of default values

### Changelog category

### Description for reviewers 

ICB infers override presence from numeric equality with a shared registry default. It cannot preserve an override when a default update makes the numbers equal. The existing API also does not directly support consumers sharing one override while using different base configurations and defaults. NBS dynconfig needs this for consumers holding different configuration snapshots.

This PR tracks override presence explicitly and exposes default updates, explicit override assignment, and optional override reads. A consumer can apply the shared override or fall back to its own base configuration.

### Complexity rationale

Adding one independent `Overridden` flag would not be sufficient because `Value`, `Default`, and override presence form one state with cross-field invariants. A reader must not observe fields from different concurrent mutations, especially during an equal-to-default override or a default update that preserves an active override. The implementation therefore serializes writers, validates optimistic multi-field reads with a sequence counter, and falls back to the writer lock when a coherent lock-free snapshot cannot be obtained.

The implementation also preserves the single atomic-load `Get()` hot path, the existing bounds rules, and the established behavior of baseline setters, wrappers, boards, and generated configuration updates. State-aware mutation results are required by the actor because setting or restoring an equal-to-default override changes only override presence: history and gauges must observe that transition even when the numeric value does not change. The additional synchronization, snapshot, mutation, actor, renderer, and test changes implement these compatibility and observability requirements.

### How controls work before this change

Static ICB controls (`TControlBoard`) and controls registered by name in DCB (`TDynamicControlBoard`) use the same `TControl`. Each control stores an effective `Value`, a registry `Default`, and bounds for submitted values. Override presence is inferred from `Value != Default`; it is not stored separately. An override can come from the HTML page, configuration processing, or another programmatic caller.

`TControl` exposes `Set(v)` and `Reset(v, lower, upper)`. The current `Set(v)` has the same baseline-reset semantics as `Reset(...)`: both set `Value = Default = v` without clamping. `Set()` preserves the existing bounds and is also used by numeric wrapper assignment `operator=(v)`; `Reset()` additionally replaces the bounds and is intended and used for initialization. The wrapper documents this requirement, but no runtime check enforces it. To update `Value` while preserving `Default`, callers use `SetFromHtmlRequest(v)`, either directly or through a board's `SetValue()`. Despite their similar names, `TControl::Set()` and a board's `SetValue()` therefore have different purposes.

| Main method or entry point | Current behavior |
| --- | --- |
| `TControl::Get()` / wrapper conversion to `i64` | Returns the effective `Value` with one atomic read. |
| `TControl::GetDefault()` / wrapper `GetDefault()` | Returns the registry `Default` with one atomic read. |
| `TControl::IsDefault()` / wrapper `IsDefault()` | Compares separately read `Value` and `Default`. Equal numbers mean no override. |
| `TControl::Set(v)` / wrapper numeric `operator=(v)` | Sets `Value = Default = v` without clamping. Removes any existing deviation from the default. `Set()` returns nothing; wrapper assignment returns `v`. |
| `TControl::Reset(d, lower, upper)` / wrapper `Reset(...)` | Sets `Value = Default = d` and replaces the bounds, without clamping `d`. Intended for initialization, as documented by the wrapper; no runtime check enforces this. |
| `TControl::SetFromHtmlRequest(v)` | Keeps `Default`. Accepts the current default directly, even outside the bounds; otherwise clamps `v` to the bounds and stores the result in `Value`. Returns the previous numeric value. A result equal to `Default` means no override. |
| `TControl::RestoreDefault()` | Copies `Default` into `Value`; returns nothing. |
| `TDynamicControlBoard::SetValue(name, v, TAtomic& previous)` | Finds the named control and calls `SetFromHtmlRequest(v)`. Writes its previous value to `previous`, then returns `IsDefault()`. For an unknown name, returns `true` and leaves `previous` unchanged; the return value is not a success flag. |
| `TControlBoard::SetValue(v, controlSlot)` | Applies `SetFromHtmlRequest(v)` to the static control in the supplied slot. Does nothing for an empty slot; returns nothing. |
| `TDynamicControlBoard::RestoreDefault(name)` / either board's `RestoreDefaults()` | Calls `RestoreDefault()` for the named control or all registered controls. An unknown dynamic name is ignored. Bulk restore iterates over controls without a transaction spanning the board. |
| Either board's `RegisterSharedControl()` / `RegisterLocalControl()` | Shared registration makes the supplied wrapper refer to an existing control when present. Local registration replaces the registered control. These operations determine object sharing, not a new configuration default. |
| Generated `UpdateControls(config)` | For each existing control processed by the generated update, a present field calls `SetFromHtmlRequest()`, and an absent field calls `RestoreDefault()`. Configuration values are not installed as new registry defaults. |

Individual values are read and written atomically, but compound updates (updates to both `Value` and `Default`) have no common writer lock or coherent state snapshot API. There is no operation that updates only the default while preserving an active override. In particular, `(Value, Default) = (20, 20)` cannot distinguish an active override of 20 from the absence of an override.

### How controls work after this change

`TControl` now stores `Overridden` independently of `Value` and `Default`. With no override, `Value` follows the registry default. With an override, `Value` remains the override value when the default changes, including when both numbers become equal. For example: default 10 → submit override 20 → update default to 20 → update default to 30 → the effective value stays 20 until the override is cleared, after which it becomes 30.

`TAdaptiveLock` serializes state mutations. Writers mark `Sequence` odd while updating the atomic fields and even when finished. `GetState()` tries once to read all state fields with an unchanged even sequence; on conflict, it reads the state under the same lock. `GetOverride()` derives its result from this coherent snapshot. Bounds are still intended to be set during initialization; `Reset()` remains unsafe to call concurrently with bounds readers.

#### Existing methods and what changes

| Existing method or entry point | Behavior with this change | Change from the current implementation |
| --- | --- | --- |
| `TControl::Get()` / wrapper conversion to `i64` | Not changed | None. |
| `TControl::GetDefault()` / wrapper `GetDefault()` | Not changed | The getter is unchanged; the new `UpdateDefault()` can change its result without removing an override. Separate calls to `Get()` and `GetDefault()` still do not form a coherent snapshot. |
| `TControl::IsDefault()` / wrapper `IsDefault()` | Returns `!Overridden`. | Reports override absence instead of numeric equality. It can return `false` when `Get() == GetDefault()`. |
| `TControl::Set(v)` / wrapper numeric `operator=(v)` | Sets `Value = Default = v` and `Overridden = false`, without clamping. Return values are unchanged. | Preserves the existing baseline-replacement behavior and explicitly clears override presence. The compound mutation is now serialized. This remains a baseline setter, not an override setter. |
| `TControl::Reset(d, lower, upper)` / wrapper `Reset(...)` | Sets `Value = Default = d`, clears override presence, and replaces the bounds. | Initializes the new state fields consistently; the documented initialization requirement and lack of default clamping remain. |
| `TControl::SetFromHtmlRequest(v)` | Preserves `Default`, applies the existing bounds rules, sets override presence, and returns the previous numeric value. | Always creates an override, including when the submitted or clamped value equals the default. Submitting the default no longer resets a control. |
| `TControl::RestoreDefault()` | Sets `Value` to the latest registry default, clears override presence, and returns `TControlMutation`. | Implements the reset directly under the state lock. Its return type changes from `void`; existing callers can ignore the result. The transition records resets that change only override presence. |
| `TDynamicControlBoard::SetValue(name, v, TAtomic& previous)` | Sets an override and writes the previous value when found. Returns `false` for a found control; returns `true` and leaves the output unchanged for an unknown name. | The flag still means override absence, not lookup success. A successful assignment now always creates an override, including assignment of the default. |
| `TControlBoard::SetValue(v, controlSlot)` | Sets an override through `SetFromHtmlRequest(v)` for a populated slot; ignores an empty slot. | The board code is unchanged, but assigning the default now creates an override instead of resetting the control. |
| `TDynamicControlBoard::RestoreDefault(name)` / either board's `RestoreDefaults()` | Not changed | Each control uses the new serialized reset. Lookup behavior and the absence of a board-wide transaction are unchanged. |
| Either board's `RegisterSharedControl()` / `RegisterLocalControl()` | Not changed | None; shared registration does not call `UpdateDefault()` or overwrite an existing control's state. |
| Generated `UpdateControls(config)` | For each existing control processed by the update, a present field creates an override, even when equal to the default; an absent field resets it. | Generated code and Console wiring are unchanged. Explicit presence now survives numeric equality with the default; configuration values do not become new registry defaults. |

#### New methods

`TControlState` contains `Value`, `Default`, and `Overridden`. `TControlMutation` contains the coherent `Before` and `After` states of one serialized mutation.

| New method | Behavior |
| --- | --- |
| `TControl::UpdateDefault(d)` / wrapper `UpdateDefault(d)` | Changes the registry default without clamping. If no override exists, also sets `Value = d`; otherwise preserves both the override value and its presence, including when equal to `d`. Returns nothing. |
| `TControl::GetState()` / wrapper `GetState()` | Returns a coherent `TControlState`. It tries one validated atomic read of the complete state, then takes the writer lock if the sequence changes or a writer is active. |
| `TControl::GetOverride()` / wrapper `GetOverride()` | Returns the active override from a coherent `GetState()` snapshot, including an equal-to-default value, or `nullopt` for a caller-specific configuration fallback. |
| `TControl::SetFromHtmlRequestWithState(v)` | Preserves the default and always sets override presence under the writer lock. Accepts the current default even outside bounds; otherwise clamps the submitted value. Returns `TControlMutation`. |
| `TControl::SetOverride(v)` / wrapper `SetOverride(v)` | Always clamps `v` to the bounds, preserves `Default`, and sets override presence even when the result equals the default. Returns `TControlMutation`. |
| `TDynamicControlBoard::SetValue(name, v, TControlMutation& mutation)` | Looks up the control and applies `SetFromHtmlRequestWithState(v)`. Returns `true` and writes the mutation when found; returns `false` and leaves the output unchanged otherwise. Unlike the existing overload, its boolean reports lookup success. |
| `TDynamicControlBoard::RestoreDefault(name, TControlMutation& mutation)` | Looks up the control and calls its `RestoreDefault()`. Returns `true` and writes the mutation when found; returns `false` and leaves the output unchanged otherwise. |

#### Impact on existing internal consumers

The updated controls preserve the existing configuration application, numeric read/write, registration, and caching logic of the internal consumers listed below. Their production code is unchanged; this PR does not migrate these callers to `UpdateDefault()` or `GetOverride()`.

| Consumer | Existing logic preserved by this PR |
| --- | --- |
| `TImmediateControlsConfigurator` and generated `TControlBoard` | Console still calls `ApplyConfig()` → `UpdateControls()`. A present field applies its value with the existing bounds rules; an absent processed field restores the registry default. Configuration values are not converted into new registry defaults. |
| PDisk | Scheduling and other numeric reads still use wrapper conversion → `TControl::Get()`. Programmatic DCB assignments retain their numeric results and bounds. The inspected sector-rate setters ignore the boolean returned by `SetValue()`, so its changed override-related result does not alter their decisions. |
| `TSchemeShard::ApplyConsoleConfigs()` | The feature-flag path still compares the incoming value with the effective control value and uses numeric wrapper assignment when different. Assignment continues to replace both value and default and clear an override; its priority is not changed to the override-preserving policy of `UpdateDefault()`. |
| `TControlWrapper` and `TMemorizableControlWrapper` | Numeric conversion, initialization, shared control identity, and the existing cache refresh policy remain unchanged. The new explicit APIs are available without changing these existing access paths. |

### Effect on existing users

Numeric reads, baseline assignment, control registration, and existing bounds remain compatible. Override classification intentionally changes: an explicitly submitted default now creates an override, `IsDefault()` reports override absence, and the existing DCB setter's boolean reflects that state. An active override remains effective when `UpdateDefault()` changes the registry default.

On the monitoring page, `Change` always creates an override. Operators and automation that previously submitted the displayed default to clear one control must use its `Restore Default` button instead; bulk `Restore Defaults` remains available. Existing change requests keep the `ControlName=value` format. The new per-control restore request uses `restoreDefault=<ControlName>`.

Console and generated control updates continue to use the existing paths. A present configuration field now creates an override even when its value equals the registry default; an absent processed field restores the default. No persistent format, Console schema, or restart behavior changes.

Components adopting the new API can use `UpdateDefault()` to publish registry defaults and `GetOverride()` to apply a shared override or fall back to their own base configuration. Existing consumers retain their current access paths.

### ICB Actor

The existing actor resolves submitted control names in static-first order, then tries DCB. Submitting the default resets one control; `Restore Default` resets both boards. History records a change when the previous value differs from the submitted number and stores the previous/new numbers. Gauges use numeric `Inc/Dec` updates, so repeated changes to one control can inflate the count and external mutations can leave it incorrect.

With this PR, Change always creates an override, including a value equal to the default. The per-control `Restore Default` action clears it explicitly and sends `restoreDefault=<ControlName>`. Change and Restore use the same static-first lookup. Names are not checked for uniqueness across boards: a dynamic control with a colliding name cannot be addressed independently through these forms.

The actor uses `TControlMutation` to record numeric changes and override-presence transitions from one operation. The history's new `Action` column distinguishes `Set override`, `Restore default`, and bulk restore, including cases where the previous/new numbers are equal. Unknown control names do not change controls or add history records.

On each HTTP request, the actor assigns gauges the number of overrides reported by the renderer. Equal-to-default overrides are counted, repeated changes do not inflate the count, and bulk restore yields zero when no writer intervenes. There is no timer or bootstrap reconciliation. External mutations become visible in gauges on the next HTTP request.

### HTML Renderer

The existing renderer reads the current value and default separately and uses numeric equality to choose highlighting and the `Changed` value. Each row offers a Change form; the page also offers bulk restore.

With this PR, each row uses one coherent `GetState()` snapshot. Highlighting and `Changed` follow `state.Overridden`; an equal-to-default override is additionally marked with `(overridden)` on a separate line. A row with an active override also offers a `Restore Default` button next to `Change`.

While rendering, it counts rows with override presence set across both tables. `GetOverriddenCount()` exposes this count to the actor for gauges, using the same snapshots as the displayed rows and adding no separate board traversal. Each row is coherent; the whole page is not an atomic snapshot of all controls.

<details>
<summary>Existing mon page</summary>

image

image

image

</details>

<details>
<summary>New mon page</summary>

image

image

`Override` == `Default` is explicitly marked with an `(overridden)` text note.

image

image

</details>

### Performance

`Get()` keeps one atomic load. Without a concurrent writer, `GetState()` and `GetOverride()` use one optimistic complete-state sample: five atomic loads, with no lock, allocation, reference counting, or retry loop. If the sequence check detects a writer, `GetState()` discards the sample and reads under the same `TAdaptiveLock` used by mutations.

Writes are expected to be rare. Longer waits when reads overlap writes are therefore an accepted tradeoff for simple synchronization and inexpensive reads in the common case. This is a design assumption; performance has not been benchmarked.

### Scope

The change covers the shared control implementation, both boards, and the monitoring page. Override state remains process-local; no persistent or configuration schema changes are introduced.

Related NBS context: https://github.com/ydb-platform/nbs/issues/6927

<details>
<summary>File summaries</summary>

| File or group | Summary |
| --- | --- |
| `lib/base/immediate_control_board_control.{h,cpp}` | Explicit override state, mutation results, and atomic reads with lock fallback. |
| `lib/immediate_control_board_wrapper.h` | Expose snapshots, optional overrides, default updates, and explicit override assignment. |
| `lib/dynamic_control_board_impl.{h,cpp}` | Return coherent mutations from named assignment and restore operations. |
| `lib/immediate_control_board_html_renderer.{h,cpp}` | Render explicit presence, the per-control restore action, and a count of displayed overrides for gauges. |
| `immediate_control_board_actor.cpp` | Handle restore and history actions; refresh gauges from the rendered state on HTTP requests. |
| Control library and actor unit tests | Cover override lifecycle, shared access, concurrent updates, HTML actions, and counters. |

</details>

### Verification

The control library and actor unit-test targets pass all 16 tests. They cover equal-to-default overrides through setters and generated config, bounds, restore visibility, history actions, static-first addressing, external changes observed by HTTP gauges, shared wrappers, and concurrent updates.

### DIFF
--- a/ydb/core/control/lib/base/immediate_control_board_control.cpp
+++ b/ydb/core/control/lib/base/immediate_control_board_control.cpp
@@ -1,5 +1,6 @@
 #include "immediate_control_board_control.h"
 #include <util/stream/str.h>
+#include <util/system/guard.h>
 
 namespace NKikimr {
 
@@ -8,30 +9,71 @@ TControl::TControl(TAtomicBase defaultValue, TAtomicBase lowerBound, TAtomicBase
       , Default(defaultValue)
       , LowerBound(lowerBound)
       , UpperBound(upperBound)
+      , Overridden(0)
+      , Sequence(0)
 {}
 
 void TControl::Set(TAtomicBase newValue) {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
     AtomicSet(Value, newValue);
     AtomicSet(Default, newValue);
+    AtomicSet(Overridden, 0);
+    AtomicIncrement(Sequence);
 }
 
 void TControl::Reset(TAtomicBase defaultValue, TAtomicBase lowerBound, TAtomicBase upperBound) {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
     AtomicSet(Value, defaultValue);
-    Default = defaultValue;
+    AtomicSet(Default, defaultValue);
+    AtomicSet(Overridden, 0);
     LowerBound = lowerBound;
     UpperBound = upperBound;
+    AtomicIncrement(Sequence);
+}
+
+void TControl::UpdateDefault(TAtomicBase newDefault) {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
+    const bool overridden = AtomicGet(Overridden);
+    AtomicSet(Default, newDefault);
+    if (!overridden) {
+        AtomicSet(Value, newDefault);
+    }
+    AtomicIncrement(Sequence);
+}
+
+TControlMutation TControl::SetOverride(TAtomicBase newValue) {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
+    const TControlState before = GetStateUnsafe();
+    newValue = Max(newValue, LowerBound);
+    newValue = Min(newValue, UpperBound);
+    AtomicSet(Value, newValue);
+    AtomicSet(Overridden, 1);
+    const TControlState after = GetStateUnsafe();
+    AtomicIncrement(Sequence);
+    return {before, after};
+}
+
+TControlMutation TControl::SetFromHtmlRequestWithState(TAtomicBase newValue) {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
+    const TControlState before = GetStateUnsafe();
+    if (newValue != before.Default) {
+        newValue = Max(newValue, LowerBound);
+        newValue = Min(newValue, UpperBound);
+    }
+    AtomicSet(Value, newValue);
+    AtomicSet(Overridden, 1);
+    const TControlState after = GetStateUnsafe();
+    AtomicIncrement(Sequence);
+    return {before, after};
 }
 
 TAtomicBase TControl::SetFromHtmlRequest(TAtomicBase newValue) {
-    TAtomicBase prevValue = AtomicGet(Value);
-    if (newValue == AtomicGet(Default)) {
-        AtomicSet(Value, newValue);
-    } else {
-        newValue = Max(newValue, LowerBound);
-        newValue = Min(newValue, UpperBound);
-        AtomicSet(Value, newValue);
-    }
-    return prevValue;
+    return SetFromHtmlRequestWithState(newValue).Before.Value;
 }
 
 TAtomicBase TControl::Get() const {
@@ -42,18 +84,60 @@ TAtomicBase TControl::GetDefault() const {
     return AtomicGet(Default);
 }
 
-void TControl::RestoreDefault() {
-    AtomicSet(Value, Default);
+// Return a coherent control snapshot, using Sequence to validate one
+// optimistic sample before falling back to StateLock.
+TControlState TControl::GetState() const {
+    // Accept the sample only if no writer was active or completed during it.
+    const TAtomicBase sequence = AtomicGet(Sequence);
+    if (!(sequence & 1)) {
+        const TControlState state = GetStateUnsafe();
+        if (sequence == AtomicGet(Sequence)) {
+            return state;
+        }
+    }
+
+    // Serialize with a writer after an overlapping mutation.
+    TGuard<TAdaptiveLock> guard(StateLock);
+    return GetStateUnsafe();
+}
+
+// Return the active override from a coherent complete state snapshot.
+std::optional<TAtomicBase> TControl::GetOverride() const {
+    const TControlState state = GetState();
+    return state.Overridden
+        ? std::optional<TAtomicBase>(state.Value)
+        : std::nullopt;
+}
+
+TControlMutation TControl::RestoreDefault() {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
+    const TControlState before = GetStateUnsafe();
+    AtomicSet(Value, before.Default);
+    AtomicSet(Overridden, 0);
+    const TControlState after = GetStateUnsafe();
+    AtomicIncrement(Sequence);
+    return {before, after};
 }
 
 bool TControl::IsDefault() const {
-    return AtomicGet(Value) == AtomicGet(Default);
+    return !AtomicGet(Overridden);
 }
 
 TString TControl::RangeAsString() const {
     TStringStream str;
     str << "[" << LowerBound << ", " << UpperBound << "]";
     return str.Str();
+}
+
+// Read one unvalidated state sample; callers must hold StateLock or verify
+// that Sequence stayed even and unchanged.
+TControlState TControl::GetStateUnsafe() const {
+    return {
+        AtomicGet(Value),
+        AtomicGet(Default),
+        static_cast<bool>(AtomicGet(Overridden)),
+    };
 }
 
 }
--- a/ydb/core/control/lib/base/immediate_control_board_control.h
+++ b/ydb/core/control/lib/base/immediate_control_board_control.h
@@ -2,14 +2,58 @@
 
 #include <util/generic/ptr.h>
 #include <library/cpp/deprecated/atomic/atomic.h>
+#include <util/system/spinlock.h>
+
+#include <optional>
 
 namespace NKikimr {
 
+// Coherent value, default, and override-presence snapshot of an ICB control.
+struct TControlState {
+    // Effective value returned by the control at the snapshot point.
+    TAtomicBase Value;
+
+    // Registry default used when no explicit override is active.
+    TAtomicBase Default;
+
+    // Override state distinguishing an explicit value from the registry default.
+    bool Overridden;
+};
+
+// State transition produced by one serialized control mutation.
+struct TControlMutation {
+    // Control state immediately before the mutation.
+    TControlState Before;
+
+    // Control state immediately after the mutation.
+    TControlState After;
+};
+
+// Immediate control with a fast effective-value read and coherent state snapshots.
 class TControl : public TThrRefBase {
+    // Effective value returned by the numeric Get() API.
     TAtomic Value;
+
+    // Registry default used when no explicit override is active.
     TAtomic Default;
+
+    // Inclusive lower bound used for explicit overrides.
     TAtomicBase LowerBound;
+
+    // Inclusive upper bound used for explicit overrides.
     TAtomicBase UpperBound;
+
+    // Override presence stored as zero or one.
+    TAtomic Overridden;
+
+    // Serialization for mutations and complete state snapshots.
+    mutable TAdaptiveLock StateLock;
+
+    // State version: odd during a mutation and even after publication.
+    TAtomic Sequence;
+
+    // Read the atomic fields without validating cross-field coherence.
+    TControlState GetStateUnsafe() const;
 
 public:
     TControl(TAtomicBase defaultValue, TAtomicBase lowerBound, TAtomicBase upperBound);
@@ -17,13 +61,29 @@ public:
     void Set(TAtomicBase newValue);
     void Reset(TAtomicBase defaultValue, TAtomicBase lowerBound, TAtomicBase upperBound);
 
+    // Update the registry default without replacing an active override.
+    void UpdateDefault(TAtomicBase newDefault);
+
+    // Set a bounded override, even when equal to the default, and return its transition.
+    TControlMutation SetOverride(TAtomicBase newValue);
+
+    // Set an HTML override; accept the current default outside bounds and return the transition.
+    TControlMutation SetFromHtmlRequestWithState(TAtomicBase newValue);
+
     TAtomicBase SetFromHtmlRequest(TAtomicBase newValue);
 
     TAtomicBase Get() const;
 
     TAtomicBase GetDefault() const;
 
-    void RestoreDefault();
+    // Return one coherent snapshot of the complete control state.
+    TControlState GetState() const;
+
+    // Return the explicit override or an empty value when the default is active.
+    std::optional<TAtomicBase> GetOverride() const;
+
+    // Restore the current default, clear the override, and return the transition.
+    TControlMutation RestoreDefault();
 
     bool IsDefault() const;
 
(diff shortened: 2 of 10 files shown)


nisqatsi (commented):
-

ydb/core/control/lib/base/immediate_control_board_control.cpp
```diff
@@ -42,12 +84,44 @@ TAtomicBase TControl::GetDefault() const {
...
+    if (!(sequence & 1)) {
+        const TControlState state = GetStateUnsafe();
+        if (sequence == AtomicGet(Sequence)) {
+            return state;
+        }
+    }
+
+    // Serialize with a writer after an overlapping mutation.
+    TGuard<TAdaptiveLock> guard(StateLock);
+    return GetStateUnsafe();
```
serbel324: This fallback in unneeded and harmful
Read still won't be able to read the state until write is finished
If multiple reads access state during the write, they will block each other
Instead I suggest to retry on seqlock fail, this way all reads may complete simultaneously

ydb/core/control/lib/base/immediate_control_board_control.cpp
```diff
@@ -8,30 +9,71 @@ TControl::TControl(TAtomicBase defaultValue, TAtomicBase lowerBound, TAtomicBase
...
+      , Sequence(0)
 {}
 
 void TControl::Set(TAtomicBase newValue) {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
     AtomicSet(Value, newValue);
     AtomicSet(Default, newValue);
+    AtomicSet(Overridden, 0);
+    AtomicIncrement(Sequence);
```
serbel324, line 22: These are all seq_cst atomic operations, seems excessive
I suggest explicit memory ordering

ydb/core/control/lib/base/immediate_control_board_control.cpp
```diff
@@ -42,12 +84,44 @@ TAtomicBase TControl::GetDefault() const {
...
+    const TAtomicBase sequence = AtomicGet(Sequence);
+    if (!(sequence & 1)) {
+        const TControlState state = GetStateUnsafe();
+        if (sequence == AtomicGet(Sequence)) {
+            return state;
+        }
+    }
+
+    // Serialize with a writer after an overlapping mutation.
+    TGuard<TAdaptiveLock> guard(StateLock);
```
serbel324, line 100: I suggest to remove this fallback completely and instead retry reading with seqlock
Anyway guard is locked until write completes, so fallback won't make write faster
And if multiple reads enter GetState() when write is in-flight, they will all mutually block each other, while with seqlock they will complete simultaneously

ydb/core/control/lib/base/immediate_control_board_control.cpp
```diff
@@ -42,12 +84,44 @@ TAtomicBase TControl::GetDefault() const {
...
+
+TControlMutation TControl::RestoreDefault() {
+    TGuard<TAdaptiveLock> guard(StateLock);
+    AtomicIncrement(Sequence);
+    const TControlState before = GetStateUnsafe();
+    AtomicSet(Value, before.Default);
+    AtomicSet(Overridden, 0);
+    const TControlState after = GetStateUnsafe();
+    AtomicIncrement(Sequence);
+    return {before, after};
```
serbel324, line 120: again, memory orders

ydb/core/control/lib/base/immediate_control_board_control.cpp
```diff
@@ -56,4 +130,14 @@ TString TControl::RangeAsString() const {
     return str.Str();
 }
 
+// Read one unvalidated state sample; callers must hold StateLock or verify
+// that Sequence stayed even and unchanged.
+TControlState TControl::GetStateUnsafe() const {
+    return {
+        AtomicGet(Value),
+        AtomicGet(Default),
+        static_cast<bool>(AtomicGet(Overridden)),
```
serbel324, line 139: again, memory ordering

ydb/core/control/lib/base/immediate_control_board_control.h
```diff
@@ -2,28 +2,88 @@
...
     TAtomicBase LowerBound;
+
+    // Inclusive upper bound used for explicit overrides.
     TAtomicBase UpperBound;
 
+    // Override presence stored as zero or one.
+    TAtomic Overridden;
+
+    // Serialization for mutations and complete state snapshots.
+    mutable TAdaptiveLock StateLock;
```
serbel324, line 50: Spinlocks are notoriously bad in highly competitive scenarios
Locks in general are bad for actor-system, so I suggest to re-evaluate this approach

serbel324 (changes requested):

PR #51867, opened 2026-09-01
Test adjacent pool waits for busy owner

kruall:
### Changelog entry 

Added coverage for adjacent executor pools with no foreign thread slots.

### Description for reviewers 

The auto-configured Batch pool has no threads and `ForcedForeignSlotCount = 0`, so only its adjacent User owner may execute its activations. The test blocks that owner, verifies that an unrelated shared thread remains runnable, and confirms that the Batch event stays pending until the owner is released and then executes on the owner.


mregrock (approved):
LGTM

PR #51877, opened 2026-09-01
Test single CPU adjacent pool owners

kruall:
### Changelog entry 

Added coverage for all adjacent pools in the single-CPU actor-system configuration.

### Description for reviewers 

The single-CPU auto configuration assigns its only shared thread to IC and marks System, User, and Batch as adjacent pools. The test waits for the IC owner to become idle before activating each pool and verifies that every activation executes on that owner.


mregrock (dismissed):
LGTM

mregrock (approved):
LGTM

mregrock (approved):
LGTM

PR #51906, opened 2026-09-01
kqp rm: pool-aware GetMemoryAvailability and optional quota (WLM side of Memory Quota 2.0, stacked on #51610)

lberserq:
Stacked on #51610

### Changelog category

* Improvement

### Description for reviewers

Implements the query-side (WLM) half of the Memory Quota 2.0 interface from #51610: `GetMemoryAvailability()` reports real node and pool headroom, and `isOptional` quota requests are refused under memory pressure instead of being treated as mandatory.

**Architecture note.** Per the agreed model the chain is MC → WLM: WLM owns pool management for memory the same way it does for CPU, receives the `QueryExecution` budget from MemoryController (at M3, as a flexible consumer instead of the static 20 % via the RB queue), slices it across its pools, and hands compute actors a pool-blind `IMemoryQuotaManager`. This PR is the "сегодня RM–CA" step of that chain: the implementation lives in a WLM-owned module (`ydb/core/kqp/workload_service/memory_quota/`, placed under `ydb/core` because `ydb/services/workload_manager` cannot be a dependency of KQP node service) behind the `IKqpResourceManager` boundary — when M3 lands, the module's budget source switches to `TEvConsumerLimit` with no caller changes. ResourceBroker is unaffected: it stays the slot-admission mechanism inside the implementation and is not a party to the interface.

The branch is `main` plus a local merge of `dq_memory_quota_2`, so the PoC's files appear in the diff. Review the last two commits only:

- `dq: fix comp_nodes test/tool factories for 3-arg Wrap* signatures` — fixes two callers #51610 itself missed (`dq/comp_nodes/ut/utils/dq_factories.cpp`, `kqp/tools/combiner_perf/factories.cpp`): a real `TDqMemoryQuota` over an unlimited `TGuaranteeQuotaManager` is wired into the factory context; `TComputeMemoryLimits` is brace-initialized explicitly so the `GetDqExecutionSettings()` default initializer is never ODR-used (keeps Wilson out of the test link). Cherry-pickable into #51610.
- `kqp: WLM-owned query memory quota module (Memory Quota 2.0, query side)` — the implementation: `IKqpResourceManager::GetTxMemoryAvailability(tx)` returns `0` while the tx's spilling cookie (node or pool) is set, otherwise `min(node headroom, named-pool headroom)` read under the RM lock together with the cookie; never negative for now (no reclaim signal exists — open question to the accounting/planning tracks). Optional quota is refused without error in the yellow zone (checked even when the channel already holds quota) or when it exceeds headroom; the mandatory path is unchanged; refusals are counted in `RM/OptionalQuotaDenied` and `RM/OptionalQuotaDeniedBytes` (bytes as requested by the task).

Tests in `kqp/rm_service/ut/kqp_rm_ut.cpp`: `OptionalQuotaDeniedAtYellowZone` (exact idle availability → yellow zone entered via real allocations → availability 0 → optional denied with counter deltas, mandatory still granted), `OptionalQuotaDeniedWhenExceedsPool`, `FreeQuotaSymmetryAfterOptionalGrant` — no internal state is poked. Suite 15/15 OK; `dq/comp_nodes/ut` 237/237 OK; `node_service`, `ut/channels`, `executer_actor`, `combiner_perf` build clean.


lberserq:
Superseded by #52082, which implements the full Memory Quota 2.0 (numeric cookie-published availability, optional/operator quota, all call sites) — a strictly better shape than this stacked draft: lock-free signed availability via TMemoryResourceCookie instead of an RM method under the lock.

Possible salvage as follow-ups on top of #52082, if useful:
- pool-dimension behaviour tests from this branch (yellow zone entered via real allocations, pool-headroom denial, quota free symmetry) — kqp_rm_ut.cpp on branch wm-mq2-quota-manager;
- per-pool refusal sensors in the workload_manager pool subgroup, as an extension of #51898.

Issue #51955, opened 2026-09-02
Cut-history: TEvCutTabletHistory is silently lost when Local's Hive pipe is down; tablet never retries


lberserq:
**Component:** core/tablet + core/mind (`EnableCutHistory` feature). Files on main: `ydb/core/mind/local.cpp:791-794`, `ydb/core/tablet/tablet_sys.cpp:1560-1626`.

Applies to any tablet type with cut-history enabled (currently DataShard via the default `CutHistoryAllowList`).

**Scenario:**
1. Cut-history is enabled; a tablet completes hard GC for an obsolete channel-0 history range.
2. `TTablet::SendCutTabletHistory()` sends `TEvCutTabletHistory` to Local and sets `CutHistoryStatus = Cut`, `WantCutHistoryAfterGc = false` (tablet_sys.cpp:1620-1621).
3. At that moment Local's pipe to Hive is disconnected (Hive restart / balancer move). `TLocalNodeRegistrar::Handle(TEvCutTabletHistory)` drops the event: `if (Connected) NTabletPipe::SendData(...)` — else nothing (local.cpp:792).
4. No path re-sends: `SendBarriersForCutHistory` early-returns while `CutHistoryStatus != None`, later snapshots only set `WantCutHistoryAfterGc` which is gated by the same check, and Hive sends no ack (`tx__cut_tablet_history.cpp` has none).

**Impact:** the stale channel-history row survives in Hive until the tablet restarts, re-feeds its history cutter, snapshots, and passes another GC round — which for a quiet tablet may be never. Hive keeps referencing a storage group that the tablet has already proven empty. The loss is entirely silent — no log or counter at the drop site.

**Related:** `TEvShrinkStoragePoolDone` is sent untracked (`hive_impl.cpp:4616`, `SendToConsolePipe(ev, false)`) with no Hive-side retry; only Console-side shrink re-initiation recovers it.

<details><summary>Candidate fixes</summary>

1. Ack-based: Hive replies to `TEvCutTabletHistory` (per tablet/channel/fromGeneration); the tablet keeps a SentCut state and re-sends on timeout / on pipe reconnect, marking Cut only on ack.
2. Cheaper: Local buffers pending cut requests while `!Connected` and flushes them on pipe (re)connect; the tablet re-sends pending cuts on the next snapshot instead of latching Cut before confirmed delivery.
[...]

PR #51961, opened 2026-09-02
Local: retain cut-history requests and replay them to Hive on reconnect

lberserq:
Issue #51955

### Changelog category

### Description for reviewers

Fixes silent loss of `TEvTablet::TEvCutTabletHistory` on the Local→Hive leg: the registrar forwards it only `if (Connected)` and otherwise destroys it, the tablet latches `CutHistoryStatus = Cut` before delivery is known, and Hive sends no ack — so a cut produced while Hive is restarting or the pipe is down leaves the stale channel-history entry in Hive until the tablet's next restart + GC + snapshot cycle, possibly never for a quiet tablet. Any tablet type with cut-history enabled is affected.

- **Retain and replay in the registrar** (`mind/local.cpp`, no protocol changes): `TLocalNodeRegistrar` records each leader tablet's cut tuples `{channel, fromGeneration, group}` (deduplicated by `{channel, fromGeneration}` — Hive keys history the same way) and replays them after every successful (re)registration, ordered after `TEvSyncTablets` on the same pipe. This covers both the `!Connected` drop and the pipe-death-in-flight race, including a Hive restart that loses an enqueued event.
- **Bounded memory, leader-only cleanup**: retention is one incarnation's cut entries per tablet; erased on the leader's `TEvTabletDead` and on node purge — a co-located follower's death deliberately does not wipe the leader's tuples. Cost is one no-op Hive tx per retained tuple per reconnect. Retention is additionally hard-capped at `MaxRetainedCutHistoryPerTablet = 1024` tuples per tablet, since the tuple space is sender-controlled; on overflow the tuple is dropped with a warning, degrading to the pre-fix behavior for that entry.
- **Why replay is safe**: `TTxCutTabletHistory` is a natural no-op for an already-erased entry (the `History` find misses; no counters, no DB writes), and a duplicate arriving after the tablet's own restart only re-cuts an entry that is still legitimately cuttable.

Test in `mind/hive/hive_ut.cpp`: `TestCutHistoryRetainedAcrossHiveReconnect` — shrink a pool for a Dummy tablet (`CutHistoryAllowList("Dummy")`), hold the registrar disconnected by blocking `TEvLocal::TEvPing` across a Hive reboot, inject the cut requests once, reconnect, and assert `TEvShrinkStoragePoolDone` arrives without restarting the tablet. Fails without the `local.cpp` change; both existing `TestShrinkStoragePool*` tests stay green.

### DIFF
--- a/ydb/core/mind/local.cpp
+++ b/ydb/core/mind/local.cpp
@@ -368,8 +368,9 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
             if (record.GetPurge()) {
                 for (const auto &x : OnlineTablets) {
                     ctx.Send(x.second.Tablet, new TEvents::TEvPoisonPill());
-                    if (x.first.second == 0) // leader
+                    if (x.first.second == 0) { // leader
                         RetainedCutHistory.erase(x.first.first);
+                    }
                 }
                 OnlineTablets.clear();
             }
@@ -920,8 +921,9 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
                 }
                 InbootTablets.erase(inbootIt);
             }
-            if (onlineIt->first.second == 0) // leader
+            if (onlineIt->first.second == 0) { // leader
                 RetainedCutHistory.erase(onlineIt->first.first);
+            }
             OnlineTablets.erase(onlineIt);
             UpdateEstimate();
             return;
@@ -932,8 +934,9 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
         });
         if (inbootIt != InbootTablets.end() && inbootIt->second.Generation <= generation) {
             MarkDeadTablet(inbootIt->first, generation, TEvLocal::TEvTabletStatus::StatusBootFailed, msg->Reason, ctx);
-            if (inbootIt->first.second == 0) // leader
+            if (inbootIt->first.second == 0) { // leader
                 RetainedCutHistory.erase(inbootIt->first.first);
+            }
             InbootTablets.erase(inbootIt);
             return;
         }
--- a/ydb/core/mind/local.cpp
+++ b/ydb/core/mind/local.cpp
@@ -123,18 +123,17 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
 
     TIntrusivePtr<TDrainProgress> LastDrainRequest;
 
-    // Retained cut-history tuples awaiting forwarding to Hive after reconnect.
-    // Hive sends no ack; TTxCutTabletHistory is a no-op for an already-erased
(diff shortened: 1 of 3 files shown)


ydb/core/mind/local.cpp
```diff
@@ -789,7 +816,28 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
...
+        auto& entries = RetainedCutHistory[tabletId];
+        if (auto* existing = FindIfPtr(entries, [&](const TCutEntry& e) {
+                return e.Channel == channel && e.FromGeneration == fromGen;
+            })) {
+            existing->GroupId = groupId;
+        } else if (entries.size() < MaxRetainedCutHistoryPerTablet) {
+            entries.push_back({channel, fromGen, groupId});
+        } else {
+            // The tuple space is sender-controlled; a misbehaving cutter must not grow
+            // this node's memory without bound. Dropping degrades to the pre-fix
```
vporyadke: В комментариях стоит описывать текущее поведение кода, а не ссылаться на какое-то там pre-fix

ydb/core/mind/local.cpp
```diff
@@ -122,6 +123,19 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
 
     TIntrusivePtr<TDrainProgress> LastDrainRequest;
 
+    // Retained cut-history tuples awaiting forwarding to Hive after reconnect.
+    // Hive sends no ack; TTxCutTabletHistory is a no-op for an already-erased
+    // entry, so replay-on-reconnect is safe.  Memory is bounded by one
+    // incarnation's cut entries (TEvTabletDead cleans up); cost is one no-op
+    // Hive tx per retained tuple per reconnect.
+    struct TCutEntry {
```
vporyadke: Зачем такое перебрасывание туда-сюда между форматами? Почему просто не хранить там TPtr'ы от ивентов (естественно, мувая их)?

vporyadke (changes requested):

ydb/core/mind/local.cpp
lberserq: Fixed in 048dd95280b.

ydb/core/mind/local.cpp
lberserq: Fixed in 048dd95280b.

lberserq: b4e6bb189ab

ydb/core/mind/local.cpp
```diff
@@ -858,6 +920,8 @@ class TLocalNodeRegistrar : public TActorBootstrapped<TLocalNodeRegistrar> {
                 }
                 InbootTablets.erase(inbootIt);
             }
+            if (onlineIt->first.second == 0) // leader
```
vporyadke: Давай не плодить тут и ниже ифов без {}

lberserq: 0bb93efe3c9

vporyadke (approved):

PR #51972, opened 2026-09-02
YDB Bench: add local log workload

kruall:
### Changelog entry 

Added the YDB CLI log workload to local YDB benchmarks.

### Description for reviewers 

Stacked on #51971.

Adds typed log workload configuration for insert, upsert, and bulk-upsert operations, safe thread-search defaults in the web builder, workload-aware throughput units, and a runnable YAML example.


kruall:
Folded into #51974 as part of native stack consolidation; the change is preserved there.

PR #52012, opened 2026-09-02
ydb-bench: bound chart rendering

kruall:
### Changelog entry 

Kept local YDB comparison charts responsive for large result sets.

### Description for reviewers 

Caps plotted series and points and reports truncation in the UI so historical runs cannot create unbounded SVG work.


kruall:

PR #52013, opened 2026-09-02
ydb-bench: bound activity replay

kruall:
### Changelog entry 

Limited the amount of local YDB activity replayed by the web UI.

### Description for reviewers 

Uses a bounded activity window with sequence cursors while retaining enough recent events for live progress.


kruall:

PR #52014, opened 2026-09-02
ydb-bench: bound workload cleanup

kruall:
### Changelog entry 

Made local YDB workload cleanup respect a bounded timeout.

### Description for reviewers 

Bounds stop and cleanup commands so a wedged YDB CLI child cannot block a benchmark run indefinitely.


kruall:

PR #52015, opened 2026-09-02
ydb-bench: show selected attempt SLO

kruall:
### Changelog entry 

Showed the latency SLO evaluated for the selected load attempt.

### Description for reviewers 

The profile result now renders the SLO metric from the selected attempt instead of a stale candidate.


kruall:

PR #52016, opened 2026-09-02
ydb-bench: align summary aggregation

kruall:
### Changelog entry 

Aligned local YDB summary aggregation with workload metric semantics.

### Description for reviewers 

Aggregates rates, counters, CPU, and latency fields according to their schemas so profile summaries remain meaningful across repetitions.


kruall:

PR #52017, opened 2026-09-02
ydb-cli: wait between topic stats windows

kruall:
### Changelog entry 

Fixed Topic workload statistics intervals to wait between collection windows.

### Description for reviewers 

Moves the sleep between snapshots so each reported interval covers real elapsed workload time.


kruall:

PR #52018, opened 2026-09-02
ydb-bench: align topic measurement windows

kruall:
### Changelog entry 

Made local Topic workload measurements use explicit warmup and measurement windows.

### Description for reviewers 

Parses the bounded Topic statistics window and derives throughput, latency, and error metrics from that interval.


kruall:

PR #52019, opened 2026-09-02
ydb-bench: align TPCC admission metrics

kruall:
### Changelog entry 

Aligned TPCC throughput and latency with admitted transaction samples.

### Description for reviewers 

Uses the same admitted transaction population for throughput, latency, and load-search decisions.


kruall:

PR #52020, opened 2026-09-02
ydb-bench: preserve adaptive TPCC warmup

kruall:
### Changelog entry 

Preserved adaptive YDB CLI warmup for TPCC local benchmarks.

### Description for reviewers 

Carries an omitted warmup through YAML, Builder, execution, and timeout calculation so TPCC can retain CLI adaptive behavior.


kruall:

PR #52021, opened 2026-09-02
ydb-bench: document topic measurement windows

kruall:
### Changelog entry 

Documented local Topic benchmark warmup and measurement windows.

### Description for reviewers 

Explains the statistics snapshots, warmup interval, measured interval, and their impact on reported metrics.


kruall:

PR #52023, opened 2026-09-02
ydb-bench: bound automatic load search

kruall:
### Changelog entry 

Bounded automatic local YDB load searches to 64 attempts per geometry stage.

### Description for reviewers 

Adds deterministic validation and runtime guards so automatic searches always have a practical attempt budget.


kruall:

PR #52024, opened 2026-09-02
ydb-bench: preserve best local YDB geometry

kruall:
### Changelog entry 

Kept the best measured result when automatic local YDB geometry scaling stops.

### Description for reviewers 

Tracks the best valid candidate across geometry stages and publishes it even when the final expanded geometry is worse.


kruall:

PR #52025, opened 2026-09-02
ydb-bench: stabilize Topic window timestamps

kruall:
### Changelog entry 

Made Topic workload window timestamps deterministic and consistent.

### Description for reviewers 

Uses explicit interval boundaries for Topic statistics so reported windows and calculated rates refer to the same period.


kruall:

PR #52026, opened 2026-09-02
ydb-bench: bound Topic measurement output

kruall:
### Changelog entry 

Rejected Topic measurements that exceed the CLI statistics output limit.

### Description for reviewers 

Validates warmup plus duration against the bounded Topic statistics history before a run starts.


kruall:

PR #52027, opened 2026-09-02
ydb-bench: stop throughput search at attempt limit

kruall:
### Changelog entry 

Made throughput search stop gracefully at the attempt limit.

### Description for reviewers 

Allows valid high-resolution throughput searches, stops after 64 measured candidates, and returns the best observed result.


kruall:

PR #52028, opened 2026-09-02
ydb-bench: bound automatic load values

kruall:
### Changelog entry 

Bounded automatic local YDB load values to the JavaScript-safe integer range.

### Description for reviewers 

Rejects unsafe configured values and saturates geometric growth before floating-point overflow can corrupt search or timeout calculations.


kruall:

PR #52029, opened 2026-09-02
ydb-bench: clarify local YDB timeout scope

kruall:
### Changelog entry 

Clarified the scope of local YDB timeouts.

### Description for reviewers 

Documents that the value is a per-command and per-cleanup budget rather than a deadline for the complete profile.


kruall:

PR #52030, opened 2026-09-02
ydb-bench: show verification cluster provenance

kruall:
### Changelog entry 

Showed the cluster provenance of verified local YDB metrics.

### Description for reviewers 

Labels fresh versus retained search-cluster verification and surfaces the distinction as comparison context without hiding valid deltas.


kruall:

PR #52115, opened 2026-09-03
kqp: fix infinite SCHEME_CHANGED retry loop in InconsistentTx write actor

lberserq:
Issue YQ-5669

### Changelog category

* Bugfix

### Description for reviewers

Fixes an infinite retry loop in the KQP write actor for `InconsistentTx` (streaming/YQ) writes after a schema operation in the database.

- (-) **Unify `STATUS_SCHEME_CHANGED` handling**: the `InconsistentTx` branch called `RetryResolve()`, but the schema version is baked into every `TEvWrite` at `Open()` time and is not refreshed by resolve — the datashard rejected the next write again, forever, and no error ever reached the session actor. Both branches now surface `RuntimeError(SCHEME_ERROR, KIKIMR_SCHEME_MISMATCH)`, so the session actor invalidates the plan and the retried execution recompiles against the new schema.
- (-) The trigger is not limited to the sink table: any scheme operation in the database invalidates scheme caches broadly. Observed on a 3-node cluster — `DROP TABLE` of an unrelated, abandoned table stopped all 77 running streaming queries writing to a different table. In `.sys/streaming_queries` they stay `RUNNING` with `RetryCount` frozen and `Issues` empty, while no new rows reach the target table and the Logbroker consumer shows no reads. Recovery requires dropping and recreating every query.
- (=) Shard splits/merges are unaffected: they emit `STATUS_WRONG_SHARD_STATE` (not `SCHEME_CHANGED`) and keep the resolve/retry path.
- (=) Complements b63a7a27ce9 (YQ-5172), which capped only the resolve-error path (dropped table); the case where resolve succeeds with a stale version is the one fixed here.

Test in `kqp/ut/tx/kqp_sink_tx_ut.cpp`: `SchemeChangedDuringWrite` — drives a genuine `InconsistentTx=true` write (injects `IsStreamingQuery=true` into the request context), ALTERs the table around an intercepted `TEvWrite`, asserts bounded `ABORTED` + "Scheme changed" instead of a hang. Verified to fail (bounded) with the actor change reverted and pass with it applied.

### DIFF
--- a/ydb/core/kqp/runtime/kqp_write_actor.cpp
+++ b/ydb/core/kqp/runtime/kqp_write_actor.cpp
@@ -1100,14 +1100,9 @@ public:
                 {"shardID", ev->Get()->Record.GetOrigin()},
                 {"sink", this->SelfId()},
                 {"issues", getIssues().ToOneLineString()});
-            // For InconsistentTx the schema version is baked into every
-            // TEvWrite at Open() time and is NOT refreshed by a successful
-            // TEvResolveKeySet round-trip.  Retrying only causes the
-            // datashard to return SCHEME_CHANGED again on the very next write,
-            // producing an infinite loop.  Surface the error so the session
-            // actor invalidates the compiled query and re-plans with the new
-            // schema.  The same RuntimeError path is already taken for the
-            // non-InconsistentTx case; unify both branches here.
+            // Schema version is baked into TEvWrite at Open() time and is not
+            // refreshed by resolve, so retrying InconsistentTx writes loops forever.
+            // Surface the error so the session actor recompiles.
             if (!InconsistentTx) {
                 UpdateStats(ev->Get()->Record.GetTxStats());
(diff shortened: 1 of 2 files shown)


lberserq:
Empirical regression-guard verification: built `ydb/core/kqp/ut/tx` and ran `SchemeChangedDuringInconsistentWrite` both ways.

- **With the fix**: PASS — `Sink[0] fatal error: Scheme changed. Table: /Root/KV`, query returns `ABORTED`.
- **Reverted `kqp_write_actor.cpp` to base (test kept)**: FAIL/CRASHED (bounded) — old path loops `RetryResolve()` until `Write retry limit exceeded`, query never completes, `WaitFuture` assertion aborts the test.

The test fails without the fix and passes with it.

nikvas0 (dismissed):

nikvas0 (approved):

PR #52178, opened 2026-09-04
Fix some heap-based VDisk allocator issues

alexvru:
### Changelog entry 

Fix some heap-based VDisk allocator issues

### Description for reviewers 

This fixes some heap-based allocator-related issues possibly leading to bugs.


mregrock (approved):
LGTM

PR #52329, opened 2026-09-06
YQ-5533 supported streaming op continuation on SS side

GrigoriyPA:
### Changelog entry 

Supported streaming op continuation on SS side

### Description for reviewers 

Currently (in main before this PR), a streaming request has two phases when it is created or modified.:

1. Metadata is updated in scheme shard
2. Activity begins to stop the previous request and launch a new version.

And if the node executing the DDL request crashes between steps 1 and 2, the operation will not complete on its own without additional user actions and will hang in an incomplete state.

The current change was made to solve the problem of incomplete operations in the event of a node restart (operations on streaming query are now divided into three phases: 1. - registration of the operation in scheme shard 2. - updating the query 3. - completion of the operation).


kardymonds (dismissed):

ydb/core/kqp/gateway/behaviour/streaming_query/queries.cpp
```diff
@@ -2315,37 +2338,6 @@ class TSyncStreamingQueryTableActor final : public TActionActorBase<TSyncStreami
...
-
-        if (State.GetStatus() != NKikimrKqp::TStreamingQueryState::STATUS_UNSPECIFIED || State.HasSchemeInfo()) {
-            // Clear query status
-            State.SetStatus(NKikimrKqp::TStreamingQueryState::STATUS_UNSPECIFIED);
-            State.ClearSchemeInfo();
-            UpdateQueryState("clear status");
-            return;
-        }
-
         Finish(Ydb::StatusIds::SUCCESS);
```
martinevsky, line 2347: > 🤖 **AI-generated review comment.** This was produced by an AI code-review agent (Claude) and is posted as a suggestion to verify, not as an authoritative finding. Please double-check before acting on it.

**`RemoveQuery()` no longer resets `STATUS_DELETING`, so a row that reaches this status without actually being deleted appears to stay there permanently.**

In the base revision (`c7f2677`) `RemoveQuery()` ended with an explicit status reset (`queries.cpp:2340-2346`):

```cpp
if (State.GetStatus() != NKikimrKqp::TStreamingQueryState::STATUS_UNSPECIFIED || State.HasSchemeInfo()) {
    // Clear query status
    State.SetStatus(NKikimrKqp::TStreamingQueryState::STATUS_UNSPECIFIED);
    State.ClearSchemeInfo();
    UpdateQueryState("clear status");
    return;
}

Finish(Ydb::StatusIds::SUCCESS);
```

In this PR that block is gone and the method falls straight through to `Finish(Ydb::StatusIds::SUCCESS)`.

The path looks reachable:

* `TSyncStreamingQueryTableActor::SyncQuery` calls `RemoveQuery()` in the `case STATUS_DELETING:` branch (`queries.cpp:2411-2414`);
* but the row itself is only removed by `TUnlockStreamingQueryRequestActor`, whose `RemoveQuery` flag is set only when `FinalStatus == Ydb::StatusIds::SUCCESS && QuerySettings.InflightOperation == TStreamingQueryConfig::TOperations::Drop` (`queries.cpp:2694-2698`).

So for a non-DROP operation (e.g. `ALTER`) over a row left in `STATUS_DELETING`, the operation reports `SUCCESS`, the query is not started, and the status stays `DELETING` for every subsequent operation as well.

Was the status reset dropped intentionally (e.g. because the new SchemeShard-side tracker now covers this), or did it get lost in the rewrite? If it is intentional, it would help to have a short comment here explaining what now guarantees the row leaves `STATUS_DELETING`.

GrigoriyPA: На самом деле не проблема, эта часть кода для совместимости со старой версией, по задумке там sync просто запустится заного

PR #52359, opened 2026-09-07
YDB-3506 Account KQP RM memory pools only for real pool limits

lberserq:
Fixes the asymmetric memory pool accounting in the KQP Resource Manager reported in YDB-3506.

### The defect

The session actor sets `PoolId = DEFAULT_POOL_ID` for every query, and the planner substitutes `MemoryPoolPercent = 100` whenever a query has no `PoolConfig`. RM then charged a pool only when `MemoryPoolPercent > 0`, but released for any non-empty `PoolId`. A query carrying a pool id with percent `-1` therefore released pool memory it never charged. With a mix of percent 100 and percent -1 queries on one key, `Used` was driven to zero and the record erased early, so the pool re-admitted its full limit while memory was still held. The record itself was tautological: 100% of the node query budget.

### The change

`TTxState::HasMemoryPoolLimit()` is now the single predicate for pool record creation, the `Y_DEFER` rollback release and `FreeResourcesImpl`. It holds only for a percent strictly inside `(0, 100)`, so the fabricated `100` and `-1` produce no pool record instead of one that can never deny anything the node budget already allows. `PoolId` and `MemoryPoolPercent` are `const`, so charge and release cannot disagree for a tx.

Because those queries no longer get a `PoolMemoryCookie`, the node level threshold is the only spilling signal left, and it was never refreshed: `SetConfigValues` stored `SpillingPercent` without applying it to `TotalMemoryResource`, whose `OverLimit` is computed in the constructor and recomputed only by `SetNewLimit`, which early returns when the limit and percent are unchanged. A runtime `spilling_percent` change was ignored until the resource broker queue limit changed or the node restarted. `SetConfigValues` now applies it.

### Tests

`ydb/core/kqp/rm_service/kqp_rm_ut.cpp`, 12 -> 17. Four are regressions, red on the pre-fix tree at their intended assertion:

- `PoolLimitNotReleasedByUnchargedQuery` - a percent -1 query releases and the next pool query is wrongly admitted.
- `PoolLimitNotReleasedByUnchargedQueryOnRollback` - same through the `Y_DEFER` path, with the broker rejecting the uncharged query.
- `PoolLimitNotChargedByHundredPercentQuery` - a percent 100 query charges and rewrites the limit of a configured pool's record.
- `SpillingPercentAppliedWithoutPoolLimit` - drives a real console config notification and asserts the lowered threshold takes effect.

`PoolLimitAppliedJustBelowHundredPercent` (percent 99) is a characterization test: it passes either way and pins the lower end of the interval so `< 100` cannot be widened silently.

### Notes for reviewers

- Nothing in the tree configures a pool at exactly 100 (the only occurrence is a parser unit test), so this removes no enforcement anyone relies on. At 100% the pool limit equals the node query limit up to a 1 byte rounding difference, and the total resource is charged first, so such a record could never deny a request the node accepted.
- `NResourcePool::TPoolSettings::IsWorkloadServiceRequired()` still tests `TotalMemoryLimitPercentPerNode > 0`, so a pool explicitly configured at 100 keeps engaging the workload service while RM does no pool accounting for it. Aligning that predicate is a workload manager decision and is left for a separate change.
- The default pool creation loop that spams the log, noted on the same ticket, is out of scope.

### DIFF
--- a/ydb/core/kqp/rm_service/kqp_rm_service.h
+++ b/ydb/core/kqp/rm_service/kqp_rm_service.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <ydb/core/protos/table_service_config.pb.h>
+#include <ydb/core/resource_pools/resource_pool_settings.h>
 #include <ydb/core/kqp/common/simple/kqp_event_ids.h>
 #include <ydb/core/kqp/counters/kqp_counters.h>
 #include <yql/essentials/minikql/computation/mkql_computation_pattern_cache.h>
@@ -84,6 +85,11 @@ public:
         return std::make_pair(Database, PoolId);
     }
 
+    bool HasMemoryPoolLimit() const {
+        return !PoolId.empty() && PoolId != NResourcePool::DEFAULT_POOL_ID
+            && MemoryPoolPercent > 0 && MemoryPoolPercent < 100;
+    }
+
     bool IsReasonableToStartSpilling() {
         return (PoolMemoryCookie && PoolMemoryCookie->SpillingPercentReached.load())
             || (TotalMemoryCookie && TotalMemoryCookie->SpillingPercentReached.load());
@@ -106,7 +112,7 @@ public:
 
         if (!PoolId.empty()) {
             res << ", PoolId: " << PoolId
-                << ", MemoryPoolPercent: " << Sprintf("%.2f", MemoryPoolPercent > 0 ? MemoryPoolPercent : 100);
+                << ", MemoryPoolPercent: " << Sprintf("%.2f", MemoryPoolPercent);
         }
 
         if (CollectBacktrace) {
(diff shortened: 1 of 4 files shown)


ydb/core/kqp/rm_service/kqp_rm_service.h
```diff
@@ -84,6 +84,11 @@ class TTxState : public TAtomicRefCount<TTxState> {
         return std::make_pair(Database, PoolId);
     }
 
+    // -1 means no pool limit, and a 100% cap is redundant with the node budget
+    bool HasMemoryPoolLimit() const {
+        return !PoolId.empty() && MemoryPoolPercent > 0 && MemoryPoolPercent < 100;
```
buhtr: Была концепция что на DEFAULT_POOL_ID тоже можно вешать лимиты. Сейчас этого нельзя делать, в DDL явно запретили. То есть если идет запрос и у него DEFAULT_POOL_ID его не надо никак лимитировать по памяти. Тут это можно явно обработать.

Hor911 (approved):

PR #52366, opened 2026-09-07
[IAM] Propagate request tracing context to AccessService

NewBediver:
### Changelog entry 

Propagate `X-Request-Id` and `X-User-IP` from incoming requests through `TEvAuthorizeTicket` and TicketParser to AccessService.

## Changes

- Added mandatory `TEvAuthorizeTicket::TTraceContext` containing immutable `PeerName` and `RequestId`.
- Removed the default trace-context constructor and updated all `TEvAuthorizeTicket` call sites.
- Propagated request IDs from gRPC, HTTP proxy, monitoring, YMQ, Kafka, and deprecated PersQueue request contexts.
- Used `GetTraceId()` as the request ID source for `IRequestCtxBaseMtSafe` implementations.
- Log missing or empty gRPC trace IDs and continue through `SetRequestId`.
- Propagated trace IDs through token refresh requests.
- Added `x-ydb-trace-id`/`x-request-id` extraction for deprecated PersQueue, preferring `x-ydb-trace-id`.
- Generate a request ID when creating `TKafkaSaslAuthActor` and reuse it unchanged for its TicketParser requests.
- Generate a new GUID for every Kafka `StartTokenRecheck` call.
- Store trace context in TicketParser token records and forward it as AccessService metadata:
  - `x-request-id`
  - `x-user-ip`
- Log `TEvAuthorizeTicket` requests with an empty request ID.
- Display the request ID on TicketParser monitoring pages.
- Extended AccessService mocks to capture tracing metadata under a dedicated mutex using `with_lock`.
- Updated tests and test request contexts to always pass explicit trace IDs.

This intentionally introduces a breaking constructor change for `TEvAuthorizeTicket`: every event must now provide both `PeerName` and `RequestId` through `TTraceContext`

### Description for reviewers 

https://nda.ya.ru/t/7I2-_pdM7pzaKk
https://nda.ya.ru/t/aue42TCS7pzaTr


ydb/core/grpc_services/base/base.h
```diff
@@ -906,7 +908,7 @@ class TGRpcRequestBiStreamWrapper
         , TraceId(GetPeerMetaValues(NYdb::YDB_TRACE_ID_HEADER))
         , AuxSettings(std::move(auxSettings))
     {
-        if (!TraceId) {
+        if (!TraceId || TraceId->empty()) {
             TraceId = UlidGen.Next().ToString();
         }
```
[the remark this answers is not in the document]
NewBediver: It's logged in ticket_parser

ydb/core/security/ticket_parser_impl.h
```diff
@@ -2385,7 +2401,8 @@ class TTicketParserImpl : public TActorBootstrapped<TDerived> {
         html << "<tr><td>Refresh Time</td><td>" << record.RefreshTime << "</td></tr>";
         html << "<tr><td>Expire Time</td><td>" << record.ExpireTime << "</td></tr>";
         html << "<tr><td>Access Time</td><td>" << record.AccessTime << "</td></tr>";
-        html << "<tr><td>Peer Name</td><td>" << record.PeerName << "</td></tr>";
+        html << "<tr><td>Peer Name</td><td>" << record.TraceContext.PeerName << "</td></tr>";
```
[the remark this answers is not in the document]
NewBediver: It has been done in this way, because RequestId can be generated by user

n00bcracker (approved):

nshestakov (approved):

ydb/core/security/ticket_parser_impl.h
```diff
@@ -1140,6 +1184,7 @@ class TTicketParserImpl : public TActorBootstrapped<TDerived> {
         auto it = userTokens.find(key);
         if (it != userTokens.end()) {
             auto& record = it->second;
+            record.TraceContext = ev->Get()->TraceContext;
```
nshestakov, line 1187: On a cache hit this unconditionally overwrites the stored `TraceContext`, including with an empty `RequestId`.

`TSecureRequestActor` (MessageBus, and any subclass that never calls `SetRequestId`) still sends `{PeerName, RequestId}` with an empty id. If that request hits a token previously populated by gRPC, a later AccessService refresh will drop `x-request-id` (and can drop `x-user-ip` if the incoming peer is empty too).

Please either skip the assignment when the incoming `RequestId` is empty, or generate a fallback id in `TSecureRequestActor::Bootstrap` before sending `TEvAuthorizeTicket` — same idea as monitoring/logout.

nshestakov (commented):
Review of the request-id / peer-name propagation into AccessService.

The production `TEvAuthorizeTicket` call sites look complete, headers are centralized in the gRPC client, and cache-hit updates of `TraceContext` are covered by tests. A few gaps around empty request ids are worth fixing: they can clobber a good cached context, and the promised gRPC empty-trace-id log never fires because a ULID is generated first.

ydb/core/grpc_services/base/base.h
```diff
@@ -906,7 +908,7 @@ class TGRpcRequestBiStreamWrapper
         , TraceId(GetPeerMetaValues(NYdb::YDB_TRACE_ID_HEADER))
         , AuxSettings(std::move(auxSettings))
     {
-        if (!TraceId) {
+        if (!TraceId || TraceId->empty()) {
```
nshestakov, line 911: The PR description says missing/empty gRPC trace ids should be logged, but this path (and the unary wrapper around L1293) generates a ULID first. TicketParser therefore never sees an empty id on the gRPC path, so the DEBUG log there does not cover this case.

Please log here before assigning the fallback — monitoring and logout already do that when they synthesize an id. Same silent fallback exists in deprecated PQ `GetRequestId()`.

ydb/core/security/ticket_parser_impl.h
NewBediver: MessageBus is legacy and not supported. Anyway, this is the same behaviour as before, I just have changed the place where this assignment performed (previous place https://github.com/ydb-platform/ydb/pull/52366/changes#diff-ebf9ac589fb496e6a36e95a81e66286238f4ab7e304e6fe5ba8deaaf7cc37056L1167)

ydb/core/grpc_services/base/base.h
NewBediver: The same comment: https://github.com/ydb-platform/ydb/pull/52366#discussion_r3950876788. We have a common log in TicketParser for such situations, so we can leave everything as it is.

PR #52367, opened 2026-09-07
ydb-bench: centralize KV and Stock workload contracts

kruall:
### Changelog entry 

Unify workload options and result metrics for local YDB benchmarks.

### Description for reviewers 

Move the existing KV and Stock definitions, CLI plans, validation and metric schemas into a shared catalog. Keep their defaults and supported operations unchanged; no new workload types are added. Reject malformed counters and avoid summarizing empty repetitions as valid results.


mregrock (approved):
LGTM

PR #52368, opened 2026-09-07
ydb-bench: improve local YDB search and execution reliability

kruall:
### Changelog entry 

Make local YDB load search and workload execution more reliable.

### Description for reviewers 

Bound candidate search, preserve the best result across geometry stages, and support independent verification with role-specific CPU measurements. Empty measurements and invalid numeric limits must not become successful candidates. Wait for all dynamic-node endpoints to appear in client discovery before running workloads, and bound CLI phases and cleanup so startup races or stuck commands do not derail a run.


mregrock (approved):
LGTM

PR #52369, opened 2026-09-07
ydb-bench: separate profile results and discovery with comparisons

kruall:
### Changelog entry 

Make benchmark results easier to find and compare in the web interface.

### Description for reviewers 

Separate profile results from the discovery process, with workload-aware KPIs, verification and partial-result states. Add local YDB baseline comparisons and search curves, and expose execution activity and commands. Extend the Builder for the existing KV and Stock settings; keep stored result schemas readable without introducing new workload types.


mregrock (approved):
LGTM

PR #52414, opened 2026-09-07
NBS-7459: own PBuffer cleanup in DirectBlockGroup

BarkovBG:
- The PBuffer cleanup barrier is now computed and sent by each DirectBlockGroup from the minimum inflight key of its own vchunks, instead of one tablet-wide minimum gathered by `TFastPathService`.
- The cleanup tick is skipped while older-generation records are inflight.


ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/direct_block_group_impl.h
```diff
@@ -291,6 +290,8 @@ class TDirectBlockGroup
 
     TDBGConnections Connections;
     TVector<TVChunkWeakPtr> VChunks;
+
+    TMap<THostIndex, ui64> LastSentBarrierByPBufferHost;
```
drbasic: почему не вектор?

drbasic (approved):

PR #52454, opened 2026-09-08
ydb-bench: expose shared threads and united pool settings

kruall:
### Changelog entry 

Allow local YDB benchmarks to configure shared threads and the united pool through YAML and the web Builder.

### Description for reviewers 

Add independent use-shared-threads and use-united-pool switches under actor-system, both disabled by default. Apply them to static and dynamic nodes, including scaled nodes and verification clusters, while keeping automatic pool sizing enabled. Preserve the settings in profile manifests and display them in run comparisons.


mregrock (approved):
LGTM

PR #52481, opened 2026-09-08
streamlookup input transform/kikimr lookup: use CA stats level

yumkam:
### Changelog entry 

...

### Description for reviewers 

Known problem: in kqp minimum possible CA StatsLevel is BASIC, there are no way to turn collecting off
Probably, we should shift levels by one? NONE, BASIC -> no collection, FULL -> same as BASIC, PROFILE->PROFILE ?
YQ-5685


yumkam:
will be integrated with linked pr

PR #52491, opened 2026-09-08
ydb-bench: collect and visualize per-attempt YDB metrics

kruall:
### Changelog entry 

Collect YDB executor-pool metrics and display them on dedicated benchmark attempt pages.

### Description for reviewers 

Persist bounded, best-effort monitoring samples per profile, separated by attempt and repetition. Charts compare all pools of a selected node, show cumulative counters as per-second deltas by default, and share a time cursor with two-decimal tooltips. Verification has its own page; raw samples remain downloadable.

Also retry transient discovery UNAVAILABLE responses within the existing startup timeout: a database can report RUNNING before its dynamic endpoints are discoverable, previously causing the benchmark to fail before measurement.


mregrock (approved):
LGTM

PR #52521, opened 2026-09-08
ydb-bench: support external ydbd binaries and version catalog

kruall:
### Changelog entry 

Allow local YDB benchmarks to use an external YDBD executable selected by path or from a version catalog in Builder.

### Description for reviewers 

Add per-profile `ydbd-binary` and a `--binaries-dir` catalog with executable files under `ydbd/<version>`. Bundled YDBD remains the default, and YDB CLI remains bundled. CLI and web runs snapshot each distinct executable path once per run and record the source path, SHA-256, and size in profile manifests. Catalog scans are bounded; hidden files are excluded.


mregrock (approved):
LGTM

PR #52539, opened 2026-09-08
ydb-bench: configure ring queues and stop completed profile refresh loops

kruall:
### Changelog entry 

Allow local YDB benchmarks to configure ring queues and prevent completed profile pages from repeatedly reloading while other profiles are running.

### Description for reviewers 

Expose `actor-system.use-ring-queue` in YAML and Builder, retain YDBD's default of `true`, and preserve the value in saved parameters and comparisons. Apply it to static, dynamic, scaled, and verification nodes.

A terminal profile previously scheduled a run-page refresh whenever its parent run remained active; remounting the same terminal profile repeated the cycle. Schedule that refresh only after observing the selected profile transition from active to terminal.


mregrock (approved):
LGTM

PR #52543, opened 2026-09-08
Fix united pool max slot accounting

kruall:
### Changelog entry 

Fixed max_threads enforcement for executor pools that use united shared threads.

### Description for reviewers 

For an auto-configured 8-CPU dynamic node, the User pool owns 4 shared slots and has max_threads=8. AllThreadsAreShared converted the basic-pool maximum to a single owner slot, while current and potential limits were measured in shared slots, so Harmonizer allowed no foreign slots and reported Current and PossibleMax above Max.

The change preserves total slot limits for all-shared pools and uses the actual owned shared-slot count instead of a boolean throughout foreign-slot, quota, and demand calculations. The resulting contract is owned_slots=4, foreign_slots=4, max_slots=8, with Current and PossibleMax bounded by Max.

### DIFF
--- a/ydb/core/driver_lib/run/auto_config_initializer_ut.cpp
+++ b/ydb/core/driver_lib/run/auto_config_initializer_ut.cpp
@@ -511,6 +511,45 @@ Y_UNIT_TEST(SharedAndUnitedAutoConfigMatrix) {
     }
 }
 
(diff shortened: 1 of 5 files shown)


mregrock (approved):
LGTM

PR #52603, opened 2026-09-09
[US] Fix node registration

NewBediver:
### Changelog entry 

Dynamic nodes started without a static configuration implicitly sent `root@builtin`, the protobuf default, in registration requests. This could prevent certificate authentication because the token took precedence over the client certificate.

This change stops sending the implicit registration token when neither a static configuration nor an authentication file is supplied, allowing the node to authenticate using its client certificate.
Registration tokens are now sent only when a static configuration is loaded or `--auth-file` / `--auth-token-file` is supplied. Static configuration includes YAML and positional protobuf configuration files (`freeArgs`). Supplying positional protobuf files preserves the previous token behavior even if those files are empty or contain no `AuthConfig`. In these explicitly configured startup modes, the existing behavior - including the `root@builtin` default when `NodeRegistrationToken` is unset - is retained.

Added functional coverage for:
- Registration and GetNodeConfig over gRPC/gRPCS using a certificate, a token, both, or neither, including authentication and authorization failures.
- Startup with YAML, positional `.pb` configuration files, separate authentication files, and no static configuration.
- Canonicalized RPC results, startup logs, and YDB server logs showing successful responses and failure reasons.

### Description for reviewers 

http://st/KIKIMR-26605

### DIFF
--- a/ydb/core/config/init/init_impl.h
+++ b/ydb/core/config/init/init_impl.h
@@ -1494,6 +1494,7 @@ public:
             cf.NodeResolveHost = cf.NodeHost;
         }
 
+        const auto& authConfig = AppConfig.GetAuthConfig();
         const TNodeRegistrationSettings settings {
             domainName,
             cf.NodeHost,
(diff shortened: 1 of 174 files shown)


CyberROFL (commented):
lgtm, но пусть @Enjection посмотрит.

ijon (approved):

Enjection (approved):

Issue #12794, opened 2024-12-19
Make partitioning settings name correct SI-correct


weres-sa:
Settings name like AUTO_PARTITIONING_PARTITION_SIZE_MB uses MiB in real.

I think it would be better to make them MB or change settings names

kkdras:
This has come up again in ydb-platform/ydb-embedded-ui#4049 and https://nda.ya.ru/t/K2U541Jz7idzmU.

The proposed UI fix is to divide `SizeToSplit` by `2^20` so that Embedded UI shows the same number as CLI and YC. That would fix the visible symptom, but the setting would still be named `_MB` while the backend interprets it as MiB.

Can we decide the unit semantics here before adding a local exception to Embedded UI?

If the setting means MB, the backend conversion should use `10^6` bytes. If the binary multiplier is intentional, the setting and its displayed unit should be MiB.

PR #52607, opened 2026-09-09
[YDBDOCS-2875] Correct AUTO_PARTITIONING_PARTITION_SIZE_MB default and units

AlejandroMokhovani:
### Changelog entry 

...

### Description for reviewers 

The documentation for the `AUTO_PARTITIONING_PARTITION_SIZE_MB` table setting stated the default value as 2000 MB (2 GB). This did not match the actual implementation: the default is `2ul << 30` bytes = 2048 MiB (2 GiB), and the value is stored in binary units (mebibytes).

This behavior may change as part of https://github.com/ydb-platform/ydb/issues/12794; nevertheless, we describe the current behavior.


weres-sa (approved):
LGTM

PR #52626, opened 2026-09-09
actors/async: fix async<T> overloads, add ActorRequest and started waiter

martinevsky:
### Changelog entry
Not for changelog: internal fixes and helpers in `ydb/library/actors/async`, no user-visible change.

### Description for reviewers
Two compile fixes:
- `WithTimeout`, `WithDeadline`, `TAsyncCancellationScope::Wrap` / `WrapShielded` and `InterceptCancellation`: the overloads that take an already created `async<T>` called `UnsafeMove()` on an lvalue, and it is `&&`-qualified, so they never compiled. Now `std::move(wrapped).UnsafeMove()`.
- `WithTaskGroup`: `std::declval<TTaskGroup<T>&>` was missing the call parentheses, so a callback with a concrete `TTaskGroup<T>&` parameter did not compile.

Two helpers in `wait_for_event.h` for the request/reply pattern from the coroutine guide:
- `AllocateWaitCookie()`: process-wide unique, never-zero cookie.
- `ActorRequest<TEvent>(recipient, request, flags)`: allocates a cookie, sends the request and returns a wait for the typed reply. Sending before awaiting is safe: a reply is a mailbox event, handled no earlier than the next turn, so the wait is always registered in time. It is an awaiter, not an `async<T>`; to bound it, wrap it in an `async<>` lambda passed to `WithTimeout` (shown in the doc comment). A late reply after the timeout reaches the state function.

Regression tests for each item in the existing `ut/*_ut.cpp` files; the ones for the compile fixes do not build without them. `ya make -tA ydb/library/actors/async/ut`: 118 OK.


ydb/library/actors/async/wait_for_event.h
```diff
@@ -84,18 +90,101 @@ namespace NActors {
...
+     * The result is an awaiter, not an async<T>, so to bound the wait wrap it in a coroutine:
+     *
+     *     auto reply = co_await WithTimeout(timeout, [&]() -> async<TEvReply::TPtr> {
+     *         co_return co_await ActorRequest<TEvReply>(recipient, new TEvRequest);
+     *     });
+     *
+     * A reply that arrives after the timeout is no longer intercepted and reaches the state function.
+     */
+    template<class TEvent>
+    inline auto ActorRequest(const TActorId& recipient, IEventBase* request, ui32 flags = 0) {
```
kruall, line 132: Предлагаю пока убрать стартер, он более спорная вещь. Прям плюсов от него не вижу, зато код становится менее интуитивно понятным

А ActorRequest переписать как-то так

```c++
    template<class TEvent>
    inline auto ActorRequest(const TActorId& recipient, IEventBase* request, ui32 flags = 0) {
        TActorId selfId = TActivationContext::AsActorContext().SelfID;
        ui64 cookie = AllocateWaitCookie();
        TActivationContext::Send(new IEventHandle(recipient, selfId, request, flags, cookie));
        return ActorWaitForEvent<TEvent>(cookie);
    }
```

kruall (changes requested):

kruall (approved):

PR #52648, opened 2026-09-09
Make StageId in TaskGraph unique across transactions

### LINKED ISSUE #52600: Make StageId in TaskGraph Unique
Currently KqpExecuter may execute several computations graphs. It places it to separate so-called "transactions" with TxId starting from 0 and enumerates Stages independently:

https://github.com/ydb-platform/ydb/blob/main/ydb/core/kqp/executer_actor/kqp_tasks_graph.cpp#L625

It causes problems with exporting execution statistics since it  has room for StageId only and stages with different TxId clash with one another

https://a.yandex-team.ru/arcadia/contrib/ydb/library/yql/dq/actors/protos/dq_stats.proto?rev=r21098954#L404

Current workaround is to export stats for TxId = 0 only 

Solution suggested is to make StageId unique for all TxIds and export all stage stats. It breaks no contracts and matched already unique TaskId numbering

Hor911:
KqpExecuter may execute several computation graphs, placed into separate "transactions" whose stages were enumerated independently, both starting from 0. NYql.NDqProto.TDqStageStats has room only for a flat StageId, so stages of different transactions clashed and ExportExecStats() worked around it by exporting stages of TxId = 0 only.

Number StageIds continuously across all transactions of the graph, so that a stage is identified by its StageId alone - matching the already unique TaskId numbering - and export the stats of every stage.

TKqpTasksGraph::MakeStageId() converts a transaction-local stage index (as stored in the physical plan protobuf) into the graph-wide stage id; TStageInfoMeta::StageIdBase / GetStageIdx() go the other way for the places that index back into the transaction body.

Fixes #52600

### DIFF
--- a/ydb/core/kqp/executer_actor/kqp_data_executer.cpp
+++ b/ydb/core/kqp/executer_actor/kqp_data_executer.cpp
@@ -659,7 +659,7 @@ private:
             const auto& tx = Request.Transactions[txIdx];
             for (ui32 stageIdx = 0; stageIdx < tx.Body->StagesSize(); ++stageIdx) {
                 const auto& stage = tx.Body->GetStages(stageIdx);
-                auto& stageInfo = TasksGraph.GetStageInfo(TStageId(txIdx, stageIdx));
+                auto& stageInfo = TasksGraph.GetStageInfo(TasksGraph.MakeStageId(txIdx, stageIdx));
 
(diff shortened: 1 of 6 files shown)


abyss7 (approved):
Нужно обратить внимание на сохранённые графы, чтобы они не попортились в момент выкатки при переходе на новый счётчик StageId

GrigoriyPA (approved):

PR #52665, opened 2026-09-09
Fix TEvBlock version handling for compatibility

alexvru:
### Changelog entry 

Fix TEvBlock version handling for compatibility

### Description for reviewers 

This patch fixes incorrect behaviour when old clients issue zero Version and it gets misprocessed.


mregrock (approved):
LGTM

PR #52670, opened 2026-09-09
ydb-bench: decouple actor-system vCPUs from CPU affinity

kruall:
### Changelog entry 

Allow local YDB benchmarks to configure static and dynamic node actor-system vCPUs independently of CPU affinity.

### Description for reviewers 

Whole-chiplet affinity can expand the requested CPU mask, causing YDB automatic pool sizing to use more CPUs than intended. Add per-role cpu-count settings to YAML and the builder, pass them to node configurations, and display them separately in reports and comparisons. Omitted settings preserve automatic CPU detection.


mregrock (approved):
LGTM

PR #52718, opened 2026-09-10
Preserve PersistentBuffer fast erase metadata across recovery

kruall:
### Changelog entry 

Prevent erased PersistentBuffer records from reappearing after restarts.

### Description for reviewers 

Fast erase persists a replacement header containing the erased LSNs and an increasing `Header.RecordLsn`. Recovery uses that version to select the newest header; freeing a sector does not erase its old contents.

When recovery found no live records for the erased generation (for example, after applying a cutoff barrier), `RestoreErases()` dropped the entire erase entry and left its sector unreserved. This also discarded `HeaderLsn`. Subsequent fast erases restarted the version at 1. On another restart, a still-valid old header with a larger version could win over the newer header, so the latest erased records were recovered as live. For example, forgetting version 100 allows a new erase at version 1 to lose to the old version 100.

Keep the erase metadata and mark its sector occupied even when there are no live records. The next erase therefore advances version 100 to 101 and releases the previous sector only after the replacement write completes. Old headers cannot outrank the new erase. The regression covers recovery without live records, sector reservation, version continuity, and replay with a stale header; the missing-generation case also retains metadata without erasing another generation's records.

This mechanism is consistent with #51089: the archived failure returned previously erased LSNs 564 and 565 after the second restart, following an empty-buffer first restart. The specific CI failure has not yet been reproduced, so its attribution to this mechanism remains unconfirmed.

### DIFF
--- a/ydb/core/blobstorage/ddisk/persistent_buffer_barriers_manager.cpp
+++ b/ydb/core/blobstorage/ddisk/persistent_buffer_barriers_manager.cpp
@@ -419,14 +419,17 @@ namespace NKikimr::NDDisk {
             auto itErase = std::upper_bound(erase.Lsns.begin(), erase.Lsns.end(), barrier.Lsn);
             erase.Lsns = std::vector<ui64>(itErase, erase.Lsns.end());
 
-            auto pbIt = persistentBuffers.find({tid, erase.Generation, dbg});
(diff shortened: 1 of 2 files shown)


Evgenik2 (approved):
LGTM

PR #52722, opened 2026-09-10
Test legacy shared pool routing and foreign slots

kruall:
### Changelog entry 

Added regression coverage for legacy shared-pool routing and foreign-slot reuse.

### Description for reviewers 

The existing shared-pool tests implicitly exercised legacy mode and stopped before checking lease reuse. This change explicitly fixes the mode to non-united, verifies it through the public interface, and confirms that returning a foreign lease allows another worker to acquire the slot.


mregrock (approved):
LGTM

PR #52726, opened 2026-09-10
Expose committed topic offsets

CyberROFL:
[KIKIMR-27030](http://st/KIKIMR-27030)

### Changelog entry 

...

### Description for reviewers 

This is the 1/7 part of https://github.com/ydb-platform/ydb/pull/52465

### DIFF
--- a/ydb/core/tx/replication/service/topic_reader.cpp
+++ b/ydb/core/tx/replication/service/topic_reader.cpp
@@ -154,6 +154,7 @@ class TRemoteTopicReader: public TActor<TRemoteTopicReader> {
             {"ev", ev->Get()->ToString()});
 
         ReadSessionId = ev->Get()->Result.ReadSessionId;
+        Send(Worker, new TEvWorker::TEvReaderStarted(ev->Get()->Result.CommittedOffset));
     }
 
     void Handle(TEvWorker::TEvCommit::TPtr& ev) {
@@ -164,7 +165,26 @@ class TRemoteTopicReader: public TActor<TRemoteTopicReader> {
             return Leave(TEvWorker::TEvGone::UNAVAILABLE);
         }
 
-        CommittedOffset = ev->Get()->Offset;
+        if (CommitInFlight.Defined()) {
+            if (ev->Get()->Offset > *CommitInFlight
+                && (!PendingCommit || ev->Get()->Offset > *PendingCommit)) {
+                PendingCommit = ev->Get()->Offset;
+            }
+            YDB_LOG_DEBUG("Commit queued while another commit is in flight",
+                {"offset", ev->Get()->Offset},
+                {"inFlightOffset", *CommitInFlight},
+                {"pendingOffset", PendingCommit.GetOrElse(*CommitInFlight)});
+            return;
+        }
+
+        StartCommit(ev->Get()->Offset);
+    }
+
+    void StartCommit(ui64 offset) {
+        Y_ABORT_UNLESS(!CommitInFlight);
+
+        CommittedOffset = offset;
+        CommitInFlight = CommittedOffset;
         Send(YdbProxy, CreateCommitOffsetRequest().release());
     }
 
@@ -179,6 +199,14 @@ class TRemoteTopicReader: public TActor<TRemoteTopicReader> {
                 {"ev", ev->Get()->ToString()});
             if (CommittedOffset) {
                 Send(ReadSession, CreateCommitOffsetRequest().release());
+            }
+            const auto committedOffset = *CommitInFlight;
+            Send(Worker, new TEvWorker::TEvCommitResult(committedOffset));
+            CommitInFlight.Clear();
+            if (PendingCommit) {
+                const auto pendingOffset = *PendingCommit;
+                PendingCommit.Clear();
+                StartCommit(pendingOffset);
             }
         }
     }
@@ -309,6 +337,8 @@ private:
     TActorId ReadSession;
     TString ReadSessionId;
     ui64 CommittedOffset = 0;
+    TMaybe<ui64> CommitInFlight;
+    TMaybe<ui64> PendingCommit;
 
     bool CreatingReadSessionInProgress = false;
     bool StoppingInProgress = false;
--- a/ydb/core/tx/replication/service/topic_reader.cpp
+++ b/ydb/core/tx/replication/service/topic_reader.cpp
@@ -165,19 +165,19 @@ class TRemoteTopicReader: public TActor<TRemoteTopicReader> {
             return Leave(TEvWorker::TEvGone::UNAVAILABLE);
         }
 
+        const auto offset = ev->Get()->Offset;
         if (CommitInFlight.Defined()) {
-            if (ev->Get()->Offset > *CommitInFlight
-                && (!PendingCommit || ev->Get()->Offset > *PendingCommit)) {
-                PendingCommit = ev->Get()->Offset;
+            if (offset > *CommitInFlight && (!PendingCommit || offset > *PendingCommit)) {
+                PendingCommit = offset;
             }
             YDB_LOG_DEBUG("Commit queued while another commit is in flight",
-                {"offset", ev->Get()->Offset},
+                {"offset", offset},
                 {"inFlightOffset", *CommitInFlight},
                 {"pendingOffset", PendingCommit.GetOrElse(*CommitInFlight)});
             return;
         }
 
-        StartCommit(ev->Get()->Offset);
+        StartCommit(offset);
     }
 
     void StartCommit(ui64 offset) {
--- a/ydb/core/tx/replication/service/topic_reader_ut.cpp
+++ b/ydb/core/tx/replication/service/topic_reader_ut.cpp
@@ -1,6 +1,7 @@
 #include "topic_reader.h"
 #include "worker.h"
 
+#include <ydb/core/testlib/basics/appdata.h>
 #include <ydb/core/tx/replication/ut_helpers/test_env.h>
 #include <ydb/core/tx/replication/ut_helpers/write_topic.h>
 #include <ydb/core/tx/replication/ydb_proxy/ydb_proxy.h>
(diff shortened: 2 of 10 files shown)


ydb/core/tx/replication/ydb_proxy/ydb_proxy.cpp
```diff
@@ -442,7 +460,8 @@ class TYdbProxy: public TBaseProxyActor<TYdbProxy> {
         auto args = std::move(ev->Get()->GetArgs());
         const auto& settings = std::get<TEvYdbProxy::TTopicReaderSettings>(args);
         auto session = std::invoke(&TTopicClient::CreateReadSession, client, settings.GetBase());
-        auto reader = RegisterWithSameMailbox(new TTopicReader(session, settings.AutoCommit_));
+        auto reader = RegisterWithSameMailbox(new TTopicReader(ev->Sender, session, settings.AutoCommit_));
+        Send(reader, new TEvents::TEvBootstrap());
```
nshestakov, line 465: `TEvBootstrap` is sent to `TTopicReader` (same mailbox as the proxy) while `TEvCreateTopicReaderResponse` goes to `TRemoteTopicReader` (another mailbox). `TEvStartTopicReadingSession` / `TEvReaderStarted` can therefore reach the worker **before** the reader handshake.

In this PR the worker ignores those events, so it does not break yet. In #52465 both `TEvReaderStarted` and handshake send `TEvCommit` when a schema barrier is pending, so this race can double-commit. Also, if a follow-up commits on `TEvReaderStarted` before `CreateTopicReaderResponse`, `ReadSession` is still empty and `Send(ReadSession, CreateCommitOffsetRequest())` goes to `TActorId{}`.

Please either delay `TEvReaderStarted` until `ReadSession` is set / handshake is sent, or make the later worker path explicitly idempotent to this order.

ydb/core/tx/replication/service/topic_reader.cpp
```diff
@@ -154,6 +154,7 @@ class TRemoteTopicReader: public TActor<TRemoteTopicReader> {
             {"ev", ev->Get()->ToString()});
 
         ReadSessionId = ev->Get()->Result.ReadSessionId;
+        Send(Worker, new TEvWorker::TEvReaderStarted(ev->Get()->Result.CommittedOffset));
```
nshestakov, line 157: Same ordering issue from the reader side: this can be sent while `CreateTopicReaderResponse` is still in flight, so `Worker` may see `TEvReaderStarted` before `TEvHandshake`. `Worker` is already set in `Handle(TEvHandshake)`, so the send itself is fine — the consumer of this event is not.

ydb/core/tx/replication/ydb_proxy/ydb_proxy_ut.cpp
```diff
@@ -698,6 +698,22 @@ Y_UNIT_TEST_SUITE(YdbProxy) {
...
+            .BeginAddConsumer()
+                .ConsumerName("consumer")
+            .EndAddConsumer();
+        auto create = env.Send<TEvYdbProxy::TEvCreateTopicResponse>(
+            new TEvYdbProxy::TEvCreateTopicRequest("/Root/topic", settings));
+        UNIT_ASSERT(create->Get()->Result.IsSuccess());
+
+        const auto reader = CreateTopicReader(env, "/Root/topic");
+        auto started = env.GetRuntime().GrabEdgeEventRethrow<TEvYdbProxy::TEvStartTopicReadingSession>(env.GetSender());
+        UNIT_ASSERT_VALUES_EQUAL(started->Sender, reader);
```
nshestakov, line 714: This test never checks `CommittedOffset`, which is the point of the PR. On a fresh topic it should be `0`. Worth also covering a non-zero consumer offset (write + commit + new reader), and asserting the field from the local proxy path.

ydb/core/tx/replication/service/topic_reader_ut.cpp
```diff
@@ -138,6 +138,46 @@ Y_UNIT_TEST_SUITE(RemoteTopicReader) {
...
+
+        const auto reader = env.GetRuntime().Register(CreateRemoteTopicReader(ydbProxy, settings));
+        env.SendAsync(reader, new TEvWorker::TEvHandshake());
+        env.GetRuntime().GrabEdgeEvent<TEvYdbProxy::TEvCreateTopicReaderRequest>(ydbProxy);
+        env.GetRuntime().Send(reader, ydbProxy,
+            new TEvYdbProxy::TEvCreateTopicReaderResponse(readSession));
+        env.GetRuntime().GrabEdgeEvent<TEvWorker::TEvHandshake>(env.GetSender());
+        env.GetRuntime().Send(reader, readSession,
+            new TEvYdbProxy::TEvStartTopicReadingSession(TString("read-session")));
+        env.GetRuntime().GrabEdgeEvent<TEvWorker::TEvReaderStarted>(env.GetSender());
```
nshestakov: `TStartTopicReadingSessionResult(TString)` leaves `CommittedOffset = 0`, and the grabbed `TEvReaderStarted` is not asserted. Please set a non-zero offset on the start event and `UNIT_ASSERT_VALUES_EQUAL` it on `TEvReaderStarted`.

Also missing: coalescing three commits (`10`, `20`, `30` while `10` is in flight should issue `30` next) and dropping a smaller pending offset.

nshestakov (commented):
Review of part 1/7 of #52465.

Direction looks right: committed offset is exposed, commits are serialized, and the topic session starts without a read request. No merge blockers for this split, but follow-ups need to treat `TEvReaderStarted` arriving **before** reader handshake as a real ordering, and the new offset field is not actually asserted in tests.

Main points are inline.

ydb/core/tx/replication/ydb_proxy/ydb_proxy.cpp
```diff
@@ -189,7 +181,11 @@ class TTopicReader: public TBaseProxyActor<TTopicReader> {
             DeferredCommit.Commit();
         }
 
-        WaitEvent(ev->Sender, ev->Cookie);
+        Y_ABORT_UNLESS(!ReadRequest);
```
nshestakov, line 185: `TRemoteTopicReader` still has `ReadQueue` and sends a new `TEvReadTopicRequest` on every `TEvPoll`. Worker usually keeps a single in-flight poll, so this may not fire today, but a second read before `DataReceived` is now a process abort instead of a second `WaitEvent`.

Consider queueing or dropping the extra request. Same for `Y_ABORT_UNLESS(ReadRequest)` on `TDataReceivedEvent` if data ever arrives while `ContinueWaiting()` is running because `!SessionStarted`.

ydb/core/tx/replication/service/topic_reader.cpp
```diff
@@ -180,6 +200,14 @@ class TRemoteTopicReader: public TActor<TRemoteTopicReader> {
             if (CommittedOffset) {
                 Send(ReadSession, CreateCommitOffsetRequest().release());
             }
+            const auto committedOffset = *CommitInFlight;
```
nshestakov, line 203: `*CommitInFlight` is unchecked. A stray `TEvCommitOffsetResponse` with an empty `TMaybe` will abort in a less obvious way — `Y_ABORT_UNLESS(CommitInFlight.Defined())` would be clearer.

Related: commits with `offset <= in-flight/pending` are dropped with no `TEvCommitResult`. Coalescing `10` then `20` is correct; a duplicate `Commit(10)` while in flight yields a single result. Follow-ups that wait for one result per `TEvCommit` (e.g. re-issue on stop/start of the same session) can stall. Either document “one result per RPC” or ack coalesced commits.

ydb/core/tx/replication/service/worker.h
```diff
@@ -65,6 +67,20 @@ struct TEvWorker {
...
 
+    struct TEvCommitResult: public TEventLocal<TEvCommitResult, EvCommitResult> {
+        size_t Offset;
+
+        explicit TEvCommitResult(size_t offset);
+        TString ToString() const override;
+    };
+
+    struct TEvReaderStarted: public TEventLocal<TEvReaderStarted, EvReaderStarted> {
+        ui64 CommittedOffset;
```
nshestakov, line 78: `TEvCommit` / `TEvCommitResult` use `size_t Offset`, this event uses `ui64 CommittedOffset`. Topic offsets should be `ui64` everywhere.

Note: worker does not handle `TEvCommitResult` / `TEvReaderStarted` yet (ok for 1/7; they are currently dropped). Inserting these ids before `EvTerminateWriter` also shifts later local event numbers — fine for `TEventLocal`, but #52465 uses a different enum order.

ydb/core/tx/replication/ydb_proxy/ydb_proxy.h
```diff
@@ -238,6 +239,7 @@ struct TEvYdbProxy {
         void Out(IOutputStream& out) const;
 
         TString ReadSessionId;
+        ui64 CommittedOffset = 0;
```
nshestakov: `TStartTopicReadingSessionResult(const TString&)` still leaves this at `0`. Local proxy then mutates the field after construction. A `(readSessionId, committedOffset)` constructor would make that path harder to miss.

`Out()` was not updated either, so logs still omit the field this PR is adding.

nshestakov (dismissed):

ydb/core/tx/replication/ydb_proxy/ydb_proxy.cpp
CyberROFL: > In ydb/core/tx/replication/ydb_proxy/ydb_proxy.cpp:463, the proxy registers TTopicReader with the same mailbox. Consequently, bootstrap cannot execute until the current handler finishes; that handler queues TEvCreateTopicReaderResponse before returning.
> 
> The subsequent start notification is therefore queued after the create response. Mailbox FIFO ordering ensures that TRemoteTopicReader first sets ReadSession and sends the worker handshake in ydb/core/tx/replication/service/topic_reader.cpp:41, then forwards TEvReaderStarted. The worker receives those notifications in that order.
> 
> Different destination mailboxes do not invalidate this dependency. Neither the proposed empty-ReadSession send nor the follow-up double-commit follows from the alleged race. Delaying the start notification is unnecessary for this path.

ydb/core/tx/replication/service/topic_reader.cpp
CyberROFL: То же самое, что выше.

ydb/core/tx/replication/service/topic_reader.cpp
CyberROFL: > At ydb/core/tx/replication/service/topic_reader.cpp:203, TMaybe dereference already checks whether it contains a value and throws if empty. An explicit assertion would clarify the invariant, but I found no legitimate unsolicited response path.

ydb/core/tx/replication/service/worker.h
CyberROFL: > The inconsistency exists in ydb/core/tx/replication/service/worker.h:63. Using ui64 consistently is sensible; there is no truncation on 64-bit builds. The current worker indeed ignores both new events. These local event IDs do not establish a wire-compatibility problem.

ydb/core/tx/replication/ydb_proxy/ydb_proxy.cpp
CyberROFL: > Two overlapping requests really do abort at ydb/core/tx/replication/ydb_proxy/ydb_proxy.cpp:184. The remote reader’s ReadQueue tracks requests; it does not serialize their dispatch. This tightens the proxy’s contract and merits documenting or testing.
> 
> However, the normal worker/writer cycle issues its next poll after processing the previous batch; the transfer writer also tracks outstanding polls. I found no normal caller producing the required overlap.
> 
> The separate speculation about data arriving without ReadRequest is unsupported: startup waiting stops after the start event when no read is pending. Later waiting requires a read request, and consuming data clears that request and returns without restarting waiting. Queueing extra requests could support concurrent callers; silently dropping them could leave callers waiting indefinitely.

nshestakov (approved):

Enjection (approved):

PR #52742, opened 2026-09-10
Parse CDC schema change records

CyberROFL:
[KIKIMR-27030](http://st/KIKIMR-27030)

### Changelog entry 

...

### Description for reviewers 

This is the 2/7 part of https://github.com/ydb-platform/ydb/pull/52465

### DIFF
--- a/ydb/core/tx/replication/service/json_change_record.cpp
+++ b/ydb/core/tx/replication/service/json_change_record.cpp
@@ -1,9 +1,15 @@
 #include "json_change_record.h"
 
 #include <ydb/core/io_formats/cell_maker/cell_maker.h>
+#include <ydb/core/protos/base.pb.h>
+#include <ydb/core/protos/replication.pb.h>
 #include <ydb/core/protos/tx_datashard.pb.h>
 
 #include <library/cpp/json/json_writer.h>
+
+#include <util/generic/algorithm.h>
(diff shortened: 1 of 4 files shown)


ydb/core/tx/replication/service/json_change_record.cpp
```diff
@@ -46,14 +53,27 @@ ui64 TChangeRecord::GetTxId() const {
     switch (GetKind()) {
         case EKind::CdcDataChange: return NService::GetTxId(JsonBody, "ts");
         case EKind::CdcHeartbeat: return NService::GetTxId(JsonBody, "resolved");
+        case EKind::CdcSchemaChange: return NService::GetTxId(JsonBody, "ts");
```
Enjection, line 56: I know we don't really care about downgrade, but just to be sure: is it handled at least minimally on the receiver side, or is it up to the operator to keep the systems in sync? My main concern is that we're going to offer async replication as a different product than it is right now. Currently it's used mostly for migrations, but we're targeting long-term replication now, so having Y_ABORT on the receiver side will become actively harmful.

Enjection: It's broader question, not just about this specific case-switch

Enjection (approved):
ack-ed, but would be great to have answer to question above

PR #52757, opened 2026-09-10
[detailed-metrics] pack detailed counters on sending node

jsjant:
### Description for reviewers 

Pack detailed counters info into new TDetailedTableCounters proto message


ydb/core/tablet/detailed_metrics/node_database_metrics_aggregator.cpp
```diff
@@ -350,6 +370,32 @@ class TNodeDatabaseMetricsAggregatorImpl : public TNodeDatabaseMetricsAggregator
         }
     }
 
+    void Pack(NProtoBuf::RepeatedPtrField<NKikimrSysView::TDetailedTableCounters>& out) override {
+        TGuard<TMutex> guard(DetailedMetricsLock());
+
+        for (auto& [_, entry] : Tables) {
```
BlazingPhoenix, line 376: Preserve unsent deltas when removing buckets
 Packing only live buckets loses increments accumulated before a bucket is removed. For a single-tablet table, pack CPU=100, report another +25, then switch TABLE → PARTITION before the next Pack(). The old bucket is destroyed, while the new bucket receives only subsequent tablet deltas, so those +25 are never transmitted. Please retain pending deltas from retired buckets until they have been packed, and add a regression test for this sequence.

ydb/core/tablet/detailed_metrics/node_database_metrics_aggregator.cpp
```diff
@@ -155,6 +156,22 @@ class TCountersBucket {
...
+        NKikimrSysView::TDbTabletCounters current;
+        current.SetType(TabletType);
+        if (ExecutorCounters.IsInitialized) {
+            ExecutorCounters.ToProto(*current.MutableExecutorCounters(), *current.MutableMaxExecutorCounters());
+        }
+        if (AppCounters.IsInitialized) {
+            AppCounters.ToProto(*current.MutableAppCounters(), *current.MutableMaxAppCounters());
+        }
+
+        NSysView::CalculateCountersDiff(&out, current, &Previous);
```
BlazingPhoenix, line 171: Distinguish a recreated bucket from a continuing delta stream.
 Previous is destroyed with the bucket. If a partition is packed, forgotten, and recreated with the same tablet/follower IDs before the next Pack(), the empty baseline causes another absolute histogram to be sent without a reset indication. Applying it as a delta turns one observation into two, and subsequent unchanged reports leave the overcount intact. The receiver never sees the intervening absence. Please
preserve the histogram baseline across recreation or explicitly signal the reset, and cover this sequence with a round-trip test.

BlazingPhoenix (approved):

PR #52782, opened 2026-09-10
Improve ydb_bench profile comparisons and Runs UI

kruall:
### Changelog entry 

Add saved profile comparisons with a baseline, configuration differences, and searchable run selection, and simplify the Runs interface.

### Description for reviewers 

Local YDB comparisons now have a list and dedicated result/configuration pages instead of an implicit persistent selection. Saved definitions use revision checks to reject stale edits; incompatible metric sources suppress baseline deltas. Move comparison selection out of Runs and ZIP import into a modal.


mregrock (approved):
LGTM

PR #52825, opened 2026-09-11
YQ-5689 Add WasmCompileController

akulaad:
Today every dynamic node that sees a new WASM UDF in the metadata snapshot starts its own AOT compile. On a platform with N nodes that means N identical LLVM runs racing to produce one artifact, with only a local, blind retry counter in front of them. This PR puts a scheduler in the middle: a new per-database tablet, `WasmCompileController`, owns the set of missing artifacts and hands out an exclusive right to compile each one to exactly one node.

Everything is behind feature flags that default to `false`, so with no configuration change the behaviour is bit-for-bit the one that exists today.

### Changelog entry 

Added the per-database `WasmCompileController` tablet, which orchestrates AOT compilation of WASM UDFs so that a module uploaded once is compiled by a single node per CPU platform instead of by every node at once. Disabled by default.

#### What the controller does

It keeps the gap set — the difference between the `modules` metadata snapshot and the keys already present in `artifacts/{cpu_spec}` — and assigns each gap to one registered worker. The `modules` side comes from the same metadata subscription the dynamic nodes use; the artifact side is read by a child `TReconcileActor`, so no heavy query ever runs inside a tablet transaction. Dedup is on `(name, kind, uid, cpu_spec)` with a default budget of one active compile per platform, and libraries are scheduled ahead of the modules that depend on them.

Codegen deliberately stays on the assigned dynamic node. WAVM object code is only valid for the `cpu_spec` of the host that produced it, and the tablet leader has no reason to sit on matching hardware; running LLVM on a tablet actor path would also be unacceptable for a control plane.

#### Points worth a closer look

- **Assignment durability.** An assignment is reserved in memory, persisted in `Execute`, and only sent to the worker in `Complete`. A leader that dies mid-transaction therefore cannot leave two owners of the same gap.
- **Assignments are self-contained.** They carry the manifest and the library uids rather than letting the worker re-read them, so the node compiles the upload the controller decided on and not whatever its own snapshot happens to show.
- **Stale versus failed.** A re-upload that invalidates an in-flight compile is reported as stale and does not count towards the poison pill; only real failures do. This required aligning `TWasmLibraryCompileActor` with `TWasmCompileActor`, which already replied `Deferred` instead of an error on a uid mismatch — without it a library re-upload would have been poisoned as a genuine failure.
- **Generations.** A node rejects an assignment from a controller generation lower than the one it registered with, and adopts a higher one, since after a tablet move the new leader's assignment can arrive before its registration reply.
- **Protocol placement.** The cross-node protocol is a separate `PROTO_LIBRARY` wrapped in `TEventPB` in a new event space, `ES_WASM_COMPILE_CTL`. The existing process-local `TEventLocal` compile responses in `udf_store/events.h` are untouched.
- **`tablet.proto`.** `Reserved47` is renamed to `WasmCompileController` and `Reserved51` is added, following the instruction in the comment at the top of the enum.
- **Schemeshard wiring** follows `StatisticsAggregator` rather than `BackupController`: the latter is missing the `FilterPrivateTablets` block in `TSubDomainInfo::Initialize()`, which looks like a bug and was not copied.

#### Two flags, and the order matters

- `FeatureFlags.EnableWasmCompileController` — schemeshard creates the tablet for new databases and backfills existing ones through `schemeshard_svp_migration`.
- `UdfStoreConfig.EnableWasmCompileController` — dynamic nodes route compiles through the controller instead of compiling straight from the snapshot.

Enable the schemeshard flag first and wait until `Params.GetWasmCompileController()` is non-zero on every served database, then enable the node one. The reverse order stops WASM UDF compilation entirely: the node no longer compiles from the snapshot and has no controller to report the gap to. That case is not silent — it logs `no compile controller to report a gap to` per dropped gap — but it is worth knowing before rollout. Rollback is turning the node flag back off; artifact layout does not change, so already written artifacts stay valid.

#### Observability

Five tablet counters in a new `counters_wasm_compile_controller.proto`, plus four gauges labelled by `cpu_spec` under `udf_store/wasm_compile`: workers registered, queue length, active assignments and modules missing an artifact. `ModulesMissingArtifact` intentionally counts gaps the scheduler refused to queue, so a platform stuck on a poisoned module does not read as healthy with an empty queue.

#### Tests

A new `ydb/services/udf_store/ut_orchestration` target with 22 tests; the existing `udf_store/ut` has no actor runtime and could not host them. Ten cover the tablet: one compile per platform, two platforms in parallel, reassignment after a worker disconnects, library before module, poison pill, per-node cap, an already existing artifact, stale failures not poisoning, and the counters. Twelve cover the node side against a real controller tablet with a fake scheme cache: discovery and registration, a database that has not been migrated yet, both states of the flag, rejection of a foreign `cpu_spec` and of an older generation, and the reporting of done, failed and stale results.

On the schemeshard side, `ut_extsubdomain` covers creating the tablet through alter, add-only validation, propagation of the id to the tenant schemeshard, and the migration of a database that was created before the feature existed.

Two things are not covered and need a live cluster: the compile actor itself and its uid predicates go through YQL and cannot run in a bare runtime, so the "N−1 nodes only load" and "a re-upload never publishes a foreign uid" halves are exercised only at the orchestration level.

#### Out of scope

Process isolation for LLVM (a wasm bomb can still take the node down — this PR removes the thundering herd, it does not put an OS boundary around codegen), cross-architecture artifacts, narrowing `modules.compile_status` to per-platform truth, and a compile duration histogram. A sys view over `artifacts/*` for per-platform status is the intended follow-up.


ydb/core/cms/console/configs_dispatcher.cpp
```diff
@@ -72,6 +72,7 @@ const THashSet<ui32> DYNAMIC_KINDS({
     (ui32)NKikimrConsole::TConfigItem::TliConfigItem,
     (ui32)NKikimrConsole::TConfigItem::PrivateDatabaseConfigItem,
     (ui32)NKikimrConsole::TConfigItem::ColumnShardConfigItem,
+    (ui32)NKikimrConsole::TConfigItem::UdfStoreConfigItem,
```
CyberROFL, line 75: Это точно релевантное данному PR изменение?

CyberROFL (commented):
По части @ydb-platform/core и @ydb-platform/schemeshard OK. Если других изменений там не будет, не призывайте меня до получения согласований от всех остальных команд.

Призывать через "re-request review".

ydb/core/cms/console/configs_dispatcher.cpp
akulaad: В pr добавляется UdfStoreConfigItem = 95 в console_config.proto и подписка TWasmCompileController на этот kind (SubscribeForConfigChanges), через которую таблетка получает лимиты и таймауты компиляции WASM из динамического конфига.

ConfigsDispatcher обслуживает только kinds из DYNAMIC_KINDS: в Handle(TEvSetConfigSubscriptionRequest) вызывается CheckKinds, и для kind вне этого списка там безусловный Y_ABORT("unexpected kind ..."). Плюс сам dispatcher подписывается у Console ровно на DYNAMIC_KINDS (CreateConfigsSubscriber), так что без этой строки нотификации до подписчика не дошли бы в принципе.

```
VERIFY failed: unexpected kind in SetConfigSubscriptionRequest handler: UdfStoreConfigItem
  ydb/core/cms/console/configs_dispatcher.cpp:1465
  CheckKinds(): requirement false failed
```

GrigoriyPA (approved):

CyberROFL (approved):

PR #52829, opened 2026-09-11
Improve Interconnect v2 read/write buffer size calculation

alexvru:
### Changelog entry 

Improve Interconnect v2 read/write buffer size calculation

### Description for reviewers 

This patch improves algorithm for scratch buffer size calculation.


mregrock (approved):
LGTM

PR #52846, opened 2026-09-11
Add multi-host management and federated views to ydb benchmark

kruall:
### Changelog entry 

Add host management, remote benchmark launching, and unified runs, topology, and cross-host comparisons to the benchmark web interface.

### Description for reviewers 

Persist peer identities and credentials, synchronize membership when joining hosts, and proxy authenticated peer requests through the current server. Host-qualified references keep runs and comparisons distinct while results remain on their owning hosts. New run uses the selected host for binary discovery, validation, drafts, and execution; remote runs can be cancelled, while remote saved comparisons remain read-only.


mregrock (approved):
LGTM

PR #52853, opened 2026-09-11
Normalize the default benchmark web route before rendering

kruall:
### Changelog entry 

Normalize the benchmark web entry URL to the Runs page when no route is specified.

### Description for reviewers 

The router treats an empty fragment as Runs, while the federated Runs renderer in #52846 requires an explicit #runs fragment after loading hosts. This mismatch leaves a bare entry URL displaying Loading indefinitely. Normalize the fragment before dispatch with replaceState, preserving query parameters and avoiding an extra history entry. Existing explicit routes remain unchanged.


kruall:
Superseded by #52870, which consolidates this fix, #52858, and the restored benchmark features on updated main. Closing this PR to keep a single review.

PR #52858, opened 2026-09-11
Restore compact New run editor

kruall:
### Changelog entry 

Restore the compact benchmark run editor with profile tabs and grouped settings.

### Description for reviewers 

The active editor still renders the legacy card-and-sidebar layout instead of the previously deployed compact form. Restore the layout without changing configuration serialization or load-search semantics, retain dataset expansion per profile, and cover Builder, YAML and validation-error controls with regression tests.

PR #52870, opened 2026-09-11
Restore benchmark configuration, compact editor and adaptive SLO search

kruall:
### Changelog entry 

Restore benchmark configuration views, compact New run controls, and exact latency-SLO selection.

### Description for reviewers 

Consolidates #52853 and #52858 with previously deployed changes missing from main. Restore the saved configuration profile/YAML tabs without applying current defaults, preserve host-aware routing and execution controls, and normalize empty routes before rendering. Latency search alternates interpolation with binary refinement to an observed adjacent-load boundary and resumes after rejected adaptive verification instead of publishing a failing selection as an independent holdout.


mregrock (approved):
LGTM

PR #52974, opened 2026-09-13
Add distributed YDB benchmark execution and configuration builder

kruall:
### Changelog entry 

Add experimental multi-host YDB benchmarks, distributed profile comparisons, and tabbed configuration builders.

### Description for reviewers 

Placement templates become detached run configurations with per-tenant settings and multiple KV or stock generators; load search applies to one CLI. Generation-scoped worker leases coordinate startup and cleanup, with bounded per-host artifacts and durable process ownership.

Comparisons now retain benchmark identity when selecting and saving distributed profiles instead of treating every profile as local-ydb. Builders expose profile rename/copy/delete actions and grouped settings; local actor-system flags remain shared while vCPU settings stay per role.


mregrock (approved):
LGTM

PR #53001, opened 2026-09-14
Fix  Y_VERIFY assert on shutdown due to access to dangling pointers

shokhor:
### Changelog entry 

Fix rare Y_VERIFY crash on server shutdown

### Description for reviewers 

View stack traces and initial analysis in LOGBROKER-10618.

### Problem

During kikimr shutdown, a `Y_VERIFY` crash occurs in `grpc::ServerAsyncReaderWriter::Finish()` when `grpc_cq_begin_op(cq_, notify_tag)` fails on a completion queue that has already been shut down.

**Root cause**: `KikimrStop()` calls `StopGRpcServers()` (which shuts down CQs and destroys `TGRpcServer` objects) **before** `ActorSystem->Stop()` (which destroys actors holding streaming request contexts). When an actor is force-destroyed, its `TGRpcRequestBiStreamWrapper` destructor runs `~TFacade()` → `Detach()` → `FinishInternal()` → `Stream.Finish()` on a dead CQ → Y_VERIFY crash.

### Changes

**1. `core/grpc_streaming/grpc_streaming.h` — `FinishInternal()` (line ~564)**

Added `!Server->IsShuttingDown()` guard before `Stream.Finish()`. When the server is shutting down, skips the gRPC finish call and calls `OnFinishDone(status)` directly to notify the actor.

**2. `core/grpc_streaming/grpc_streaming.h` — `OnWriteDone()` (line ~512)**

Added early `return` if `Server->IsShuttingDown()` to avoid issuing the next gRPC write/finish operation on a dead CQ.

**3. `core/driver_lib/run/run.cpp` — `StopGRpcServers()` (line ~241)**

Commented out `wrapper->Servers.clear()`. The `TGRpcServer` objects are now destroyed when `GRpcServersWrapper` (a `shared_ptr` member of `TKikimrRunner`) is released in `~TKikimrRunner()`, which runs after `ActorSystem.Destroy()` — ensuring the `Server` pointer remains valid during all actor destruction.

**4. `core/grpc_streaming/grpc_streaming_ut.cpp` — Test**

Added `FinishAfterServerShutdownDoesNotCrash` unit test that creates a streaming request, starts server shutdown, and verifies no crash occurs during actor destruction.

### Risk Analysis

**Change 1 — `FinishInternal()` guard**:
- **Risk**: Low. `Server->IsShuttingDown()` reads an atomic flag already accessed in `MaybeClone()` and `Start()`. The guard only skips `Stream.Finish()` when the server is shutting down — the stream is being torn down anyway, so the client is already gone.
- **Behavior change**: When `ShuttingDown_` is set, the actor receives `TEvNotifiedWhenDone` via `OnFinishDone()` instead of via the gRPC finish callback. The actor already handles this event in its `StateWork` handler.

**Change 2 — `OnWriteDone()` guard**:
- **Risk**: Low. `OnWriteDone()` is a CQ callback that fires when a `Stream.Write()` or `Stream.WriteAndFinish()` completes. Without the guard, the code proceeds to issue the *next* gRPC operation (`Stream.Write`, `Stream.WriteAndFinish`, or `Stream.Finish`) on the CQ — which crashes with Y_VERIFY if the CQ is being shut down. The guard skips only the next gRPC operation.
- **Behavior change**: When `Server->IsShuttingDown()` is true, `OnWriteDone()` returns early after sending `TEvWriteFinished` to the actor and clearing `FlagWriteActive`. The next queued write (if any) is not issued to gRPC. If `FlagFinishCalled` was set, `OnFinishDone()` is not called — the finish is silently skipped. The actor will be destroyed shortly anyway during `ActorSystem->Stop()`.

**Change 3 — Remove `Servers.clear()`**:
- **Risk**: Low. `server->Stop()` is still called, which shuts down CQs and sets `ShuttingDown_`. The `TGRpcServer` objects stay alive longer (until `~TKikimrRunner`) but their internal resources (CQs, server, threads) are already released by `Stop()`. The destructor `~TGRpcServer()` only asserts `Ts.empty()` and clears `Services_`, both safe after `Stop()`.
- **Memory**: The `TGRpcServer` objects remain in memory for the brief window between `StopGRpcServers()` and `~TKikimrRunner()`. This is negligible — shutdown is already in progress.
- **Behavior change**: None. The servers are stopped (not accepting new requests) at the same point as before. Only the destruction is deferred.

**Pre-existing latent bug**: The `Server` pointer in `TGRpcStreamingRequest` (and `Server_` in `TGRpcRequestImpl`) was already dereferenced in destructors (`DeregisterRequestCtx`) before this fix. The `Servers.clear()` removal fixes this pre-existing use-after-free for all three at-risk locations:
1. `~TGRpcStreamingRequest()` line 148 — `Server->DeregisterRequestCtx(this)`
2. `FinishInternal()` line 564 — `Server->IsShuttingDown()` (new)
3. `TGRpcRequestImpl::DestroyRequest()` line 165 — `Server_->DeregisterRequestCtx(this)`


ydb/core/grpc_streaming/grpc_streaming.h
```diff
@@ -507,6 +507,12 @@ class TGRpcStreamingRequest final
             }
         }
 
+        // same as FinishInternal - do not issue gRPC operations
+        // on a completion queue that is (or is about to be) shut down.
+        if (Server->IsShuttingDown()) {
+            return;
+        }
```
nshestakov: This early return also skips `OnFinishDone()`, which is **not** a CQ operation. `OnFinishDone()` sets `FlagFinishDone`, calls `DecRequest()`, updates counters, and `DeregisterRequestCtx`.

If the completed write was already a `WriteAndFinish` (`wasWriteAndFinish`), or `nextStatus` is set, we should still call `OnFinishDone(status)` here. Otherwise:
- limiter tokens leak (`DecRequest` is skipped);
- the request stays in `Requests_` until the destructor, so `Stop()` waits for `GRpcShutdownDeadline`;
- a dequeued `next` is dropped with no `TEvWriteFinished`.

Guard only the `Stream.*` calls, e.g.:

```cpp
if (next && nextStatus) {
    if (!Server->IsShuttingDown()) {
        Stream.WriteAndFinish(next->Message, next->Options, *nextStatus, OnWriteDoneTag.Prepare());
    } else {
        OnFinishDone(status);
    }
} else if (next) {
    if (!Server->IsShuttingDown()) {
        Stream.Write(next->Message, next->Options, OnWriteDoneTag.Prepare());
    }
} else if (nextStatus) {
    if (!Server->IsShuttingDown()) {
        Stream.Finish(*nextStatus, OnFinishDoneTag.Prepare());
    } else {
        OnFinishDone(status);
    }
} else if (wasWriteAndFinish) {
    OnFinishDone(status);
}
```

ydb/core/grpc_streaming/grpc_streaming.h
```diff
@@ -547,7 +553,14 @@ class TGRpcStreamingRequest final
...
-        if (finish) {
+        // when the server is shutting down its completion queues are (or are about to be) dead.
+        // Issuing Stream.Finish() on a dead CQ fails with "grpc_cq_begin_op(cq_, notify_tag) failed"
+        // and crashes the process. This happens when an actor holding the IStreamCtx
+        // (TFacade) is force-destroyed after the gRPC server has been stopped
+        // (e.g. ActorSystem::Stop() destroys the request proxy and its deferred
+        // streaming requests). Skip the gRPC finish in that case; the stream is
+        // being torn down anyway.
+        if (finish && !Server->IsShuttingDown()) {
             Stream.Finish(status, OnFinishDoneTag.Prepare());
```
nshestakov: The PR description says that when the server is shutting down we skip `Stream.Finish()` and call `OnFinishDone(status)` directly. The code only skips `Stream.Finish()`.

Please do what the description promises:

```cpp
if (finish) {
    if (!Server->IsShuttingDown()) {
        Stream.Finish(status, OnFinishDoneTag.Prepare());
    } else {
        OnFinishDone(NYdbGrpc::EQueueEventStatus::ERROR);
    }
}
```

`TEvNotifiedWhenDone` is sent from `OnDone` (`AsyncNotifyWhenDone`), not from `OnFinishDone` — the description should be corrected as well.

ydb/core/driver_lib/run/run.cpp
```diff
@@ -239,7 +239,14 @@ void StopGRpcServers(std::weak_ptr<TGRpcServersWrapper> grpcServersWrapper, bool
...
-    wrapper->Servers.clear();
+    // Do not destroy TGRpcServer objects here.
+    // KikimrStop() calls StopGRpcServers() before ActorSystem->Stop(), so destroying the
+    // servers here would leave dangling TServer* pointers in any
+    // TGRpcStreamingRequest objects that are still alive when the actor system
+    // destroys their holding actors. The servers are destroyed later when
+    // GRpcServersWrapper (a shared_ptr member of TKikimrRunner) is released in
+    // ~TKikimrRunner(), which runs after ActorSystem.Destroy().
+    // wrapper->Servers.clear();
 }
```
nshestakov, line 249: Please delete `Servers.clear()` instead of leaving it commented out.

Side effect of keeping the objects: `TGRpcServer::Stop()` can now run twice on the same instance (`TGRpcServersManager::Stop` and then `KikimrStop`). Previously the second pass iterated an empty vector. Confirm that a second `Server_->Shutdown()` / `cq->Shutdown()` is safe, or make `Stop()` idempotent.

Also, `TGRpcServersManager::Start()` still does `wrapper->Servers = factory()`, which destroys the old servers while actors may still hold `TServer*`. For `KikimrStop` this is fine; for cert-rotation restart the UAF just moves from `Stop()` to `Start()`.

ydb/core/grpc_streaming/grpc_streaming_ut.cpp
```diff
@@ -484,6 +538,46 @@ Y_UNIT_TEST_SUITE(TGRpcStreamingTest) {
...
+            // in flight on the server side.
+            Y_UNUSED(stream);
+
+            // Wait until the server-side THangActor has been created and attached
+            // to the stream. This guarantees the IStreamCtx (TFacade) is held by
+            // the actor before we shut the server down, so the facade destructor
+            // will run after the completion queue is dead.
+            THangActor::Attached.WaitI();
+        }
+
```
nshestakov: The comment says "Keep the client stream alive so the RPC stays in flight", but the scope ends before `server.Reset()`. The client is already gone by the time CQs are shut down, so this is a weaker repro than LOGBROKER-10618 (in-flight Topic write).

Hold `ClientContext` / `stream` across `server.Reset()`. Even better, have the server write at least once before shutdown.

`THangActor::Attached` is a process-wide `TManualEvent` and is never reset — an in-process retry will return from `WaitI()` before the new actor attaches. Prefer a per-test event or `Reset()`.

nshestakov (changes requested):
The direction is right: do not call `Stream.Finish()` on a dead CQ, and do not destroy `TGRpcServer` while actors still hold `TServer*`. That matches the LOGBROKER-10618 stack, and `~TKikimrRunner()` really does `ActorSystem.Destroy()` before `GRpcServersWrapper` is released.

Please address the points below before merge. The `OnWriteDone` / `FinishInternal` guards are broader than needed and do not match the PR description (which says `OnFinishDone()` is invoked when finish is skipped).

**Also not in the diff, but related:** `Write()` / `Read()` still issue `Stream.Write` / `Stream.Read` with no `IsShuttingDown()` check. Combined with the `OnWriteDone` early return (which clears `FlagWriteActive`), a still-running actor can write again during `TGRpcServer::Stop()` and hit the same `Y_VERIFY`.

nshestakov (approved):

ydb/core/grpc_streaming/grpc_streaming.h
```diff
@@ -507,12 +533,32 @@ class TGRpcStreamingRequest final
...
+                with_lock (WriteLock) {
+                    Flags &= ~FlagWriteActive;
+                }
+                OnFinishDone(status);
+            }
         } else if (next) {
-            Stream.Write(next->Message, next->Options, OnWriteDoneTag.Prepare());
+            if (!Server->IsShuttingDown()) {
+                Stream.Write(next->Message, next->Options, OnWriteDoneTag.Prepare());
+            }
```
nshestakov, line 555: When `next` is dequeued, `FlagWriteActive` is intentionally left set (write chaining). If we skip `Stream.Write` here, nothing clears that flag and `OnFinishDone()` is not called.

`FinishInternal()` then sees `FlagWriteActive` and sets `finish == false`, so the limiter / `DeregisterRequestCtx` / counters are skipped while the actor is still alive. `TGRpcServer::Stop()` waits for `RequestsInProgress() == 0` *before* shutting down CQs, so this turns a drainable streaming session into a full `GRpcShutdownDeadline` wait (default 30s).

Previously the CQ was still alive at this point, the next write went out, and the actor could `Finish`. Now `IsShuttingDown()` is true from `StopService()`, so the chain is cut.

Please do the same bookkeeping as the `next && nextStatus` branch: clear `FlagWriteActive`, drop `WriteQueue`, and call `OnFinishDone()` if `FlagFinishCalled` is already set. That also lets a later `~TFacade` → `FinishInternal()` complete shutdown.

ydb/core/grpc_streaming/grpc_streaming.h
```diff
@@ -547,8 +593,20 @@ class TGRpcStreamingRequest final
...
+        // and crashes the process. This happens when an actor holding the IStreamCtx
+        // (TFacade) is force-destroyed after the gRPC server has been stopped
+        // (e.g. ActorSystem::Stop() destroys the request proxy and its deferred
+        // streaming requests). Skip the gRPC finish in that case; the stream is
+        // being torn down anyway. OnFinishDone() is still invoked so the limiter
+        // token is released, the request is deregistered and counters are updated.
         if (finish) {
-            Stream.Finish(status, OnFinishDoneTag.Prepare());
+            if (!Server->IsShuttingDown()) {
+                Stream.Finish(status, OnFinishDoneTag.Prepare());
```
nshestakov, line 606: `IsShuttingDown()` is only a snapshot. `KikimrStop()` runs on the main thread concurrently with actors, so this window is real:

1. this thread observes `IsShuttingDown() == false`
2. main thread: `StopService()` → wait inflight → `cq->Shutdown()`
3. this thread calls `Stream.Finish` / `Write` / `Read` on a dead CQ → the same `Y_VERIFY`

`Start()` already uses `ProtectShutdown()` for this: while the guard is held, `IsUnsafeToShutdown()` prevents CQ shutdown. Please wrap `Stream.Read` / `Write` / `WriteAndFinish` / `Finish` the same way (here, in `Write()`/`Read()` after the flag check, and in the `OnWriteDone` `Stream.*` calls). A bare `if (!IsShuttingDown())` is not enough.

ydb/core/grpc_streaming/grpc_streaming.h
```diff
@@ -424,6 +437,19 @@ class TGRpcStreamingRequest final
...
+        // about to be) shut down. The actor is still alive here (it is calling
+        // Write), so deliver a failure completion:
+        // this lets the actor's state machine proceed without
+        // blocking on a TEvWriteFinished that will never arrive. The gRPC
+        // client is not notified (the CQ is dead anyway).
+        if (Server->IsShuttingDown()) {
+            auto event = MakeHolder<typename IContext::TEvWriteFinished>();
+            event->Success = false;
+            ActorSystem.Send(Actor, event.Release());
+            return false;
```
nshestakov, line 450: This sends `TEvWriteFinished(false)` **and** returns `false`. That disagrees with the `FlagFinishCalled` contract used everywhere else (including `Read()` just above): `false` means the call was not accepted, so no completion event.

`Write()` docs also say each accepted write produces exactly one `TEvWriteFinished`. Callers such as `TReadSessionActor::WriteToStreamOrDie` push into `ActiveWrites` then `Die()` on `false`; a later `TEvWriteFinished` is a second completion for a write that was rejected.

Please either:
- behave like `FlagFinishCalled` (return `false`, no event), or
- accept the write and guarantee exactly one event

not both. Also check `FlagFinishCalled` *before* this shutdown branch, otherwise a write after `Finish()` emits a spurious `TEvWriteFinished` during shutdown.

ydb/core/driver_lib/run/run.cpp
```diff
@@ -239,7 +239,13 @@ void StopGRpcServers(std::weak_ptr<TGRpcServersWrapper> grpcServersWrapper, bool
...
     }
 
-    wrapper->Servers.clear();
+    // Do not destroy TGRpcServer objects here.
+    // KikimrStop() calls StopGRpcServers() before ActorSystem->Stop(), so destroying the
+    // servers here would leave dangling TServer* pointers in any
+    // TGRpcStreamingRequest objects that are still alive when the actor system
+    // destroys their holding actors. The servers are destroyed later when
+    // GRpcServersWrapper (a shared_ptr member of TKikimrRunner) is released in
+    // ~TKikimrRunner(), which runs after ActorSystem.Destroy().
```
nshestakov, line 248: Correct for the `KikimrStop` path.

Still open: `TGRpcServersManager::Start()` does `wrapper->Servers = factory()`, which destroys these objects while actors may still hold `TServer*`. The UAF just moves from `Stop()` to the next `Start()` (pile / gRPC restart).

Also, the new lifetime depends on `ActorSystem.Destroy()` running **in the body** of `~TKikimrRunner()`. `GRpcServersWrapper` is declared *after* `ActorSystem`, so without that explicit `Destroy()` C++ would free the wrapper first. Worth a follow-up so this does not come back.

ydb/core/grpc_streaming/grpc_streaming_ut.cpp
```diff
@@ -484,6 +543,47 @@ Y_UNIT_TEST_SUITE(TGRpcStreamingTest) {
...
+        options.SetGRpcShutdownDeadline(TDuration::MilliSeconds(100));
+
+        auto server = MakeHolder<TGRpcTestServer<THangActor>>(std::move(options));
+
+        auto channel = grpc::CreateChannel(server->GRpcEndpoint, grpc::InsecureChannelCredentials());
+        auto stub = NStreamingTest::TStreamingService::NewStub(channel);
+
+        // The server actor (THangActor) now holds the IStreamCtx and never
+        // finishes the stream. Keep the client stream alive across server.Reset()
+        // so the RPC stays in flight on the server side (with an in-flight write)
```
nshestakov, line 568: Nit: the comment says the RPC stays in flight "with an in-flight write", but `THangActor` intentionally never writes so that `FlagWriteActive` is clear. Please fix the comment.

This test covers the original `FinishInternal` crash, not the queued-write `OnWriteDone` hole. A second case that writes at least once (or queues a second write) before `server.Reset()` would lock that in.

`THangActor::Attached` is still a process-wide `TManualEvent`. `Reset()` at the end helps, but a failure before that makes an in-process retry return from `WaitI()` too early. Prefer a per-test event.

nshestakov (changes requested):
## Review

The direction is right: do not call `Stream.Finish()` / `Write()` / `Read()` on a dead CQ, and do not destroy `TGRpcServer` while actors still hold `TServer*`. That matches LOGBROKER-10618, and `~TKikimrRunner()` really does `ActorSystem.Destroy()` before `GRpcServersWrapper` is released. The second commit also fixes `OnFinishDone` on the finish paths, makes `Stop()` idempotent, and holds the client stream across `server.Reset()`.

Please address the `OnWriteDone` `next`-only branch before merge. Dropping the queued write without clearing `FlagWriteActive` / calling `OnFinishDone()` will stall graceful drain of streaming sessions (Topic write/read with a write queue) until `GRpcShutdownDeadline` (30s). There is also a remaining TOCTOU around `IsShuttingDown()` vs `Stream.*` that can still hit the same `Y_VERIFY`.

### Already good
- Removing `Servers.clear()` so `TGRpcServer` outlives actor destruction on the `KikimrStop` path
- Idempotent `TGRpcServer::Stop()` (second pass from `KikimrStop` after `TGRpcServersManager::Stop`)
- `FinishInternal` calling `OnFinishDone()` when finish is skipped
- `FinishAfterServerShutdownDoesNotCrash` + fixture member order (`GRpcServer` before `Server`)

Inline comments have the details.

ydb/core/grpc_streaming/grpc_streaming_ut.cpp
shokhor: > A second case that writes at least once (or queues a second write) before server.Reset() would lock that in.

Тест, который бы вызывал  ассерт в OnWriteDone, не удаётся написать.

ydb/core/grpc_streaming/grpc_streaming.h
shokhor: Валидный коммент, но реализация выглядит затруднительно - значительный риск соорудить дедлок.

ydb/core/driver_lib/run/run.cpp
shokhor: UAF  в gRPC restart - не регрессия, выглядит значительным рефакторингом, не в контексте этой задачи.

nshestakov (approved):

PR #53011, opened 2026-09-14
tablet_flat: cut channel history reassigned by a storage pool shrink after MoveData

lberserq:
### Changelog category

* Bugfix

### Description for reviewers

After a storage pool shrink (channel reassign → `TEvMoveData` → restart), flat-executor tablets never cut the reassigned channel history, so the old group was never released: the boot after MoveData still saw old-generation blobs.

- **Redo replay.** `NBoot::TAlter::Apply` applies the same reassigned-channel check as `ReadAlterLog`, so a schema that lives only in redo is rewritten.
- **Vacuum.** After its second snapshot the vacuum waits until GC has collected everything deleted before it, then completes on a third, final snapshot. The existing `TEvSnapshotConfirmed` advances the GC boundary without another full snapshot. `TExecutor::DriveVacuumGc` sends the GC rounds an idle tablet would not send itself, including when snapshot confirmation arrives.
- **Alter log.** `TLogicAlter::SnapToLog` also hands `ObsoleteLog` to the snapshot commit's `GcDelta.Deleted`, so obsolete schema bodies are collected in the current generation.

Tests in `flat_executor_ut.cpp`: `RestartAfterMoveDataCutsReassignedHistory`, `SchemaOnlyRestartAfterMoveDataCutsReassignedHistory`, plus delayed-GC and delayed-snapshot-confirmation cases. They assert three snapshots and check that the schema and rows survive a restart with the old channel history removed.

### DIFF
--- a/ydb/core/tablet_flat/flat_executor.cpp
+++ b/ydb/core/tablet_flat/flat_executor.cpp
@@ -1044,6 +1044,19 @@ void TExecutor::FollowerGcApplied(ui32 step, TDuration followerSyncDelay) {
         Counters->Percentile()[TExecutorCounters::TX_PERCENTILE_FOLLOWERSYNC_LATENCY].IncrementFor(followerSyncDelay.MicroSeconds());
 }
 
+void TExecutor::DriveVacuumGc(bool collect) {
+    VacuumLogic->OnCollectedGarbage(OwnerCtx());
+    if (collect && VacuumLogic->NeedGC()) {
+        GcLogic->SendCollectGarbage(ActorContext());
+        // A channel with nothing left to collect sends nothing and brings no result, so re-check right away.
+        VacuumLogic->OnCollectedGarbage(OwnerCtx());
+    }
+    // N.B. PassAway may have already been called
+    if (Owner && VacuumLogic->NeedLogSnaphot()) {
+        MakeLogSnapshot();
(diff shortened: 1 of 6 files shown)


ydb/core/tablet_flat/flat_executor.cpp
```diff
@@ -1053,9 +1066,7 @@ void TExecutor::CheckCollectionBarrier(TIntrusivePtr<TBarrier> &barrier) {
                 Owner->CompletedLoansChanged(OwnerCtx());
             }
         }
-        if (VacuumLogic->NeedGC()) {
-            GcLogic->SendCollectGarbage(ActorContext());
-        }
+        DriveVacuumGc(true);
```
vporyadke, line 1070: Мой робот утверждает, что вызов MakeLogSnapshot отсюда приведёт к созданию вложенного коммита и падению по ensure-ке здесь https://github.com/ydb-platform/ydb/blob/baec32b39bd5a0c12996b36cddb8a559b3710890/ydb/core/tablet_flat/flat_exec_commit_mgr.h#L172

lberserq: Исправлено в 54cf9840107

vporyadke (approved):

PR #53024, opened 2026-09-14
YQ-5696 added timestamp value validation for kafka records

GrigoriyPA:
### Changelog entry 

Added timestamp value validation for kafka records


ydb/public/sdk/cpp/src/library/kafka/kafka_records.cpp
```diff
@@ -734,6 +736,14 @@ i32 TKafkaRecordBatch::Size(TKafkaVersion _version) const {
     return _collector.Size;
 }
 
+void TKafkaRecordBatch::ValidateTimestampDelta(TKafkaRecord::TimestampDeltaMeta::Type delta) const {
+    if ((delta > 0 && BaseTimestamp > std::numeric_limits<i64>::max() - delta) ||
+        (delta < 0 && BaseTimestamp < std::numeric_limits<i64>::min() - delta))
+    {
+        ythrow yexception() << "Kafka record timestamp overflow";
```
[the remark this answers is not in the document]
GrigoriyPA: Maybe out-of-scoupe (fixed one path in this PR)

nshestakov: This only runs from `TKafkaRecordBatch::Read`. Produce with batching enabled uses `ReadKafkaBatchHeader` + raw v2 storage (`ParseProduceRecords` / `CanStoreRawRecords`) and never parses records, so `BaseTimestamp == INT64_MAX` + `TimestampDelta == 1` is ingested and fails later in `TKafkaBatchCutter` / fetch.

Apache Kafka does not treat `baseTimestamp + timestampDelta` overflow as a decode error (Java wrap). It validates timestamps **before append** vs broker time (KIP-937) and returns **32 `INVALID_TIMESTAMP`**. YDB maps this `yexception` to **42 `INVALID_RECORD`**, and only on the non-raw parse path.

Please validate (or parse) records on ingest before the raw-storage path, and consider mapping overflow/out-of-range to `INVALID_TIMESTAMP` rather than a generic parse failure.

ydb/public/sdk/cpp/src/library/kafka/kafka_records.cpp
```diff
@@ -655,6 +655,7 @@ void TKafkaRecordBatch::Read(TKafkaReadable& _readable, TKafkaVersion _version)
         NPrivate::Read<RecordsMeta>(body, _version, Records);
         for (auto& record : Records) {
             record.OwnPayload();
+            ValidateTimestampDelta(record.TimestampDelta);
```
nshestakov: Same validate is duplicated in the compressed branch. One loop over `Records` after the `if/else` would be enough.

Also: this still does not cover produce when records are not read at all.

ydb/public/sdk/cpp/src/library/kafka/kafka_records.h
```diff
@@ -4,6 +4,7 @@
 
 #include <optional>
 #include <utility>
+#include <vector>
```
nshestakov, line 8: Unrelated to timestamp validation — `std::vector` was already used in this header. Fine as IWYU, but it is not required by this change.

nshestakov (commented):
## Summary

This is **C++ signed-overflow hardening** for `baseTimestamp + timestampDelta`, not Apache Kafka timestamp validation (KIP-937). The overflow check itself is correct and the boundary tests are good, but it does not run on the main produce path.

## Apache Kafka behavior

In magic v2 a record timestamp is `baseTimestamp + timestampDelta` (`DefaultRecord.readFrom`):

```java
long timestampDelta = ByteUtils.readVarlong(buffer);
long timestamp = baseTimestamp + timestampDelta;
```

Java overflow is **defined** (wraps around `Long.MIN/MAX`). Kafka does **not** fail decode because the sum overflows. After that:

1. **CreateTime** — broker checks a window vs broker time (`log.message.timestamp.before.max.ms` / `after.max.ms`, KIP-937). Error **32 `INVALID_TIMESTAMP`**, the whole batch is rejected **before** append.
2. **LogAppendTime** — timestamp is overwritten with broker time; CreateTime checks are skipped.
3. `NO_TIMESTAMP = -1` skips the range check.

KIP-937 is about overflow in `record.timestamp - now`, not about `base + delta` at parse time.

| | Apache Kafka | This PR |
|---|---|---|
| Decode overflow | wrap, batch is readable | `yexception` |
| Produce error | `INVALID_TIMESTAMP` (32) | `INVALID_RECORD` (42), and only if records are parsed |
| When rejected | before append | in `Read()`, not on raw-batch ingest |
| Window vs broker time | yes (KIP-937) | no |
| LogAppendTime | overwrite, no CreateTime check | same overflow check |

YDB kafka proxy never uses `INVALID_TIMESTAMP`. Produce maps `yexception` to `INVALID_RECORD`.

## Main issue: raw produce bypasses the check

With batching enabled (typical path) `ParseProduceRecords` reads only the header and stores a v2 batch as raw bytes. `ValidateTimestampDelta` is never called.

A batch with `BaseTimestamp = INT64_MAX` and `TimestampDelta = 1` is **accepted and stored**. It fails later:

- `TKafkaBatchCutter::Cut` → `ReadKafkaRecordBatch` → exception on already persisted data
- fetch / viewer / SDK decode

Kafka does not put an invalid timestamp into the log. Here a poison batch is ingested and breaks the read path.

The same `BaseTimestamp + TimestampDelta` add remains in:

- `ydb/core/kafka_proxy/actors/kafka_produce_actor.cpp` (`ui64 createTime = batch.BaseTimestamp + record.TimestampDelta`)
- `ydb/core/persqueue/pqtablet/batching/batch_cutter.cpp`

Offset already has a safe helper (`TryRecordOffset`) in the cutter. Timestamp does not — the check is only in the decoder, so ingest with raw storage is uncovered.

## Smaller notes

- `#include <vector>` in `kafka_records.h` is unrelated; `std::vector` was already used there.
- Compressed/uncompressed `Read` both loop `OwnPayload` + validate; one loop after the `if/else` is enough.
- `MaxTimestamp` is not checked against record timestamps.
- There is no CreateTime window vs broker now. If the ticket meant Kafka-style timestamp validation, this is not it.
- Tests cover overflow/boundaries and NONE/GZIP/ZSTD, but only full `Read` round-trip — not produce with raw storage and not the cutter.

## Verdict

Reasonable as UBSan/`base + delta` hardening in C++. Not Kafka-compatible: Kafka does not reject this at parse time, rejects later with code 32, and only vs broker time.

To make this useful in YDB, run the check **on produce before raw persist** (or parse records on the batching path too), map it to a Kafka error that matches intent, and stop doing unchecked `BaseTimestamp + TimestampDelta` in the produce actor / cutter. Otherwise this only tightens the decoder unit test and does not close real ingest.

ydb/public/sdk/cpp/src/library/kafka/kafka_records.cpp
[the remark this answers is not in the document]
GrigoriyPA: Fixed

GrigoriyPA: Fixed

ydb/public/sdk/cpp/src/library/kafka/kafka_records.h
GrigoriyPA: Оставил, иначе IDE подсвечивает файл

nshestakov (dismissed):

ydb/core/viewer/viewer_topic_data.cpp
```diff
@@ -1,6 +1,7 @@
 #include "viewer_topic_data.h"
 #include "log.h"
 #include <library/cpp/protobuf/json/proto2json.h>
+#include <ydb/public/sdk/cpp/src/library/kafka/kafka_records.h>
```
CyberROFL: Сортировка поехала.

CyberROFL (dismissed):
Можно поправить отдельно, но поправить надо обязательно.

Shfdis (dismissed):

yumkam (dismissed):

kuzin57 (dismissed):

ydb/core/viewer/viewer_topic_data.cpp
GrigoriyPA: Исправил

ydb/public/sdk/cpp/src/library/kafka/kafka_records.cpp
GrigoriyPA: Исправил

CyberROFL (approved):

alexvru (approved):

yumkam (approved):

kuzin57 (approved):

Shfdis (approved):

PR #53112, opened 2026-09-15
Fix XDC oversize

alexvru:
### Changelog entry 

Fix XDC oversize

### Description for reviewers 

This fixes problem that can lead to node assertion failure when peer sends a message that is too large through XDC.


ydb/library/actors/interconnect/interconnect_uring_engine.cpp
```diff
@@ -279,7 +291,13 @@ namespace NActors {
             void ApplyBytesReadCopy(const char *data, size_t num, size_t poolBufSize) {
                 BytesReceived += num;
                 NSan::Unpoison(data, num);
-                Deserializer.Push(TRcBuf::Copy({data, num}), this, SessionId);
+                try {
```
mregrock, line 294: Copypaste

mregrock (changes requested):

ydb/library/actors/interconnect/interconnect_uring_engine.cpp
alexvru: You're absolutely right!

mregrock (approved):

PR #53113, opened 2026-09-15
Fix blocks cache bug in VDisk

alexvru:
### Changelog entry 

Fix blocks cache bug in VDisk

### Description for reviewers 

This fixes incorrect assertion failure while processing blocks cache.


mregrock (approved):
LGTM

PR #53123, opened 2026-09-15
Polish benchmark configuration and run interfaces

kruall:
### Changelog entry 

Improve benchmark forms, run reports, and configuration readability.

### Description for reviewers 

Align distributed Builder fields, actor-system controls and dialogs with the existing UI; keep New run actions visible while scrolling. Compact run headers, move refresh into navigation, and group distributed configuration by topic while preserving the inner YAML tab.

Reuse the distributed YAML serializer for template drafts and emit block sequences instead of JSON arrays. Resolve host names in configuration views, restore run-policy guidance, hide inactive comparison reset controls, and align comparison actions to the right.


mregrock (approved):
LGTM

PR #53128, opened 2026-09-15
Fix executor statistics race during shared pool switching

kruall:
### Changelog entry 

Fixed a data race when collecting actor system statistics during shared executor pool switching.

### Description for reviewers 

Monitoring and harmonizer threads could read the executor-owned statistics pointer while SwitchPool changed it. Publish a separate atomic pointer and select one pool for the entire collection update, preserving the executor's ordinary pointer on the event path. Relaxed ordering is sufficient because the per-pool statistics objects are initialized before threads start and remain at stable addresses.

Addresses YDBBUGS-775 and YDBBUGS-778.


mregrock (approved):
LGTM

PR #53147, opened 2026-09-15
[YDB-3530] memory controller: re-register a consumer kind instead of aborting, add unregister

lberserq:
### Changelog category

* Improvement

### Description for reviewers

Replaces the abort on a repeated `TEvConsumerRegister` for the same kind, so a consumer service that restarts no longer takes the node down.

- **Re-registration instead of abort.** `Y_ABORT_UNLESS(..., "Consumer kinds should be unique")` is gone. The new registrant gets a fresh consumer object, so the replaced one cannot keep writing into MC through its old pointer; a takeover by another actor logs a warning and increments `Stats/ConsumerTakeovers`.
- **`TEvConsumerUnregister`.** A consumer can take its bytes out of the accounting, as `TEvMemTableUnregister` already does for memtables, and the removed kind's gauges are zeroed. An unregister from an actor that does not own the kind is ignored.

Overlaps with [#53013](https://github.com/ydb-platform/ydb/pull/53013), which relaxes the same check for `QueryExecution` only.

Tests in `memory_controller/memory_controller_ut.cpp`: `ConsumerTakeoverRebindsKind`, `ConsumerUnregisterDropsAccounting`.


ydb/core/memory_controller/memory_controller.cpp
```diff
@@ -324,12 +326,50 @@ class TMemoryController : public TActorBootstrapped<TMemoryController> {
 
     void Handle(TEvConsumerRegister::TPtr &ev, const TActorContext& ctx) {
         const auto *msg = ev->Get();
-        auto consumer = Consumers.emplace(msg->Kind, MakeIntrusive<TMemoryConsumer>(msg->Kind, ev->Sender));
-        Y_ABORT_UNLESS(consumer.second, "Consumer kinds should be unique");
+        auto [it, inserted] = Consumers.emplace(msg->Kind, MakeIntrusive<TMemoryConsumer>(msg->Kind, ev->Sender));
```
abyss7, line 329: Через новый путь перехвата можно подменить встроенные потребители контроллера: MemTable и ColumnTablesPortionsMetaDataCache. Может для них abort оставить?

lberserq: fixed in bdcf7189f1121237f0089bc399ae8099c34a8d42

lberserq:
Вошло в main вместе с #53165 (383e0ee8892): он стоял на вершине стека и влит одним squash-коммитом, файлы memory controller в main совпадают с головой его ветки.

PR #53156, opened 2026-09-15
Improve benchmark comparisons, builder and filtering

kruall:
### Changelog entry 

Improve benchmark configuration comparisons, distributed run setup, host activity and filtering.

### Description for reviewers 

Compare configuration fields by placement node, tenant and CLI instead of opaque JSON cells, distinguishing missing values from absent objects. Move distributed template selection into the Builder while preserving matching CLI settings and confirming removed or retargeted clients. Use the same activity query for local and peer hosts, replacing the local placeholder. Unify automatic filters, inclusive date labels and reserved Reset button space across run and comparison pages.


mregrock (approved):
LGTM

PR #53166, opened 2026-09-15
[detailed-metrics] wire processor aggregator to SVP

jsjant:
### Description for reviewers 

Include TDetailedTableCounters into TEvSendDbCountersRequest and send actual info from SysView Service to SysView Processor

In SysView Processor attach to a `ydb_detailed` counters group and publish it


ydb/core/sys_view/processor/db_counters.cpp
```diff
@@ -432,6 +497,9 @@ void TSysViewProcessor::Handle(TEvPrivate::TEvApplyCounters::TPtr&) {
     for (auto it = NodeCountersStates.begin(); it != NodeCountersStates.end(); ) {
         auto& state = it->second;
         if (state.FreshCount > 1) {
+            if (auto* aggregator = DetailedAggregator.Get()) {
+                aggregator->DropNode(it->first);
```
BlazingPhoenix, line 501: Resynchronize histogram baselines after node expiry

DropNode() discards the node’s live histogram balance, but the sender retains its Previous baseline and continues sending deltas. After a temporary disconnection, an unchanged histogram therefore stays zero at the processor; moving an observation between buckets then subtracts from zero and wraps to UINT64_MAX, corrupting table.datashard.used_core_percents. Subsequent reports cannot restore the missing baseline. Please send live histograms as absolute snapshots or introduce baseline resynchronization, and add a test where the same sender resumes reporting after DropNode().

jsjant: I added a 0 clamp for such case

BlazingPhoenix (approved):

PR #53174, opened 2026-09-15
Fix data race in KikimrStop

kruall:
### Changelog entry 

Fix a data race during gRPC server shutdown.

### Description for reviewers 

The gRPC factory runs on an actor thread and writes EnabledGrpcService while KikimrStop reads it without synchronization. Initialize this configuration flag before starting the manager instead of writing it on every factory invocation.

Related issue: [YDBBUGS-781](https://st.yandex-team.ru/YDBBUGS-781).


mregrock (dismissed):
LGTM

mregrock (approved):
LGTM

PR #53178, opened 2026-09-15
Add SDK OIDC providers

### LINKED ISSUE #53357: [SSO] Add OIDC support to the C++ SDK
We want applications using the YDB C++ SDK to authenticate through external identity providers such as Keycloak. The SDK should support existing access tokens, Client Credentials Grant for service applications, and Device Authorization Grant for interactive user sign-in.

These flows should share a common configuration and provider factory. The SDK should handle endpoint discovery, token expiration, refresh, and temporary failures, while allowing applications to provide token storage and user interaction through dedicated interfaces. Authentication should work asynchronously and shut down cleanly when cancelled. Tests should cover successful sign-in, error handling, and token renewal.

NewBediver:
### Changelog entry 

Add OIDC/OAuth authentication providers to the C++ SDK for external identity providers, including static access tokens, Client Credentials Grant, and Device Authorization Grant.

This adds a common `CreateOidcProviderFactory` entry point backed by three separate provider implementations:

- **Static:** uses a supplied access token and checks its expiration without refreshing it.
- **Client:** obtains tokens through Client Credentials Grant using `client_secret_basic`.
- **Device:** reports the verification URL and user code through `IAuthAcceptor`, then polls the token endpoint until authorization completes or expires.

The protocol implementation discovers endpoints from the issuer's OpenID configuration, validates HTTPS URLs and the returned issuer, and adds the `openid` scope when it is absent. It handles token expiration, refresh token rotation, transient failures, and device polling responses such as `authorization_pending` and `slow_down`.

`ITokenCacher` allows applications to supply token persistence. Client and device providers read cached tokens at startup and save newly acquired or refreshed tokens. Static credentials save the supplied token. Cached tokens remain unprefixed; the SDK sends access tokens as `Bearer <token>` in `x-ydb-auth-ticket`.

A shared `ICredentialsProvider` adapter owns the background worker and retains the selected provider until the worker finishes. Asynchronous credential requests support cancellation and response-queue delivery, including provider destruction from a completion callback. Client identities hash credential material so secrets and tokens are not exposed in the identity string.

### Description for reviewers 

https://nda.ya.ru/t/Fjdf_-L37rM46k


ydb/public/sdk/cpp/tests/unit/client/oidc/ya.make
```diff
@@ -0,0 +1,19 @@
+UNITTEST()
```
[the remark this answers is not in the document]
NewBediver: No, you should use UNITTEST for these tests

ydb/public/sdk/cpp/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## v3.23.0
 
+* Added OIDC/OAuth authentication via `CreateOidcProviderFactory`, supporting static access tokens, Client Credentials Grant, and Device Authorization Grant, with token refresh and interfaces for token caching and interactive sign-in.
```
NewBediver, line 3: I’m not sure we need to add this; the AI reviewer suggested it

ydb/public/sdk/cpp/src/client/types/credentials/oidc/credentials.cpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+TCredentialsProviderPtr TFactory::CreateProvider(std::weak_ptr<ICoreFacility> facility) const {
+    return CreateProviderImpl(std::move(facility), false);
+}
+
+std::string TFactory::GetClientIdentity() const {
+    return Identity;
+}
```
NewBediver, line 72: I don’t quite understand the purpose of this method. Currently, different Identities are created for each provider (since I’m generating a GUID), is that okay?

ydb/public/sdk/cpp/src/client/types/credentials/oidc/protocol.cpp
```diff
@@ -0,0 +1,378 @@
...
+THttpJob::~THttpJob() {
+    Cancellation.Cancel();
+    if (Ready()) {
+        Worker.join();
+    } else {
+        // DNS/connect/TLS setup does not observe cancellation. The worker owns
+        // all request data, so it can finish without keeping the provider alive.
+        Worker.detach();
+    }
+}
```
NewBediver, line 129: It’s not clear. If the job stops, the AI persistently suggests doing a detach, even though I’m just doing a join. How critical is it to do/not do a detach here?

ydb/public/sdk/cpp/src/client/types/credentials/oidc/provider_base.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+TCredentialsProviderAdapter::~TCredentialsProviderAdapter() {
+    Provider->Stop();
+    if (Worker.get_id() == std::this_thread::get_id()) {
+        Worker.detach();
+    } else {
+        Worker.join();
+    }
+    Provider->CancelDeliveries();
+}
```
NewBediver, line 339: The same question with detach here

ydb/public/sdk/cpp/src/client/types/credentials/oidc/credentials.cpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+TCredentialsProviderPtr TFactory::CreateProvider() const {
+    with_lock (Mutex) {
+        if (!Provider) {
+            Provider = CreateProviderImpl(std::weak_ptr<ICoreFacility>{}, true);
+        }
+        return Provider;
+    }
```
Shfdis: Provider initialisation should be non blocking

ydb/public/sdk/cpp/src/client/types/credentials/oidc/credentials.cpp
Shfdis: no, this is not. The client identity is used for process-wide caching of providers. So this method should return a string containing some real client identity for the cache to work as intended, for example username.
In case of OIDC this should be the sub claim, if it works the same way as JWT

ydb/public/sdk/cpp/src/client/types/credentials/oidc/private.cpp
```diff
@@ -0,0 +1,139 @@
...
+    const auto it = map.find(name);
+    return (it == map.end()) ? nullptr : &it->second;
+}
+
+ui64 Seconds(const NJson::TJsonValue& value, const std::string& field, bool allowZero) {
+    if ((!value.IsInteger() && !value.IsUInteger()) || (value.IsInteger() && value.GetInteger() < 0)) {
+        throw TError("invalid " + field, false, {});
+    }
+    const auto seconds = value.GetUInteger();
+    if ((!seconds && !allowZero) || seconds > std::numeric_limits<i64>::max() / 2 / 1'000'000) {
```
Shfdis: please explain. I don't get this restriction

ydb/public/sdk/cpp/src/client/types/credentials/oidc/private.cpp
```diff
@@ -0,0 +1,139 @@
...
+        throw TError("invalid " + field, false, {});
+    }
+    return seconds;
+}
+
+NUri::TUri ParseUrl(const std::string& value, bool issuer) {
+    const std::string role = issuer ? "issuer" : "endpoint";
+    NUri::TUri url;
+    if (value.empty() ||
```
Shfdis: this looks terrifying. Please split the lines. Mb use switch case

ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/types/credentials/oidc/credentials.h
```diff
@@ -0,0 +1,111 @@
...
+
+#include <util/datetime/base.h>
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <variant>
+#include <vector>
+
+namespace NYdb::inline Dev {
```
Shfdis: in this file there are NYdb:: TOauthToken, TTokenCache, and multiple other top-leve structs which are used only in OIDC contexts. Put them into OIDC namespace

Shfdis (commented):
also the provider is inherently overcomplicated:
 credentials adapter
  └── persistent provider worker thread
      └── per-request HTTP worker thread
          └── blocking TKeepAliveHttpClient

I'd much rather mimic existing oauth2 pattern: one worker performs synchronous HTTP and is joined during destruction. Accept that shutdown may wait for the configured HTTP timeout.
This approach also resolves the weird thread detachment on destruction

ydb/public/sdk/cpp/CHANGELOG.md
Shfdis: you need to add this. But above the previous version mark
* entry


## v3.23.0

ydb/public/sdk/cpp/src/client/types/credentials/oidc/private.cpp
NewBediver: Seconds are converted to signed chrono microseconds. Reserve half the range for deadline arithmetic and doubling the polling interval without overflow.

Also seconds can be empty (zero value), so we has function argument allowZero to ignore this or not.

ydb/public/sdk/cpp/src/client/types/credentials/oidc/protocol.cpp
```diff
@@ -0,0 +1,313 @@
...
+    try {
+        // Shutdown waits for this synchronous request. Keep HTTP cancellation
+        // subscriptions local to the request instead of retaining them until shutdown.
+        NThreading::TCancellationTokenSource requestCancellation;
+        TKeepAliveHttpClient client(host, url.GetPort(),
+            std::min(SocketTimeout, remaining), std::min(ConnectTimeout, remaining), false, false, true);
+        status = (form != nullptr)
+            ? client.DoPost(path, body.Print(), &response, headers, nullptr, requestCancellation.Token())
+            : client.DoGet(path, &response, headers, nullptr, requestCancellation.Token());
+        Cancellation.ThrowIfCancellationRequested();
```
Shfdis, line 134: don't we need to include port in host field?

ydb/public/sdk/cpp/include/ydb-cpp-sdk/client/types/credentials/oidc/credentials.h
```diff
@@ -0,0 +1,113 @@
...
+    TFlowConfig FlowConfig;
+
+    FLUENT_SETTING(std::shared_ptr<ITokenCacher>, Cacher);
+    FLUENT_SETTING(std::shared_ptr<IAuthAcceptor>, Acceptor);
+};
+
+void ValidateOidcConfig(const TOidcConfig& config);
+
+// Deterministic credential fingerprint; excludes cacher/acceptor instances.
+std::string GetOidcClientIdentity(const TOidcConfig& config);
```
Shfdis: do we actually need those methods to be public?

NewBediver: No, we don't. Moved to private.h

Shfdis (approved):

PR #53180, opened 2026-09-15
Speed up benchmark run history with SQLite indexing and pagination

kruall:
### Changelog entry 

Speed up benchmark run lists and comparison selection with server-side filtering and pagination.

### Description for reviewers 

Run-list requests previously scanned and parsed the complete file history. Add a rebuildable SQLite metadata index with incremental updates while keeping manifests and artifacts authoritative in files. Merge bounded cursor pages across hosts and preserve comparison selections across pages. Retain the legacy list API; paginated federation requires updated peers and reports partial failures without advancing past unavailable hosts.


mregrock (approved):
LGTM

PR #53058, opened 2026-09-14
Add durable operation UID admission and replay in SchemeShard

Enjection:
### Changelog entry

Add durable SchemeShard UID admission and replay for full backup, incremental backup, and restore operations.

### Description for reviewers

Retries after a lost response now return the retained operation ID. SchemeShard stores the UID and exact original DDL in the same local transaction that admits the operation; cleanup and Forget release the UID.

All seven UID-capable operation kinds share `TOperationUidAdmission`, `FindOperationByUid`, and one `OperationsByUid` hash map keyed by `(existing operation-ID kind, UID)`. The separate export, import, index-build, and SET NOT NULL UID indexes are removed. Existing operation records remain the source of persisted identity; their handlers own persistence and cleanup. Import's child-index progress and recovery lookups also use the shared index.

Legacy semantics are preserved: import/export replay with their existing domain checks, while index builds and SET NOT NULL reject duplicates. SET NOT NULL's UID remains memory-only and is lost on restart. The consolidation adds no persistent registry or migration of legacy records.

The shared capability registry covers all seven kinds, with optional SQL/KQP mappings. New scheme submissions use the legacy 128-byte UID limit and character policy, compare owner before exact DDL, and scope keys by SchemeShard tablet and operation kind. Their metadata has no database binding. Internal metadata uses the existing proposal messages. The unrelated restore fixes have already merged upstream; table creation and topic alteration are unchanged.

Validation on `57b14a918de`: **271 tests passed** with `hya make -T -ttt --build=relwithdebinfo`:

- Complete suites: `ut_full_backup` (72), `ut_backup_collection` (80), `ut_incremental_restore` (53), `ut_incremental_restore_reboots` (32), and `common/ut` (8).
- Legacy UID compatibility: export (1), import (2), index build (1), and SET NOT NULL (1).
- Indexed-import progress and completion (8), plus reboot and pipe-reset variants (12).
- SchemeShard schema golden test (1), against the rebuilt server.
- `git diff --check`: clean.

The two new regression variants give a full backup and an index build the same UID, restart SchemeShard, then forget either kind and verify that the other remains bound before and after another restart. The released UID can be reused independently. Existing tests also cover replay/body conflicts, ownership, concurrent admission, rollback, lookup misses, Forget races, and restore finalization.

[...]

PR #53245, opened 2026-09-16
fix: restore TxChangePathState after SchemeShard restart and roll back its Propose

Enjection:
### Changelog entry

Fix SchemeShard restart with an in-flight restore path-state operation and roll back staged restore path and transaction state when a proposal is rejected.

### Description for reviewers

`TxChangePathState` derives from `TSubOperationWithContext`, but restart reconstruction used its context-free factory and reached an aborting state-selection overload. Pass the operation context when reconstructing it.

Register path and new transaction state with the existing memory rollback mechanism in both `TChangePathStateOp::Propose` and the restore control-plane proposal. Otherwise a later rejected part leaves a changed path and orphan transaction state in memory.

The restart regression verifies a `TxChangePathState` row is persisted, reboots SchemeShard, then requires restore completion and normal path states for the restored table and both backup source tables. Two rollback variants stage each affected operation before rejecting a later part and check that the path and transaction state are restored.

Original validation: on upstream, the restart regression aborts in the context-free `SelectStateFunc`; both rollback variants leave the path in `EPathStateOutgoingIncrementalRestore`. With the fixes, all three full suites (`ut_backup_collection`, `ut_incremental_restore`, `ut_incremental_restore_reboots`) pass: 155 tests. `git diff --check` passes. All production edits are byte-identical to the corresponding hunks extracted from #53058.

Rebased onto upstream/main `e0bab8b3266`, including the merged logging build fix #53280. `git range-diff` confirms the PR patch is unchanged, and `git diff --check` passes. Previously, all three full suites passed (156 tests) with the exact logging correction now included upstream. CI validation of the new head is pending.


Enjection:
hmm, something went wrong on fresh main

CyberROFL (approved):

PR #53277, opened 2026-09-16
Wait for config v2 feature flag in ConfigGRPCService tests

kruall:
### Changelog entry 

Fix flaky configuration gRPC tests.

### Description for reviewers 

The dynamic config RPC can return SUCCESS before the gRPC node applies SwitchToConfigV2. A subsequent ReplaceConfig then reaches BSC without the enable switch and fails with UNSUPPORTED. Wait for the runtime flag in EnableConfigV2 before proceeding, with a bounded timeout.

Related mute issue: [#52240](https://github.com/ydb-platform/ydb/issues/52240).


mregrock (approved):
LGTM

PR #53280, opened 2026-09-16
fix: restore SchemeShard build after logging migration

Enjection:
### Changelog entry

Fix the SchemeShard build after the structured logging migration.

### Description for reviewers

#53163 added a `LOG_E` call in `CollectBackupTablePaths` after #53042 had been tested. Merging #53042 removed the macro definition but retained that new call, breaking the build on main with an invalid stream expression and an unused `ctx` parameter.

Convert the remaining call to `YDB_LOG_ERROR_CTX`, matching the surrounding logging code and using `ctx` explicitly. This is a one-line build fix with no restore behavior changes.

Validation: the failure is confirmed in both relwithdebinfo and release-asan CI on #53244. This exact correction previously allowed all three SchemeShard backup/restore suites to pass locally while validating #53244 and #53245. The standalone commit now passes `hya make --build=relwithdebinfo --no-bazel-remote-store ydb/core/tx/schemeshard` and `git diff --check`. The PR was opened before local verification to unblock other developers.


ijon (approved):

maximyurchuk (approved):

ydb/core/tx/schemeshard/schemeshard_incremental_restore_scan.cpp
```diff
@@ -611,7 +611,7 @@ class TSchemeShard::TTxProgressIncrementalRestore : public NTabletFlatExecutor::
                 std::pair<TString, TString> paths;
                 TString err;
                 if (!TrySplitPathByDb(tablePath, bcPath.GetDomainPathString(), paths, err)) {
-                    LOG_E("Failed to split backup table path: " << err);
+                    YDB_LOG_ERROR_CTX(ctx, "Failed to split backup table path", {"error", err});
```
CyberROFL, line 614:
```suggestion
                    YDB_LOG_ERROR_CTX(ctx, "Failed to split backup table path",
                        {"error", err});
```

PR #53288, opened 2026-09-16
Enable shared actor threads by default

kruall:
### Changelog entry 

Enable shared actor threads by default when using automatic actor system configuration.

### Description for reviewers 

Set `UseSharedThreads` to default to `true` and honor the default when selecting application pools. Preserve manual executor configurations and the explicit `UseSharedThreads: false` opt-out.


mregrock (approved):
LGTM

PR #53382, opened 2026-09-17
Fix races in scans in columnshards

kirillvasilenko:
# What

Makes the scans in columnshards race-free (more or less by design).

Fixes:
- YDBBUGS-686
- YDBBUGS-695
- YDBBUGS-713
- YDBBUGS-739

Deletes old hacks around races.

# How

Double strike does not let races any hope to get into the scans in Columnshards anymore.

## Strike 1: no concurrent access to a source anymore 

The whole idea is in ydb/core/tx/columnshard/engines/reader/common_reader/iterator/fetching.cpp
TStepAction::DoExecuteImpl()

Plus, we use TExecutionResult (can be Done or Pending) instead of bool as the return value from IFetchingStep.

### Before

```mermaid
sequenceDiagram
    participant A as TStepAction<br/>(conveyor worker A)
    participant S as IFetchingStep
    participant D as IDataSource
    participant W as async work<br/>(blob read, memory grant)
    participant B as TStepAction<br/>(conveyor worker B)

    A->>S: ExecuteInplace(source)
    S->>D: sync work on the source
    S->>W: start async work
    Note over W: the work may complete immediately
    W->>B: schedule continuation TStepAction
    par worker A is still inside ExecuteInplace
        S->>D: more sync work on the source
        S-->>A: return false (wait)
    and worker B already runs the continuation
        B->>S: ExecuteInplace(source)
        S->>D: sync work on the same source
    end
    Note over D: two workers access one source at the same time
```

What was IFetchingStep.ExecuteInPlace(source) doing before:
1. Did some local sync stuff on the source
2. Ran some async tasks (that were doing stuff on the source too!!!)
3. Did some more local sync stuff on the source (that were concurrent to the async tasks it ran at the 2d step)

So, the async tasks accessed the source concurrently with the sync stuff from step 3.

### After

```mermaid
sequenceDiagram
    participant A as TStepAction<br/>(conveyor worker A)
    participant S as IFetchingStep
    participant D as IDataSource
    participant J as IAsyncJob
    participant W as async work<br/>(blob read, memory grant)
    participant B as TStepAction<br/>(conveyor worker B)

    A->>S: ExecuteInplace(source)
    S->>D: sync work on the source
    S->>D: more sync work on the source
    S-->>A: TExecutionResult::Pending(job)
    Note over A: result.IsPending(), nothing else to do on the source
    A->>J: Start()
    J->>W: start async work
    A-->>A: return, worker A is done with the source
    W->>B: schedule continuation TStepAction
    B->>S: ExecuteInplace(source)
    S->>D: sync work on the source
    Note over D: one worker at a time, always
```

What do IFetchingStep.ExecuteInPlace(source) do now:
1. Do some local sync stuff
2. Do more local sync stuff
3. If it wants to run some async stuff it return IAsyncJob (that can run that async stuff) 
4. The client code sees that result.IsPending() (not Done), and runs it

So, no concurrent access to a source anymore.

## Strike 2:  Introducing concurrent access to IDataSource is impossible (more or less) anymore

The main idea is to use unique_ptr(TDataSourceLease) instead of shared_ptr(IDataSource).

It is unique_ptr(TDataSoucreLease) that gets passed around between fetching steps.

TDataSourceLease is here
ydb/core/tx/columnshard/engines/reader/common_reader/iterator/source.h:443

Why not just unique_ptr(IDataSource)? Plain reader works in "intervals" and a single data_source can be in several intervals simultaneously, so they do need to share a single source, but they need it read only.

So, if we use the same IDataSource in several places concurrently, it is shared_ptr(const IDataSource) now.

Most of the code now uses just IDataSource&.

As a bonus, I got rid of weak_ptrs we used to have here. Replaced them depending on the case with IDataSoucre& or unique_ptr(TDataSourceLease).

### Example of how it works in practice, TAllocateMemoryStep

```mermaid
sequenceDiagram
  participant step_action as TStepAction
  participant step as TAllocateMemoryStep
  participant job as TFetchingStepAllocation.TStartJob
  participant limiter as NGroupedMemoryManager::TMemoryLimiterActor
  participant allocation as TFetchingStepAllocation<br/>: NGroupedMemoryManager::IAllocation
  participant next as next TStepAction

  step_action ->> step: ExecuteInplace(IDataSource&, cursor)
  activate step
    Note over step: computes the size
    step -->> step_action: Pending(TStartJob(size, cursor, stage = FETCHING))
  deactivate step

  step_action ->> job: Start(std::move(SourceLease))
  activate job
    Note over job: builds TFetchingStepAllocation around the lease
    job ->> limiter: TEvStartTask(..., TFetchingStepAllocation, ...)
  deactivate job
  Note over step_action: returns false, the task is done with the source

  Note over limiter: waits until the FETCHING budget has room, then TAllocationInfo::Allocate
  limiter ->> allocation: OnAllocated(TAllocationGuard)
  activate allocation
    Note over allocation: RegisterAllocationGuard, Step.Next()
    allocation ->> next: TStepAction(std::move(SourceLease), std::move(Step))
    Note over allocation,next: the lease moves on, the allocation is finished
  deactivate allocation
```

# Checked

I ran 
- transfer/ut/column_table
- columnshard/ut_rw

under tsan twice, no sanitizer reports


ydb/tests/stability/tests/tsan.supp
```diff
@@ -1,3 +1,3 @@
 # https://github.com/ydb-platform/ydb/issues/41296
-# Benign races on IDataSource / TExecutionContext metrics: source processing hops conveyor threads.
-race:AddExecutionDuration
+# The races on IDataSource / TExecutionContext are fixed: a source is processed by exactly one owner at a time,
+# so no suppressions are needed for columnshard scans any more. The file stays because deployments reference it.
```
dorooleg: Давай сделаем его пустым

dorooleg (approved):

dorooleg (changes requested):
Там еще сборка упала

ydb/tests/stability/tests/tsan.supp
kirillvasilenko: Сделал

dorooleg (dismissed):

dorooleg (approved):

PR #53420, opened 2026-09-17
Use YQL type strings and bridge-only WASM UDF manifests

akulaad:
### Changelog entry

WASM UDF manifests now declare argument and result types as YQL type strings and use the bridge ABI for all calls, including objects and TypeConfig. This is a breaking change: legacy JSON type descriptors and `calling_convention` fields are rejected; existing manifests and guest modules must be migrated.

### Description for reviewers

Parse manifest signatures with the YQL type parser and preserve exact types, including nested Optional states and Decimal precision/scale. Validate supported bridge types, dictionary keys, scalar result kinds, nullability, and a maximum type nesting depth of 32 edges.

Use bridge handles throughout function and configured-object lifecycles, including lazy creation, compartment-generation changes, and cleanup after errors. Keep typed views distinct when Optional values share their payload's identity. Remove the obsolete ABI path and SDK definitions, and migrate the existing guest modules, example manifests, and documentation.

Based directly on upstream `main`. RPC/CLI, ReefProfile, compiler infrastructure, and the other integration-branch changes are outside this PR.

Validation:

- Both guest module roots successfully cross-compiled with `clang20-emscripten-wasm64` (`/tmp/manifest-main-guests.log`).
- Validated all 15 changed JSON manifests; `git diff --check` passed.
- Host configure and build completed successfully with 375 tests passing and 9 configure warnings (`/tmp/manifest-main-tests.log`).
- GitHub Actions checks are running on the upstream `main` baseline.


yql/essentials/ast/yql_type_string.cpp
```diff
@@ -206,6 +207,23 @@ class TTypeParser {
 
 private:
     TAstNode* ParseType() {
+        if (TypeDepths_.size() > MaxDepth_) {
```
akulaad: yql/essentials не трогаем

martinevsky (approved):

PR #53438, opened 2026-09-17
Fix erasure_species as number case in migrator

mregrock:
### Changelog entry 

Fix erasure_species as number case in migrator

### Description for reviewers 

...


Evgenik2 (approved):
LGTM

PR #53476, opened 2026-09-18
Allow deprecated gRPC client to send x-ydb-auth-ticket.

Alek5andr-Kotov:
### Changelog entry 

The deprecated gRPC client can now pass an authentication token using the `x-ydb-auth-ticket` header.

### Description for reviewers 

...


nshestakov (approved):

ydb/public/lib/deprecated/client/grpc_client.h
```diff
@@ -31,13 +31,15 @@ namespace NKikimr {
 
         class TGRpcClient {
             TGRpcClientConfig Config;
+            TString AuthToken;
```
nshestakov: This field is never read. `SwitchToLocation` takes the token from `TImpl::SecurityToken`, not from here. Please drop it from `TGRpcClient` and only keep the copy inside `TImpl`.

ydb/public/lib/deprecated/client/grpc_client.cpp
```diff
@@ -307,6 +321,11 @@ namespace NKikimr {
             return Config;
         }
 
+        void TGRpcClient::SetAuthToken(const TString& token) {
+            AuthToken = token;
+            Impl->SetAuthToken(token);
+        }
```
nshestakov, line 328: `AuthToken` on `TGRpcClient` is assigned here and never used. The write is also unsynchronized, so any future reader would be a data race.

Forward to `Impl` only:

```cpp
void TGRpcClient::SetAuthToken(const TString& token) {
    Impl->SetAuthToken(token);
}
```

ydb/public/lib/deprecated/client/grpc_client.cpp
```diff
@@ -207,6 +215,12 @@ namespace NKikimr {
                 return Channel->GetState(false);
             }
 
+void SetAuthToken(const TString& token) {
+    with_lock (Mutex) {
+        AuthToken = token;
+    }
+}
```
nshestakov: Please indent this method like the rest of `TImpl` — Copilot Autofix dropped it at column 0.

Locking only in the setter is correct: `Issue` / `IssueStream` / pending-queue drain already call `processor->Start()` under this `Mutex`. Do not also take `Mutex` inside `Start()` — `TMutex` is not recursive and that would deadlock.

ydb/public/lib/deprecated/client/grpc_client.cpp
```diff
@@ -75,6 +76,9 @@ namespace NKikimr {
 
             private:
                 void Start() override {
+                    if (!Impl->AuthToken.empty()) {
+                        Context.AddMetadata(NYdb::YDB_AUTH_TICKET_HEADER, Impl->AuthToken);
+                    }
```
nshestakov: Same metadata block is duplicated in `TStreamRequestProcessor::Start()`. Consider a single `TImpl` helper (e.g. `AddAuthMetadata(grpc::ClientContext&)`) and call it from both `Start()` methods.

ydb/public/lib/deprecated/kicli/kikimr.cpp
```diff
@@ -311,6 +316,9 @@ class TKikimr::TGRpcImpl : public TKikimr::TImpl {
         }
         config.Locator = newLocation;
         TAutoPtr<NGRpcProxy::TGRpcClient> gRpcClient = new NGRpcProxy::TGRpcClient(config);
+        if (!SecurityToken.empty()) {
+            gRpcClient->SetAuthToken(SecurityToken);
+        }
         gRpcClient.Swap(GRpcClient);
```
nshestakov, line 322: Nit / residual race: `SetSecurityToken` updates the current `GRpcClient`, while `SwitchToLocation` builds a new client and copies `SecurityToken` with no shared lock. A token refresh concurrent with `ChooseProxy` can leave the new client with a stale or empty header.

The protobuf token path was already unsynchronized, so this is not a first-order regression. If logbroker rotates the token on a live client, the copy onto the new client should be atomic with the swap.

nshestakov (changes requested):
The idea is right: the deprecated gRPC client should send `x-ydb-auth-ticket` the same way the SDK does. Unary, streaming, and endpoint switching are covered, and after the autofix the token write is under the same `Mutex` that `Start()` already runs with.

Please fix the unused outer `AuthToken` field and the broken indentation from Copilot Autofix before merging. Other notes are nits.

ydb/public/lib/deprecated/client/grpc_client.cpp
Alek5andr-Kotov: исправил в https://github.com/ydb-platform/ydb/pull/53476/commits/2194a38483c7bf2ccea842192e65df58f0da42e1

ydb/public/lib/deprecated/client/grpc_client.cpp
Alek5andr-Kotov: исправил в https://github.com/ydb-platform/ydb/pull/53476/commits/2194a38483c7bf2ccea842192e65df58f0da42e1

ydb/public/lib/deprecated/client/grpc_client.cpp
Alek5andr-Kotov: исправил в https://github.com/ydb-platform/ydb/pull/53476/commits/2194a38483c7bf2ccea842192e65df58f0da42e1

ydb/public/lib/deprecated/client/grpc_client.h
Alek5andr-Kotov: исправил в https://github.com/ydb-platform/ydb/pull/53476/commits/2194a38483c7bf2ccea842192e65df58f0da42e1

ydb/public/lib/deprecated/kicli/kikimr.cpp
Alek5andr-Kotov: http://st/LOGBROKER-10669

nshestakov (approved):

PR #53480, opened 2026-09-18
Support id range expiration while agent was out

alexvru:
### Changelog entry 

Support id range expiration while agent was out

### Description for reviewers 

Subj.


ydb/core/blob_depot/group_metrics_exchange.cpp
```diff
@@ -130,6 +130,8 @@ namespace NKikimr::NBlobDepot {
     void TBlobDepot::UpdateThroughputs(bool reschedule) {
         static constexpr TDuration Window = TDuration::Seconds(3);
 
+        UpdateAgentsBlockingGC();
```
mregrock, line 133: This method is called for every TEvPushMetrics message and scans all agents. With N agents, this results in O(N^2) background work on the tablet, even without user activity. Consider moving this call inside if (reschedule) so it only runs once every 3 seconds.

ydb/core/blob_depot/agent.cpp
```diff
@@ -291,6 +342,98 @@ namespace NKikimr::NBlobDepot {
...
+            Y_ABORT_UNLESS(channel.NextBlobSeqId);
+
+            // Everything this agent could still be holding on this channel -- free ids and writes in flight alike --
+            // was issued below NextBlobSeqId, so ordering it to drop that whole step range covers all of it.
+            ui32& step = agent.ExpiredSteps[channelIndex];
+            step = Max(step, TBlobSeqId::FromSequentalNumber(channelIndex, generation, channel.NextBlobSeqId - 1).Step);
+
+            // Ids issued from now on have to survive that trim, exactly as TData::HandleTrash arranges when it
+            // invalidates a step; otherwise a freshly allocated id could share the step we just told them to drop.
+            auto next = TBlobSeqId::FromSequentalNumber(channelIndex, generation, channel.NextBlobSeqId);
```
mregrock, line 418: Copypaste with https://github.com/ydb-platform/ydb/blob/35cbfd6d5e95a1b14ebd345199500309ab49a602/ydb/core/blob_depot/data_trash.cpp#L100

ydb/core/blob_depot/agent.cpp
```diff
@@ -291,6 +342,98 @@ namespace NKikimr::NBlobDepot {
...
+        std::vector<ui32> expired;
+
+        for (auto& [nodeId, agent] : Agents) {
+            if (agent.Connection) {
+                continue; // it is back
+            } else if (!agent.SupportsIdRangeExpiry) {
+                continue; // we have no way to make this agent drop the ids, so we must keep them reserved
+            }
+
+            bool holdsRanges = false;
```
mregrock, line 370: Consider extracting TAgent::HasGivenIdRanges() to share this check with UpdateAgentsBlockingGC()

alexvru: I agree with the intent, but I’d classify it as a low-priority cleanup.

PR #53503, opened 2026-09-18
NBS-7459: report belated writes when the quorum is not reached

BarkovBG:
A write that lands on a PBuffer after the client has been answered was reported to the vchunk only when the quorum had been reached. If the quorum was never reached, the copy still ends up on the PBuffer and nobody learns about it, so it is not erased until a cleanup barrier covers it.

Such a write is now reported as belated in both cases. Groundwork for the follow-up PR that keeps the record in the dirty map until every host has answered.


ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/write_request.cpp
```diff
@@ -166,11 +166,7 @@ void TWriteRequestExecutor::OnIndirectWriteResponse(
 
     CompletedWrites = CompletedWrites.Include(completedWritesOfCurrentResponse);
 
-    if (IsQuorumReached()) {
-        ReplyOrNotifyBelated(MakeError(S_OK), completedWritesOfCurrentResponse);
-        return;
-    }
-
+    ReplyOrNotifyBelated(completedWritesOfCurrentResponse);
     SendAdditionalDirectWrites();
```
drbasic: а зачем SendAdditionalDirectWrites() если ответили на запрос?

vazhem (commented):
## AI review

`ReplyOrNotifyBelated` now reports a successful PBuffer copy as belated whenever the client has already been answered, not only when that copy completed a quorum. That matches `WriteFinished`: a failed write is forgotten, so a later ack has to go through the belated erase queue or the copy sits until barrier GC. The DirectWrite timeout test covers the ticket scenario.

Copies already in `CompletedWrites` at the error reply still take the forget-and-barrier path. `OnIndirectWriteResponse` is half the patch and has no matching test. The removed early return after quorum on the indirect path is a no-op: `SendAdditionalDirectWrites` returns when `IsReplied`.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/write_request.cpp
```diff
@@ -511,7 +509,11 @@ void TWriteRequestExecutor::OnRequestTimeout()
         LogTitle.GetWithTime().c_str(),
         ExtendedDebugState().c_str());
 
-    ReplyOrNotifyBelated(MakeError(E_TIMEOUT, "Write request timeout"), {});
+    if (IsReplied) {
+        return;
+    }
+
+    Reply(MakeError(E_TIMEOUT, "Write request timeout"));
```
vazhem, line 520: **AI review (Minor):** error reply does not belated-notify CompletedWrites

Timeout (and the other error `Reply`s: quorum unreachable) still hand already-confirmed hosts only to `WriteFinished`. That function `RemovePBuffer`s when `confirmed < quorum`, and `UpdateBelatedEraseQueue` never sees those hosts. Only acks after `IsReplied` call `NotifyBelated`.

A timeout with one OK and then one late OK therefore erases only the late copy promptly; the copy that won the race with the timer waits for barrier GC. The old timeout path passed `{}` into `ReplyOrNotifyBelated`, so this is not a regression, but it is the same leak this PR is closing for post-reply acks.

After `Reply(E_TIMEOUT)` (and the quorum-unreachable `Reply`s), `NotifyBelated(CompletedWrites)` would put the already-confirmed hosts on the belated queue. If that is intentionally #53505, say so here.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/write_request_ut.cpp
```diff
@@ -97,6 +97,37 @@ Y_UNIT_TEST_SUITE(TWriteRequestTest)
...
 
+    Y_UNIT_TEST_F(
+        ShouldNotifyBelatedWriteWithoutQuorum,
+        TWriteRequestTestFixture)
+    {
+        Init();
+
+        auto writeRequest = CreateRequestExecutor(
+            MakeWriteTestRequestHeaders(Range, BlockSize),
+            EWriteMode::DirectWrite);
```
vazhem, line 108: **AI review (Minor):** add an IndirectWrite belated test

The new test only runs `EWriteMode::DirectWrite`. The other half of the patch is `OnIndirectWriteResponse`: it always calls `ReplyOrNotifyBelated` and no longer returns early on quorum.

Timeout, then `ManyPBufferCallback(CreateOneOkResponse(THostIndex{1}))`, and the same `AllCompletedWrites` asserts would cover that path.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/write_request.cpp
BarkovBG: https://github.com/ydb-platform/ydb/blob/4d7fb068584972e0102c079896a33bc551c4a73a/ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/write_request.cpp#L177-L200

BarkovBG: чуть порефакторил код чтобы было понятнее

drbasic: ReplyOrNotifyBelated но получается что ReplyOrNotifyBelated может не сделать ни Reply ни NotifyBelated

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/write_request.cpp
BarkovBG: это решает другой пр https://github.com/ydb-platform/ydb/pull/53505

но он еще не готов к ревью

drbasic (dismissed):

drbasic (dismissed):

drbasic (approved):

PR #53516, opened 2026-09-18
Fix Hive TTxKillNode performance on large clusters

UgnineSirdis:
### Changelog entry 

...

### Description for reviewers 

Problem. `TObjectDistributions::UpdateDistribution` and its use of the `SortedDistributions` set were at the top of the profile of the `TTxKillNode` transaction, even when the killed node had no tablets. On a cluster with 2500 nodes, killing an empty node was slow. This is a common case for NBS, which has most of the nodes without tablets.

Cause. `TTxKillNode` calls `ObjectDistributions.RemoveNode`, which visits every object distribution in the Hive. Every registered node is present in every matching distribution with count 0, so the walk covers all objects. For each object, UpdateDistribution unconditionally extracted the object from the `std::multiset<TObjectDistribution>` and reinserted it, two O(log M) tree operations whose comparator recomputed `GetImbalance()` on every step. This reordering was almost always wasted: removing a zero-count node does not change an object's imbalance while other zero-count nodes remain.

Fix. `SortedDistributions` is now a `std::multimap<double, TObjectDistribution>` keyed by the object's imbalance as of its last insertion. `UpdateDistribution` mutates the object in place, compares the new imbalance with the cached key, and only extracts and reinserts when the value actually changed. `AddNode` and `RemoveNode` pass the iterator directly instead of looking each object up by id again. The `ImbalancedObjects` counter and the balancer's `GetMaxImbalance / GetObjectToBalance` are unaffected.


vporyadke (approved):

qkrorlqr (changes requested):
having a nonempty Description would've been really helpful

CyberROFL (approved):

PR #53653, opened 2026-09-21
[NBS-7655] Remove watermarks and ahead state

drbasic:
### Changelog entry 

Use Behind as the sole source of DDisk freshness, with reads limited to the prefix before the first behind range. Remove obsolete watermark/ahead fields, persist vChunk config and dirty-map state atomically, and show current Fresh state in Overview. Compatibility with the old persisted format is not preserved.


ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/vchunk.cpp
```diff
@@ -1139,61 +1156,81 @@ void TVChunk::PersistNextPendingConfig()
...
-    Y_ABORT_UNLESS(pending.Config.IsValid());
+    Y_ABORT_UNLESS(config.GetVChunkIndex() == VChunkConfig.GetVChunkIndex());
+    Y_ABORT_UNLESS(config.IsValid());
 
-    auto onPersisted =
-        PartitionDirectService->UpdateVChunkConfig(pending.Config);
+    const ui32 stateGeneration = BlocksDirtyMap->GetCurrentGeneration();
+    auto dirtyMapState = BlocksDirtyMap->MakeFutureState(config, IsTouched());
+    const THostMask freshDDisks = GetFreshDDisks(dirtyMapState);
+    auto onPersisted = PartitionDirectService->UpdateVChunkState(
```
vazhem, line 1201: **AI review (Major):** Config persist is not exclusive with dirty-map persist

`PersistNextPendingConfig` sends `UpdateVChunkState(config, MakeFutureState(...))` without taking `DirtyMapStatePersisting`. `DoPersistDirtyMap` only looks at that flag, and flush / copy-progress / cleanup still call it while a config persist is in flight.

`MakeFutureState` is a prediction against the *current* map. Until `OnConfigPersisted` runs `ApplyConfig`, the new DDisk is not in `DesiredDDisks` and `GetStateForPersist()` will not save its Behind (and returns an empty proto when no current DDisk is Fresh). A `DoPersistDirtyMap` snapshot taken in that window is therefore the pre-promotion map.

`OnConfigPersisted` then calls `StatePersisted` and `ApplyConfig`, but not `DoPersistDirtyMap`. `TDDiskState::Init` never calls `OnBehindChanged`, so adding a full-Behind disk does not raise `StateGeneration`. After a later dirty-map completion does `StatePersisted(Max(...))`, `NeedPersist()` is false and the wrong proto is left in the local DB.

Serialize one persist at a time for the vChunk (config or dirty map), and after `ApplyConfig` persist again if in-memory Behind is not the proto that just committed. `Init` of a non-empty Behind should bump the generation the same way `AddBehind` does, so `CheckEraseAbility`'s `StateGeneration == 0` shortcut cannot treat a newly promoted disk as “no red blocks”.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct_tablet/part_updatevchunkstate.cpp
```diff
@@ -0,0 +1,76 @@
...
+    TTxPartition::TUpdateVChunkState& args)
+{
+    Y_UNUSED(ctx);
+
+    TPartitionDatabase db(tx.DB);
+    for (const auto& request: args.UpdateStateRequests) {
+        if (!request.VChunkConfig.Empty()) {
+            db.StoreVChunkConfig(request.VChunkConfig);
+        }
+        db.StoreDirtyMapState(request.VChunkIndex, request.DirtyMapState);
```
vazhem, line 41: **AI review (Major):** Last `StoreDirtyMapState` wins and can drop new-disk Behind

`ExecuteUpdateVChunkState` always `StoreDirtyMapState` for every request, in queue order. Sharing one queue does not protect the combined update: a following dirty-map-only request for the same vChunk overwrites it, including an empty proto.

That is the tablet-side half of the race with `DoPersistDirtyMap` during config persist, and the batching test actually enqueues `UpdateDirtyMapState(vChunk 0)` behind `UpdateVChunkConfig(vChunk 0)`. Coalesce per vChunk in the pending list: if any request carries a config, do not let a later config-less snapshot replace its dirty-map payload. A config-less empty proto meaning “all repaired” is only safe when that vChunk’s in-memory map has already applied the config.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/dirty_map/dirty_map.cpp
```diff
@@ -39,6 +40,9 @@ TBlocksDirtyMap::TBlocksDirtyMap(
         DDiskStates.emplace_back(ArenaAllocator, blockCount);
     }
 
+    for (THostIndex host: DesiredDDisks) {
```
vazhem, line 43: **AI review (Minor):** Load of empty/short Behind cannot reconstruct a promoted DDisk

The constructor now sets `DesiredDDisks` from the loaded config, `Init`s those disks as fully Operational, then `UpdateConfig` sees no `added` disks. `TDDiskState::Load` always replaces Behind, including with empty. Together, empty-or-short persisted state cannot reconstruct a DDisk that was promoted in the config; it comes up Operational.

That encoding is consistent with “empty proto means nothing to repair”, which is why the persist clobber is fatal rather than self-healing. Either keep persist exclusive so a promoted disk’s Behind cannot disappear, or treat a config DDisk with no Behind entry as fully behind when the vChunk is touched (the old watermark-0 path).

vazhem (commented):
## AI review

The behind-only model is the right shape: one map, a readable prefix, copy completion as a dirty-map persist rather than a config watermark, and Fresh in Overview taken from live dirty-map stats. Combining vChunk config and Behind into one tablet transaction is also the right durability story.

That story is not closed. `PersistNextPendingConfig` and `DoPersistDirtyMap` can both be in flight, the tablet queue last-write-wins per vChunk, and `GetStateForPersist()` of the *current* map does not include a DDisk that exists only in `MakeFutureState`. A later dirty-map-only update can replace the predicted Behind with an empty slot. `TDDiskState::Init` does not bump `StateGeneration`, so `NeedPersist()` stays false and the clobber is never rewritten. After restart the constructor treats every disk already present in the loaded config as Operational, so the new DDisk is served as healthy with no data.

vazhem (approved):

PR #53672, opened 2026-09-21
Validate YAML configurations by related sections

Enjection:
### Changelog entry

Speed up YAML configuration validation by resolving distinct combinations of related sections.

### Description for reviewers

Merge equal intermediate states while preserving selector correlations and incompatibility rules. Binary decision diagrams share label-value branches and represent complements by an edge bit. Indexed equality matching avoids scanning the domain for every selector. Periodic reclamation removes unreachable regions and stale operation-cache entries.

Resolve coupled validators and shared transform inputs together, including feature flags with gRPC settings. Custom validators receive complete configurations by default. Preserve database policy and field diagnostics. Validation and resolution failures propagate without retrying full enumeration; the first reported error may differ from legacy ordering.

For one label with N equality selectors converging to the same configuration, region processing takes expected O(N log N) time and O(N) peak memory. General cost depends on intermediate configuration states and decision-diagram size; there is no general polynomial bound.

Validation:

- 223 YAML tests passed in optimized and AddressSanitizer builds. Coverage includes legacy acceptance/rejection parity, shared inputs, cross-section validation, unknown fields, detached-node lifetime, 10,000 selectors writing either the base or a common non-base value, reclamation with correlated labels and incompatibility rules, and exclusion of unused binary domain codes.
- 22 console regressions passed, covering YAML validation, diagnostics, and database-selector policy.
- Non-base convergence benchmark: 1,000 selectors used 38 MiB peak RSS and 0.022 s validation time; 10,000 used 120 MiB and 0.288 s. Shuffled order at 10,000 used 120 MiB and 0.336 s. RSS includes the input YAML, parsing, diagnostics and validation.
- Five separate-process runs per production configuration, including reading, parsing, diagnostics, and validation: original 0.390 s median / 0.396 s maximum; two additional independent labels 0.429 s median / 0.447 s maximum. All accepted.
- Production config unchanged; its paths and distinctive name fragments are absent from the PR.


ydb/library/yaml_config/yaml_config.h
```diff
@@ -148,4 +153,10 @@ class IConfigSwissKnife {
 
 std::unique_ptr<IConfigSwissKnife> CreateDefaultConfigSwissKnife();
```
CyberROFL, line 154: Оффтоп, но может переименовать это в `CreateDefaultConfigValidator`?

ydb/library/yaml_config/public/yaml_config.cpp
```diff
@@ -1358,6 +1359,283 @@ void ResolveUniqueDocs(
...
+        if (value < 2) {
+            return 1 - value;
+        }
+        if (auto it = Negations.find(value); it != Negations.end()) {
+            return it->second;
+        }
+        // Recursion can reallocate Nodes.
+        auto node = Nodes[value];
+        TMap<size_t, TId> children;
+        for (size_t i = 0; i < Labels[node.Label].second.size(); ++i) {
```
CyberROFL: Пишут, что тут N^2 памяти:
> Not() expands an equality selector’s complement into an explicit child for every other label value. EnumerateDistinctProjections() computes this complement for every selector, and Intern() retains each child map in both Unique and Nodes.
>
> Consequently, one label with N values and N equality selectors consumes Θ(N²) memory—even when every selector redundantly writes the base configuration and only one configuration state exists. With 10,000 values, this retains approximately 100 million complement edges
twice. This introduces a substantial memory regression over legacy enumeration and can exhaust console memory.
>
> Represent complements using complemented edges or a default child with sparse exceptions. Add coverage for many values of one label; the existing scaling test covers many labels with small domains.

Enjection: Я улучшил это место, но там кажется ещё можно выжать, может потом отдельным PR/

ydb/library/yaml_config/public/yaml_config.cpp
```diff
@@ -1358,6 +1359,309 @@ void ResolveUniqueDocs(
...
+                const auto child = it != children.end() && it->first == branch ? (it++)->second : defaultChild;
+                if (child != canonicalDefault) {
+                    exceptions[branch] = child;
+                }
+            }
+            children = std::move(exceptions);
+            defaultChild = canonicalDefault;
+        }
+        auto key = std::tuple{label, defaultChild, children};
+        auto [it, inserted] = Unique.emplace(std::move(key), Nodes.size());
```
CyberROFL: > With one label, N equality selectors, base value base, and every selector writing selected, only two configurations remain active. Nevertheless, Unique and Nodes retain every growing prefix of selected label values and its complement. Their exception maps require Θ(N²)
total storage.
>
> The 10,000-selector test still writes base in every overlay, so it avoids this case. The new lifetime regression uses only two selectors and does not address memory growth.
>
> Reclaim obsolete regions and cache entries, or structurally share branch storage. Add scaling coverage where every selector writes the same value different from the base

CyberROFL (commented):
Замечание ниже выглядит валидно: это деградация по памяти (O(N) -> O(N^2)).

ydb/library/yaml_config/public/yaml_config.cpp
Enjection: Относительно исходного скорее не деградация, с полным резолвом, но в целом выглядит не очень, согласен

CyberROFL (approved):

PR #53710, opened 2026-09-21
[NBS-7661] Add resize implementation

vazhem:
### Changelog entry 

...

### Description for reviewers 

[NBS-7661] Add resize implementation


vazhem (commented):
## AI review

Claim, then persist, then reply OK, then poison is the right order for the fast path: the new block count is picked up on the next boot, and a same-size update does not restart the tablet. Two holes should be fixed before merge.

The grow claim is a `Queries` allocate. BSC turns that into `DefineDirectBlockGroup` with `NumDDisks` taken from the pool's fail-domain count. After add-host or remove-host that count is no longer the group, so the resize releases extra disks or allocates new ones, and the success path never reads the listing back.

A failed or busy claim replies `ERROR_UPDATE_IN_PROGRESS` and drops the operation. SchemeShard leaves the alter in `ConfigureParts` and does not send the config again, so the resize waits until SchemeShard restarts. The new unit tests resend from the edge actor, which is not that path.

Copilot's note on `FinishVolumeGrow` is still open. `DetachStorage` leaves the vhost endpoint registered, and the next `StartEndpoint` only reattaches, so the device keeps the old block count. `ShouldRestartTabletAndEndpointOnResize` records the `StartEndpoint` argument and then writes through the load-actor adapter, so it does not catch that.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct_tablet/volume_grow.cpp
```diff
@@ -0,0 +1,298 @@
...
+        ctx,
+        NKikimrServices::NBS_PARTITION,
+        "%s Growing %lu -> %lu blocks",
+        LogTitle.GetWithTime().c_str(),
+        currentBlockCount,
+        newBlockCount);
+
+    auto request = std::make_unique<
+        TEvBlobStorage::TEvControllerAllocateDDiskBlockGroup>();
+    request->Record = MakeDDiskClaimRecord(
```
vazhem, line 222: **AI review (Major):** grow claim rewrites the group's disk count

`MakeDDiskClaimRecord` sends `Queries`. In `TTxAllocateDDiskBlockGroup::Transform` each query becomes `DefineDirectBlockGroup` with `NumDDisks` and `NumPersistentBuffers` set from the pool's `NumFailDomainsPerFailRealm`, not from the group. The apply loop then walks `Max(numDDisks, record size)`: indexes past `numDDisks` get 0 chunks and are released, empty indexes below `numDDisks` allocate a new DDisk.

That matches a group only while it still has exactly the pool layout. Add-host grows `NumDDisks` to `LiveHostCount + 1`. Remove-host `DeleteDDisks` packs the BSC record down to the live disks (the tablet keeps the removed slot). A later resize therefore drops the added hosts or recreates removed ones. `HandleGrowAllocateResult` on OK ignores `Responses` and persists the new `VolumeConfig` anyway, so the tablet's connections stay on the pre-resize listing.

Keep `Queries` for the first allocate. For a grow, send one `DefineDirectBlockGroup` per DBG whose `NumDDisks` and `NumPersistentBuffers` are that group's current live counts and whose `NumChunksPerDDisk` is the new target — the add-host request shape, without changing the host count. Validate the listing the way `HandleAddHostAllocationResult` does before `TStoreVolumeConfig`.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct_tablet/volume_grow.cpp
```diff
@@ -0,0 +1,298 @@
...
+            LogTitle.GetWithTime().c_str(),
+            NKikimrProto::EReplyStatus_Name(record.GetStatus()).c_str(),
+            record.GetErrorReason().c_str());
+        ReplyUpdateVolumeConfig(
+            ctx,
+            VolumeGrowInFlight->ReplyTo,
+            VolumeGrowInFlight->Cookie,
+            VolumeGrowInFlight->TxId,
+            NKikimrBlockStore::ERROR_UPDATE_IN_PROGRESS);
+        VolumeGrowInFlight.reset();
```
vazhem, line 265: **AI review (Major):** IN_PROGRESS drops the grow and SchemeShard will not retry

On any non-OK allocate result this replies `ERROR_UPDATE_IN_PROGRESS` and clears `VolumeGrowInFlight`. The same reply is returned when an add or a remove is already in progress, and that request is not saved.

SchemeShard accepts only `OK` and `ERROR_UPDATE_IN_PROGRESS`. On `IN_PROGRESS`, `TConfigureParts::HandleReply` returns without `ActivateTx` and without sending the config again; the comment there is that the tablet will finish the alter later. The volume actor forwards the status once and erases the request. A partition restart does not help either: the in-flight grow was only in memory, and SchemeShard already has its answer.

`ShouldRetryGrowAfterBscAllocateFailure` and `ShouldRetryGrowAfterBscPipeConnectFailure` resend `UpdateVolumeConfig` from the test edge. SchemeShard will not. Keep the grow (the way add-host persists its intent and replays it from `CompleteLoadState`) and reply `OK` on the saved sender only after the claim and `TStoreVolumeConfig` have both committed. `IN_PROGRESS` is a progress report, not a terminal status, unless that later `OK` is sent without a new request.

PR #53713, opened 2026-09-21
wlm: decouple wlm from kqp_proxy

buhtr:
### Description for reviewers 

 Extract the WLM classifier state out of `TKqpProxyService::ResourcePoolsCache` into a per-node actor `TResourcePoolsCacheActor` with a synchronously-lookup-able `IGateway`. Unblocks BulkUpsert and other non-KQP callers that need classification but cannot subscribe as actors.


ydb/services/workload_manager/gateway/gateway.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+namespace {
+
+struct TGateways {
+    std::weak_ptr<NPrivate::TWorkloadManagerGateway> Default;
+    TMutex Lock;
+    std::unordered_map<ui32, std::weak_ptr<NPrivate::TWorkloadManagerGateway>> ByNodeId;
+};
+
+TGateways Gateways;
```
Hor911, line 71: Global vars are a very bad practice

PR #53749, opened 2026-09-22
[NBS-7705] Promote least loaded host

drbasic:
### Changelog entry 

1. VChunk should promote the least loaded host.
2. Show DDisk placement imbalance in DBG config table

### Description for reviewers 

...


ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/vchunk.cpp
```diff
@@ -1211,6 +1211,7 @@ void TVChunk::PersistNextPendingConfig()
         (const TPersistResultFuture& f) mutable
         {
             if (f.GetValue() != EPersistResult::Success) {
+                // Config persistence must never fail.
```
vazhem, line 1214: **AI review (Minor):** Release the DDisk reservation when persist does not finish

`AllocateDDiskForPromote` records the host before the config persist finishes, and `CommitDDiskPromotion` runs only from `OnConfigPersisted`. This callback returns on any result other than `Success`, including `EPersistResult::Cancelled` (fast-path stop and partition delete). The interface comment on `CommitDDiskPromotion` says that path releases the reservation.

The host stays in `PendingDDiskAllocations` for the rest of the DBG lifetime. Other vchunks of that DBG, including other regions, keep counting it. `AllocateDDiskForPromote` also aborts when the vchunk index is already present, so a later promote of the same index crashes the tablet once `Persisting` is clear or the vchunk exists again.

The success path has the same hole. If `weakSelf` has expired, `OnConfigPersisted` is skipped. `TRegion::Stop` completes without waiting for this persist and then drops the vchunk.

Release the reservation on both exits. On failure, hop to the DBG executor and call `CommitDDiskPromotion` there.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/mon_page/mon_render_overview.cpp
```diff
@@ -172,17 +239,24 @@ TDbgConfigTableData TDbgConfigTableData::BuildAndFill(
 {
     TDbgConfigTableData result;
     result.DefaultConfigs.resize(dbgs.size());
+    result.DDiskCounts.resize(dbgs.size());
     for (const auto& dbg: dbgs) {
         Y_ABORT_UNLESS(dbg.Index < dbgs.size());
         result.Headers[dbg.Index % VChunkPerRegionCount].DbgIds.push_back(
             dbg.Index);
         result.DefaultConfigs[dbg.Index] = BuildDefaultConfigCache(dbg);
+        result.DDiskCounts[dbg.Index].Configured.resize(dbg.Connections.size());
```
vazhem, line 248: **AI review (Minor):** Imbalance histogram counts removed host slots

`DDiskCounts` has one bin per `dbg.Connections` entry. `DoBuildMonSnapshot` emits every `GetSlotCount()` slot, and `MarkSlotDead` leaves that slot in the list until the next load compacts it. After vchunks demote the removed host, its bin is zero. `AllocateDDiskForPromote` will not select it, because the host is disabled. `CalculateDDiskImbalance` still targets an even spread over that bin, so the header keeps reporting moves onto a slot that cannot take them.

A newly added live host is a real zero and should stay in the histogram. Leave dead slots out of the counts. `TConnectionSnapshot` does not say which slots are dead, so the snapshot needs that bit. The `hostCount < DirectBlockGroupHostCount` guard does not cover this: a removed slot makes the vector longer.

The cell text is the touched percentage only (`Imb: N%`). Label it. The tooltip is the only place that distinguishes Config from Touched.

vazhem (commented):
## AI review

Promotion reserves the least-loaded enabled host on the DBG and keeps that host in the load total until the new config is applied. The concurrent test matches the accounting: an existing vchunk already sits on host 3, so the first promote takes host 4, the second takes host 3, and after the first commit a third promote takes host 4 again. The move-count formula matches the overview cases (6 of 15, 3 of 9, 5 of 15).

The reservation is not released when persist does not call `OnConfigPersisted`. The overview histogram also counts removed connection slots, so after remove-host it keeps asking to move DDisks onto a slot the allocator will not use. The reservation rollback belongs in this PR.

vazhem (approved):

PR #53755, opened 2026-09-22
merge storage issues relating to dc outage

vporyadke:
### Changelog entry 

...

### Description for reviewers 

[YDBRELIABILITY-66](http://st/YDBRELIABILITY-66)


ydb/core/health_check/health_check.cpp
```diff
@@ -3124,15 +3136,75 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
...
+            for (const auto& vDiskId : record.IssueLog.location().storage().pool().group().vdisk().id()) {
+                if (TString key = GetFailRealmKey(vDiskId)) {
+                    ++failedDisksInRealm[key];
+                }
+            }
+        }
+        auto hasLostFailRealm = [&](const TSelfCheckContext::TIssueRecord& record) {
+            const auto& vDiskIds = record.IssueLog.location().storage().pool().group().vdisk().id();
+            return AnyOf(vDiskIds, [&](const TString& vDiskId) {
+                return failedDisksInRealm[GetFailRealmKey(vDiskId)] > 1;
```
CyberROFL, line 3186: > The predicate treats any two failing VDisks in one fail realm as a lost data center, regardless of erasure type. A normal block-4-2 group has one fail realm containing all eight VDisks. Consequently, two isolated disk failures on different nodes in the same data center enter lostRealmRecords, merge into one issue, and lose their node locations—even though the data center remains operational. This changes ordinary failure reporting beyond the outage case covered by the new tests.
>
> Gate this classification on mirror-3-dc, using the group’s erasure metadata, and add a block-4-2 regression case that preserves separate node locations.

ydb/core/health_check/health_check.cpp
```diff
@@ -3233,18 +3327,26 @@ class TSelfCheckRequest : public TActorBootstrapped<TSelfCheckRequest> {
...
                 handled.splice(handled.end(), similar, similar.begin());
             }
         }
         records.splice(records.end(), handled);
     }
 
+    void MergeLevelRecords(TMergeIssuesContext& context, TList<TSelfCheckContext::TIssueRecord>& records) {
+        TList<TSelfCheckContext::TIssueRecord> lostRealmRecords;
+        ExtractLostFailRealmRecords(records, lostRealmRecords);
+        MergeSimilarRecords(context, lostRealmRecords, EDiskMergeScope::DataCenter);
```
CyberROFL: > The new data-center merge exposes a broken reason chain when groups share a failed PDisk. For example, group A loses disks on P1 and P2 in one realm, while group B loses only its disk on P1. A’s VDisk issues merge across nodes; B’s remains separate.
>
> The subsequent PDisk pass processes B first, moving P1’s records behind P2. Processing A then merges P1 into P2 and deletes P1’s issue without updating B’s reference. RemoveUnlinkIssues removes that dangling reference, leaving B’s VDisk issue without its PDisk cause. Previously, the per-node VDisk merge kept the shared P1 failures together.
>
> Retain PDisk issues referenced by other parents, or redirect all affected references when merging. Add a mixed-failure regression test with PDisk information—the new helper explicitly clears it.

PR #53769, opened 2026-09-22
Add ICB control enabling unlimited viewer trace verbosity

serbel324:
### Changelog entry 

Add an ICB atomic to disable hard limit for the tracing detalization level of the viewer queries

### Description for reviewers 

...


ydb/core/viewer/json_pipe_req.cpp
```diff
@@ -168,19 +169,29 @@ void TViewerPipeClient::BuildParamsFromFormData(TStringBuf data) {
...
+    const ui8 viewerTraceMaxVerbosity = limitTraceVerbosity
+        ? TComponentTracingLevels::DynamicNodesOnly
+        : NWilson::TTraceId::MAX_VERBOSITY;
 
     auto request = GetRequest();
     NWilson::TTraceId traceId;
     TString traceparent = request.GetHeader("traceparent");
     if (traceparent) {
-        traceId = NWilson::TTraceId::FromTraceparentHeader(traceparent, viewerTraceMaxVerbosity);
+        traceId = NWilson::TTraceId::FromTraceparentHeader(traceparent, viewerTraceDefaultVerbosity);
```
CyberROFL: > With ViewerControls.LimitTraceVerbosity=0, a request containing both a valid traceparent and X-Trace-Verbosity: 15 still receives verbosity 4. FromTraceparentHeader assigns the supplied default, and the subsequent !traceId guard skips processing the requested verbosity. Consequently, clients propagating an existing trace cannot use the new unlimited verbosity capability without dropping their parent trace context.
>
> Compute the requested, capped verbosity before choosing between continuing and creating a trace, and use it in both paths. Preserve verbosity 4 when no explicit verbosity is supplied. Add coverage for requests containing both headers.

CyberROFL (approved):

PR #53787, opened 2026-09-22
[NBS-7705] Balance ddisks manualy

drbasic:
### Changelog entry 

Added manual DDisk balancing controls to the Direct Block Group monitoring page.

### Description for reviewers 

- Added Balance buttons for individual DBGs and for each row of DBGs. Requests from the monitoring page use the touched VChunk strategy.
- Added a planner that selects source and target hosts using DDisk counts, pending allocations, and enabled host status. Each VChunk is selected at most once per request.
- A selected VChunk promotes a DDisk on the target host. Excess DDisks are then demoted from the most loaded eligible host.
- Moved imbalance calculation into the balance model. Monitoring now displays values calculated by the DBG for configured and touched VChunks, excluding disabled hosts.

Verified with the model, Direct Block Group, and monitoring unit test suites.


ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/direct_block_group_impl.cpp
```diff
@@ -284,6 +292,29 @@ void TDirectBlockGroup::CommitDDiskPromotion(const TVChunkConfig& config)
...
+THostMask TDirectBlockGroup::SelectDDiskForDemote(THostMask candidates) const
+{
+    Y_ABORT_UNLESS(ExecutorThreadChecker.Check());
+
+    if (candidates.Empty()) {
+        return THostMask::MakeEmpty();
+    }
+
+    const auto ddiskCountByHost =
+        CountDDisksByHost(EDDiskBalanceStrategy::Configured, candidates);
```
vazhem, line 315: **AI review (Major):** Touched balance can copy a replica and then delete it

`BalanceDDisks` promotes `targetHost` and leaves removal to `DemoteUnnecessaryHostsIfNeeded`. `SelectDDiskForDemote` always ranks the healthy replicas with `EDDiskBalanceStrategy::Configured` and never receives the source host the planner chose.

The monitoring buttons send `Touched`. A host can be the least loaded touched target and still be the heaviest host once untouched VChunks are counted. After the new replica becomes healthy it is a demote candidate, so demotion deletes the replica that was just copied and the touched imbalance stays where it was.

Touched counts `{1, 4, 3, 2, 2}` on five hosts need one move, host 1 → host 0. If host 0 is heavier in the configured counts, the plan promotes host 0 and demote picks host 0. Demoting host 1 instead yields `{2, 3, 3, 2, 2}`, which is balanced.

`ShouldPromoteTargetDDiskForBalance` calls `BalanceDDisks(0, 3)` on the mock, and the mock returns `candidates.First()`. Host 0 is both the requested source and the lowest index, so that test does not catch this.

Demote the planned source while it still has a healthy replica. Use the hottest remaining replica only as a fallback, ranked with the same strategy the plan used.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/direct_block_group_impl.cpp
```diff
@@ -2178,10 +2221,114 @@ TDBGDumpResponse TDirectBlockGroup::DoDebugPrintDirtyMap() const
...
+    return ddiskCountByHost;
+}
+
+bool TDirectBlockGroup::IsBalancingAllowed(
+    const TVChunk& vChunk,
+    EDDiskBalanceStrategy strategy) const
+{
+    Y_ABORT_UNLESS(ExecutorThreadChecker.Check());
+
+    return MatchesBalanceStrategy(vChunk, strategy) &&
```
vazhem, line 2287: **AI review (Major):** A second click stacks another full promotion

`IsBalancingAllowed` skips a VChunk only when `PendingDDiskAllocations` already contains it. `DoBalanceDDisks` never inserts the target it just chose. The in-memory config is also unchanged until the promote persist finishes, so `CountDDisksByHost` does not see that target.

The overview buttons do not wait for the previous request. A second POST runs `DoBalanceDDisks` again, plans the same moves, and queues another `PromoteHost`. Each prepare reads the latest config, so the promotions accumulate and one VChunk starts that many full copies. `CommitDDiskPromotion` already drops the pending entry when a config persist completes.

Record the chosen target in `PendingDDiskAllocations` before calling `BalanceDDisks`. The VChunk is then skipped, and the target counts toward host load.

vazhem (commented):
## AI review

The planner and `CalculateDDiskImbalance` agree on how many disks must move, and the overview page shows both figures and posts a touched-strategy request. The move that runs afterwards is a different one. `BalanceDDisks` only promotes the target; demotion later ranks replicas by every configured VChunk and can delete the replica that was just copied, so a touched balance makes no progress. A second click before that promote is persisted stacks another promotion on the same VChunk, because the target is never reserved.

ydb/core/nbs/cloud/blockstore/libs/storage/partition_direct/model/ddisk_balance.cpp
```diff
@@ -0,0 +1,202 @@
...
+std::array<TVector<const TVChunkConfig*>, MaxHostCount> CollectVChunksByHost(
+    const TVector<const TVChunkConfig*>& vChunks,
+    const std::array<size_t, MaxHostCount>& toMove)
+{
+    std::array<TVector<const TVChunkConfig*>, MaxHostCount> vchunksByHost;
+
+    // Keep registration order within each priority group.
+    for (size_t priority = 0; priority < 2; ++priority) {
+        for (const auto* config: vChunks) {
+            if ((config->GetEnabledDDisks().Count() >= 4) != (priority == 1)) {
```
vazhem: **AI review (Minor):** Sub-quorum VChunks are preferred as move sources

The preferred source group is every VChunk with fewer than 4 enabled DDisks. That cutoff is `QuorumDirectBlockGroupHostCount + 1` written as a literal, and it includes VChunks that are below quorum.

`GetUnnecessaryDDisksForDemote` removes a replica only when at least quorum + 1 disks are healthy. Promoting a VChunk that has two enabled disks adds a replica and leaves the planned source in place, after `toMove` for that host has already been consumed.

Skip candidates that cannot drop a replica after the promote, and compare with `QuorumDirectBlockGroupHostCount + 1`.

yudov-maksim (approved):

PR #53509, opened 2026-09-18
YQ-5689 Add UDF C++ SDK and CLI commands

akulaad:
### Changelog entry

Add the draft C++ UDF client and experimental `ydb experimental udf upload`, `list`, `describe`, and `delete` commands for managing WASM modules and libraries through the UDF service.

### Description for reviewers

Build on the UDF server API merged in #53376. Add a draft C++ UDF client with bounded-memory file uploads and a queued gRPC `WritesDone` operation to complete upload streams. Expose write modes, expected UID/checksum conditions, pagination, and per-platform compilation status through the draft SDK and experimental CLI, with JSON/YAML output where supported.

Compilation state is platform-specific: upload and list do not expose a global status, and list no longer accepts a global status filter. `describe` remains the readiness API and reports status per `cpu_spec`. The server-side artifact-state and protobuf transition is in [#53827](https://github.com/ydb-platform/ydb/pull/53827); compile timestamps will be exposed by this client after those protobuf fields reach `main`.

The client lives under `ydb-cpp-sdk/client/draft/ydb_udf.h`. The CLI is available only in the separately built experimental binary as `ydb experimental udf`; the official `ydb` binary does not register the command.

Add functional coverage for CLI round trips, write preconditions, YAML type preservation, pagination, and multi-chunk/large uploads. Include an upload-and-wait script and usage documentation updated for the experimental command path.

Validation:

- Built `ydb/apps/ydb`, `ydb/apps/ydb/experimental/ydb`, the draft SDK library, and the UDF functional target in `relwithdebinfo`: 905/905 targets.
- Rebuilt the changed draft SDK and experimental CLI with `-DUSER_CXXFLAGS=-Werror` after removing the global compile-state surface: 705/705 targets.
- Ran `ydb/tests/functional/udf_store`: 25 tests passed before the latest API-only adjustment.
- Verified the official CLI rejects `ydb udf` and the experimental CLI accepts `ydb experimental udf --help`.
- `git diff --check`, shell syntax, Python syntax, and stale stable-path checks passed.

PR #53827, opened 2026-09-22
Move UDF compile state to per-platform artifacts

akulaad:
An upload can compile successfully on one CPU platform and fail on another. This change completes the server-side transition to per-platform compilation state described in [RFC 002](https://github.com/ydb-platform/ydb-rfc/pull/344).

- Store status, error, start time and finish time in each `artifacts/{cpu_spec}` row.
- Stop using and updating compile state in the global `modules` row. The legacy physical columns and `.sys/udf_modules` contract remain unchanged for a separate security-owned change.
- Make module and library compile actors persist PENDING, COMPILING, READY and FAILED in their platform artifact table.
- Make `DescribeModule.platforms[]` read the persisted artifact state and timestamps instead of combining a global status with volatile controller state.
- Reserve removed protobuf fields and remove the global upload result and list status filter.
- Restore cold-start loading from ready artifact rows and track ready library artifacts by uid on each platform.
- Update direct upload and functional-test helpers for the new schema.

Backward compatibility for the UDF API is intentionally out of scope: the feature is not in production.

SDK and CLI surface changes are kept in [#53509](https://github.com/ydb-platform/ydb/pull/53509).

Validation:

- `./ya make -r ydb/services/udf_store ydb/core/udf_api ydb/core/sys_view/udf_modules`
- `./ya make --build relwithdebinfo -tA ydb/services/udf_store/ut_storage` (2 tests OK)
- Python syntax check for the updated functional helpers

The full `ut_orchestration` target was started separately but cancelled while compiling its large transitive graph; CI should run it in full.


ydb/services/udf_store/wasm_compile_actor.cpp
```diff
@@ -496,8 +473,8 @@ void TWasmCompileActor::FailAndPersist(const TString& message) {
         ReplyError(message);
         return;
     }
-    Step_ = EStep::UpdateMetaFailed;
-    ExecuteQuery(NTableQuery::BuildUpdateCompileStatusQuery(ModulesTablePath_), false);
+    Step_ = EStep::MarkArtifactFailed;
+    ExecuteQuery(NTableQuery::BuildMarkArtifactFailedQuery(ArtifactTablePath_), false);
```
martinevsky: > 🤖 **AI-generated review comment** — produced by Claude Opus 5.5 (`claude-opus-5-5`), posted at the request of a human reviewer. Please verify before acting.

**Every failed compile attempt is persisted as FAILED, although the controller still retries the key.**

`FailAndPersist` runs `BuildMarkArtifactFailedQuery` on every failed attempt (here and in `wasm_library_compile_actor.cpp:316-323`). The controller, however, only gives up after `WasmMaxCompileAttempts` (default 3): `attempt.Poisoned = attempt.FailCount >= MaxCompileAttempts` (`compile_controller/controller_impl.cpp:926-931`); otherwise the key is re-queued and the next `MarkCompiling` accepts `'failed'` (`table_query.cpp:230-238`).

So FAILED is not terminal, which contradicts the contract in `table_query.h:61-62`:

> Records a terminal failure of a compiling artifact. Retry policy belongs to the controller: do not use this for a failure that is still being retried.

Since `DescribeModule.platforms[]` now reports the persisted row as-is, a platform about to be retried shows `FAILED`, and the CI recipe in `ydb/core/udf_api/README.md` ("Fail on any failed platform") aborts on a transient first-attempt error that would have ended `READY`. In the base revision, `FAILED` only came from the controller's `Failed` flag, which is set for poisoned keys only.

Reproduced on the storage layer with a scratch unit test in `ut_storage` (EnsurePending → MarkCompiling → MarkFailed → MarkCompiling → Publish for the same uid):
```
after attempt 1: status=failed error='transient attempt 1 error'
retry start:     status=compiling
after attempt 2: status=ready
```
(The actor↔controller retry loop itself was checked by reading only.)

Possible fixes: pass an "is last attempt" flag with `TEvAssignCompile` and persist `FAILED` only then (otherwise go back to `PENDING`, keeping `compile_error` for diagnostics), or let the controller write the terminal state when it poisons the key. Note that `test_wasm_chunk_query_error_is_persisted` currently relies on `FAILED` appearing after the first attempt, so its wait would need to cover all attempts.

martinevsky (approved):

PR #53882, opened 2026-09-23
Introduce fresh chunk reservation

alexvru:
### Changelog entry 

Introduce fresh chunk reservation

### Description for reviewers 

First step of correct VDisk space projection feature update.


ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
```diff
@@ -2254,6 +2264,54 @@ void TPDisk::AttachSharedUringRouter(const TYardInit& evYardInit, TEvYardInitRes
...
+            // Leave both alone rather than free a chunk something is still working on.
+            continue;
+        }
+        // The chunk goes back to the free pool, so from now on it must be treated as holding
+        // someone else's data. DATA_RESERVED chunks were never marked.
+        if (TPDisk::IS_SHRED_ENABLED && !state.IsDirty) {
+            state.IsDirty = true;
+            isDirtyMarked = true;
+        }
+        Mon.UncommitedDataChunks->Dec();
```
mregrock, line 2293: Copypaste with 
https://github.com/ydb-platform/ydb/blob/995ade752ffdd7489d035593c98be96de856a7ac/ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp#L1715-L1723

mregrock (changes requested):

ydb/core/blobstorage/vdisk/hullop/blobstorage_hull.h
```diff
@@ -68,6 +70,12 @@ namespace NKikimr {
                 const TBarrierIngress& ingress,
                 const TActorContext& ctx);
 
+        // Calls `func(levelIndex, record, type)` for every database the admission has records for.
+        template <typename TFunc>
+        void ForEachFreshRecord(const TFreshAdmission& admission, TFunc&& func) const;
+        // Starts a Fresh compaction of the database, should one be due.
+        void CompactFreshIfRequired(EHullDbType type, const TActorContext& ctx);
```
mregrock: This method hides another namespace-level template with the same name
https://github.com/ydb-platform/ydb/blob/e2ed3953e820e59330c0ba3287fd1fc80cbbac73/ydb/core/blobstorage/vdisk/hullop/blobstorage_hull.cpp#L660

Code will probably fail on compile. Anyway, it will affect some calls inside, like
https://github.com/ydb-platform/ydb/blob/e2ed3953e820e59330c0ba3287fd1fc80cbbac73/ydb/core/blobstorage/vdisk/hullop/blobstorage_hull.cpp#L694

ydb/core/blobstorage/pdisk/blobstorage_pdisk_impl.cpp
```diff
@@ -2271,10 +2336,15 @@ bool TPDisk::YardInitForKnownVDisk(TYardInit &evYardInit, TOwner owner) {
 
     ownerData.OwnerRound = evYardInit.OwnerRound;
     TOwnerRound ownerRound = evYardInit.OwnerRound;
+    // The owner is starting over, so anything it reserved but never committed is unreachable.
+    // Release it before reporting what it owns, or it would come back as a leak it cannot fix.
+    ReleaseUncommittedChunks(owner);
```
mregrock: This call returns chunks to the free pool during YardInit. However, termination of the front actor does not guarantee that the old Hull has terminated. It can still send a delayed ChunkForget after a chunk has been reserved again by the new VDisk incarnation. Since the owner ID remains the same and ChunkForget only validates OwnerRound for DDisk, the stale request can free the new reservation and cause subsequent writes to fail. We should validate the round for ordinary VDisk requests at execution time as well.

ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
```diff
@@ -1423,6 +1629,22 @@ namespace NKikimr {
...
+            if (FreshGate) {
+                // One command writes a barrier and keep flags for blobs alike, and is admitted to both at once.
+                const auto& record = ev->Get()->Record;
+                TFreshAdmission admission;
+                admission.Barriers.AddIndexOnly(record.HasCollectGeneration() ? 1 : 0);
+                admission.LogoBlobs.AddIndexOnly(record.KeepSize() + record.DoNotKeepSize());
+                auto refuse = [&] {
+                    ReplyError({NKikimrProto::OUT_OF_SPACE, "out of space"}, ev, ctx, TAppData::TimeProvider->Now());
+                };
+                if (!AdmitToFresh(ev, std::move(admission), TOutOfSpaceLogic::FreshRefuseAtColorForGC(), true, ctx,
```
mregrock: When both Fresh reservation and the heap allocator are enabled, blocks/barriers reserve ordinary PDisk chunks, but their compactor writes to separately allocated heap stripes. These reservations remain unused until compaction finishes. Near the space limit, this can reject GC with OUT_OF_SPACE even when an existing heap stripe has enough space for the barrier SST. Admission should account for the allocator actually used by the compactor.

mregrock (changes requested):

ydb/core/blobstorage/vdisk/skeleton/blobstorage_skeleton.cpp
alexvru: This is the point of current design.

mregrock (approved):

